PR #64, opened 2022-08-08
cli serve, first pass at sandbox preflights

paulbellamy:
### What

Add `cli serve` command

It serves a jsonrpc API on `http://localhost:8080/rpc` which can be called with something like:
```sh
$ curl -v
    -X POST
    --header "content-type: application/json"
    http://localhost:8080/rpc
    --data-binary '{"jsonrpc":"2.0", "id":1, "method":"call", "params":{"id": "1", "func":"add", "args":[5, 3]}}'
```

This call command is similar to doing a cli invoke with:

```sh
$ cli invoke --contract 1 --fn add --arg 5 --arg 3
``` 

This is just a first pass at what the preflight request could look like. But it would need to include much more data in the response, so this will definitely change.

### Why

Fixes https://github.com/stellar/soroban-cli/issues/58

In particular, the idea is to let you run your dapp locally against the sandbox (The dev would do this separately). The API from `cli serve` should exactly match pubnet/futurenet. In one terminal/script you'd run `cli serve`, then in another you'd launch your web app server (e.g. `npm run dev`). As you click around in your browser the web app will submit txns and preflights to the cli server, simulating the real thing.

### Questions

- [ ] Should the `call` method be renamed to `invoke` to match the cli?
- [ ] Doesn't support submitting txns against the local sandbox, yet.
- [ ] Consider if/how to integrate this work with remote core backends instead of just local sandbox
- [ ] Consider how to share more code between this and the cli invoke subcommand.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
-    - run: find . -type f -name '*.rs' -print0 | xargs -I {} -0 rustfmt --check "{}"
+    - run: find . -type f -name '*.rs' -print0 | xargs -I {} -0 rustfmt --edition 2021 --check "{}"
 
   build-and-test:
     strategy:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
-    - run: find . -type f -name '*.rs' -print0 | xargs -I {} -0 rustfmt --edition 2021 --check "{}"
+    - run: cargo fmt --all --check
 
   build-and-test:
     strategy:
--- a/src/invoke.rs
+++ b/src/invoke.rs
@@ -1,18 +1,18 @@
-use std::{fmt::Debug, fs, io, io::Cursor, rc::Rc};
+use std::{fmt::Debug, fs, io, rc::Rc};
 
 use clap::Parser;
 use soroban_env_host::{
     budget::CostType,
     storage::Storage,
     xdr::{
-        Error as XdrError, HostFunction, ReadXdr, ScHostStorageErrorCode, ScObject, ScSpecEntry,
-        ScSpecFunctionV0, ScStatus, ScVal,
+        Error as XdrError, HostFunction, ReadXdr, ScHostStorageErrorCode, ScObject, ScStatus, ScVal,
     },
     Host, HostError, Vm,
 };
 
 use hex::FromHexError;
 
+use crate::contractspec;
 use crate::snapshot;
 use crate::strval::{self, StrValError};
 use crate::utils;
@@ -89,7 +89,7 @@ impl Cmd {
         let h = Host::with_storage(storage);
 
         let vm = Vm::new(&h, [0; 32].into(), &contents).unwrap();
-        let input_types = match Self::function_spec(&vm, &self.function) {
+        let input_types = match contractspec::function_spec(&vm, &self.function) {
             Some(s) => s.input_types,
             None => {
                 return Err(Error::FunctionNotFoundInContractSpec);
@@ -140,20 +140,5 @@ impl Cmd {
 
         snapshot::commit(ledger_entries, Some(&storage.map), &self.ledger_file)?;
         Ok(())
-    }
-
-    pub fn function_spec(vm: &Rc<Vm>, name: &str) -> Option<ScSpecFunctionV0> {
-        let spec = vm.custom_section("contractspecv0")?;
-        let mut cursor = Cursor::new(spec);
-        for spec_entry in ScSpecEntry::read_xdr_iter(&mut cursor).flatten() {
-            if let ScSpecEntry::FunctionV0(f) = spec_entry {
-                if let Ok(n) = f.name.to_string() {
-                    if n == name {
-                        return Some(f);
-                    }
-                }
-            }
-        }
-        None
     }
 }
--- a/src/invoke.rs
+++ b/src/invoke.rs
@@ -142,7 +142,7 @@ impl Cmd {
         Ok(())
     }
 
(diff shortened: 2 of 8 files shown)


src/jsonrpc.rs
```diff
@@ -0,0 +1,62 @@
...
+pub enum Id {
+    /// Null
+    Null,
+    /// Numeric id
+    Number(u64),
+    /// String id
+    Str(String),
+}
+
+/// JSON-RPC request object as defined in the [spec](https://www.jsonrpc.org/specification#request-object).
```
leighmcculloch, line 16: I don't think we need to do this now, but it may be worth considering if we can leverage an existing jsonrpc crate, such as https://docs.rs/jsonrpc.

paulbellamy:
per discussion w @leighmcculloch. The goal here is for the HTTP API to be identical between this (for local sandbox) and core (for production). To that end, we probably want the api parts here to be shared with core so that they stay in sync. @graydon Would you prefer the API parts here move to the https://github.com/stellar/rs-soroban-env crate? or a new crate of it's own (rs-soroban-api, or something)? Or just leave them here for a while and copy-paste it back-and-forth?

src/jsonrpc.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Parameter values of the request.
+    pub params: Option<T>,
+}
+
+/// JSON-RPC Response object as defined in the [spec](https://www.jsonrpc.org/specification#request-object).
+/// TODO: Figure out a cleaner way to do this.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub enum Response<T, E> {
```
leighmcculloch, line 35: This looks very similar to the builtin `Result<T, E>` type.

src/serve.rs
```diff
@@ -0,0 +1,244 @@
...
+        func: String,
+        args: Option<Vec<Value>>,
+        args_xdr: Option<Vec<String>>,
+    },
+}
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
+        // let context = HostContext::default();
+        // eprintln!("Hello World!");
```
leighmcculloch, line 69: 👀 ❓

src/serve.rs
```diff
@@ -0,0 +1,244 @@
...
+                            "code":-32603,
+                            "message": "Internal server error",
+                            },
+                        })
+                        .to_string()
+                    })
+                },
+            );
+
+        let addr: SocketAddr = ([127, 0, 0, 1], self.port.unwrap_or(8080)).into();
```
leighmcculloch, line 125: 💡 Can you make the port configurable? It's likely to be a pain point as 8080 can be often already in use.

.github/workflows/rust.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
-    - run: find . -type f -name '*.rs' -print0 | xargs -I {} -0 rustfmt --check "{}"
+    - run: find . -type f -name '*.rs' -print0 | xargs -I {} -0 rustfmt --edition 2021 --check "{}"
```
leighmcculloch, line 24: We can use this now:
```suggestion
    - run: cargo fmt --all --check
```

src/invoke.rs
```diff
@@ -142,7 +142,7 @@ impl Cmd {
         Ok(())
     }
 
-    fn function_spec(vm: &Rc<Vm>, name: &str) -> Option<ScSpecFunctionV0> {
+    pub fn function_spec(vm: &Rc<Vm>, name: &str) -> Option<ScSpecFunctionV0> {
```
leighmcculloch, line 145: Should move the fn into a new module, such as `contractspec.rs`.

src/jsonrpc.rs
```diff
@@ -0,0 +1,62 @@
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+
+#[derive(Debug, PartialEq, Clone, Hash, Eq, Deserialize, Serialize, PartialOrd, Ord)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
```
leighmcculloch, line 6: I think we should also add `#[serde(rename_all = "camelCase")]` to all values. We did that in the XDR lib, which is used in JSON in the CLI, so it'd keep both using the same form.

leighmcculloch (approved):
Nice. I left some comments, none critical to do before this is merged. Let me know if you want me to merge it now or wait.

The duplication looks fine between `invoke` and `serve`. I wouldn't try and consolidate them at this time, although it is valuable to keep their behavior as similar as possible when it comes to the actual invoke.

leighmcculloch:
> per discussion w @leighmcculloch. The goal here is for the HTTP API to be identical between this (for local sandbox) and core (for production). To that end, we probably want the api parts here to be shared with core so that they stay in sync. @graydon Would you prefer the API parts here move to the https://github.com/stellar/rs-soroban-env crate? or a new crate of it's own (rs-soroban-api, or something)? Or just leave them here for a while and copy-paste it back-and-forth?

@graydon @paulbellamy Could we get on a call and discuss this relationship that core and the jsonrpc server will have? There will be some overlap, but it's unclear to me that it is truly the same thing, because I don't think core will support all endpoints the jsonrpc server could likely grow to provide.

src/jsonrpc.rs
paulbellamy: Yeah, I looked at jsonrpsee, but I didn't want to drag in another dependency right now. Might be worth it long term, though.

src/jsonrpc.rs
paulbellamy: Yep, I tried adding a From/Into pair, but it was awkward with the extra "id" field.

paulbellamy:
Yeah, I expect the jsonrpc server will be a superset, with some of the endpoints initially provided by core (e.g. preflight), and some permanently supported by core (e.g. txsub).

PR #70, opened 2022-08-11
Add shell autocomplete using clap_complete crate

### LINKED ISSUE #69: CLI: Generate shell autocompletion configuration

It would be nice if there were shell completions to help quickly enter commands and explore command options. Coming from Kubernetes/kubectl, I find autocompletion indispensable once you have a decent number of commands/subcommands/args in your CLI tool.

Initially `bash` support would probably be sufficient, though `clap` completions are supported via https://crates.io/crates/clap_complete which supports common shells (bash, zsh, etc.).

- no completions
- manually generated completion scripts

Neither of these alternatives seem like good options to me.

I have a basic working implementation using `clap_complete` I can wrap up into a PR if this sounds like a desirable feature.

fdesjardins:
### What

This commit adds a `completion` subcommand that will generate autocomplete code for shells supported by `clap_complete`.

### Why

To enable users to enter commands and arguments more quickly and explore the CLI.

### Known limitations

N/A

### Usage & Testing

Build the CLI using `cargo build` then you should see the `completion` command after running `cargo run`. Running `cargo run completion --help` will display more detailed usage instructions.

To test the functionality run (for bash): `source <(cargo run completion bash)` and tab-completion for `soroban-cli` should be enabled in the current shell.

I've also been using an alias `sb` which can use completions by running `complete -o default -F _soroban-cli sb`.

If you use the suggested installation instructions in the `--help` completions should stay up-to-date with the `soroban-cli` if a new version is installed.

Closes #69

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,6 +140,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "3.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ead064480dfc4880a10764488415a97fdd36a4cf1bb022d372f02e8faf8386e1"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "3.2.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1110,6 +1119,7 @@ version = "0.0.2"
 dependencies = [
  "base64",
  "clap",
+ "clap_complete",
  "hex",
  "num-bigint",
  "serde",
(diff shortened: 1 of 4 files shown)


leighmcculloch (commented):
Thanks @fdesjardins!!!

I pushed a couple small tweaks (running cargo fmt, and using the clap_complete arg enum instead of redefining our own).

It looks all good to me, but I ran into an issue getting completion to work, so I want to return to that and make sure it works for me, then I'll merge.

fdesjardins:
Thanks @leighmcculloch, I originally had non-positional args then decided to make it positional to reduce typing a bit. :laughing: I'd also wondered about using the clam_complete enum but my newness to Rust got me there.

Let me know if you want to see any other changes, I have some free time to work on things so I won't mind at all.

leighmcculloch:
For some reason completion doesn't work for me, but my setup is a little bespoke. @sisuresh confirmed it works on his machine 🎉.

leighmcculloch (approved):

PR #71, opened 2022-08-11
Support G- strkey public keys as arguments

### LINKED ISSUE #53: CLI: Map strkeys (e.g. G...) to FixedBinary<32>
People will naturally want to use Stellar's strkey format for keys as inputs to functions or arguments to things. We should map this to `FixedBinary<32>` or to some other type, so that people can continue to use those keys. ~Possibly now is the right time to finish implementing the `PublicKey` type that is still in the XDR.~

cc @paulbellamy @jayz22 @jonjove 

I'm revisiting this because of this question in Discord by @fsodano:
https://discord.com/channels/897514728459468821/966788672164855829/1003810451408306246

2opremio:
Closes #53

### DIFF
--- a/src/strval.rs
+++ b/src/strval.rs
@@ -7,6 +7,8 @@ use soroban_env_host::xdr::{
     ScSpecTypeOption, ScSpecTypeTuple, ScSpecTypeVec, ScStatic, ScVal, ScVec, VecM,
 };
 
+use stellar_strkey::StrkeyPublicKeyEd25519;
+
 #[derive(Debug)]
 pub enum StrValError {
     UnknownError,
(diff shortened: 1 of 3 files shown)


src/strval.rs
```diff
@@ -61,12 +63,24 @@ pub fn from_string(s: &str, t: &ScSpecTypeDef) -> Result<ScVal, StrValError> {
...
-                        .try_into()
-                        .map_err(|_| StrValError::InvalidValue)?,
-                ))),
+                _ =>
+                // it could be a G- strkey
+                {
+                    match StrkeyPublicKeyEd25519::from_string(s) {
+                        Ok(key) => {
+                            let b = &key.0;
+                            ScVal::Object(Some(ScObject::Binary(
```
paulbellamy, line 72: Shouldn't this be an ScPublicKey? Or maybe that type is missing from the ScSpecTypeDef?

2opremio: I am using `stellar_strkey::StrkeyPublicKeyEd25519` I can't find `ScPublicKey`.

leighmcculloch: ScPublicKey and ScHash aren't being used anywhere at the moment, even within the host. All the examples that do have keys use `Binary`.

leighmcculloch (approved):
Nice 🎉 Thanks @2opremio!

PR #87, opened 2022-08-25
Update cli serve api

paulbellamy:
### What

Update the `cli serve` api

### Why

To match the new API design, and also include `footprint` and `cost` fields for `simulateTransaction`.

Follow-up from https://github.com/stellar/soroban-cli/issues/58 and https://github.com/stellar/soroban-cli/issues/78

### Known limitations

- General parsing, validation, and return values are all pretty janky for now. I've left `TODO` where there is obvious work needed.

### DIFF
--- a/src/serve.rs
+++ b/src/serve.rs
@@ -4,6 +4,7 @@ use clap::Parser;
 use hex::FromHexError;
 use serde_json::{json, Value};
 use soroban_env_host::{
+    budget::Budget,
     storage::{AccessType, Footprint, Storage},
     xdr::{
         Error as XdrError, FeeBumpTransactionInnerTx, HostFunction, OperationBody, ReadXdr,
@@ -144,7 +145,7 @@ fn reply(
 
 fn simulate_transaction(txn_xdr: &str, ledger_file: &PathBuf) -> Result<Value, Error> {
     // Parse and validate the txn
-    let ops = match TransactionEnvelope::from_xdr_base64(txn_xdr)? {
+    let ops = match TransactionEnvelope::from_xdr_base64(txn_xdr.to_string())? {
         TransactionEnvelope::TxV0(envelope) => envelope.tx.operations,
         TransactionEnvelope::Tx(envelope) => envelope.tx.operations,
         TransactionEnvelope::TxFeeBump(envelope) => {
@@ -208,7 +209,7 @@ fn simulate_transaction(txn_xdr: &str, ledger_file: &PathBuf) -> Result<Value, E
 
     let snap = Rc::new(snapshot::Snap { ledger_entries });
     let storage = Storage::with_recording_footprint(snap);
-    let h = Host::with_storage(storage);
+    let h = Host::with_storage_and_budget(storage, Budget::default());
 
     // TODO: Check the parameters match the contract spec, or return a helpful error message
     let mut complete_args = vec![
@@ -229,11 +230,11 @@ fn simulate_transaction(txn_xdr: &str, ledger_file: &PathBuf) -> Result<Value, E
     let mut cost = serde_json::Map::new();
     cost.insert(
         "cpu_insns".to_string(),
-        Value::String(budget.cpu_insns.get_count().to_string()),
+        Value::String(budget.get_cpu_insns_count().to_string()),
     );
     cost.insert(
         "mem_bytes".to_string(),
-        Value::String(budget.mem_bytes.get_count().to_string()),
+        Value::String(budget.get_mem_bytes_count().to_string()),
     );
     // TODO: Include these extra costs. Figure out the rust type conversions.
     // for cost_type in CostType::variants() {
--- a/src/serve.rs
+++ b/src/serve.rs
@@ -220,30 +220,28 @@ fn simulate_transaction(txn_xdr: &str, ledger_file: &PathBuf) -> Result<Value, E
 
     let res = h.invoke_function(HostFunction::Call, complete_args.try_into()?)?;
 
-    // Calculate the budget usage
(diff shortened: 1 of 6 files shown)


src/serve.rs
```diff
@@ -55,186 +51,221 @@ pub enum Error {
...
+        );
+        // TODO: Include these extra costs. Figure out the rust type conversions.
+        // for cost_type in CostType::variants() {
+        //     m.insert(cost_type, b.get_input(*cost_type));
+        // }
+        m
+    });
 
+    // Calculate the storage footprint
     let (storage, _, _) = h.try_finish().map_err(|_h| {
```
leighmcculloch, line 242: Instead of calling get_budget above, you should use the second parameter returned in `try_finish` which will be a `Budget`. See this code as an example:


https://github.com/stellar/soroban-cli/blob/13a51da46ee7d9553554f703ef28d83443367646/src/invoke.rs#L125-L137

src/serve.rs
```diff
@@ -55,186 +51,221 @@ pub enum Error {
...
     })?;
+    let mut read_only: Vec<String> = vec![];
+    let mut read_write: Vec<String> = vec![];
+    let Footprint(m) = storage.footprint;
+    for (k, v) in m {
+        let dest = match v {
+            AccessType::ReadOnly => &mut read_only,
+            AccessType::ReadWrite => &mut read_write,
+        };
+        dest.push(base64::encode(k.to_xdr()?));
```
leighmcculloch, line 255:
```suggestion
        dest.push(k.to_xdr_base64()?);
```

src/serve.rs
```diff
@@ -55,186 +51,221 @@ pub enum Error {
...
-    func: &String,
-    args: &[Value],
-    args_xdr: &[String],
-    ledger_file: &PathBuf,
-) -> Result<ScVal, Error> {
-    let contract_id: [u8; 32] = utils::contract_id_from_str(contract)?;
+fn simulate_transaction(txn_xdr: &str, ledger_file: &PathBuf) -> Result<Value, Error> {
+    // Parse and validate the txn
+    let decoded = base64::decode(txn_xdr).map_err(|_| Error::Xdr(XdrError::Invalid))?;
+    let ops = match TransactionEnvelope::from_xdr(decoded)? {
```
leighmcculloch, line 148:
```suggestion
    let ops = match TransactionEnvelope::from_xdr_base64(txn_xdr)? {
```

src/serve.rs
```diff
@@ -55,186 +51,221 @@ pub enum Error {
...
+        TransactionEnvelope::TxV0(envelope) => envelope.tx.operations,
+        TransactionEnvelope::Tx(envelope) => envelope.tx.operations,
+        TransactionEnvelope::TxFeeBump(envelope) => {
+            let FeeBumpTransactionInnerTx::Tx(tx_envelope) = envelope.tx.inner_tx;
+            tx_envelope.tx.operations
+        }
+    };
+    if ops.len() != 1 {
+        return Err(Error::Xdr(XdrError::Invalid));
```
leighmcculloch, line 157: It would be ideal to return to this later and add unique errors for each of these error cases, but I understand we probably don't have time for that now. Can you follow up with that at a later time?

src/serve.rs
```diff
@@ -55,186 +51,221 @@ pub enum Error {
...
+    // snapshot::commit(ledger_entries, Some(&storage.map), ledger_file)?;
 
-    Ok(res)
+    Ok(json!({
+        "cost": cost,
+        "footprint": {
+            "readOnly": read_only,
+            "readWrite": read_write,
+        },
+        "xdr": base64::encode(res.to_xdr()?),
```
leighmcculloch, line 267:
```suggestion
        "xdr": res.to_xdr_base64()?,
```

leighmcculloch (approved):
👍🏻 Couple small suggested changes but looks good. One request for a follow up change when you have time.

src/serve.rs
paulbellamy: Oh cool. I think I copied it from an old version of invoke. Will update.

src/serve.rs
paulbellamy: The error messages in general need a *bunch* of improvement.

paulbellamy: Added https://github.com/stellar/soroban-cli/issues/90 to track error message improvements.

paulbellamy:
Ok, updated the `soroban-env-host` to have the base64 xdr update. Had to change the type signature of `snapshot::commit` due to metering changes here https://github.com/stellar/rs-soroban-env/pull/307

src/serve.rs
```diff
@@ -55,186 +52,218 @@ pub enum Error {
...
+    // binary. The cap says binary, but other implementations use symbol.
+    let method: String = if let ScVal::Object(Some(ScObject::Bytes(bytes))) = method_xdr {
+        bytes
+            .try_into()
+            .map_err(|_| Error::Xdr(XdrError::Invalid))?
+    } else if let ScVal::Symbol(bytes) = method_xdr {
+        bytes
+            .try_into()
+            .map_err(|_| Error::Xdr(XdrError::Invalid))?
     } else {
```
leighmcculloch, line 203: I thought we had added conversions for these that were easier, like scval.try_into()? for string, but I might be wrong.

leighmcculloch (approved):

src/serve.rs
paulbellamy: Oh maybe it wasn't in the older version of the sdk, but they're there now?

paulbellamy: Hmmm... Edit still says there's no ScVal -> String TryInto implementation.

PR #91, opened 2022-08-25
Add json output option for gen command

tamirms:
### What

There is an existing `gen` command which currently generates rust bindings for a soroban smart contract. This PR adds an option to generate a JSON representation for a soroban smart contract spec.

### Why

Fixes https://github.com/stellar/soroban-cli/issues/56

### Known limitations

[N/A]

### DIFF
--- a/src/gen.rs
+++ b/src/gen.rs
@@ -1,10 +1,6 @@
-use std::{fmt::Debug, fs, io, io::Cursor};
+use std::fmt::Debug;
 
 use clap::{ArgEnum, Parser};
-use soroban_env_host::{
-    xdr::{ReadXdr, ScSpecEntry},
-    Host, HostError, Vm,
-};
 use soroban_spec::gen::{json, rust};
 
 #[derive(Parser, Debug)]
@@ -31,16 +27,8 @@ pub enum Error {
     GenerateRustFromFile(rust::GenerateFromFileError),
     #[error("format rust error: {0}")]
     FormatRust(syn::Error),
-    #[error("host")]
-    Host(HostError),
-    #[error("xdr error: {0}")]
-    Xdr(soroban_env_host::xdr::Error),
-    #[error("io")]
-    Io(io::Error),
-    #[error("serialize json error: {0}")]
-    Json(serde_json::Error),
-    #[error("contractnotfound")]
-    ContractSpecNotFound,
+    #[error("generate json from file: {0}")]
+    GenerateJsonFromFile(json::GenerateFromFileError),
 }
 
 impl Cmd {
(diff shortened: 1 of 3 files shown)


src/gen.rs
```diff
@@ -17,38 +21,59 @@ pub struct Cmd {
...
+    #[error("host")]
+    Host(#[from] HostError),
+    #[error("xdr error: {0}")]
+    Xdr(#[from] soroban_env_host::xdr::Error),
+    #[error("io")]
+    Io(#[from] io::Error),
+    #[error("json error: {0}")]
+    Json(#[from] serde_json::Error),
+    #[error("contractnotfound")]
+    ContractSpecNotFound,
```
leighmcculloch, line 43: This code intentionally didn't use the #from attribute because the intent was to map specific error situations and not broad error categories.

src/gen.rs
```diff
@@ -17,38 +21,59 @@ pub struct Cmd {
...
-        match syn::parse_file(&code_raw) {
-            Ok(file) => {
-                let code_fmt = prettyplease::unparse(&file);
-                println!("{}", code_fmt);
-                Ok(())
-            }
-            Err(e) => {
-                println!("{}", code_raw);
-                Err(Error::FormatRustError(e))
+        let code_fmt = prettyplease::unparse(&syn::parse_file(&code_raw)?);
```
leighmcculloch, line 58: `syn::parse_file` is very finicky and will fail to parse in situations that are difficult to debug. This code intentionally printed out the raw code without formatting if this happens so we can debug it in those cases.

src/gen.rs
tamirms: I removed the uses of #from

src/gen.rs
tamirms: I reverted all my changes to generate_rust

src/gen.rs
```diff
@@ -47,8 +64,31 @@ impl Cmd {
...
+            let json_entries: Result<Vec<json::Entry>, Error> =
+                ScSpecEntry::read_xdr_iter(&mut cursor)
+                    .map(|spec_entry| {
+                        json::Entry::try_from(&spec_entry.map_err(Error::Xdr)?).map_err(Error::Xdr)
+                    })
+                    .collect();
+            println!(
+                "{}",
+                serde_json::to_string(&json_entries?).map_err(Error::Json)?
+            );
```
leighmcculloch, line 88: We should be able to call into the soroban-spec crate to do most of this work. I realize when we added the JSON logic we didn't add the same functions the rust version had, so I just added them in https://github.com/stellar/rs-soroban-sdk/pull/498, and will change this to use them and merge.

PR #96, opened 2022-08-28
Allow mixing args and xdr args

2opremio:
### What

It allows mixing `--arg` and `--arg-xdr` arguments.

Additionally we will now complain when more arguments that needed are provided (up until now we allowed providing more arguments and just using the first needed)

Implementation detail: when using Clap's `#derive` API, the position of the arguments isn't saved so I had to expand the call to `Root::parse()` in order to save the arguments matches (which contains the ordering) and pass it down to the `invoke` subcommand.

### Why

Addresses part of https://github.com/stellar/soroban-cli/issues/22

### Known limitations

N/A

### DIFF
--- a/src/invoke.rs
+++ b/src/invoke.rs
@@ -32,12 +32,7 @@ pub struct Cmd {
     #[clap(long = "arg", value_name = "arg", multiple = true)]
     args: Vec<String>,
     /// Argument to pass to the function (base64-encoded xdr)
-    #[clap(
-        long = "arg-xdr",
-        value_name = "arg-xdr",
-        multiple = true,
-        conflicts_with = "args"
-    )]
+    #[clap(long = "arg-xdr", value_name = "arg-xdr", multiple = true)]
     args_xdr: Vec<String>,
     /// Output the cost execution to stderr
     #[clap(long = "cost")]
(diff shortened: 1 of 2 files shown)


src/invoke.rs
```diff
@@ -97,27 +100,42 @@ impl Cmd {
...
+            return Err(Error::TooManyArguments);
+        }
+
+        let parsed_args: Vec<ScVal> = all_indexed_args
+            .iter()
+            .zip(inputs.iter())
+            .map(|(arg, input)| match &arg.1 {
+                Arg::ArgXDR(s) => match base64::decode(s) {
                     Err(_) => Err(StrValError::InvalidValue),
                     Ok(b) => ScVal::from_xdr(b).map_err(StrValError::Xdr),
```
paulbellamy, line 128: Unrelated Nit/Suggestion: We can actually use `ScVal::from_xdr_base64` now, and it will decode and parse in one step.

src/invoke.rs
```diff
@@ -97,27 +100,42 @@ impl Cmd {
...
+            .unwrap_or_default()
+            .zip(self.args_xdr.iter())
+            .map(|(a, b)| (a, Arg::ArgXDR(b.to_string())))
+            .collect();
+        let mut all_indexed_args: Vec<(usize, Arg)> = [indexed_args, indexed_args_xdr].concat();
+        all_indexed_args.sort_by(|a, b| a.0.cmp(&b.0));
+
+        if all_indexed_args.len() > inputs.len() {
+            return Err(Error::TooManyArguments);
+        }
```
paulbellamy, line 120: Suggestion: Might be nice to do a too few arguments one as well?

paulbellamy (approved):
Sick.

PR #105, opened 2022-08-30
Print event at the end of invocation

jayz22:
### What

Add a print statement at the end to print out the `events` from `host`.  cc @sisuresh 

### Why

To make the contract event example more meaningful. It's much better to show the event being printed than describing it as part of the host that can be retrieved. 

### Known limitations

The print relies on the `Debug` trait of `xdr::ContractEvent`, which isn't pretty. We might want to add the proper `Display` impl of it at the `xdrgen` later.

### DIFF
--- a/src/invoke.rs
+++ b/src/invoke.rs
@@ -3,6 +3,7 @@ use std::{fmt::Debug, fs, io, rc::Rc};
 use clap::Parser;
 use soroban_env_host::{
     budget::{Budget, CostType},
+    events::HostEvent,
     storage::Storage,
     xdr::{
         Error as XdrError, HostFunction, ReadXdr, ScHostStorageErrorCode, ScObject,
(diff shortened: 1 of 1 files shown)


jayz22:
Here is what is looks like:
Screen Shot 2022-08-30 at 5 26 13 PM

leighmcculloch:
Could the output be JSON? The XDR should serializable to JSON pretty printed.

jayz22:
Updated to using Json pretty printing.

sisuresh:
What does the output look like with pretty printing? The last time I used serde's pretty printing, each byte of an array was on a new line, which made it hard to read.

jayz22:
@sisuresh Yeah you are right, it does span many lines. I'm neutral on print vs pretty print. I can change it to print if you prefer.

jayz22:
Switched json pretty string to regular string as it fits the screen better:
Screen Shot 2022-08-30 at 6 12 29 PM

sisuresh:
Yeah I think the regular string is better.

jayz22:
> Not something we have to do right now, but probably will need to as part of the RPC work. 

@leighmcculloch Would this change be part of the `env` or the `cli`. If it's in the `env` I can create an issue for it.

leighmcculloch:
This would be in the CLI / RPC, and independent of the host/env.

PR #108, opened 2022-08-31
Make read accept strval or xdr, and output strval, json, or xdr

leighmcculloch:
### What
Make read accept strval or xdr, and output strval, json, or xdr.

### Why
To make it easier to use, and to include its use in examples.


paulbellamy:
The `--output` thing is nice. Would be good for this to be consistent across other subcommands (e.g. `invoke`)

PR #110, opened 2022-08-31
Improve error reporting

2opremio:
The serve subcommand still has room for improvement, but I will tackle that later


src/error.rs
```diff
@@ -0,0 +1,82 @@
...
+    xdr::{Error as XdrError, ScVal},
+    HostError,
+};
+use soroban_spec::gen::{json, rust};
+
+use hex::FromHexError;
+use std::io;
+
+#[derive(thiserror::Error, Debug)]
+pub enum CmdError {
```
leighmcculloch, line 13: Moving these errors into a single place makes it difficult to ascertain what error conditions exist for each sub-command. The errors were listed in each subcommand as a goal in improving the ability to understand the error conditions of each command we add. What's the motivation for moving all the errors into a single file?

leighmcculloch (changes requested):

src/error.rs
2opremio: Multiple commands use the same errors, contract id parsing, ledger file opening etc.

leighmcculloch: There are some duplicates, but I don't think we need to have complete code deduplication. As this tool grows this error list will grow, and I think it will become difficult to manage a large list of unrelated errors.

leighmcculloch: I think we can improve the error reporting without deduplicating the errors into a single list?

2opremio: I suppose I can defactor it.

2opremio: PTAL

src/deploy.rs
```diff
@@ -23,25 +22,61 @@ pub struct Cmd {
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error("io")]
-    Io(#[from] io::Error),
-    #[error("xdr")]
+    #[error(transparent)]
+    // TODO: the Display impl of host errors is pretty user-unfriendly
+    //       (it just calls Debug). I think we can do better than that
+    Host(#[from] HostError),
```
leighmcculloch, line 28: Instead of fixing the host error here, we should fix this in the host, so that we address the problem in all the places it's going to show up (here in the CLI, in Rust tests of contracts, etc). cc @graydon @jayz22
```suggestion
    #[error(transparent)]
    Host(#[from] HostError),
```

src/gen.rs
```diff
@@ -13,24 +13,24 @@ pub struct Cmd {
...
-
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     #[error("generate rust from file: {0}")]
-    GenerateRustFromFile(rust::GenerateFromFileError),
+    RustFromFile(rust::GenerateFromFileError),
     #[error("format rust error: {0}")]
     FormatRust(syn::Error),
     #[error("generate json from file: {0}")]
     GenerateJsonFromFile(json::GenerateFromFileError),
```
leighmcculloch, line 23:
```suggestion
    GenerateRustFromFile(rust::GenerateFromFileError),
    #[error("format rust error: {0}")]
    FormatRust(syn::Error),
    #[error("generate json from file: {0}")]
    GenerateJsonFromFile(json::GenerateFromFileError),
```
Can we keep the Generate prefix for the Rust, or remove the Generate prefix from the Json one? They're the same type of error and benefit from looking the same. It's unclear what benefit we get by shortening one.

src/invoke.rs
```diff
@@ -48,8 +47,8 @@ pub struct Cmd {
 pub enum Error {
     #[error("parsing argument {arg}: {error}")]
     CannotParseArg { arg: String, error: StrValError },
-    #[error("parsing XDR argument {arg}: {error}")]
-    CannotParseXDRArg { arg: String, error: XdrError },
+    #[error("parsing XDR key {arg}: {error}")]
+    CannotParseXdrArg { arg: String, error: XdrError },
```
leighmcculloch, line 51:
```suggestion
    #[error("parsing XDR arg {arg}: {error}")]
    CannotParseXdrArg { arg: String, error: XdrError },
```
Looks like this was maybe changed by mistake?

leighmcculloch (commented):
Thanks @2opremio, this is great. The new error messages are much easier to understand, and I noticed you replace a lot of automatic error conversion with specific unique errors for each situation, and the code, while longer, is now much easier to understand and the consistent across the board is going to help a lot as this app scales.

There's a couple small things I mention inline. I'll fix them and merge.

leighmcculloch (approved):

PR #121, opened 2022-09-08
Add publish dry run to build on release branches

leighmcculloch:
### What
Add publish dry run to build on release branches.

### Why
There is nothing that checks the release process will succeed before we perform the release, fixing problems after starting the publish process is painful. This will help catch problems earlier when we are bumping versions.


leighmcculloch:
This build won't pass because the CLI is currently dependent on unreleased versions of its deps, which is exactly the type of thing this change is intended to detect and block. Since I'm just adding this change to CI now for future use, I'm closing this change, and opening the same change with a different branch name so that the check won't run now and we can merge it now. See #122.

Issue #42, opened 2022-07-15
CLI: Invoke a contract on network


paulbellamy:
### What problem does your feature solve?

When deploying/managing a contract the dev may wish to call some setup functions (e.g. `contract.init()`). This way they can issue "management" commands and other contract calls to their live contract.

### What would you like to see?

`$ stellar-contract-cli --network testnet call <contract-id> init '<args>'`
(or something)

### What alternatives are there?

Don't do it? Make them use laboratory for now?

leighmcculloch:
The format of the command given the CLI today would be something like:

```
soroban-cli --network testnet invoke --id <contract-id> --fn init --arg ...
```

Where no `--network` option indicates sandbox, and any `--network` option indicates a remove (even locally remote) network.

leighmcculloch:
After giving this issue some more thought in conjunction with #41 and #142, the form of the command would be something like:

1. `soroban-cli invoke --horizon-url <URL> --account <G...> --id <CONTRACTID> --fn <FN> --arg ...`
   - Make a HTTP request to Horizon to get the sequence number of the account.
   - Build a transaction with:
      - Source account set to the given account.
      - Sequence number set to the next sequence number for the source account.
      - One `INVOKE_HOST_FUNCTION` operation that calls the invoke function and includes the given WASM. 
   - Prompt the user to enter the S secret key.
   - Sign the transaction with the S secret key.
   - Submit the transaction to the transaction submission endpoint of the Horizon URL.

2. `soroban-cli invoke --rpc-url <URL> --account <G...> --id <CONTRACTID> --fn <FN> --arg ...`
   - Make a HTTP request to the RPC server to get the sequence number of the account.
   - Build a transaction with:
      - Source account set to the given account.
      - Sequence number set to the next sequence number for the source account.
      - One `INVOKE_HOST_FUNCTION` operation that calls the invoke function and includes the given WASM. 
   - Prompt the user to enter the S secret key.
   - Sign the transaction with the S secret key.
   - Submit the transaction to the `sendTransaction` endpoint of the RPC server.

@paulbellamy @tomerweller Questions:

1. Does the above align with your expectations?
1. Does this align with plans in the RPC server?
1. What will be the preferred flow for invoking, via an RPC server or via Horizon? (So that I can prioritize the respective implementation.)

Issue #39, opened 2022-07-15
CLI: CLI Event Streaming Tool for local sandbox


paulbellamy:
Idea is a local CLI tool to stream events for devs, such as something like:
```
$ cli events --contract-id 01
{"type": "Transfer", "data": {...}}
```

Relevant CAP: https://github.com/stellar/stellar-protocol/issues/1280
Soroban-RPC API: https://docs.google.com/document/d/1TZUDgo_3zPz7TiPMMHVW_mtogjLyPL0plvzGMsxSz6A/edit#heading=h.t03hji406t76

### Background & Planning work

- [ ] Design a storage format and file for events in the local sandbox
  - `.soroban/events.json` with a list of events serialized to json might be fine. or xdr?
  - The storage and transfer mechanism here should support linux/mac/windows
- [ ] Add the command to the CLI
  - `soroban-cli events`
  - Default behaviour is to print the latest N messages from the sandbox. (or all messages? see what kubectl logs does).
  - Figure out a nice human-friendly format to print the events in.
- [ ] stream events from the sandbox as they come in.
  - `soroban-cli events -f`
- [ ] add `--id CONTRACT_ID` flag for per-contract filtering (or `--id` to match other functions?)
- [ ] Add `getEvents` method support in `soroban serve`

## Nice-to-haves

- [ ] allow output in json (strval) format, as well as xdr?

paulbellamy:
Blocked on https://github.com/stellar/stellar-core/issues/3487

paulbellamy:
Deferring this to sept/october, as I don't think it is critical for the first e2e example

tomerweller:
> HTTP api for this in cli http server (or move to new issue)

+1 on this. 

This almost seems more important than streaming to the standard output. Standard output is great for debug but, if the developer wants to use the events, wouldn't they prefer an HTTP endpoint? 

More so, assuming that RPC servers use the same API spec, this should provide a smooth transition from local testing to production RPC, no?

leighmcculloch:
I think this functionality is important for developer testing their applications in sandbox, but it should be a small lift in addition to the HTTP endpoint since they'd both be backed by the same functionality.

leighmcculloch:
> $ cli events --contract-id 01

This subcommand structure makes sense I think for a deployed contract, but less so for a contract running in sandbox. For a contract running in the sandbox and being invoked via `soroban-cli invoke`, I think the events should just be dumped to stdout as part of the output of the invoke command.

[...]

PR #137, opened 2022-09-16
invoke: add --event and --event-file flags

### LINKED ISSUE #128: CLI: Print events during invoke if --events option is present
I think we should output events, including any `DebugEvents` the system has generated to stdout, or somewhere, when the `invoke` command is used.

This behavior was previously described here: https://github.com/stellar/soroban-cli/issues/39#issuecomment-1238517053

2opremio:
### What

Add new event-related flags to the invoke subcommand:

* --events: enable printing host/debug events
* --events-file: write events to a file instead of stdout

### Why

1. Close #128
2. Enable saving events to a file/pipe which in turn will be used by `soroban-cli serve` to subscribe to local events (a new, matching `--events-file` flag will be added to to `sorobab-cli serve` for that purpose). This is part of https://github.com/stellar/soroban-cli/issues/39

### Known limitations

N/A

### DIFF
--- a/src/invoke.rs
+++ b/src/invoke.rs
@@ -1,4 +1,9 @@
-use std::{fmt::Debug, fs, io, rc::Rc};
+use std::{
+    fmt::Debug,
+    fs::{self, OpenOptions},
+    io::{self, Write},
+    rc::Rc,
+};
 
 use clap::Parser;
 use hex::FromHexError;
@@ -42,6 +47,12 @@ pub struct Cmd {
     /// File to persist ledger state
     #[clap(long, parse(from_os_str), default_value(".soroban/ledger.json"))]
     ledger_file: std::path::PathBuf,
+    /// Output file to write events
+    #[clap(long, parse(from_os_str))]
+    events_file: Option<std::path::PathBuf>,
+    /// Write events (to stdout unless --events_file is set)
+    #[clap(long)]
+    events: bool,
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -66,6 +77,13 @@ pub enum Error {
         filepath: std::path::PathBuf,
         error: io::Error,
     },
+    #[error("opening file {filepath}: {error}")]
+    CannotOpenEventsFile {
+        filepath: std::path::PathBuf,
+        error: io::Error,
+    },
+    #[error("writing event: {error}")]
+    CannotWriteEvent { error: io::Error },
     #[error("committing file {filepath}: {error}")]
     CannotCommitLedgerFile {
         filepath: std::path::PathBuf,
@@ -250,11 +268,27 @@ impl Cmd {
             }
         }
 
-        for (i, event) in events.0.iter().enumerate() {
-            eprintln!("Event #{}:", i);
-            match event {
-                HostEvent::Contract(e) => eprint!("{}", serde_json::to_string(&e).unwrap()),
-                HostEvent::Debug(e) => eprint!("{}", e),
+        if self.events {
+            let mut out: Box<dyn io::Write> = Box::new(io::stdout());
+            if let Some(f) = &self.events_file {
+                let file = OpenOptions::new()
+                    .create_new(true)
+                    .write(true)
+                    .append(true)
+                    .open(f)
+                    .map_err(|e| Error::CannotOpenEventsFile {
+                        filepath: f.to_path_buf(),
+                        error: e,
+                    })?;
+                out = Box::new(file);
+            }
+
+            for event in events.0.iter() {
+                let event_str = match event {
+                    HostEvent::Contract(e) => serde_json::to_string(&e).unwrap(),
+                    HostEvent::Debug(e) => format!("{}", e),
+                };
+                writeln!(out, "{}", event_str).map_err(|e| Error::CannotWriteEvent { error: e })?;
             }
         }
 
--- a/src/invoke.rs
+++ b/src/invoke.rs
@@ -277,13 +277,13 @@ impl Cmd {
                     .append(true)
                     .open(f)
                     .map_err(|e| Error::CannotOpenEventsFile {
-                        filepath: f.to_path_buf(),
+                        filepath: f.clone(),
                         error: e,
                     })?;
                 out = Box::new(file);
             }
 
-            for event in events.0.iter() {
+            for event in events.0 {
                 let event_str = match event {
                     HostEvent::Contract(e) => serde_json::to_string(&e).unwrap(),
                     HostEvent::Debug(e) => format!("{}", e),


leighmcculloch:
> 2\. Enable saving events to a file/pipe which in turn will be used by `soroban-cli serve` to subscribe to local events (a new matching `--events-file` flag will be added to to `sorobab-cli` for that purpose)

The dependency of serve on invoke via a file/pipe seems unnecessary, and unintuitive, and breaks the "these are separate runtimes" model.

I don't think someone running `soroban-cli serve` should see events from `soroban-cli invoke` commands. Invoke and serve should be independent of each other, other than sharing the same `ledger.json` file for data storage if they so choose.

If a transaction is submitted to `soroban-cli serve` via a HTTP call then it's events should be accessible via the HTTP interface, because the events were generated within that runtime.

But if a contract is invoked via `soroban-cli invoke` the only place those events should go are to the command line (or to file) for the use of the invoker, not for the use of the HTTP service, because they are both different runtimes and events are ephemeral.

cc @paulbellamy

src/invoke.rs
```diff
@@ -250,11 +268,27 @@ impl Cmd {
...
+                    })?;
+                out = Box::new(file);
+            }
+
+            for event in events.0 {
+                let event_str = match event {
+                    HostEvent::Contract(e) => serde_json::to_string(&e).unwrap(),
+                    HostEvent::Debug(e) => format!("{}", e),
+                };
+                writeln!(out, "{}", event_str).map_err(|e| Error::CannotWriteEvent { error: e })?;
```
leighmcculloch, line 291: This looks like it will produce a file that contains invalid JSON, because debug events are not being outputted as JSON, so I don't think this file will be very usable.

It's quite unclear what the use of writing the events to file are to me, and I think we could leave it out of the CLI.

2opremio: I wrote the intention in the PR's description.

tomerweller:
@2opremio if a developer doesn't supply an `--events-file` then events won't get plumbed to the `serve` http server?

leighmcculloch:
@paulbellamy @2opremio The intent with the invoke command was that it could also invoke contracts on remote hosts by building a tx and submitting it to it. Originally that remote host was just going to be a Horizon. But if you'd like the invoke command to invoke a transaction in an RPC server I think that makes sense too. That's how I think it would make the most sense to integrate invoke with a running RPC server (or the serve command which is just an RPC server emulated). In this situation the runtime would be inside the serve, so the events would be generated inside the serve.

Does that achieve what you're wanting to do?

paulbellamy:
> breaks the "these are separate runtimes" model.

I think that's the fundamental disagreement here. I think they should be the same "runtime". If you have an rpc server running against a network, and submit a new txn there via "invoke". You'd expect it to be the same runtime. I don't see why the sandbox should be different. `ledger.json` is equivalent to a "network".

src/invoke.rs
```diff
@@ -42,6 +47,12 @@ pub struct Cmd {
     /// File to persist ledger state
     #[clap(long, parse(from_os_str), default_value(".soroban/ledger.json"))]
     ledger_file: std::path::PathBuf,
+    /// Output file to write events
+    #[clap(long, parse(from_os_str))]
```
paulbellamy, line 51:
```suggestion
    #[clap(long, parse(from_os_str), default_value(".soroban/events.json"))]
```
wdyt?

leighmcculloch:
Got it, interesting. 

I think a developer will see the events file like a log file. For example, debug events (called logs in the SDK) are logged as a non-roundtrippable string because the intent is for the developer to write this code:
```
log!("uh oh: {}, value: {}", symbol!("hi"), 2);
```
and to see a human readable formatted string, like this:
```
uh oh: Symbol(hi), value: I32(2)
```

To make the output roundtrippable we'd have to output the debug event as a JSON object which would make logs not human readable as intended.

I think this means at the very least we need to separate debug events from contract and system events when configuring output destinations, so that we can continue to output human readable logs without needing them to be roundtrippable. 

The second thing is that when I look at the CLI interface and see an events file it isn't intuitive to me that it is a communication mechanism. Even the ledger.json file is dubious to describe as a communication mechanism. From the perspective of the invoke I would describe it as just an input and output. Having two applications read and write from the same file also seems like it will have it's challenges with corruption, etc. 

I think it is very reasonable to expect that the invoke command can affect the serve runtime and to generate events there, but I think the way it should do that is by submitting a transaction to the serve http endpoint, and then streaming back the results and events from the RPC server in the same way it will do so when invoke a contract with a real remote RPC server.

We could maybe even make this seamless to the developer by writing the serve URL to a file in `.soroban/` and having invoke default to a http request to serve when it sees that file there.

paulbellamy:
re: having soroban-cli invoke do a jsonrpc request to serve if it is running. Seems... complicated to get reliable, and like it could break in strange ways. Doing an internal jsonrpc request seems like a hacky way to re-use code. But, if we need to in order to prevent data corruption, then fine. Maybe a better solution would be to share the implementation code, and have both processes use something like sqlite that would ensure consistency? Another dependency, though, which sucks.

As it is now, I don't expect things in the `.soroban` to be user-friendly/readable, so I'm not worried about the format of things there (though we could change that in future if we want).

leighmcculloch:
> having soroban-cli invoke do a jsonrpc request to serve if it is running. Seems... complicated to get reliable

This is the way that the dapp will communicate with the RPC server and the serve emulator, so we need a reliable http server in any case?

leighmcculloch:
> a pipe would avoid file consistency issues, at the cost of historical and user readability

Regarding pipes in general: Keep in mind that soroban-cli will be used on five targets, three operating systems across two architectures and pipes come with their own set of challenges such a deadlock, and may not behave the same on all operating systems.

Would the pipe be where the soroban-cli sends the JSON-RPC request? That sounds fine if you prefer that, although I think http will present less issues, and I think this creates additional work unnecessarily. 

If the pipe will be a side channel for events only, this seems like a not ideal architecture, and again creates additional work unnecessarily.

The serve already exposes a transaction endpoint, and we're already adding logic to invoke to send a transaction to FutureNet in #42, so in minimal case where a developer points invoke at their local serve it's a zero additional work on top of other work we already have planned.

src/invoke.rs
2opremio: I should transform the debug events to JSON though

2opremio:
> If the pipe will be a side channel for events only, this seems like a not ideal architecture, and again creates additional work unnecessarily.

That was the idea, using the pipe as an events one-way communication channel from `soroban-cli invoke` to  `soroban-cli serve`

If you feel strongly about using an http endpoint instead, that's fine, but please let's make a choice and move forward. We have already spent too much time on this.


@leighmcculloch as I understand it, what you mean is

1. The user starts `soroban-cli serve` (listening on, say `localhost:8000`) , which incorporates a new endpoint (say `eventsSend`) which can be used to send events
2. The user subscribes to events subscribes to events by calling the `eventsSubscribe` endpoint on the running on `soroban-cli serve` from (1)
3. The user runs `soroban-cli invoke --events-server localhost:8000 ...`.
4. The user recieves events notifications from `soroban-cli serve` (which have been sent from `soroban-cli invoke` using the new `eventsSend` endpoint).


Please confirm.

Having to deal with `--events-server` reduces usability (the pipe location can be implicit) and adding extra http machinery for this seems like a waste of time. However, we have spent way more time discussing this, so I am willing to cave in.

leighmcculloch:
@2opremio That's not what I'm suggesting.

I'm suggesting we build no side-channel for events. I'm suggesting we build what we plan to build already for invoking contracts on remote RPC servers, which serve is a local emulator for, i.e. https://github.com/stellar/soroban-cli/issues/42.

By invoking a contract on the remote RPC server (i.e. soroban-cli serve), events should naturally be generated within the soroban-cli serve runtime, and the invoke command should be able to subscribe to events like it would any remote RPC server or dapp.

To break it down:

- Use runs:
   ```
   soroban-cli invoke --rpc-url http://localhost:8000 --account <G...> --id <CONTRACTID> --fn <FN> --arg ...
   ```
- Contract is invoked within soroban-cli serve, in the same way the dapp will invoke a contract within serve.
- soroban-cli serve will produce events in the same way it will when the dapp invokes contracts on serve.
- soroban-cli invoke collects events from serve like a dapp would.

i.e. I'm suggesting we build nothing unique / bespoke for sandbox invoke with sandbox serve, and instead rely on features we already have to build for the dapp to serve integration.

2opremio:
I don’t fully understand what needs to be done from your description. Do you mean that #39 shouldn’t be implemented at all?

Please be clear on the mechanics of `—rpc-url`. Does this mean `serve` will also implement the sandbox emulation?

leighmcculloch:
> Does this mean serve will also implement the sandbox emulation?

`serve` is an implementation of the soroban-rpc. It exposes the same API and is backed by the ledger.json, or what we have been calling the sandbox.

Issue #142, opened 2022-09-17
CLI: Faucet sub-command


leighmcculloch:
Horizon and the soroban-rpc (and soroban-cli) all provide faucet endpoints.

- Horizon provides a friendbot backed endpoint which is pre-existing.

- Soroban-rpc (and soroban-cli) provide a `requestAirdrop` endpoint (https://github.com/stellar/soroban-cli/issues/100).

We should add a command that developers can use to create a new account on test networks and on soroban-cli serve.

The command should take the form:

```
soroban-cli account new [--id G...]
```

 - A new `account` subcommand would be added.
 - A new `new` subcommand of the account subcommand would be added.
 - An `--id` would be optional:
    - If provided the command attempts to create and fund that account.
    - If not provided we could randomly generate one.
 - The command would output:
    - Success / failure. (Exit code non-zero for failure.)
    - In the case that `--id` was not provided, it should also output the `S...` secret key that was generated.

A developer would need to use this command before doing things like using the `deploy` command (captured in #41). For deployed networks a user could use this command or use the Stellar laboratory. For local networks or for `soroban-cli serve` this command would be the primary way to get an account.

leighmcculloch:
@paulbellamy I think this story is required for Meridian and we don't currently have an issue focused on it. Wdyt? We'll need someone to pick this up.

paulbellamy:
(For meridian we used laboratory to fund the account on futurenet)

PR #144, opened 2022-09-20
Add sandbox command to contain deploy, invoke, read

leighmcculloch:
### What
Add sandbox command to contain the sandbox versions of deploy, invoke, read.

### Why
I took a quick look at how a remote deploy command would be added (#41) and it soon became overwhelming thinking about how to have such a command coexist with the code in the sandbox deploy command. 

At some point someone had expressed interest in grouping sandbox command under a sandbox super command so as to clearly separate them from commands that invoke remote services. I think it might have been @paulbellamy who had dropped this idea in a doc a long time ago.

We never did it, but I think it will make it easier to implement the following tasks if we have the sandbox variations under `sandbox` and put the remote variations of the commands under another subcommand such as `remote`:
- #41
- #42
- #142

This change makes no functional changes and is purely moving code into a directory (so it is in a Rust module under sandbox), and inserting the sandbox command in between them.

Close https://github.com/stellar/soroban-cli/issues/165

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,13 +1,11 @@
 use clap::{AppSettings, CommandFactory, FromArgMatches, Parser, Subcommand};
 
 mod completion;
-mod deploy;
 mod gen;
 mod inspect;
-mod invoke;
 mod jsonrpc;
 mod network;
-mod read;
+mod sandbox;
 mod serve;
 mod snapshot;
 mod strval;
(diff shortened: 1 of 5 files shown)


leighmcculloch:
I didn't include `serve` in this change because while I assume it's probably live under `sandbox`, I didn't want to move the file right now since other people might be working on it.

PR #167, opened 2022-09-30
Update deploy subcommand to use `create_contract_from_source()`

2opremio:
### What

Update deploy subcommand so that when it's used against an RPC server uses `create_contract_from_source()` instead of `create_contract_from_ed25519()`

Also, print the contract ID once created.

I used this CPP test as a reference: https://github.com/stellar/stellar-core/blob/2dccaae1ec9db8e82580bc5e7b84d404d1836f3c/src/transactions/test/InvokeHostFunctionTests.cpp#L135

### Why

#152 

### Known limitations

I haven't really been able to test this end-to-end against an RPC server because `soroban-cli serve` still only seems to support `HostFunction::InvokeContract`

### DIFF
--- a/src/deploy.rs
+++ b/src/deploy.rs
@@ -203,6 +203,9 @@ impl Cmd {
             self.network_passphrase.as_ref().unwrap(),
             &key,
         )?;
+
+        println!("Contract ID: {}", hex::encode(contract_id.0));
+
         client
             .request("sendTransaction", rpc_params![tx.to_xdr_base64()?])
             .await?;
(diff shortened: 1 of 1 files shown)


src/deploy.rs
```diff
@@ -214,7 +215,8 @@ impl Cmd {
                 .await?;
             match response.status.as_str() {
                 "success" => {
-                    println!("{}", response.status);
+                    println!("{}", response.status.as_str());
+                    println!("Contract ID: {}", hex::encode(contract_id.0));
```
leighmcculloch, line 219: Could we print the contract id out before the transaction is submitted?

leighmcculloch (approved):

PR #170, opened 2022-09-30
Extract soroban_rpc crate

leighmcculloch:
### What
Extract soroban_rpc crate

### Why
To tidy up the deploy code. This code was written by @paulbellamy and I'm just extracting it from his Pr in progress into a separate Pr so that we can land it sooner: https://github.com/stellar/soroban-cli/pull/166.

Thanks @paulbellamy!!!


src/rpc/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+
+// TODO: this should also be used by serve
+#[derive(serde::Deserialize, serde::Serialize, Debug)]
+pub struct TransactionStatusResponse {
+    pub id: String,
+    pub status: String,
+    // TODO: add results
+}
+
+pub struct Rpc {
```
paulbellamy, line 49: `Client`?

paulbellamy (approved):
Suggestion, but looks fine for now.

src/rpc/mod.rs
leighmcculloch: 1c4b21a

PR #178, opened 2022-10-04
Update the sdk to 864a309b

paulbellamy:
### What

Update the sdk to 864a309b.

### Why

So we can build/deploy/invoke a working contract without environment version errors in https://github.com/stellar/soroban-example-dapp/pull/20

### Known limitations

[TODO or N/A]


leighmcculloch (commented):
🎉 Thanks!

PR #185, opened 2022-10-05
Fix binary name in instructions in README

willemneal:
### What

Change example binary name

### Why

In the Cargo.toml the binary is given the shorter name so when installed is the one added to your path.

### Known limitations

N/A


leighmcculloch (approved):
Thank you! I missed updating the readme when I changed the binary in the toml.

PR #205, opened 2022-10-07
invoke: Fix (and enforce) getContractData invocation

2opremio:
### What

Now that the Go RPC Server supports `getContractData` ( https://github.com/stellar/go/pull/4652 ), we can enforce it when running invoke.

This means that `--wasm` won't be allowed when running `soroban invoke --rpc-url foo`

Also, it fixes the expected contract code type.

### Why

Allowing `soroban invoke --rpc-url foo --wasm bar` was just a workaround 

### Known limitations

N/A

### DIFF
--- a/src/invoke.rs
+++ b/src/invoke.rs
@@ -5,8 +5,8 @@ use clap::Parser;
 use hex::FromHexError;
 use soroban_env_host::xdr::{
     InvokeHostFunctionOp, LedgerFootprint, Memo, MuxedAccount, Operation, OperationBody,
-    Preconditions, ScStatic, ScVec, SequenceNumber, Transaction, TransactionEnvelope,
-    TransactionExt, VecM,
+    Preconditions, ScContractCode, ScStatic, ScVec, SequenceNumber, Transaction,
+    TransactionEnvelope, TransactionExt, VecM,
 };
 use soroban_env_host::{
     budget::{Budget, CostType},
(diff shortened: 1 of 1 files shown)


src/invoke.rs
```diff
@@ -275,29 +272,19 @@ impl Cmd {
...
-            let contract_data = client
-                .get_contract_data(
-                    &hex::encode(contract_id),
-                    ScVal::Static(ScStatic::LedgerKeyContractCode),
-                )
-                .await?;
-
-            match ScVal::from_xdr_base64(contract_data.xdr)? {
-                ScVal::Object(Some(ScObject::Bytes(bytes))) => bytes.to_vec(),
-                scval => return Err(Error::UnexpectedContractCodeDataType(scval)),
```
leighmcculloch, line 299: Can you make --wasm optional rather than a conflict to provide? There are edge cases where we might actually want to provide the wasm on the command line rather than rely on it being downloaded. For e.g. the token contract. It will also mean this change won't be a breaking change in the final hour.

2opremio: Can you elaborate on the edge case and why downloading the contract is not an option?

2opremio: Regarding breaking changes, we see not at a point of considering backwards compatibility when invocation support was added 3 days ago.

2opremio: Almost every PR has been a breaking change in the past few days (starting with multiple unconsensuated flag renamings)

2opremio:
Let's get this merged and unblock get Contract Data. I will happily reintroduce the wasm flag if need be.

src/invoke.rs
leighmcculloch: In the last day we started trying to avoid breaking changes that would affect the crowdfund example dapp's scripts and the e2e scripts. That's what I was referring to.

leighmcculloch:
> Let's get this merged and unblock get Contract Data so that it can be included in testing ASAP.

I don't this change is necessary. The code prior to this change already supports getting the wasm from contract data when the `--wasm` flag isn't passed. All this change did was remove the ability to provide the wasm if the developer wanted. I'll revert it.

PR #211, opened 2022-10-09
Change cli serve defaults and path to match quickstart docker

paulbellamy:
### What

Change `serve` subcommand to default to `localhost:8000/soroban/rpc`.

### Why

This is where quickstart serves things by default. Having them be the same makes switching back and forth simpler, as you don't need to change the URL, just the network passphrase.

### Known limitations

[TODO or N/A]


leighmcculloch (approved):

leighmcculloch:
@paulbellamy let me know if you want me to ship new versions of the CLI.

PR #214, opened 2022-10-10
Embed the token spec

leighmcculloch:
### What
Try embedding the token spec

### Why
Do you don't need to provide the .wasm file for the token contract.

Close https://github.com/stellar/soroban-cli/issues/172

### Status

- [x] Test that this works with sandbox
- [x] Test that this works with the rpc server


leighmcculloch:
The testing approach I'm using is this:
```
$ soroban token create --name tok --symbol tok
d55b5a3a5793539545f957f7da783f7b19159369ccdb19c53dbd117ebfc08842
```
```
$ soroban invoke --id d55b5a3a5793539545f957f7da783f7b19159369ccdb19c53dbd117ebfc08842 \
--fn decimals
7
```

leighmcculloch:
@paulbellamy @2opremio I'm unable to get token create to work with `soroban-rpc` or `soroban serve`. I might just be making a silly mistake, but I'm not comfortable merging this unless we feel confident it works with all the cases you're using. Are you able to verify if this change is good with those, or share how you run them?

2opremio:
I’ll try it tomorrow, but I wouldn’t merge it until the token spec crate exports the full spec (instead of reconstructing it here).

That is, unless Paul plans to use this for the demo on Wednesday.

leighmcculloch:
Tested successfully with soroban-rpc:

```
$ soroban token create --name tok --symbol tok \
    --rpc-url http://127.0.0.1:8000/soroban/rpc \
    --network-passphrase 'Standalone Network ; February 2017' \
    --secret-key S...
success
success
6d6e5b2963d3c53557c66fd4afcd323e67d94fd419ee3368995f8eb470bf6db2
```
```
$ soroban invoke --id 6d6e5b2963d3c53557c66fd4afcd323e67d94fd419ee3368995f8eb470bf6db2 \
    --fn balance \
    --arg '{"object":{"vec":[{"symbol":[67,111,110,116,114,97,99,116]},{"object":{"bytes":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}}]}}' \
    --rpc-url http://127.0.0.1:8000/soroban/rpc \
    --network-passphrase 'Standalone Network ; February 2017' \
    --secret-key S...
"0"
```

leighmcculloch:
I checked with @paulbellamy and he doesn't need this soon, so I don't plan to merge this until later this week.

2opremio:
I've created https://github.com/stellar/rs-soroban-sdk/pull/724 to expose the full spec directly from `soroban_token_spec`

leighmcculloch:
There's a risk that the token spec crate is dependent on a different env meta version, or different env in general, than the one imported into the CLI. I'd like to see if there's a way to ensure they don't differ.

leighmcculloch:
@paulbellamy @2opremio In terms of timing I think we can merge this change now. The PR is updated to use the function introduced in https://github.com/stellar/rs-soroban-sdk/pull/724. Please take a look.

leighmcculloch:
Putting this back into draft. I started adding a test for this change, and discovered that the change was broken when I updated to using the SDK function for the contract spec. Investigating.

leighmcculloch:
I added back using the local spec xdr since that works, and I'd like to merge this because it's blocking another change, and then follow up with a fix to upstream and then circle back on updating this to use it.

leighmcculloch:
Fix is in https://github.com/stellar/rs-soroban-sdk/pull/732, so once that is merged I can update this PR again to use it.

2opremio (approved):

PR #215, opened 2022-10-10
Fix the url the invoke client connects to for the rpc

leighmcculloch:
### What
Change the URL the invoke client connects to, to not append the rpc path and rely on the user providing the full path at the command line. Also prints the served path to the command line so people know what it is.

### Why
https://github.com/stellar/soroban-cli/pull/211 changed the path that get served to be /soroban/rpc instead of /api/v1/jsonrpc, to match soroban-rpc.

However, the soroban invoke command still sends requests to /api/v1/jsonrpc. This doesn’t seem to matter for the real soroban-rpc server, but the invoke command doesn’t work with the local serve anymore.

### DIFF
--- a/src/rpc/mod.rs
+++ b/src/rpc/mod.rs
@@ -88,7 +88,7 @@ impl Client {
     }
 
     fn client(&self) -> Result<HttpClient, Error> {
-        let url = self.base_url.clone() + "/api/v1/jsonrpc";
+        let url = self.base_url.clone();
         let mut headers = HeaderMap::new();
         headers.insert("X-Client-Name", "soroban-cli".parse().unwrap());
         let version = VERSION.unwrap_or("devel");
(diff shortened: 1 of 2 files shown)


sreuland:
interesting, the routing of `/soroban/rpc` url in quickstart is only done at the nginx layer with 'starts with' matching rule equivalent to any request starting with '/soroban/rpc' triggers a match and is forwarded to back-end soroban-rpc endpoint which that service ignores url request paths, it doesn't parse context from url, it just parses the POST body. so `http://localhost:8000/soroban/rpc` and `http://localhost:8000/soroban/rpcxyzlskdjfas` are both routed to the endpoint which is why `api/v1/jsonrpc` at end of url didn't cause an issue, not sure that was intentional, but is in place for now.

it looks like url route matching is more strict in 'serve', and the client change adheres to it and gets intended interop from https://github.com/stellar/soroban-cli/pull/211 between either service. As a follow-on, we should probably keep both services in sync on rpc url behavior, so, we either change quickstart to direct match or vice versa, 'serve' to use 'starts with'.

leighmcculloch:
@sreuland Ah very interesting. Thanks for identifying where the difference in behavior is from.

I think this means this change safe to make.

@paulbellamy Do you agree? Okay with me merging this?

PR #219, opened 2022-10-18
Create source account

leighmcculloch:
### What
Create source account when invoking with an account that does not exist.

### Why
For consistency with SDK tests that do the same.

The source account used in an invoke can only ever be created. There is no valid situation where an invoke can occur with an account that does not exist. For this reason there is little value in allowing such a situation to be executed. Requiring a user to always create the account first would seem tedious when the main goal of invoke is to just run the contract. We should therefore make sure that when the contract runs that all the parameters of the execution environment model a real execution. One invariant is that the invoker exists. Therefore this change makes it so that the account always exists.


leighmcculloch:
We need to iterate on this a little, but I think this is sufficient to start adding more .wasm driven tests. At some point we're going to hit a cliff where this becomes too slow, but I'm not sure where that is yet. When we reach that point, we may want to compile all the test wasms to files and commit the .wasm files. The env repo does this, although for different reasons that are not relevant to this repo.

.github/workflows/rust.yml
```diff
@@ -47,7 +47,8 @@ jobs:
     - run: rustup update
     - run: rustup target add ${{ matrix.target }}
     - run: cargo clippy --all-targets --target ${{ matrix.target }}
-    - run: cargo test --target ${{ matrix.target }}
+    - run: make build-test-wasms
```
leighmcculloch, line 50: This line builds the .wasm files ahead of running the tests.

.github/workflows/rust.yml
```diff
@@ -47,7 +47,8 @@ jobs:
...
+    - run: cargo test --workspace --target ${{ matrix.target }}
```
leighmcculloch, line 51: The change to add `--workspace` causes the tests for all crates in the workspace to be run instead of only the tests for the default members.

tests/invoke_sandbox.rs
```diff
@@ -26,3 +26,20 @@ fn invoke_token() -> Result<(), Box<dyn std::error::Error>> {
...
+fn source_account_exists() -> Result<(), Box<dyn std::error::Error>> {
+    let tmp = TempDir::new().unwrap();
+    let ledger = tmp.child("ledger.json");
+
+    let mut cmd = Command::cargo_bin("soroban")?;
+    cmd.arg("invoke");
+    cmd.arg("--ledger-file").arg(ledger.as_os_str());
+    cmd.arg("--id").arg("1");
+    cmd.arg("--wasm")
+        .arg("target/wasm32-unknown-unknown/test-wasms/test_invoker_account_exists.wasm");
```
2opremio, line 40: Optional: Since the test doesn't explicitly enforce building the test contracts, it may be a good idea to give a firendly/specific error if the file does not exist.

leighmcculloch: Good idea. I added more descriptive panic message in f4b29b2. Note that this is still a large wall of text, because any test failure usually results in a pretty big wall of text, such is the Rust way as far as I can tell.

tests/invoke_sandbox.rs
```diff
@@ -26,3 +26,25 @@ fn invoke_token() -> Result<(), Box<dyn std::error::Error>> {
...
+    let tmp = TempDir::new().unwrap();
+    let ledger = tmp.child("ledger.json");
+
+    let mut cmd = Command::cargo_bin("soroban")?;
+    cmd.arg("invoke");
+    cmd.arg("--ledger-file").arg(ledger.as_os_str());
+    cmd.arg("--id").arg("1");
+    cmd.arg("--wasm").arg(WASM);
+    cmd.arg("--fn").arg("invkexists");
+    cmd.assert().success().stdout("true\n");
```
sreuland, line 47: how did this get true, didn't see the test create source account first, is this b/c sandbox has `GAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAWHF` and cli hardcodes that as default `--account` , if so, may be worthwhile to include pre-req/assumptions in test comments.

leighmcculloch: Have a look in the `src/invoke.rs` file in this PR. This PR is introducing the automatic creation of the source account, whatever it is.

test-wasms/invoker_account_exists/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use soroban_sdk::{testutils::Accounts, Env};
+
+    use crate::{Contract, ContractClient};
+
+    #[test]
+    fn test_invoker() {
```
sreuland, line 30: does this test execute the [updates on invoke path](https://github.com/stellar/soroban-cli/pull/219/files#diff-7933b1f789a38779ee96fb34ae9b41170279f5a2bc35af08fb399fe0e93af89cR380), I'm still learning on rust..

can the test exercise both outcomes, i.e. verify source account doesn't exist and if worthwhile, test the baseline, source account did exist also?

tests/invoke_sandbox.rs
sreuland: doh! ok, got it, `invoke` has different implementations per the target(rpc vs sandbox), this account creation behavior is only done within path for sandbox target.

Issue #197, opened 2022-10-07
CLI: Redesign


leighmcculloch:
There's a few issues with the CLI's top-level surface area and how the commands are organized. We have got issues open for those specific things (#165 #173), but I think we need to take a step back and rethink the design of the CLI.

This is a tracking issue, for tracking the problem and coordinating, but the design itself is in the below Google Document. The document is publicly accessible and can be commented on by anyone.

https://docs.google.com/document/d/1zV_F3WMPbxOuXxAknGFocXLOB5X_LcHBF4LG-AF8BUo

willemneal:
Newest proposal is here: https://docs.google.com/document/d/16lJhyMU6cqRzNtMclr9yXd978MJgbZ0El9UhxbTz_Ts/edit?usp=sharing

leighmcculloch:
Closing this issue as we're moving ahead with the proposal, and issues can be opened for specific areas for implementation.

PR #220, opened 2022-10-18
Add first integration test

2opremio:
### What

It tests deploying and invoking a contract against the RPC server

### Why

There are no integration tests. Also, see stellar/quickstart#72 

### Known limitations

The way in which the quickstart container is spawned could be nicer (it may be a good idea to do it from the tests themselves).


tests/integration_test.rs
```diff
@@ -0,0 +1,55 @@
+#[cfg(test)]
+use assert_cmd::Command;
+
+#[test]
+fn deploy_and_invoke_contract_against_rpc_server() {
+    // This test assumes a fresh standalone network rpc server on port 8000
+
+    let mut cmd = Command::cargo_bin("soroban").unwrap();
+    // TODO: How to reuse the env variables between deploy and invoke
```
2opremio, line 9: @leighmcculloch I would love your input on this. I tried creating a factory function building a base command but it didn't work since I cannot derreference the command and return it.

tests/integration_rpc_server.rs
```diff
@@ -0,0 +1,55 @@
...
+            "SOROBAN_SECRET_KEY",
+            "SC5O7VZUXDJ6JBDSZ74DSERXL7W3Y5LTOAMRF7RQRL3TAGAPS7LUVG3L",
+        )
+        .env(
+            "SOROBAN_NETWORK_PASSPHRASE",
+            "Standalone Network ; February 2017",
+        )
+        .args(&[
+            "deploy",
+            "--wasm=tests/fixtures/soroban_hello_world_contract.wasm",
```
leighmcculloch, line 22: It looks like this test is dependent on a `soroban-examples` .wasm file. A while ago we removed dependencies on .wasm files in other repos because we found too often they would change and break and the approach was too brittle. Instead we've moved to generating .wasm files in the repos that require them. Take a look at the rs-soroban-env repo which has an implementation of this: https://github.com/stellar/rs-soroban-env/tree/main/soroban-test-wasms. The idea is any test vectors the CLI needs should be generated by code in the same repo as the test.

tests/integration_rpc_server.rs
```diff
@@ -0,0 +1,55 @@
+#[cfg(test)]
+use assert_cmd::Command;
+
+#[test]
+fn deploy_and_invoke_contract_against_rpc_server() {
+    // This test assumes a fresh standalone network rpc server on port 8000
```
leighmcculloch, line 6: We need a way to run tests without running tests that require a running a stellar-core.

I see a couple ways we could do that:
- Have an env variable that provides the base URL of the RPC server. If it is set, use that. If it is not set, use `Command::cargo_bin` to run `soroban serve`, and then point to it. This has some huge advantages:
   - We can run tests twice, once with the serve, once with the real rpc server. Two tests for the price of one.
   - Random dev can run the tests without the heavy rpc server / stellar-core. We maintain agility.
   - People running the rpc server locally on a different port can do so.

- Add a cfg ignore to the tests `#[cfg_attr(not(feature = "tests_e2e_rpc_server"), ignore)]`. This has the advantages that the tests are always compiled, and when running tests a line will be outputted saying a test is being skipped. This is similar to "Skip" in Go tests.

- Add a cfg to the top of the file `#![cfg(feature = "tests_e2e_rpc_server")]`. This more convenient, but the tests won't compile at all unless the feature is enabled.

I prefer the first option because it involves us always running the tests in some form. Wdyt?

.github/workflows/rust.yml
```diff
@@ -47,7 +47,13 @@ jobs:
     - run: rustup update
     - run: rustup target add ${{ matrix.target }}
     - run: cargo clippy --all-targets --target ${{ matrix.target }}
-    - run: cargo test --target ${{ matrix.target }}
+    # unit tests
+    - run: cargo test --bins --target ${{ matrix.target }}
+    # integration tests
+    # TODO: we should probably manage the container in the test itself
+    # TODO: do a proper functional check on the container startup instead of waiting 15 seconds
+    - run: docker run -d -p 8000:8000 stellar/quickstart:soroban-dev --standalone --enable-soroban-rpc --protocol-version 20 && sleep 15
```
leighmcculloch, line 55: While this will work today, adding a dependency on stellar-core to the CLI tests will probably prevent many PRs from being merged on this repo. That's because stellar-core is rarely updated and released with the new version of the env interface, but the env interface is under active development, so any time you update the env without also getting the core team to build a new deb, the tests dependent on the quickstart will break.

I imagine we need to iterate on this test setup, so could we move the quickstart tests into a separate job that won't block PRs from merging?

.github/workflows/rust.yml
```diff
@@ -47,7 +47,13 @@ jobs:
     - run: rustup update
     - run: rustup target add ${{ matrix.target }}
     - run: cargo clippy --all-targets --target ${{ matrix.target }}
-    - run: cargo test --target ${{ matrix.target }}
+    # unit tests
+    - run: cargo test --bins --target ${{ matrix.target }}
+    # integration tests
+    # TODO: we should probably manage the container in the test itself
+    # TODO: do a proper functional check on the container startup instead of waiting 15 seconds
```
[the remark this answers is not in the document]
leighmcculloch: You can probably get something equivalent with a little bash. This is what we use in other projects:
https://github.com/stellar/starlight/blob/919096552e44eec5ec3fd6278922396926239874/.github/workflows/sdk.yml#L59-L62

.github/workflows/rust.yml
leighmcculloch, line 55: If we're going to use the quickstart image in integration/e2e tests, we should enable artificial acceleration mode. You can enable it by appending this option: `--enable-core-artificially-accelerate-time-for-testing`.

(Artificial acceleration made a big difference to the tests of the Starlight project where we do the same thing: https://github.com/stellar/starlight/blob/main/.github/workflows/sdk.yml#L59-L62.)

leighmcculloch (commented):
Thanks for getting a start on this, and for using the assert_cmd/fs packages to test from the outside in 👏🏻. We already have one integration test

A few concerns and suggestions inline. A while ago we moved away from embedding .wasm files from other repos. See inline for more details on why. The dependence on stellar-core is also going to block PRs. I left some suggestions inline, but @2opremio @sreuland I'm available to help with planning this out in a synchronous discussion in Discord or in a call if that's helpful.

Also fyi all tests in the `tests/` repo are integration tests – that's where integration tests go in Rust projects – so I don't think we need to rename `invoke_sandbox.rs` to `integration_sandbox.rs`, and we could drop the prefix on the rpc server file too.

tests/integration_rpc_server.rs
leighmcculloch: @2opremio I'm actually in process of adding a .wasm fixture (via a crate) for #219. Would you like me to take the lead on the foundation for that? or do you want to do it? If you want to do it, I'll hold off on #219 until you're done.

2opremio:
> Also fyi all tests in the `tests/` repo are integration tests – that's where integration tests go in Rust projects – so I don't think we need to rename `invoke_sandbox.rs` to `integration_sandbox.rs`, and we could drop the prefix on the rpc server file too.

I am using the prefix to separate them when invoking them from the github actions (see `cargo test --test 'integration_*' --target ${{ matrix.target }}`). If you know of a better way to run only the integration tests (i.e. tests in the `tests` directory) I can remove the prefix.

tests/integration_rpc_server.rs
2opremio: I agree it's better to track the source code than the wasm. But ... can't the contract source code also get outdated? Plus ... doesn't building the contract complicate the testing process too?

I can wait for #219 and see how it's done

2opremio:
I want to run them in every PR, but run them separately, so that we can clearly see whether it's integration tests that are failing.

.github/workflows/rust.yml
2opremio: We've been running integration tests on Horizon (which depend on Core) for more than a year without many problems. They break once in a while but it hasn't been a problem IMO.

Regarding the `env` changes, bear in mind this is a very basic test. Do you expect the env to break deploying and invoking a basic contract?

tests/integration_rpc_server.rs
leighmcculloch: @2opremio I went ahead and used this pattern in the other PR. I'm still tidying it up, but that's the idea, take a look, wdyt?: https://github.com/stellar/soroban-cli/pull/219.

tests/integration_rpc_server.rs
2opremio: > We need a way to run tests without running tests that require a running a stellar-core

For this particular test, I disagree. The intention is to have an end to end integration test that checks everything works.

.github/workflows/rust.yml
2opremio: I will add it

tests/integration_rpc_server.rs
2opremio: > * Have an env variable that provides the base URL of the RPC server. If it is set, use that. If it is not set, use `Command::cargo_bin` to run `soroban serve`, and then point to it.

This is a good idea but it won't work today. Also, it's not about testing one thing or the other. We should test both against a real RPC Server and `soroban serve` (which unfortunately doesn't work today AFAICR since it doesn't accept contract creation transactions).

tests/integration_rpc_server.rs
2opremio: I took a look, it seems simple enough. I am happy to migrate the test to that later on. Let's get this merged first if you don't mind.

2opremio:
> add a env var to point it to an external one if needed

We can already do that, it's just a matter of starting `soroban serve` or the full RPC Server first

tests/integration_rpc_server.rs
sreuland: It feels like we are conflating integration and unit test concerns in a single repo , ideally should be able to trigger the same coverage over functional code from either test scope, unit tests w/mocking or integration. But integration tests can happen less often(nightly, ad-hoc, PR merge) and from a separate dedicated 'qa' repo, where they spin up all the live architecture rather than duplicating that boilerplate code in each functional component repo like cli, env. 

as part of soroban manual e2e testing, had spun out an issue to think about integration tests from an external repo, initially created [#375](https://github.com/stellar/quickstart/issues/375) in quickstart, but it really could be a separate integration test repo and keep out of quickstart which is meant for other purposes.

tests/integration_rpc_server.rs
leighmcculloch: @sreuland probably, but given the largest risk with a change to the CLI is not changing the internal behavior, but the external visible behavior, I think the leaning into tests that exercise the CLI from the outside-in will have a larger impact on us providing a stable UX.

tests/integration_rpc_server.rs
leighmcculloch: @sreuland I agree, but I think there's more than one layer of integration testing for us to consider so I think we should do that, plus some other things locally in this repo. For example:

1. Integration testing all the way through to the soroban-rpc server, that includes a stellar-core captive-core, maybe we can call that end-to-end testing:

   @sreuland If you're saying this type of test should happen in a separate repo, I agree. I think it's going to slow down development in this repo for the reasons described in [comment](https://github.com/stellar/soroban-cli/pull/220#discussion_r998562880). Especially as long as stellar-core is a component these tests are dependent on. 

   @tsachiherman @graydon @ire-and-curses See @sreuland's comment above. We should discuss effort to setup a QA style end-to-end test repo that runs all components of the entire Soroban experience with a test of test vectors as a top-down view of compatibility.

2. Integration testing from the sense of outside-in testing the CLI using real wasm files:

   I think we can do that in this repo. The performance cost is much smaller. Incompatibilities are much easier and faster to resolve. It would be faster to iterate on code protected by these types of tests than the tests in (1). We've found this model work well for the env repo, although I acknowledge it is not perfect.

3. Integration testing from the sense of outside-in with mocked inputs:

   I think we can do that in this repo, although I think (2) may be easier for us to implement in the short term because mocking will make it harder to change the overall design, and with #197 there might be a lot of changes.

---

@2opremio I think it's fine if you want to continue with exploring an integration test in this repo that uses quickstart in the interim until the above idea is worked out, but I think it is important that:
- Any test dependent on a running rpc server should be skipped by default. Typically skipping tests are done via a feature as I describe here https://github.com/stellar/soroban-cli/pull/220#discussion_r998560044.
- That the existing integration tests in the tests/ repo aren't skipped as part of ☝🏻.

tests/integration_rpc_server.rs
sreuland: agreed, the coverage is needed and executed as often as called for, but does that mean the external coverage and it's complexity has to be implemented in this repo? I [commented about de-coupling integration test and unit test concerns from repos a bit more](https://github.com/stellar/soroban-cli/pull/220#discussion_r998729023), if that aligns, maybe external testing can be implemented as e2e cases once in a dedicated 'qa' repo instead of variations of its complexity inside each repo. Still retain the ability to execute the tests as often as needed from 'qa' repo per different environments like GHA workflow(CI) or local per configs.

tests/integration_rpc_server.rs
leighmcculloch: Regarding @sreuland's idea for a test repo, that's captured in https://github.com/stellar/jump-cannon-deliverables/issues/25 that he just opened recently.

tests/integration_test.rs
leighmcculloch: I think this should work:
```rust
fn cmd() -> Command {
    let mut cmd = Command::cargo_bin("soroban").unwrap();
    cmd.env("SOROBAN_RPC_URL", "http://localhost:8000/soroban/rpc")
        .env(
            "SOROBAN_SECRET_KEY",
            "SC5O7VZUXDJ6JBDSZ74DSERXL7W3Y5LTOAMRF7RQRL3TAGAPS7LUVG3L",
        )
        .env(
            "SOROBAN_NETWORK_PASSPHRASE",
            "Standalone Network ; February 2017",
        );
    cmd
}

#[test]
fn deploy_and_invoke_contract_against_rpc_server() {
    let mut deploy = cmd();
    let deploy = deploy
        .args(&[
            "deploy",
            "--wasm=tests/fixtures/soroban_hello_world_contract.wasm",
            "--salt=0",
        ]);

   // ...
```

Another option is to provide the list of envs:
```rust
const COMMAND_ENVS: &[(&str, &str)] = &[
    ("SOROBAN_RPC_URL", "http://localhost:8000/soroban/rpc"),
    (
        "SOROBAN_SECRET_KEY",
        "SC5O7VZUXDJ6JBDSZ74DSERXL7W3Y5LTOAMRF7RQRL3TAGAPS7LUVG3L",
    ),
    (
        "SOROBAN_NETWORK_PASSPHRASE",
        "Standalone Network ; February 2017",
    ),
];

#[test]
fn deploy_and_invoke_contract_against_rpc_server() {
    let mut cmd = Command::cargo_bin("soroban").unwrap();
    let deploy = cmd
        .envs(COMMAND_ENVS.iter().copied())
        .args(&[
            "deploy",
            "--wasm=tests/fixtures/soroban_hello_world_contract.wasm",
            "--salt=0",
        ]);

   // ...
```

A factory is simpler imo, but just mentioning the alternative in case it is preferred.

2opremio: Ahh, simple enough, thanks! I didn't manage to return a `Command` 🤷

2opremio:
I gave a try to using feature flag for e2e tests.

However, it causes everything to rebuild, making everything really slow. If I later run unit tests, then everything is rebuilt again.


It seems like feature flags completely reset the build cache.

leighmcculloch:
TIL you can unignore ignored tests, that's a very nice way of doing it without requiring a feature. 👏🏻 I don't think I've seen anyone do that before, but the fact that it removes the need for the feature but provides the same capabilities is awesome.

(I do not see rebuilds when I run with and without the feature at commit d56e59a. The rebuilds could be to do with a few other unrelated things. If you're using rust-analyzer with VSCode, it can sometimes be a good idea to tell rust-analyzer to use a different `target/` directory, otherwise the two will silently invalidate each others caches. Rust-analyzer may do this without showing you anything visible to indicate it is doing that. I'll drop some suggestions for dev env setup in the #rust Slack channel shortly.)

.github/workflows/rust.yml
```diff
@@ -49,6 +49,23 @@ jobs:
     - run: cargo clippy --all-targets --target ${{ matrix.target }}
     - run: cargo test --target ${{ matrix.target }}
 
+  e2e-test:
```
leighmcculloch, line 52: FYI this test won't be required because it isn't included in the `complete` job's required jobs, which I think is good for now because this test will probably be brittle.

tests/integration_rpc_server.rs
leighmcculloch: The follow up to this is in https://github.com/stellar/soroban-cli/pull/222.

PR #223, opened 2022-10-19
Check container is ready in e2e-test ci job

leighmcculloch:
### What
Check container is ready in e2e-test ci job by checking if the history ledgers is greater than 0.

### Why
I noticed a TODO in here when working in the same file. We did this in the starlight project by adding a simple curl and check on the number of history ledgers ingested. Not sure if that is sufficient here, but I think it might be sufficient.


.github/workflows/rust.yml
```diff
@@ -63,8 +63,8 @@ jobs:
       - run: rustup update
       - run: rustup target add ${{ matrix.target }}
       # TODO: we should probably manage the container in the test itself
-      # TODO: do a proper functional check on the container startup instead of waiting 15 seconds
       - run: docker run -d -p 8000:8000 stellar/quickstart:soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing --protocol-version 20 && sleep 15
```
2opremio, line 66:
```suggestion
      - run: docker run -d -p 8000:8000 stellar/quickstart:soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing --protocol-version 20
```

.github/workflows/rust.yml
```diff
@@ -63,8 +63,8 @@ jobs:
...
+      - run: while ! [ "$(curl -s --fail localhost:8000 | jq '.history_latest_ledger')" -gt 0 ]; do echo waiting; sleep 1; done
```
2opremio, line 67: you may want to merge this with the prior command

leighmcculloch: I'd rather keep it separate so that it is clear in the GitHub Action when the container has started vs the build is waiting for the container to start. That seemed to work well here: 
https://github.com/stellar/starlight/blob/919096552e44eec5ec3fd6278922396926239874/.github/workflows/sdk.yml#L59-L62

2opremio: No strong opinions, I think it's easier to read if done in the same step (spawn the docker container and make sure it's working).

2opremio (approved):

PR #224, opened 2022-10-19
Move the git revision logic to a crate

leighmcculloch:
### What
Move the git revision logic to a crate:
https://crates.io/crates/crate-git-revision

### Why
So that the logic can be applied to other repositories.


graydon (approved):
LGTM!

PR #229, opened 2022-10-21
Publish binaries

leighmcculloch:
### What

Publish binaries.

### Why

For installs without building from source.

This doesn't have a huge use, given that most people can install soroban-cli using cargo, and most people who need to install it will probably already have cargo installed because it's required for developing contracts.

But this has use in some places where it is difficult to install from source. For example, when installing soroban-cli into a repl.it, the build consumes so much disk space that it often fails to install.

### Known limitations

There's plenty of things we can improve about this. For example, we should publish binaries for Windows and Linux-aarch64. Also, we will probably subsume this into the shared publish.yml workflow that lives in the stellar/actions repo, but for now this will do.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,3 +10,53 @@ jobs:
     uses: stellar/actions/.github/workflows/rust-publish.yml@main
     secrets:
       CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  upload:
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          # TODO: Re-enable when soroban-cli can compile for Linux aarch64.
+          # - os: ubuntu-latest
+          #   target: aarch64-unknown-linux-gnu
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Note that this was already used to retroactively publish binaries for v0.1.2 in https://github.com/stellar/soroban-cli/pull/227.

leighmcculloch:
> * why tar if it's just a single file?

I thought we might want to add a CHANGELOG and a LICENSE file. I didn't want to have to change expectations around what was in the file in the future.

I also wasn't sure if we need to include all the license files for dependencies?

I don't feel strongly though. If we don't want to include sidecar files, a raw binary would be perfectly fine.

> * what's wrong with the aarch64 build?

`ring` is a dependency that gets pulled in because the platform team used `rusttls`. It seems that `ring` can't build on Linux aarch64.

I think we should probably revisit some of the dependencies that are being pulled in for http stuff in soroban-cli.

leighmcculloch:
Removed the taring in [74dcf9b](https://github.com/stellar/soroban-cli/pull/229/commits/74dcf9b53b0f618704f0e0c652e6be9070493f95).

PR #235, opened 2022-10-26
chore: reorg integration tests into single binary and add helper types to improve tests

willemneal:
This PR reorganizes the tests into a subdirectory with a single `main` that will be built when running tests.

It also adds a `Soroban` struct, which wraps assert command. Potentially we could expose the CLI commands themselves with a builder interface and this wouldn't be needed. For now it's a good start and made the tests more concise.


tests/it/util.rs
```diff
@@ -0,0 +1,119 @@
...
+        ""
+    } else {
+        ".wasm"
+    };
+    PathBuf::from("target/wasm32-unknown-unknown/test-wasms").join(format!("{name}{wasm}"))
+}
+
+pub struct Soroban {
+    inner: Command,
+}
```
leighmcculloch, line 24: ❗ I think this type makes tradeoffs that compromise too much on the 1-1 mapping we currently have between the commands people will use, and the commands visible in the tests. The type introduces indirection. Without the type reading a test looks like a command that's being tested, so writing tests is in every also dogfooding. It's easy to look at the test and see that yes this is the exact command being executed. With the type the command being executed is being obfuscated and we invent a language of sorts the test writer is then interacting with rather than the raw command.

Before:
```rust
     const WASM: &str = "target/wasm32-unknown-unknown/test-wasms/test_invoker_account_exists.wasm";
     assert!(
         Path::new(WASM).is_file(),
         "file {WASM:?} missing, run 'make test-wasms' to generate .wasm files before running this test"
     );

     let tmp = TempDir::new().unwrap();
     let ledger = tmp.child("ledger.json");

     let mut cmd = Command::cargo_bin("soroban")?;
     cmd.arg("invoke");
     cmd.arg("--ledger-file").arg(ledger.as_os_str());
     cmd.arg("--id").arg("1");
     cmd.arg("--wasm").arg(WASM);
     cmd.arg("--fn").arg("invkexists");
     cmd.assert().success().stdout("true\n");
```

After:
```rust
     Soroban::invoke()?
         .ledger_file(temp_ledger_file())
         .contract_id("1")
         .wasm(wasm_file("test_invoker_account_exists"))?
         ._fn("invkexists")
         .assert()
         .success()
         .stdout("true\n");
```

tests/it/e2e_rpc_server.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::Ok;
+
+use crate::util::{wasm_file, Soroban};
+
+// e2e tests are ignore by default
+#[test]
+#[ignore]
+fn e2e_deploy_and_invoke_contract_against_rpc_server() -> Result<(), anyhow::Error> {
```
leighmcculloch, line 8: ❔ I'm not opposed to using anyhow in applications, but what do we get out of adding that dependency here? It appears we're only replacing `dyn std::error::Error` with `anyhow::Error`, and replacing `Result::Ok` with `anyhow::Ok`.

tests/it/main.rs
```diff
@@ -0,0 +1,3 @@
+mod e2e_rpc_server;
```
leighmcculloch, line 1: 🎉 Thanks for consolidating the integration tests into a single binary. This will scale better as we add more tests, since there will be less linker runs.

tests/it/e2e_rpc_server.rs
```diff
@@ -0,0 +1,30 @@
...
+    // This test assumes a fresh standalone network rpc server on port 8000
+
+    Soroban::new_standalone()?
+        .arg("deploy")
+        .wasm(wasm_file("test_hello_world"))?
```
leighmcculloch, line 13: ❓ Why is the `tests/fixtures/soroban_hello_world_contract.wasm` file committed as part of the PR? I'm assuming that was by accident because we don't have fixtures excluded in the gitignore.

tests/it/util.rs
```diff
@@ -0,0 +1,119 @@
...
+pub fn wasm_file(name: &str) -> PathBuf {
+    let path = Path::new(name);
+    let wasm = if path
+        .extension()
+        .map_or(false, |ext| ext.eq_ignore_ascii_case("wasm"))
+    {
+        ""
+    } else {
+        ".wasm"
+    };
```
leighmcculloch, line 18: ❗ Let's make this work one way, and expect the .wasm to always be present or not. I don't think we need to support both.

tests/it/util.rs
```diff
@@ -0,0 +1,119 @@
...
+    where
+        T: AsRef<OsStr>,
+    {
+        let path = PathBuf::from(file.as_ref());
+        if path.is_file() {
+            self.arg("--wasm");
+            self.arg(path);
+            Ok(self)
+        } else {
+            Err(anyhow::anyhow!("File not found: {}. run 'make test-wasms' to generate .wasm files before running this test", path.display()))
```
leighmcculloch, line 82: 🎉 It would be awesome to preserve the `wasm_file` helper, and to move this "is file" check and error (which can just be a panic, we don't gain a lot from an error in the helper) to the wasm_file helper.

leighmcculloch (commented):
@willemneal Thanks for improving the integration tests. I have a couple questions (❔❓), and a couple asks (❗).

tests/it/e2e_rpc_server.rs
leighmcculloch: Ah I see, it was added for the wasm_file fn. I think that the file-doesn't-exist case can continue to panic. There's little advantage as far as I know to generating errors in tests as they're presented in the same format/output as panics are in the test output. In a step-through debugger a panic is also captured typically which makes debugging test setup/utility failures easier.

tests/it/e2e_rpc_server.rs
willemneal: Yep I'll remove that.

tests/it/e2e_rpc_server.rs
willemneal: Removed Results in favor of panics.

tests/it/util.rs
willemneal: I just reorganized it so that there is a trait for creating the commands.  Thus turning into:

```rust
 Sandbox::new()
        .arg("invoke")
        .arg("--ledger-file")
        .arg(temp_ledger_file())
        .arg("--id=1")
        .arg("--wasm")
        .arg(test_wasm("test_invoker_account_exists"))
        .arg("--fn=invkexists")
        .assert()
        .success()
        .stdout("true\n");
```

willemneal: Is there a reason for breaking up the cmd from `.assert...`?

tests/e2e_rpc_server.rs
```diff
@@ -1,87 +0,0 @@
...
-        "--id=8af3f0c5c2c4b5a3c6ac67b390f84d9db843b48827376f42e5bad215c42588f7",
-        "--fn=symbol",
-    ]);
-
-    invoke
-        .assert()
-        .stdout("[88,76,77]\n")
-        .stderr("success\n")
-        .success();
-}
```
leighmcculloch, line 87: ❗ This test appears to have been removed. I don't see it in the new file. Are we able to keep it?

leighmcculloch (commented):
👏🏻 Nice. Thanks for improving these, this looks really good to me.

There's just one thing (❗) below that I think is an easy fix.

tests/e2e_rpc_server.rs
willemneal: This was moved as part of the rewrite, but didn't get renamed.

leighmcculloch (approved):
🎉 Looks great.

PR #236, opened 2022-10-27
Add optimize sub-command using wasm-opt

leighmcculloch:
### What
Add optimize sub-command using wasm-opt.

### Why
I saw @brson and @Aimeedeer's [blog post] about the [`wasm-opt`] crate today and after a quick test it appears to work very smoothly. Including optimization into the soroban-cli rather than [instructing people] to install other tools seems like it fits our philosophy of making writing effecient contracts the default.

This PR introduces an ultra simple sub-command (`soroban optimize`) that runs `wasm-opt` across the given .wasm file with the `-Oz` mode, which is the same mode we currently recommend.

I imagine we might also want to automatically run this across .wasm files as part of the deploy subcommand, although I thought doing that at this time, given that we are considering redesigning the CLI (#197), might be a bit premature.

[blog post]: https://brson.github.io/2022/10/26/creating-wasm-opt-rust-bindings-with-cxx
[`wasm-opt`]: https://crates.io/crates/wasm-opt
[instructing people]: https://soroban.stellar.org/docs/tutorials/build-optimized


leighmcculloch:
I have a concern about merging this:
- The `wasm-opt` crate requires a C compiler to also be installed. That might mean that `cargo install`ing this crate might not work in all the cases that it works today. It is unlikely to be a big problem, but I'm not sure.

leighmcculloch:
```
❯ soroban optimize --wasm target/wasm32-unknown-unknown/release/soroban_auth_contract.wasm
Reading: target/wasm32-unknown-unknown/release/soroban_auth_contract.wasm (4414 bytes)
Writing to: target/wasm32-unknown-unknown/release/soroban_auth_contract.wasm...
Optimized: target/wasm32-unknown-unknown/release/soroban_auth_contract.wasm (3522 bytes)

```

src/main.rs
```diff
@@ -40,6 +41,8 @@ enum Cmd {
     Invoke(invoke::Cmd),
     /// Inspect a WASM file listing contract functions, meta, etc
     Inspect(inspect::Cmd),
+    /// Optimize a WASM file
+    Optimize(opt::Cmd),
```
leighmcculloch, line 45: Maybe this command should be named `opt` rather than `optimize`?

dtolnay:
> * The `wasm-opt` crate requires a C compiler to also be installed. That might mean that `cargo install`ing this crate might not work in all the cases that it works today. It is unlikely to be a big problem, but I'm not sure.

You already depend on `ring` (via `rustls`, via `jsonrpsee-http-client`) which uses a C compiler.

Cargo.toml
```diff
@@ -44,6 +44,7 @@ ed25519-dalek = "1.0.1"
 jsonrpsee-http-client = "0.15.1"
 jsonrpsee-core = "0.15.1"
 regex = "1.6.0"
+wasm-opt = "0.110.1"
```
leighmcculloch, line 47: @willemneal this appears to be very similar to importing binaryen which you added in a similar way in #226, but rust-analyzer appears to still work with wasm-opt where-as I know you mentioned in #226 binaryen caused it to break and you had some concerns with it being unmaintained. Wdyt of using wasm-opt instead?

src/opt.rs
```diff
@@ -0,0 +1,53 @@
...
+        OptimizationOptions::new_optimize_for_size_aggressively()
+            .run(&self.wasm, wasm_out)
+            .map_err(Error::OptimizationError)?;
+
+        let wasm_out_size = std::fs::metadata(&wasm_out)
+            .map_err(Error::ReadingFile)?
+            .len();
+        println!(
+            "Optimized: {} ({} bytes)",
+            wasm_out.to_string_lossy(),
```
2opremio, line 47: printing out the wasm seems overkill considering it's saved to a file

src/main.rs
2opremio: I like optimize better

Cargo.toml
willemneal: I much prefer it! Binaryen was unmaintained and clearly a lot of work including using cxx went into making wasm-opt!

src/opt.rs
leighmcculloch: This line is printing out the filename, not the file contents. The bytes count that gets printed out is useful because I think as a developer whenever I optimize something I'm interested in knowing what the outcome was.

Cargo.toml
leighmcculloch: @willemneal Fyi I'm still interested in #226, which goes a step further as having an entire build process built in. Independent of that I think we'll still want a command like this, to optimize a wasm file. So I think #236 and #226 should co-exist.

willemneal: Great! I'm about to finish it up.  I'll strip out the binaryen stuff and I'll rebase to this one to reuse wasm-opt.

brson:
I have published version 0.110.2 of this crate that contains [a single Binaryen backport](https://github.com/brson/wasm-opt-rs/issues/105) that slightly reduces output size when the input module contains no "memories" section.

leighmcculloch:
> I have published version 0.110.2 of this crate that contains [a single Binaryen backport](https://github.com/brson/wasm-opt-rs/issues/105) that slightly reduces output size when the input module contains no "memories" section.

Ah nice. 

@brson Thank you!

paulbellamy:
Initially, I'm a bit on the fence about this.

Argument for including the `soroban optimize` command:
- It is a generic wasm command which is applicable regardless of contract SDK used.
- We may wish to integrate it into the deploy process automatically
- We already have `soroban inspect` which is a wasm-specific command
- Batteries-included approach to efficient contract development

Argument for excluding the `soroban optimize` command:
- It is a build-step process, which should be dealt with as part of the build pipeline (rust/make/npm/whatever).
- `soroban-cli` should not modify the wasm it is receiving (e.g. optimizing in deploy)
- Reinforces the C compiler dependency
- Automatically optimizing as part of deployment could make repeatable builds and source-code verification harder? (are repeated wasm-opt calls idempotent?)

But, overall, from writing that down, I think this SGTM.

src/opt.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::fmt::Debug;
+use wasm_opt::{OptimizationError, OptimizationOptions};
+
+#[derive(Parser, Debug)]
+pub struct Cmd {
+    /// WASM file to optimize
+    #[clap(long, parse(from_os_str))]
+    wasm: std::path::PathBuf,
+    /// Path to write the optimized WASM file to (defaults to same location as --wasm)
+    #[clap(long, parse(from_os_str))]
```
paulbellamy, line 11: Maybe a safer default is `stdout`, instead of overwriting the input file?

paulbellamy: Or some default rename `foo.opt.wasm`? `foo.min.wasm` (to match `foo.min.js` convention)?

paulbellamy (approved):

src/opt.rs
leighmcculloch: Agreed, this is a really good point, the CLI should not default to an action that is destructive.

Writing to stdout might break peoples shells because it's binary data and not all shells handle that well.

Writing to a alternative file by default seems like a good default to me. I've named it `.optimized.wasm` to match the terminology of the commands name.

leighmcculloch:
> * It is a build-step process, which should be dealt with as part of the build pipeline (rust/make/npm/whatever).

For npm projects this makes sense. For Rust projects, there's no way to hook post-build steps into a project.

The reason to include `wasm-opt` into `soroban` isn't to take over the build pipeline (although #226 is probably getting into that territory), but to provide easy access to wasm-opt. Wasm-opt can be installed in a variety of ways that complicate system setup. Since devs already need to have soroban installed, we can say here's this default set of tools that is builtin that you can use to optimize.

But maybe it is being too opinionated.

> * Automatically optimizing as part of deployment could make repeatable builds and source-code verification harder? (are repeated wasm-opt calls idempotent?)

Good point. Your comment also got me thinking that automatically modifying contracts during deploy could be surprising, and if there was ever a bug there's no opportunity for a developer to test out the artifact before it deploys.

leighmcculloch:
> * are repeated wasm-opt calls idempotent?

I did some tests with this and it appears no. I managed to get a .wasm file even smaller by running it multiple times.

willemneal:
@leighmcculloch there is a `--converge` option on the CLI which should result in the same binary as manually rerunning it until it stops decreasing.

leighmcculloch:
Thanks @willemneal. I'm experimenting with that option in https://github.com/stellar/soroban-cli/pull/238.

PR #238, opened 2022-10-31
Enable converge on wasm-opt optimization

leighmcculloch:
### What
Enable converge on wasm-opt optimization

### Why
https://github.com/stellar/soroban-cli/pull/236#issuecomment-1297325311


paulbellamy (approved):
Not a big size win, but making the `soroban optimize` subcommand idempotent is a win, IMO.

PR #239, opened 2022-10-31
Fix the bin name in the completion command

fdesjardins:
### What

Change `soroban-cli` to `soroban` for the `completion` command

### Why

Completion got broken when the bin name changed to `soroban`. This PR fixes that.


leighmcculloch (approved):
👏🏻 Thanks for fixing this.

PR #251, opened 2022-11-09
Restructure repository

tsachiherman:
### What

Restructure repository so that it would support both rust and go projects.

### Why

The Soroban-cli and the soroban-rpc are logically dependent of each other, and would ideally live at the same place.
That would allow us to move the entire soroban testing that we have scattered across the Soroban-cli and go into a single location.

### Known limitations

N/A


willemneal:
Also need to update README.md with updated bin name.

leighmcculloch:
This change is addressing a few issues together. One of those issues is particularly interesting to me: the pain of our existing use of `[patch.*]` with managing transitive dependencies that are also in development. I'm investigating that independently in https://github.com/stellar/rs-soroban-env/issues/567.

tsachiherman:
> Also need to update README.md with updated bin name.

yes, you're correct. I'll fix that later on today.

willemneal:
One quick question: now that the rust crate is a workspace member, why is the top level not just a virtual manifest? That is not contain anything other than the workspace deps and members? It seems that a lot of things are repeated in the two Cargo.toml's.

I just tested it locally and if I remove everything above the workspace info in the root Cargo.toml everything works the same.

tsachiherman:
> One quick question: now that the rust crate is a workspace member, why is the top level not just a virtual manifest? That is not contain anything other than the workspace deps and members? It seems that a lot of things are repeated in the two Cargo.toml's.
> 
> I just tested it locally and if I remove everything above the workspace info in the root Cargo.toml everything works the same. Also could move the `build.rs` to the inner crate's directory.

Not opposing to doing that - but previous attempts I've made ended up with some of the tests failed to compile.
How do you feel about applying your changes on top of mine ?

willemneal:
Sure PR incoming!

leighmcculloch:
Some thoughts:

- Agreed, a lot of the CLI files should be moved into the CLI directory rather than sit in the top-level.

- The top-level `tests` directory contains Rust tests. Should it? Should Go tests go in there too? Maybe that tests directory should be moved inside the soroban-cli directory, or it should be split in half so tests taht use both live in that directoryh while tests that are only CLI live inside that directory.

- There's an assumption being made that changes will always be applied to the CLI and RPC at the same time. If we continue to work on a model where the RPC and RPC Simulator are duplicates of each other, that makes sense. If the RPC Simulator lives inside the RPC and is just embedded in the CLI, then the two efforts can be disconnected.

- What's the update process for the RPC + CLI? Living in the same repo I assume CI will require them to be compatible with each other at the same time. But the update process for both are extremely different, with different dependencies.

   It will take longer to get the RPC up to date if it is dependent on data from stellar-core. Where-as the CLI can otherwise likely move much faster.
   
   For example:
   [soroban-dependencies-update-speed excalidraw](https://excalidraw.com/#json=G3SyVr4dnJxcB_H4_sKwY,s246Ci9qGT_uK5IQyXXegQ)

- > That would allow us to move the entire soroban testing that we have scattered across the Soroban-cli and go into a single location.
   
   How does this align with the system tests that @sreuland has been talking about? 

In terms of how we manage multiple languages within the one repo: We could explore using language agnostic monorepo build systems. It feels like that is what `deptool` is, more lightweight than most. Some that come to mind are:
- Bazel – https://bazel.build
- Buck – https://buck.build
- Pants – https://www.pantsbuild.org

I only have experience with Bazel, and it's quite a beast. I remember having more problems with it, than I thought it was solving, but I know a lot of people have success with it. Buck and Pants I have no experience with, but just know of.

leighmcculloch:
In terms of how we manage multiple languages within the one repo: We could explore using language agnostic monorepo build systems. It feels like that is what `deptool` is, more lightweight than most. Some that come to mind are:
- Bazel – https://bazel.build
- Buck – https://buck.build
- Pants – https://www.pantsbuild.org

I only have experience with Bazel, and it's quite a beast. I remember having more problems with it, than I thought it was solving, but I know a lot of people have success with it. Buck and Pants I have no experience with, but just know of.

tsachiherman:
> 

Lots of good points, let me go through in my own arbitrary order:
* Versioning and releases: As you've suggested, releases off this repository would be more frequent and would be triggered as result of either core or soroban specific changes. That means that the end user might be getting an "upgraded" soroban-cli without any meaningful changes to the cli itself. That is unfortunate, but not harmful in any way.
* In term of dependency tools - I haven't explored the ones that you've mentioned yet. I'd like to make the minimal changes here, without completing the development of the deptool. I'll defer that to a subsequent task and/or use of the tools you've suggested. In your response, you've identified that our needs here are quite minimal - which I completely agree with.
* Moving some of the tests that are under `/tests` into `/cmd/soroban-cli/tests` is something I'm perfectly fine with. I think that we still do want to have `/tests` that would exercise some soroban-cli and soroban-rpc related testing.
* Once this PR is merged, I'll be working on making the required changes to the quickstart, so that it would work seamlessly from there as well. Ideally, I would be able to extend the cross-repo dependency model to there as well.
* @sreuland was working on creating a repo that would focus on running functional tests across multiple targets ( sandbox, quickstart ); that isn't directly conflicting with this change.
* I expect some changes to the release process would be required.

Makefile
```diff
@@ -2,13 +2,19 @@ all: check build test
 
 export RUSTFLAGS=-Dwarnings -Dclippy::all -Dclippy::pedantic
 
-install:
+# update the Cargo.lock every time the Cargo.toml changes.
+Cargo.lock: Cargo.toml
+	cargo update
```
leighmcculloch, line 7: ❓ Why do we need to run `cargo update` everytime the toml file changes? That will result in updating all transitive dependencies, which is unlikely the intent of someone making a change to Cargo.toml. If someone wants to update the transitive dependencies they can run `cargo update` manually.

For context on why I'm interested in this: We usually try to be very much in control and intentional with updating dependencies, because from a security point-of-view doing so can be a risk.

README.md
```diff
@@ -17,5 +17,5 @@ All values passed to `--arg` are the JSON representation of SCVals.
 ## Example
 
 ```
-soroban invoke --id <HEX_CONTRACTID> --wasm <WASMFILE> --fn <FUNCNAME> --arg 32 --arg 4
+soroban-cli invoke --id <HEX_CONTRACTID> --wasm <WASMFILE> --fn <FUNCNAME> --arg 32 --arg 4
```
leighmcculloch, line 20: ❓ Why change the soroban binary from `soroban`, back to `soroban-cli`? We recently made the change to remove the `-cli` which came from both internal feedback and external community feedback.

leighmcculloch (commented):
I don't think my concern with binding RPC + CLI development was addressed, but also, I don't think my concern needs to block this PR.

I defer to @paulbellamy @2opremio as it affects you folks much more going forward.


I only reviewed the soroban-cli files, and from them the only thing I see here that would block merging this, and both could be addressed after merging anyway, is the two things I note below about cargo update and the soroban-cli rename.

Makefile
tsachiherman: I added this when I found that git versioning changes that I've applied to the `Cargo.toml` file were not taking affect until the `Cargo.lock` file was regenerated. I think that every time we update our dependencies, we'll need to do that.
I'm not really sure if this is required or not, and would be perfectly fine dropping this if it causes any issues.

paulbellamy: Updating dependencies automatically (unintended) seems risky to me as well.

paulbellamy (approved):
LGTM, pending binary rename, and `cargo update --workspace` changes.

2opremio:
@tsachiherman something I am missing here is the `soroban-rpc` github workflow ( https://github.com/stellar/go/blob/46b0e5c507899980f2a7ae96727ddcd90f5fce73/.github/workflows/soroban-rpc.yml ) . What's the plan in that respect?

2opremio:
@tsachiherman also, all the soroban-rpc method handlers are missing, why is that?

PR #253, opened 2022-11-11
chore: make root Cargo.toml virtual and move build script to inner crate

willemneal:


willemneal:
Seems that locally I had already built the test binaries but can confirm that they fail to rebuild.

PR #254, opened 2022-11-15
First pass proposal at `soroban profile` subcommands

paulbellamy:
Fixes https://github.com/stellar/soroban-cli/issues/173, see that for background context.

The goal here is to make managing keys and networks easier than it currently is. The "profile" concept combines a secret key with all the other config to connect to a network into a named entity.

For example, instead of:
```sh
$ soroban token create \
  --name tok --symbol tok \
  --rpc-url http://localhost:8000/soroban/rpc \
  --network-passphrase 'Local Sandbox Stellar Network ; September 2022'\
  --secret-key SDA2NVFCPEBVYD7L2JQKBIFRNZYO6YTUUHLBKJOX4YQ5ORQRG3KPSLCE
```
You could do:
```sh
# Create a new profile
$ soroban profile new --name local  \
  --rpc-url http://localhost:8000/soroban/rpc \
  --network-passphrase 'Local Sandbox Stellar Network ; September 2022' \
  --secret-key SDA2NVFCPEBVYD7L2JQKBIFRNZYO6YTUUHLBKJOX4YQ5ORQRG3KPSLCE

# Use the new profile automatically
$ soroban profile use local
$ soroban token create --name tok --symbol tok

# Or explicitly
$ soroban --profile local token crate --name tok --symbol tok
```

In this proposal the profiles are stored in `.soroban/profiles.json`.
A default profile is always created, called `sandbox`, which uses the local sandbox.
If a secret key is not specified (or for the sandbox), a random one is generated and stored.

## Other Ideas

- Instead of `soroban profile use` and `soroban profile current`, have a single `soroban profile default` command which either prints or sets depending if an argument is passed in.
- Make this a more general config-merging/defaults mechanism.
  - Paul: Not sure how that works with clap, though.
- Provide presets (or pre-configured?) for futurenet etc.
  - Paul: Unclear how this works with the `--rpc-url` though.

### Required

- [x] implement profile subcommands:
	- [x] current
	- [x] ls
	- [x] new
	- [x] use
- [x] generate secret key for new profiles, if not provided
- [ ] auto-create sandbox profile
	- [ ] generate and store the secret key on first use
- [ ] wire them into existing subcommands
	- [x] invoke
	- [ ] token
	- [ ] deploy
	- [ ] read
- [ ] Decide if we need to support overriding the flags (e.g. --rpc-url) inline in the `invoke` (and other) subcommands, or not.

### Nice-to-have

- [ ] maybe generalize it into config defaults system
- [ ] maybe use toml for the file? It would be more ergonomic for people hand-editing...


paulbellamy:
@willemneal , @jhammond2012, @chadoh, thoughts?

willemneal:
My first question is there a reason to use json and not toml for the config file? I have a preference toward config files with comments.

src/profile/new.rs
```diff
@@ -0,0 +1,100 @@
...
+        conflicts_with = "rpc-url",
+        env = "SOROBAN_LEDGER_FILE",
+        help_heading = HEADING_SANDBOX,
+    )]
+    ledger_file: std::path::PathBuf,
+
+    /// RPC server endpoint
+    #[clap(
+        long,
+        requires = "network-passphrase",
```
willemneal, line 42: Being new to the ecosystem, what it he point of this passphrase if the endpoints are unique?

Also this feels error prone since it depends on a long string.

leighmcculloch: Everything for a network includes the network passphrase in the signature payload. It makes things signed for Testnet invalid for Pubnet, and vice versa. It also allows folks to run their own networks, and even if you use the same keys across all, the things you sign will only be valid for the network they are intended.

This is similar to a "chain ID" in some other networks.

src/main.rs
```diff
@@ -33,6 +35,24 @@ const HEADING_RPC: &str = "OPTIONS (RPC)";
 struct Root {
     #[clap(subcommand)]
     cmd: Cmd,
+
```
willemneal, line 38: I would avoid adding arguments here unless needed globally.

Instead I would recommend making a new args struct that containing these and then using something like [config[(https://github.com/mehcode/config-rs). Then in any command that needs it you add 

```
#[flatten]
profile::Args
```

This avoids having to pass the extra args along e.g.
```
pub fn run(profiles_file: &std::path::PathBuf) -> Result<(), Error> {
```
becomes
```
pub fn run(&self) -> Result<(), Error> {
  self.profiles_file
```

chadoh:
I like this design! I think it solves a problem now, while being minimal enough to adapt to whatever new features emerge down the road. 

## Nitpick on the API:

```bash
$ soroban profile new --name local  \
  --rpc-url http://localhost:8000/soroban/rpc \
  --network-passphrase 'Local Sandbox Stellar Network ; September 2022' \
  --secret-key SDA2NVFCPEBVYD7L2JQKBIFRNZYO6YTUUHLBKJOX4YQ5ORQRG3KPSLCE
```

I'd rather see this mimic `git remote` commands, using `add` instead of `new` (not strongly opinionated about that one) and a positional argument for the name. So it would look like:

```bash
$ soroban profile add local  \
  --rpc-url http://localhost:8000/soroban/rpc \
  --network-passphrase 'Local Sandbox Stellar Network ; September 2022' \
  --secret-key SDA2NVFCPEBVYD7L2JQKBIFRNZYO6YTUUHLBKJOX4YQ5ORQRG3KPSLCE
```

I like this mostly because it gets rid of one double dash 🙃 

But I also like that all of the double dash args are the exact same ones that will be filled-in when using this profile.

Also, git remotes seem like a reasonable analog for network profiles.

## Thoughts on "other ideas"

> Instead of `soroban profile use` and `soroban profile current`, have a single `soroban profile default` command which either prints or sets depending if an argument is passed in.

I like the use/current version better. I imagine myself making aliases for specific environments, `alias sbn-loc="soroban --profile local"`, etc.

> Make this a more general config-merging/defaults mechanism.

I also like the idea of making it a general config-merging/defaults mechanism, but I don't think it needs to start there. This specific case could help inform that more general solution later.

> Provide presets (or pre-configured?) for futurenet etc.

This doesn't fully answer Paul's concern about `--rpc-url`, but:

My ideal way of running local nodes would feel a bit like [Postgres.app](https://postgresapp.com/), with a little Soroban icon sitting in my menu bar, and some essential GUI shipped along with it (block explorer, wallet, maybe something like the Stellar playground, maybe more), which I could access via its dropdown menu.

When I install such a thing, it could offer to add CLI profiles for each network I spin up, setting the RPC node to the local one.

For now, we could update the localnet/futurenet docs, having people add a profile before they run the `docker` commands.

willemneal:
> I would recommend a separate file per keypair, as described in [this NEAR CLI redesign proposal](https://gov.near.org/t/new-near-cli-design-proposal/3616#key-management-12).

@chadoh. Currently it's a set of network configs, but you raise a good point with how we are going to handle a keystore. The near-cli-rs [project works with the MacOs keychain](https://github.com/near/near-cli-rs/blob/master/src/common.rs#L1035). I think we should open a new issue to track implementing a keystore.

willemneal:
Expanding on Chad, I think that barring a custom profile, I think the others could be hard coded defaults.  Either that or just written to the config with something like `--init`?

leighmcculloch:
> * Instead of `soroban profile use` and `soroban profile current`, have a single `soroban profile default` command which either prints or sets depending if an argument is passed in.

I liked the `use` language, it feels intuitive, but `current` feels a bit odd – not sure why though. `default` is pretty common and is what `rustup` uses, and folks are likely to be familiar with it since they'll be using it for contract dev too. I think either works.

leighmcculloch:
> Provide presets (or pre-configured?) for futurenet etc.

+1. I think we could provide defaults for a local instance too with some general defaults that matches what quickstart image uses.

leighmcculloch:
An idea for usability: Instead of requiring the network passphrase, we can get the network passphrase from the RPC server.

leighmcculloch:
If someone wants to use multiple keys with a single network, should they be able to reuse networks, or configure networks separately to keys? Maybe "profiles" should be an "identity" + "network", where identities and networks are configured separately?

paulbellamy:
> My first question is there a reason to use json and not toml for the config file? I have a preference toward config files with comments.

+1 toml is a better choice. I just didn't want to install/learn a toml-parsing crate at this point.

paulbellamy:
> > I would recommend a separate file per keypair, as described in [this NEAR CLI redesign proposal](https://gov.near.org/t/new-near-cli-design-proposal/3616#key-management-12).
> 
> @chadoh. Currently it's a set of network configs, but you raise a good point with how we are going to handle a keystore. The near-cli-rs [project works with the MacOs keychain](https://github.com/near/near-cli-rs/blob/master/src/common.rs#L1035). I think we should open a new issue to track implementing a keystore.

re: keystore, using the os keychain is cute. Not sure what windows support for that is like. My main goal is to ultimately support hardware wallets, as that seems like the "proper" solution for devs to be deploying/invoking contracts from.

paulbellamy:
> I'd rather see this mimic `git remote` commands, using `add` instead of `new` (not strongly opinionated about that one) and a positional argument for the name. So it would look like:

Debated this when writing it. So whichever makes more sense to you, imo. Picking another tool as inspiration seems like a great idea, and git is probably a good choice.

> Also, git remotes seem like a reasonable analog for network profiles.

👍 

> Instead of `soroban profile use` and `soroban profile current`, have a single `soroban profile default` command which either prints or sets depending if an argument is passed in.

Yep, I debated that when I was writing this as well.

> I like the use/current version better. I imagine myself making aliases for specific environments, `alias sbn-loc="soroban --profile local"`, etc.

There's pretty good arguments for and against a global flag.
global = uniform and easy to alias (also for scripts)
local = only present when needed. Cleaner and more minimal.

FWIW, when writing bash scripts for other chains, I've often ended up doing something like:

```bash
NETWORK="${1:-local}"

contract() {
  soroban --network "$NETWORK" contract "$@"
}

contract deploy --wasm foo.wasm
contract invoke foo
```
Which seems easier with the global approach, which is why I opted for that here.

> My ideal way of running local nodes would feel a bit like [Postgres.app](https://postgresapp.com/), with a little Soroban icon sitting in my menu bar, and some essential GUI shipped along with it (block explorer, wallet, maybe something like the Stellar playground, maybe more), which I could access via its dropdown menu.

Sounds cool.

paulbellamy:
Closed in favour of the network/identity proposal by @willemneal  and @chadoh

PR #258, opened 2022-11-17
Fix soroban-cli => soroban

leighmcculloch:
### What
Rename `soroban-cli` to `soroban`.

### Why
We removed the `-cli` suffix on the command a while ago because we had feedback from external and internal folks that we should shorten the command. It got recently accidentally added back in #251.

### Merging
This PR is intended to be merged to `main` after #257.


tsachiherman:
I wonder how it passed all the tests without that. Looks good.

Issue #61, opened 2022-08-05
CLI: CLI Event Streaming Tool for remote core (futurenet)


paulbellamy:
Extend the cli event streaming tool from #39 to work with a remote core (e.g. futurenet).

This depends on https://github.com/stellar/go/issues/4554 from soroban rpc first to provide the service endpoint.

- [x] Allow filtering by contract id

### Nice-to-haves

- [x] Allow filtering by other parts of the topic
- [ ] Allow config to push events to a message broker (e.g. sqs?)
  - How would devs like to use this? Maybe package it as a docker image with ENV-var flags?

paulbellamy:
Blocked on: https://github.com/stellar/soroban-cli/issues/39

jcx120:
For now, output will be in console.

Shaptic:
Closed by #298.

PR #260, opened 2022-11-17
[Draft] Add scaffolding to output smart contract events.

Shaptic:
**Note**: While this is a draft PR, I'd love feedback on the design and Rust-isms.

### What
This adds a new `events` sub-command with the following help text:

```
Watch the network for contract events

USAGE:
    soroban events [OPTIONS] --start-ledger <START_LEDGER> --end-ledger <END_LEDGER>

OPTIONS:
    -s, --start-ledger <START_LEDGER>
            The first ledger sequence number in the range to pull events
            https://developers.stellar.org/docs/encyclopedia/ledger-headers#ledger-sequence

    -e, --end-ledger <END_LEDGER>
            The last (and inclusive) ledger sequence number in the range to pull events

        --format <FORMAT>
            Formatting options for outputted events [default: pretty] [possible values: pretty,
            plain, json]

    -h, --help
            Print help information

OPTIONS (RPC):
        --rpc-url <RPC_URL>    RPC server endpoint [env: SOROBAN_RPC_URL=]

OPTIONS (SANDBOX):
        --ledger-file <PATH>    Sandbox file from which to pull events from [env:
                                SOROBAN_LEDGER_FILE=]

FILTERS:
        --id <CONTRACT_IDS>...    A set of (up to 5) contract IDs to filter events on
        --topic <TOPICS>...       A set of (up to 5) topic filters to filter events on
```

The options should be self-explanatory and are based on the [current API design](https://docs.google.com/document/d/1TZUDgo_3zPz7TiPMMHVW_mtogjLyPL0plvzGMsxSz6A/edit#bookmark=id.xot5781go3at) for the `getEvents(...)` RPC endpoint:

 * specify a ledger range you want events for
 * (optionally) pick some contracts and/or topics to filter on
 * (optionally) pick an output format
 * specify either an RPC server or sandbox URL

It's output is something like this:

image

### Why
We should be able to retrieve smart contract events for certain ledger ranges. 

See also #61.

### Known limitations
We still need an actual RPC implementation to test against.


cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::rpc::Event;
+use crate::utils;
+use crate::{rpc, rpc::Client};
+use crate::{HEADING_RPC, HEADING_SANDBOX};
+
+#[derive(Parser, Debug)]
+#[clap()]
+pub struct Cmd {
+    /// The ledger range to pull events from
+    start_ledger: u32,
```
willemneal, line 18: Is this a block height? Also should it be possible to have open ranges, e.g. `start_ledger: 10` --> til the end?

Lastly is there a max range?

Shaptic: The [ledger sequence number](https://developers.stellar.org/docs/glossary/ledger/) is akin to the block height, yeah.

According to the [referenced API spec](https://docs.google.com/document/d/1TZUDgo_3zPz7TiPMMHVW_mtogjLyPL0plvzGMsxSz6A/edit#heading=h.t03hji406t76), open ranges aren't possible (yet, for v0), and the max range is a `TODO` item at the moment. It's likely that there will be one!

willemneal: Ah well then perhaps it would be helpful to add because currently it's unclear what the type of this is from the help text.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Sandbox file from which to pull events from
+    #[clap(
+        long,
+        parse(from_os_str),
+        value_name = "PATH",
+        env = "SOROBAN_LEDGER_FILE",
+        help_heading = HEADING_SANDBOX,
+        conflicts_with = "rpc-url",
+    )]
+    sandbox: Option<std::path::PathBuf>,
```
willemneal, line 38: Other commands have used `ledger_file`

cmd/soroban-cli/src/events.rs
willemneal: I think even including the link would be great. Perhaps this one:

https://developers.stellar.org/docs/encyclopedia/ledger-headers#ledger-sequence

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,192 @@
...
+            utils::contract_id_from_str(raw_contract_id).map_err(|e| {
+                Error::CannotParseContractId {
+                    contract_id: raw_contract_id.clone(),
+                    error: e,
+                }
+            })?;
+        }
+
+        let mut events: Vec<Event> = Vec::new();
+        if self.rpc_url.is_some() {
```
willemneal, line 117: if let is not stable so this could be:

```rust
if let Some(rpc_url) = self.rpc_url.as_ref() {
```

Which removes the need to unwrap in the branch. Same goes for the next case.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,192 @@
...
+        let scval = ScVal::from_xdr_base64(topic)?;
+        write!(&mut stdout, "\n            {:?}", scval)?;
+    }
+    set_white(&mut stdout)?;
+    write!(&mut stdout, "\n  Value: ")?;
+    set_green(&mut stdout)?;
+    let scval = ScVal::from_xdr_base64(&event.value)?;
+    writeln!(&mut stdout, "{:?}", scval)?;
+
+    set_white(&mut stdout)?;
```
willemneal, line 180: Can just return the result.
```suggestion
    set_white(&mut stdout)
```

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,192 @@
...
+    Ok(())
```
willemneal, line 181:
```suggestion
```

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    set_white(&mut stdout)?;
+    write!(&mut stdout, ")\n  Contract: ")?;
+    set_green(&mut stdout)?;
+    write!(&mut stdout, "{}", event.contract_id)?;
+
+    set_white(&mut stdout)?;
+    write!(&mut stdout, "\n  Topics:")?;
+    set_green(&mut stdout)?;
+    for topic in event.topic.iter() {
```
willemneal, line 170:
```suggestion
    for topic in &event.topic {
```

cmd/soroban-cli/src/events.rs
Shaptic: Nice! Done :+1:

cmd/soroban-cli/src/events.rs
Shaptic: This one gave me trouble, since it's an `io::Error` rather than a `Box` error? Which is confusing because it works with the `?` syntax...

cmd/soroban-cli/src/events.rs
Shaptic: Ah, good call. This is common nomenclature in the Stellar world, but cross-references are always helpful for people exploring this for the first time!

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,191 @@
...
+use crate::utils;
+use crate::{rpc, rpc::Client};
+use crate::{HEADING_RPC, HEADING_SANDBOX};
+
+#[derive(Parser, Debug)]
+#[clap()]
+pub struct Cmd {
+    /// The ledger range to pull events from
+    start_ledger: u32,
+    end_ledger: u32,
```
leighmcculloch, line 20: :thought_balloon: The start and end ledger will be the first place in the CLI that there are positional arguments. Every existing command has only option arguments. I don't think we need to be religious about it at this point, there might be commands where there is a definite primary participant who could be the positional argument. In this case, these values don't feel like are that though. End ledger feels like it should be optional at the least.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,191 @@
...
+        parse(from_os_str),
+        value_name = "PATH",
+        env = "SOROBAN_LEDGER_FILE",
+        help_heading = HEADING_SANDBOX,
+        conflicts_with = "rpc-url",
+    )]
+    ledger_file: Option<std::path::PathBuf>,
+
+    /// A set of (up to 5) contract IDs to filter events on
+    #[clap(long = "ids", multiple = true)]
```
leighmcculloch, line 42: ❗ 
```suggestion
    #[clap(long = "id", multiple = true)]
```
In other subcommands we've named multi-value options without the plural, since that on the command line they'll be expressed one at a time. For example, if it is named `ids`, it will look like this:
```
soroban events --ids 1 --ids 2 --ids 3 ...
```

For consistency with existing multi-value options I think we should rename it to `id`:

```
soroban events --id 1 --id 2 --id 3
```

leighmcculloch, line 42: 💡 I haven't used it before, but I think you can use `max_values(5)` here so that `clap` will ensure the max number of values doesn't exceed 5.
```suggestion
    #[clap(long = "ids", multiple = true, max_values(5))]
```

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,191 @@
...
+    #[clap(long, multiple = true)]
+    topics: Vec<String>,
+
+    /// Formatting options for outputted events
+    #[clap(
+        long,
+        possible_values = ["console", "json"],
+        default_value = "console"
+    )]
+    format: String,
```
leighmcculloch, line 55: :bulb: Fyi you can use an `enum` for this field type, which will mean elsewhere in the code you can access the value as an enum rather than as a string. It will ensure everywhere we use the enum we exhaustively consider all paths, etc.

Here's an example for how to do it:
https://github.com/stellar/soroban-tools/blob/e4d3ed4d85181eabf745681b69b039aad028cb8f/cmd/soroban-cli/src/gen.rs#L15-L25

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,191 @@
...
+    },
+
+    #[error("too many contracts IDs (max 5)")]
+    TooManyContractIds {},
+
+    #[error("too many topic filters (max 5)")]
+    TooManyTopicFilters {},
+
+    #[error("you must specify either an RPC server or sandbox filepath")]
+    TargetRequired {},
```
leighmcculloch, line 76: :bulb: Fyi variants that don't have values don't need the `{}`.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,191 @@
...
+            events = rpc_event.events;
+        } else if let Some(path) = self.ledger_file.as_ref() {
+            // TODO: Get events from the sandbox.
+            if !path.exists() {
+                return Err(Error::InvalidSandboxFile {
+                    path: path.to_str().unwrap().to_string(),
+                });
+            }
+        } else {
+            return Err(Error::TargetRequired {});
```
leighmcculloch, line 130: :bulb: Fyi if you remove the `{}` from where this error is defined, you can call it without the `{}` here too.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        for event in &events {
+            if self.format == "json" {
```
leighmcculloch, line 134: ❗ We should avoid unwrap in the application if we can since the output is pretty scary to a user. Here's an example of what we do elsewhere for JSON generation failures:
https://github.com/stellar/soroban-tools/blob/e4d3ed4d85181eabf745681b69b039aad028cb8f/cmd/soroban-cli/src/read.rs#L179-L184

leighmcculloch (commented):
👏🏻 Very nice. Thanks for adding such clear error messages for all the error cases.

Some comments inline. A couple small things I think we should change are marked with ❗. Other suggestions 💡 are things I noticed that would be more idiomatic of Rust, or other things you have as an option.

cmd/soroban-cli/src/events.rs
Shaptic: I knew this comment was coming :sunglasses: I adopted a philosophy of "optional arguments should be optional" some time ago, so I always defer to using positional arguments for things that are required. The end ledger may become optional, yeah, but the spec needs to be updated first. (I suppose we could make a `getLatestLedger`-esque call as an alternative, but that'd be a post-v0 enhancement imo.) Anyway, like you, I don't want to be religious about it, just make the case for making things that are required a little more explicit. I'm happy to change this to `--start-ledger` and `--end-ledger`, but wanted this thread to happen, first.

Thinking about this a bit, though:

>  there might be commands where there is a definite primary participant who could be the positional argument

it might actually make sense to have `<rpc-url-or-sandbox>` as a positional argument.

cmd/soroban-cli/src/events.rs
Shaptic: Oh nice. I was doing it as `--ids 1 2 3` which is why the plural made sense; I didn't realize you can repeat the argument name like that.

cmd/soroban-cli/src/events.rs
leighmcculloch: > it might actually make sense to have `<rpc-url-or-sandbox>` as a positional argument.

Something to consider it's difficult to make positional arguments consistent across different commands. Take the `<rpc-url-or-sandbox>` option. If it was positional, would it be positional on every command? That also implies it can't be provided some other way like an ENV var.

> I adopted a philosophy of "optional arguments should be optional" some time ago

This makes for some awkward commands. For example, if end ledger is optional and start ledger is mandatory, it's quite difficult to understand just by reading the following command what it does.

```
soroban events --end-ledger 20 15
```

cmd/soroban-cli/src/events.rs
Shaptic: Good idea! String comparisons are silly :smile:

Shaptic: `r#output` seems like magic :magic_wand:

cmd/soroban-cli/src/events.rs
leighmcculloch: I think what you're doing here is fine with `?;`.

The `?` syntax converts the `io::Error` into an `Error`. It automatically calls `.into()` on the error. Whereas when you just return the function there's no auto-conversion.

cmd/soroban-cli/src/events.rs
leighmcculloch: Oh, you don't need the `r#`. If you have a identifier that is the same word as a keyword, you can put `r#` on the front of it to force it to be an identifier. You shouldn't need to use it in your case.

cmd/soroban-cli/src/events.rs
Shaptic: Hmm, fair enough. In my mind, the RPC server will never return junk, but maybe I give users too much credit :laughing:

cmd/soroban-cli/src/events.rs
Shaptic: Neat!

Shaptic:
Thanks for the feedback, @leighmcculloch! I've incorporated all of your suggestions besides the one pertaining to the short philosophical thread, [above](https://github.com/stellar/soroban-tools/pull/260#discussion_r1025786113), for the sake of discussion :)

cmd/soroban-cli/src/events.rs
Shaptic: That's true! Though you'd write `events 15 --end-ledger 20` for clarity. But still awkward. Your note about envvars is very compelling, too :+1: anyway, fixed in latest commit

PR #263, opened 2022-11-20
Added type description to errors when using --arg

waldmatias:
### What

Adding what type (Symbol, Byte, etc) a specified argument via the --arg option is trying to be parsed into, and if a parsing error occurs, include it in the error output to provide a hint as to why the arg is unable to be parsed. 

### Why

Currently, there's a very general and unhelpful error if an error occurs when parsing a val to type. 

### Known limitations

A simple change, basically to show more error information when specifing --arg.


cmd/soroban-cli/src/strval.rs
```diff
@@ -120,21 +122,21 @@ pub fn from_json(v: &Value, t: &ScSpecTypeDef) -> Result<ScVal, StrValError> {
         }
         (ScSpecTypeDef::I32, Value::Number(n)) => ScVal::I32(
             n.as_i64()
-                .ok_or(StrValError::InvalidValue)?
+                .ok_or(StrValError::InvalidValue(Some(ScSpecType::I32)))?
```
leighmcculloch, line 125: 💡 I think if we pass in `t`, which is an `ScSpecTypeDef`, instead of writing out an `ScSpecType` enum value, we'll get even more information in the cases that types have sub types, generics, etc.

cmd/soroban-cli/src/strval.rs
```diff
@@ -15,7 +16,7 @@ use crate::utils;
 #[derive(Debug)]
 pub enum StrValError {
     UnknownError,
-    InvalidValue,
+    InvalidValue(Option<ScSpecType>),
```
leighmcculloch, line 19: 💭 An optional is fine for now, but I suspect we can probably replace any of the `None` cases because in every case we expect to know what the `ScSpecTypeDef` is. This doesn't need to block this PR, and is something we can follow up with.

leighmcculloch (approved):
This is great 👏🏻, and really improves the experience of interacting with args. Great idea.

There's one thing we can do to possibly further improve the output that I note below, but I think this already goes a long way to improve the situation, so we could merge this as is and consider follow ups to continue to improve.

cmd/soroban-cli/src/strval.rs
waldmatias: The thinking behind `None` is simply because as of today there are some cases where `I don't know` what the best error output (info) would be. The more I get to know Soroban and its type def, I could certainly add the correct error handling to these cases, therefore resulting in `None`, and `Option<>` being replaced for something better :)   -- such as having a `mod hints` that could give you a-la-Rust error messages: _"You passed value X but a Symbol is required. Did you try passing X in the following format :Y? Remember Symbols are 10chars and your is Z chars."_

cmd/soroban-cli/src/strval.rs
waldmatias: I agree, but using `ScSpecTypeDef` added complexity for `Vec`, `Map`, `Tuple`, etc and also custom types and I wanted a first-approach working solution.

leighmcculloch: I had a look and switched it to using def. It doesn't result in additional functional, but it should make ti easier to maintain because we already have the def in all cases so we don't need to hardcode types and instead can pass the def value through.

PR #265, opened 2022-11-23
 soroban-rpc: sync up with go repository

tsachiherman:
### What

sync up with go repository.

### Why

Need to sync up before removing it from the go repo.

### Known limitations

N/A


.github/workflows/soroban-rpc.yml
```diff
@@ -0,0 +1,52 @@
...
+    name: Integration tests
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        go: [1.19.1]
+    runs-on: ${{ matrix.os }}
+    env:
+      SOROBAN_RPC_INTEGRATION_TESTS_ENABLED: true
+      SOROBAN_RPC_INTEGRATION_TESTS_CAPTIVE_CORE_BIN: /usr/bin/stellar-core
+      PROTOCOL_19_CORE_DEBIAN_PKG_VERSION: 19.4.0-1075.39bee1a2b.focal
```
sreuland, line 20: should this be bumped to most recent soroban core version:

19.5.1-1111.eba1d3de9.focal~soroban

go.mod
```diff
@@ -8,26 +8,20 @@ require (
 	github.com/sirupsen/logrus v1.4.1
 	github.com/spf13/cobra v0.0.0-20160830174925-9c28e4bbd74e
 	github.com/spf13/viper v0.0.0-20150621231900-db7ff930a189
-	github.com/stellar/go v0.0.0-20221028081946-9308f9531041
```
sreuland, line 11: just for suggestion, a comment about the version string, i.e. which part is git hash, maybe a brief how-to/example for changing to a newer version of go repo?

tsachiherman: I don't know if I can add that comment in here ( i.e. inside the go.mod file ), but I can add a comment in the README file. Beside that, I'm hoping that the deptool would be able to help us with that ( once it's ready for prime time ).
When I did the update, I used
go get github.com/stellar/go@9308f9531041

but I clearly realize it's not the best approach and we need better tooling here.

.github/workflows/soroban-rpc.yml
tsachiherman: yes, I'll update that.

sreuland (approved):
since it is mainly lift and move code, haven't gone through line by line, the CI tests should be passing to verify net change, looks like the rpc integration tests are having an issue in the pr workflow checks.

tsachiherman:
> 

Yes, I'm working on addressing these; shouldn't be any material change, though.

PR #266, opened 2022-11-28
deptool: update the dependency scanning tool to support workspace versioning

tsachiherman:
### What

Update the dependency scanning tool to support workspace versioning

### Why

We started using a cargo.toml workspace versioning on some of our repositories. This implies that the version for the package itself could be in either the root workspace cargo.toml or in the package itself.

### Known limitations

N/A


cmd/deptool/analyze.go
```diff
@@ -62,87 +69,277 @@ func analyzedDependency(depInfo projectDependency) analyzedProjectDependency {
...
+}
+
+func findCargoVersionForCommit(repo *git.Repository, pkgName string, commit *object.Commit) (string, bool, error) {
+	treeRoot, err := commit.Tree()
+	if err != nil {
+		return "", false, err
+	}
+	rootCargoFile, err := treeRoot.File("Cargo.toml")
+	internalWorkspacePackage := false
+	if err == nil {
```
2opremio, line 262: I would check for the error instead, to avoid the nesting. It will make the function much more readable.

cmd/deptool/analyze.go
```diff
@@ -62,87 +69,277 @@ func analyzedDependency(depInfo projectDependency) analyzedProjectDependency {
...
+					continue
+				} else if strings.HasPrefix(lineParts[0], "version") && curPkgName == pkgName {
+					version := strings.ReplaceAll(strings.TrimSpace(lineParts[1]), "\"", "")
+					return version, false, nil
+				}
+			}
+		}
+		// fall-back to package specific versioning.
+	}
+	if internalWorkspacePackage {
```
2opremio, line 311: Same here

cmd/deptool/analyze.go
2opremio: You could do

```
if err != nil && !internalWorkspacePackage {
 // print/return error
}

// handle it without error
```

Alternatively you can move the existing if blocks into functions of their own, since they are pretty large.

2opremio (approved):
Only an (optional) code style comment

PR #268, opened 2022-11-28
git: update .gitIgnore

tsachiherman:
### What

update the .gitignore so that git would ignore .DS_Store and .vscode/* files.

### Why

This would allow the developers run `git status --porcelain` effectively and confirm that no changes were introduced.

### Known limitations

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target/
 .soroban/
+.DS_Store
+.vscode/


2opremio (approved):
Could you also add `.idea/` (Jetbrains IDEs) to the file?

Thanks!

.gitignore
```diff
@@ -1,2 +1,4 @@
 target/
 .soroban/
+.DS_Store
```
leighmcculloch, line 3: You may want to put this into a global gitignore so that by default you don't find these files in your git repos. I use that and have found it handy: https://github.com/leighmcculloch/dotfiles/blob/c490042da914da097a856ef0281e3d17a3debc04/files/gitconfig#L7

leighmcculloch (approved):

tsachiherman:
> Could you also add `.idea/` (Jetbrains IDEs) to the file?
> 
> Thanks!

I think that I'll go with @leighmcculloch suggestion and use a global .gitignore file; it makes it easier across repos.

PR #269, opened 2022-11-29
soroban-rpc: Use Rust to obtain preflights

2opremio:
Migrated from https://github.com/stellar/go/pull/4692


tsachiherman:
@2opremio could you hold off merging this PR until after the coming release ? The features here aren't required for this release, so I think we could reduce the surface-area for potential bugs by holding this for another week.

cmd/soroban-rpc/lib/preflight/.gitignore
```diff
@@ -0,0 +1,2 @@
+Cargo.lock
```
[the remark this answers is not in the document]
2opremio: AFAIK, it's common practice to not commit Cargo.lock for libraries.

cmd/soroban-rpc/lib/Makefile
```diff
@@ -0,0 +1,14 @@
+
```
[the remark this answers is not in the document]
2opremio: It is needed to obtain the default target, it could also be a script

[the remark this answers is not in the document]
2opremio: In other words, what's the alternative? (we can't simply remove the Makefile)

2opremio:
> @2opremio could you hold off merging this PR until after the coming release ? The features here aren't required for this release, so I think we could reduce the surface-area for potential bugs by holding this for another week.

We can't merge as is anyways, since the PR is missing a way to supply the Rust preflight library with ledger entries.

2opremio:
unrelated tests don't work on windows because MSVC fails to build the rust library (which I am not even sure why it's attempted)

I will keep looking into this

2opremio:
`Soroban Tools e2e / System tests` is failing due to https://github.com/stellar/system-test/pull/18 (CC @sreuland )

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -49,45 +130,85 @@ func NewSimulateTransactionHandler(logger *log.Entry, coreClient *stellarcore.Cl
...
+		li := C.CLedgerInfo{
+			network_passphrase: C.CString(networkPassphrase),
+			sequence_number:    C.uint(latestLedger),
+			protocol_version:   20,
+			timestamp:          C.uint64_t(time.Now().Unix()),
+			// Current base reserve is 0.5XLM (in stroops)
+			base_reserve: 5_000_000,
+		}
+
+		handle := C.uintptr_t(cgo.NewHandle(snapshotSourceHandle{readTx, logger}))
```
paulbellamy, line 185: I guess this doesn't need a matching free?

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -265,6 +278,74 @@ func TestSimulateTransactionSucceeds(t *testing.T) {
...
+
+	assert.NoError(t, err)
+	txB64, err := tx.Base64()
+	require.NoError(t, err)
+	request := methods.SimulateTransactionRequest{Transaction: txB64}
+	var response methods.SimulateTransactionResponse
+	err = client.CallResult(context.Background(), "simulateTransaction", request, &response)
+	assert.NoError(t, err)
+	assert.Empty(t, response.Error)
+}
```
paulbellamy, line 347: We should probably make some positive assertions about the result as well. Like footprint is not empty, etc.

cmd/soroban-rpc/lib/preflight/.gitignore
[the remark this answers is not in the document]
paulbellamy: But this isn't a rust library we're shipping to the general public. And the extra certainty from the lockfile might help avoid having to debug weird CI failures in future.

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+        // TODO: use a proper error
+        let entry = LedgerEntry::from_xdr_base64(res_str)
+            .map_err(|_| ScHostStorageErrorCode::UnknownError)?;
+        unsafe { FreeGoCString(res) };
+        Ok(entry)
+    }
+
+    fn has(&self, key: &LedgerKey) -> Result<bool, HostError> {
+        let key_xdr = key.to_xdr_base64().unwrap();
+        let key_cstr = CString::new(key_xdr).unwrap();
```
paulbellamy, line 56: `unwrap` vs `?` here?

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            )),
+            Err(_) => preflight_error(
+                "panic during preflight_host_function() call: unknown cause".to_string(),
+            ),
+        },
+        Ok(r) => r,
+    }
+}
+
+fn preflight_host_function_or_maybe_panic(
```
paulbellamy, line 156: Intuitively, I feel like pushing the panic-handling down as low as possible, and using proper rust error handling would make this code easier to follow? But that might not be worth the effort...

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: You are absolutely right

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: 👍

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -2,16 +2,97 @@ package methods
...
+#cgo linux,arm64 LDFLAGS: -L${SRCDIR}/../../../../target/aarch64-unknown-linux-gnu/release/ -lpreflight -ldl -lm
+*/
+import "C"
+
+type snapshotSourceHandle struct {
+	readTx db.LedgerEntryReaderTx
+	logger *log.Entry
+}
+
+// SnapshotSourceGet takes a LedgerKey XDR in base64 string and returns its matching LedgerEntry XDR in base64 string
```
leighmcculloch, line 40: > base64

💡  Do we need to base64 encode/decode these values? I would have thought we could pass the byte slices of the ledger keys without doing the base64 dance.

I don't know cgo well enough to know if this is guaranteed, but if there's no base64 encode/decode there's less allocations, maybe less need to free Go strs. We might be able to remove `FreeGoCString`.

It also looks like there's a lot of string dancing happening (e.g. unsafe string conversions) that could be avoided if we used byte slices.

cmd/soroban-rpc/lib/preflight/src/lib.rs
leighmcculloch: ❗ +1 It would be ideal if we could eliminate all the `.unwrap()` calls.

If we remove the base64 back and forth we should reduce some of the error cases needed to handle, since we wouldn't need to dance around Rust's unicode-only string types.

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+
+fn storage_footprint_to_ledger_footprint(
+    foot: &storage::Footprint,
+) -> Result<LedgerFootprint, xdr::Error> {
+    let mut read_only: Vec<LedgerKey> = Vec::new();
+    let mut read_write: Vec<LedgerKey> = Vec::new();
+    for (k, v) in &foot.0 {
+        match v {
+            AccessType::ReadOnly => read_only.push((**k).clone()),
+            AccessType::ReadWrite => read_write.push((**k).clone()),
```
leighmcculloch, line 99: 💡 We don't need to change this now, but fyi I think you can skip cloning the footprint in this conversion. If you change the `foot` parameter to be a move instead of a borrow (i.e. remove the `&`) you should be able to destruct its contents. Moving the footprint should be fine because the `storage` value in the `preflight_host_function_or_maybe_panic` function is only used for this purpose.

This might be a rabbit hole though because of the Rc<>'s.

🤔 @graydon It's pretty inconvenient that the Storage footprint map returned from `Host::try_finish` contains Rc<>'s, especially since users usually interact with it after calling `try_finish` and deconstructing the host, implying that the data is being completely handed over to the user. Maybe there's nothing good we can do about that in the env crate.

leighmcculloch (commented):
I had a look through. Most of my comments don't need to block merging this (❔ ℹ️ 💬 💡). I think we should address the error handling (❗), but even that could happen after merging so we can get ahead with using this logic, but ideally we don't leave those .unwrap calls for long. 

ℹ️ – Fyi
❔ – Question
💡 – Suggestion
💬 – Discussion we should have
❗ – Important

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -39,6 +40,12 @@ type LedgerEntryUpdaterTx interface {
 	Done() error
 }
 
+type LedgerEntryReaderTx interface {
```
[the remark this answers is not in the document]
2opremio: That was the idea, yes

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -2,16 +2,97 @@ package methods
...
+		h.logger.Errorf("SnapshotSourceGet(): GetLedgerEntry() failed: %v", err)
+		return nil
+	}
+	if !present {
+		return nil
+	}
+	out, err := xdr.MarshalBase64(entry)
+	if err != nil {
+		h.logger.Errorf("SnapshotSourceGet(): failed to marshal ledger entry from store: %v", err)
+		return nil
```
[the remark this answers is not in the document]
2opremio: I didn't want to panic in Go callbacks from Rust. I will test it and see if panicking works.

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: > If we remove the base64 back and forth we should reduce some of the error cases needed to handle

That would require:
1. creating (and book-keeping) C stubs for ledger entries 
2. Converting the C entries version back a forth Go<->C<->Rust

I would very much like to avoid that

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -2,16 +2,58 @@ package methods
...
+/*
+#include "../../lib/preflight.h"
+#include <stdlib.h>
+// This assumes that the Rust compiler should be using a -gnu target (i.e. MinGW compiler) in Windows
+// (I (fons) am not even sure if CGo supports MSVC, see https://github.com/golang/go/issues/20982)
+#cgo windows,amd64 LDFLAGS: -L${SRCDIR}/../../../../target/x86_64-pc-windows-gnu/release/ -lpreflight -ldl -lm -static -lws2_32 -lbcrypt -luserenv
+// You cannot compile with -static in macOS (and it's not worth it in Linux, at least with glibc)
+#cgo darwin,amd64 LDFLAGS: -L${SRCDIR}/../../../../target/x86_64-apple-darwin/release/ -lpreflight -ldl -lm
+#cgo darwin,arm64 LDFLAGS: -L${SRCDIR}/../../../../target/aarch64-apple-darwin/release/ -lpreflight -ldl -lm
+// In Linux, at least for now, we will be dynamically linking glibc. See https://github.com/2opremio/soroban-go-rust-preflight-poc/issues/3 for details
```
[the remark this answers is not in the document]
2opremio: > What's the consequence of this? Will this means the Linux binaries are only usable on that same distribution and version of Linux?

Yes, that's what I understand.

The only alternative I saw was  linking statically with musl but it didn't quite work, due to the broken unwinding of Rust panics https://github.com/2opremio/soroban-go-rust-preflight-poc/issues/3#issuecomment-1317976714

.github/workflows/soroban-rpc.yml
```diff
@@ -25,19 +29,69 @@ jobs:
       - uses: ./.github/actions/setup-go
         with:
           go-version: ${{ matrix.go }}
-      - run: go test -race -cover -timeout 25m -v ./cmd/soroban-rpc/...
+      - run: |
+          go generate ./cmd/soroban-rpc/...
```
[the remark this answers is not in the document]
2opremio: We would need to commit the library for every supported architecture, I would like to avoid that but if `go generating` proves hard to maintain, we can implement it.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -2,16 +2,97 @@ package methods
...
+// I (fons) did try linking statically against musl but it caused problems catching (unwinding) Rust panics.
+#cgo linux,amd64 LDFLAGS: -L${SRCDIR}/../../../../target/x86_64-unknown-linux-gnu/release/ -lpreflight -ldl -lm
+#cgo linux,arm64 LDFLAGS: -L${SRCDIR}/../../../../target/aarch64-unknown-linux-gnu/release/ -lpreflight -ldl -lm
+*/
+import "C"
```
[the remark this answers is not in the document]
2opremio: That's a good idea, but I don't think we should do it until another go project needs this.

Also, I don't how hard it would be to do the Rust-library compiling and linking when providing a library. My guess is that users of the library won't be able to use it in a straightforward way. Although, if the sqlite go bindings work this should probably work too.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -13,15 +13,17 @@ import (
 
 	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal"
 	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/config"
-	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/ledgerentry_storage"
+	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/db"
+	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/ledgerentry_writer"
```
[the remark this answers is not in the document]
2opremio: Cool, happy to rename it

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: Uhmm, good point, I will look into it.

cmd/soroban-rpc/lib/preflight/.gitignore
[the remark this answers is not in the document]
2opremio: I will commit it then!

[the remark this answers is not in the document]
2opremio: Wait, there is no Cargo.lock file to commit anymore (I think that, after I started working on this, we introduced workspaces, which uses the toplevel Cargo.lock file).

I will just remove the .gitignore file

cmd/soroban-rpc/lib/Makefile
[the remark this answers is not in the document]
2opremio: I've reworked the way in which it's built and moved everything to the top Makefile. PTAL.

The only thing I have not been able to work out is how to avoid compiling libpreflight twice and differently through the `build` (which builds libpreflight through `cargo build`) and `libpreflight target (which does `cd cmd/soroban-rpc/lib/preflight && cargo build --target $(CARGO_BUILD_TARGET) --release`).

I think it's much better overall though.

.github/workflows/soroban-rpc.yml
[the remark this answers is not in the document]
2opremio: BTW, I have just removed the `go generate` comment and resorted to just use the top makefile (it's cleaner since `go generate` was simply invoking Make). See https://github.com/stellar/soroban-tools/pull/269#discussion_r1088883689

cmd/soroban-rpc/internal/methods/simulate_transaction.go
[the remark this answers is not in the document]
2opremio: It seems to work, so I incorporated it

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: I think I've improved it, PTAL

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: I looked into passing binary-encoded values but:

* it doesn't reduce the unwrap calls (since you still need to handle encoding/decoding errors )
* it doesn't reduce the explicit allocation/deallocation calls (FreeGoCString and so on), since they need to be replaced by analog ones.

In fact, if we don't want to handle length parameters (which is a pain) we would still pass `char *` parameters around.

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: > you should be able to destruct its contents

how, sorry, but Rc's are super confusing

2opremio: Like this?

```rust
fn storage_footprint_to_ledger_footprint(
    foot: storage::Footprint,
) -> Result<LedgerFootprint, xdr::Error> {
    let mut read_only: Vec<LedgerKey> = Vec::new();
    let mut read_write: Vec<LedgerKey> = Vec::new();
    for (k, v) in foot.0 {
        match v {
            AccessType::ReadOnly => read_only.push(Rc::try_unwrap(k).unwrap()),
            AccessType::ReadWrite => read_write.push(Rc::try_unwrap(k).unwrap()),
        }
    }
    Ok(LedgerFootprint {
        read_only: read_only.try_into()?,
        read_write: read_write.try_into()?,
    })
}
```

.github/workflows/rust.yml
```diff
@@ -64,8 +64,11 @@ jobs:
     - run: cargo clippy --all-targets --target ${{ matrix.target }}
     - run: make build-test-wasms
     - if: startsWith(matrix.target, 'x86_64')
-      run: cargo test --workspace --target ${{ matrix.target }}
-
+      # specify directories explicitly to avoid building the preflight library (otherwise it will fail with missing symbols)
+      run: |
+        for I in cmd/soroban-cli cmd/soroban-cli/tests/fixtures/test-wasms/hello_world cmd/soroban-cli/tests/fixtures/test-wasms/invoker_account_exists; do
+          cargo test --target ${{ matrix.target }} $I
+        done
```
leighmcculloch, line 71: You can prevent the preflight crate from being tested by adding the following to its Cargo.toml:
```
[lib]
test = false
doctest = false
```
I think if you add that, you probably don't need to do this loop.

.github/workflows/soroban-rpc.yml
```diff
@@ -25,19 +29,69 @@ jobs:
...
+          sudo apt-get install -y gcc-10-aarch64-linux-gnu
+          echo 'CC=aarch64-linux-gnu-gcc-10' >> $GITHUB_ENV
+
+      - name: Build Soroban RPC reproducible build
+        run: |
+          CARGO_BUILD_TARGET=${{ matrix.rust_target }} make build-libpreflight
+          CGO_ENABLED=1 GOARCH=${{ matrix.go_arch }} go build -trimpath -buildvcs=false ./cmd/soroban-rpc
+          touch empty
+          ls -lh soroban-rpc
+          file soroban-rpc
```
leighmcculloch, line 82: What's the `touch empty` for?

cmd/soroban-rpc/lib/preflight/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[lib]
+crate-type = ["staticlib"]
+
+[dependencies]
+libc = "0.2.2"
+soroban-env-host = {workspace = true, features = ["vm"]}
+
+[profile.release]
+panic = 'unwind'
```
leighmcculloch, line 13: ❗ This profile configuration is generating warnings:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/leighmcculloch/Code/soroban-tools/cmd/soroban-rpc/lib/preflight/Cargo.toml
workspace: /Users/leighmcculloch/Code/soroban-tools/Cargo.toml
```

The warning is indicating it is ignored because profiles are only read out of the workspace Cargo.toml, not crate Cargo.toml files.

Unfortunately if we need one crate to have a different profile to other crates, then we'll need to create a new profile, such as `release-with-panic-unwind`, and then manually build this crate with that different profile.

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn FreeGoCString(str: *const libc::c_char);
+}
+
+struct CSnapshotSource {
+    handle: libc::uintptr_t,
+}
+
+impl SnapshotSource for CSnapshotSource {
+    fn get(&self, key: &LedgerKey) -> Result<LedgerEntry, HostError> {
+        let key_xdr = key.to_xdr_base64().unwrap();
```
leighmcculloch, line 38: @graydon @jayz22 What's the right way for code like the code here to create a `HostError`? As far as I can tell creating a `HostError` is really difficult unless you have a borrow of a `Host` and you can use it's helper functions to construct the right value. In this code there's no `Host` available for the code to use to construct an error, and so we're ending up with a lot of panics instead.

How do you recommend we map errors in this code?

cmd/soroban-rpc/lib/preflight/src/lib.rs
leighmcculloch: > * creating (and book-keeping) XDR C stubs for ledger entries

I'm not sure what this means. I wasn't suggesting passing C structs back and forth, but rather just the XDR binary, instead of XDR binary that is also base64 encoded. I believe it would mean not needing to use `CString`. But yes there would still be failure cases for `.to_xdr()` instead of `.to_xdr_base64()`.

Regardless, we should be able to return a `HostError` in each of these error cases instead of panicking. The `HostError` can be a little awkward to work with, but we can create simple ones with the `Sc` error codes, which I see us doing in some lines above (e.g. `.map_err(|_| ScHostStorageErrorCode::UnknownError)?`).

We can ask @graydon @jayz22 about how to attach more information to the host error too.

2opremio: > > * creating (and book-keeping) XDR C stubs for ledger entries
> 
> I'm not sure what this means. I wasn't suggesting passing C structs back and forth, but rather just the XDR binary, instead of XDR binary that is also base64 encoded. I believe it would mean not needing to use `CString`. But yes there would still be failure cases for `.to_xdr()` instead of `.to_xdr_base64()`.

1. The `preflight_host_function()` caller needs base64-encoded results
2. If we pass binary-buffers through C, we need to pass both the buffer and the length separately, which is a pain, particularly for return values (e.g. `extern char *SnapshotSourceGet(uintptr_t handle, char *ledger_key);`)

So, for simplicitly, I think it's better to just use base64 everywhere.

> Regardless, we should be able to return a `HostError` in each of these error cases instead of panicking. The `HostError` can be a little awkward to work with, but we can create simple ones with the `Sc` error codes, which I see us doing in some lines above (e.g. `.map_err(|_| ScHostStorageErrorCode::UnknownError)?`).
> 
> We can ask @graydon @jayz22 about how to attach more information to the host error too.

Happy to use `HostError` but I didn't see how to include generic information (and the existing errors wouldn't convey what happened). I will ask for help.

> Or is there no consequence of a panic in this code? I'm not sure how a panic here will impact the integrity or the Go program, or the impact to unfreed memory.


The Go program should be fine, since the panic shouldn't propagate there (we catch the panic explicitly).

As for the memory, I think that we can only leak memory by not calling `FreeGoCString()`, which could happen if `res_cstr.to_str().unwrap();` panics. I will try to return a `HostError` there.

.github/workflows/rust.yml
2opremio: I may end up adding tests to the library. The loop is there because trying to run the tests is only problematic in some architectures (e.g. windows, if I recall properly)

.github/workflows/soroban-rpc.yml
2opremio: I don't recall, it was there before the PR got started. I will remove it

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: Yeah, but unfortunately `ScStatus` isn't expressive enough.

cmd/soroban-rpc/lib/preflight/Cargo.toml
2opremio: Thanks, I've done that

2opremio:
@leighmcculloch I have removed all `unwrap`s where I thought was reasonable. PTAL

cmd/soroban-rpc/internal/methods/simulate_transaction.go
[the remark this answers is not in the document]
tsachiherman: I would like to suggest a third option:
create a go wrapper library, but place it within soroban-rpc subtree ( for now ).
If we'll need to use it anywhere else, it would be a matter of moving it around.

That would allow us to properly unit test that, while keeping this part of the soroban-rpc codebase for now

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    handle: libc::uintptr_t,
+}
+
+impl SnapshotSource for CSnapshotSource {
+    fn get(&self, key: &LedgerKey) -> Result<LedgerEntry, HostError> {
+        let key_xdr = key
+            .to_xdr_base64()
+            .map_err(|_| ScStatus::UnknownError(Xdr))?;
+        let key_cstr = CString::new(key_xdr).map_err(|_| ScStatus::UnknownError(General))?;
+        let res = unsafe { SnapshotSourceGet(self.handle, key_cstr.as_ptr()) };
```
tsachiherman, line 43: dealing with multiple allocators is quite error prone. In fact, I don't know if this usage would work reliably, since GO might decide to GC the string ( since there are no more references to it ).
I think that the better solution here is to avoid mixing allocators altogether : 
make SnapshotSourceGet(C.uintptr_t, char*, void* val, C.uintptr_t in_val_len)
when called with  val == null and in_val_len == 0, the SnapshotSourceGet would return the required size.
then, a second call to SnapshotSourceGet with a staging buffer and a size would allow the SnapshotSourceGet to fill out that buffer.
This would also work around the need of base64 encoding.

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -353,3 +261,109 @@ func runMigrations(db *sql.DB, dialect string) error {
...
+	}
+	latestLedgerStr := results[0]
+	latestLedger, err := strconv.ParseUint(latestLedgerStr, 10, 32)
+	if err != nil {
+		return 0, err
+	}
+	return uint32(latestLedger), nil
+}
+
+func (l *ledgerReaderTx) GetLedgerEntry(key xdr.LedgerKey) (bool, xdr.LedgerEntry, error) {
```
tsachiherman, line 309: I'm not sure if I'm following this correctly; while executing a smart contract, the host would request a specific key.
we would then:
1. pull the value from the database ( and get a byte slice )
2. convert the byte slice into xdr.LedgerEntry
3. convert the xdr.LedgerEntry back to byte slice
4. encode the byteslice as base64
5. decode the base64 into bytes ( rust )
6. decode the bytes into xdr ( rust )

that's seems to be a lot of conversions. maybe we would have a simpler version of this method that just get the bytes and pass these, as-is, to the rust code ? i.e.
1. pull the value from the database ( and get a byte slice )
2. decode the bytes into xdr ( rust )

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -353,3 +261,109 @@ func runMigrations(db *sql.DB, dialect string) error {
...
+	// care whether we commit it or roll it back as long as we close it
+	return l.tx.Rollback()
+}
+
+func GetLatestLedgerSequence(db DB) (uint32, error) {
+	tx, err := db.NewLedgerEntryReaderTx()
+	if err != nil {
+		return 0, err
+	}
+	defer tx.Done()
```
tsachiherman, line 350: Error return value of `tx.Done` is not checked (errcheck)

cmd/soroban-rpc/internal/db/db.go
```diff
@@ -353,3 +261,109 @@ func runMigrations(db *sql.DB, dialect string) error {
...
+	return tx.GetLatestLedgerSequence()
+}
+
+func GetLedgerEntryAndLatestLedgerSequence(db DB, key xdr.LedgerKey) (bool, xdr.LedgerEntry, uint32, error) {
+	tx, err := db.NewLedgerEntryReaderTx()
+	if err != nil {
+		return false, xdr.LedgerEntry{}, 0, err
+	}
+	defer tx.Done()
```
tsachiherman, line 359: Error return value of `tx.Done` is not checked (errcheck)

tsachiherman (commented):
few comments so far; I'll keep reviewing..

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: >  I don't know if this usage would work reliably, since GO might decide to GC the string ( since there are no more references to it ).

It works reliably because we are calling `C.CString()` on the Go side to allocate it. That's the idiomatic way to pass a string to C from Go. In case you haven't, I suggest you take a look at how to work with CGO.

>  think that the better solution here is to avoid mixing allocators altogether  ...

We are not mixing allocators, due to what I mentioned above.

Using a buffer+length is possible (CGO provides `C.CBytes` and `C.GoBytes`) but it's inconvenient due to bookeeping the length (see my answers to @leighmcculloch 's comments). If base64-encoding ends up being a bottleneck I will be happy to optimize that.

cmd/soroban-rpc/internal/db/db.go
2opremio: You are right but this requires changing the database interface to return XDR-encoded ledger entries and passing buffers+lengths between Go and Rust.

I would like to only do that when/if needed due to performance reasons.

2opremio: For future reference, see https://stackoverflow.com/questions/64574831/how-to-return-struct-from-cgo

2opremio: Happy to create a ticket once we merge this to do this in the future.

.github/workflows/e2e.yml
```diff
@@ -14,7 +14,8 @@ jobs:
         scenario-filter: ["^TestDappDevelop$/^.*$"]
     runs-on: ubuntu-latest
     env:
-      SYSTEM_TEST_DOCKER_REPO_TAG: "stellar/system-test:1.0.2"
+      # TODO: restore stellar/systen-test once the tests are fixed
```
paulbellamy, line 17:
```suggestion
      # TODO: restore stellar/system-test once the tests are fixed
```

paulbellamy (approved):

.github/workflows/soroban-rpc.yml
```diff
@@ -25,19 +29,68 @@ jobs:
       - uses: ./.github/actions/setup-go
         with:
           go-version: ${{ matrix.go }}
-      - run: go test -race -cover -timeout 25m -v ./cmd/soroban-rpc/...
+      - run: |
+          make build-libpreflight
```
tsachiherman, line 33: please add a step to the golangci-lint.yml as well, before running the linter.
Otherwise, the linter would not be able to compile the code and generate some nonsense errors.

2opremio: I don't think the linter needs the binary library. What makes you think that? (I didn't see any errors in the linting output)

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -22,7 +25,17 @@ var (
...
 )
 
-// createInvokeHostOperation creates a dummy InvokeHostFunctionOp. In this case by installing a contract code.
+func getHelloWorldContract(t *testing.T) []byte {
+	_, filename, _, _ := runtime.Caller(0)
+	testDirName := path.Dir(filename)
+	contractFile := path.Join(testDirName, "../../../../target/wasm32-unknown-unknown/test-wasms/test_hello_world.wasm")
+	ret, err := os.ReadFile(contractFile)
+	if err != nil {
+		t.Fatalf("test_hello_world.wasm can't be found (%v) please run `make build-test-wasms` at the project root directory", err)
```
tsachiherman, line 34: not always. Yes, there was an error reading the file, and "file not found" is very likely the reason.
could you change the text to "unable to read test_hello_world.wasm due to %v" ? Suggesting to run `build-test-wasms` is probably a good solution in the case that the file not found.

.github/workflows/soroban-rpc.yml
tsachiherman: it printout "could not import C (cgo preprocessing failed) (typecheck)". I'm hoping that if it would be able to compile it, then it would stop complaining about that.

2opremio: Ah, I hadn't seen that.

2opremio: That fixed it

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: This is tracked at https://github.com/stellar/soroban-tools/issues/367

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: Part of this is tracked at https://github.com/stellar/soroban-tools/issues/367

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: This is tracked at https://github.com/stellar/soroban-tools/issues/367

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn free_preflight_result(result: *mut CPreflightResult) {
+    if result.is_null() {
+        return;
+    }
+    unsafe {
+        if !(*result).error.is_null() {
+            let _ = CString::from_raw((*result).error);
```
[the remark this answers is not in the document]
2opremio: AFAIU, It brings the string back from `raw` to being memory managed by Rust, resulting in it being deallocated at the end of the call

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+
+struct CSnapshotSource {
+    handle: libc::uintptr_t,
+}
+
+impl SnapshotSource for CSnapshotSource {
+    fn get(&self, key: &LedgerKey) -> Result<LedgerEntry, HostError> {
+        let key_xdr = key
+            .to_xdr_base64()
+            .map_err(|_| ScStatus::UnknownError(Xdr))?;
```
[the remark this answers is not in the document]
2opremio: Yep, I'll be happy to update the code once that PR is merged.

cmd/soroban-rpc/lib/preflight/src/lib.rs
graydon: @leighmcculloch the `Rc<>`s here may be multiply owned. They're to enable the use of `MeteredMap` which treats its elements as shallow-clonable ("like" `RawVal` but, in this case, not `RawVal`) rather than deep-clonable. Recall that `MeteredMap` at this point is an immutable type that duplicates itself entirely whenever modified, and that this is at least partly necessary for both gas metering and transactional semantics when aborting a function -- we should roll back to a previous storage map. So copying storage maps should be "cheap" ish -- O(N) pointer-copies and Rc bumps -- rather than deep copies. Thus the `Rc<>` values are often likely to be multiple references to truly shared data, and a `try_unwrap` is might work but might fail, in which case you need to clone-out.

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: > Is it possible to use CString with Go byte slices converted to a string? The best of both worlds: we don't have to manage length, we don't have to do base64 enc/dec.

In general, you can’t. The problem is that CStrings are ended in zero and thus, you cannot represent byte slices containing zeros as a CString.

In particular, you cannot represent binary-marshaled XDR as a CString.

PR #270, opened 2022-11-29
cmd/soroban-rpc: getEvents method (Horizon-backed)

paulbellamy:
### What

Fixes https://github.com/stellar/go/issues/4554
Replaces https://github.com/stellar/go/pull/4694

### Why

Moved the work from the go repo to the soroban-tools repo

### Todo

- [x] Double-check the cursor pagination works as expected with horizon. The unit tests are mocks, but the mocks might be wrong.
- [x] Needs an update to the go repo dependency once https://github.com/stellar/go/pull/4704 is finished & merged.


cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -0,0 +1,475 @@
...
+	return nil
+}
+
+func (t TopicFilter) Matches(event []xdr.ScVal) bool {
+	for _, segmentFilter := range t {
+		if segmentFilter.wildcard != nil {
+			switch *segmentFilter.wildcard {
+			case "*":
+				// one-segment wildcard
+				if len(event) == 0 {
```
sreuland, line 180: just for consideration, i think could move this guard check on lines 180 - 183 up to top of loop, after 175, and then delete the duplicate check from lines 191-193.

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -0,0 +1,475 @@
...
+		return fmt.Errorf("invalid event id %s", input)
+	}
+
+	// Parse the first part (toid)
+	idInt, err := strconv.ParseInt(parts[0], 10, 64)
+	if err != nil {
+		return errors.Wrapf(err, "invalid event id %s", input)
+	}
+	parsed := toid.Parse(idInt)
+	id.ID = &parsed
```
sreuland, line 270: super minor, is it worthwhile to only update the attributes of `id` instance at the end when err is nil, otherwise could return err here but have partially populated `id` instance with some attributes set.

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -0,0 +1,475 @@
...
+	cursor := EventID{
+		ID:         toid.New(request.StartLedger, int32(0), 0),
+		EventOrder: 0,
+	}
+	if request.Pagination != nil && request.Pagination.Cursor != "" {
+		if err := cursor.Parse(request.Pagination.Cursor); err != nil {
+			return nil, errors.Wrap(err, "invalid pagination cursor")
+		}
+	}
+	err := a.ForEachTransaction(cursor.ID, finish, func(transaction horizon.Transaction) error {
```
sreuland, line 303: just for readability, maybe pull up the inlined function to be declared as package private func instead and reference to it here?

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -0,0 +1,475 @@
...
+		if topicFilter.Matches(v0.Topics) {
+			return true
+		}
+	}
+	return false
+}
+
+type TopicFilter []SegmentFilter
+
+func (t *TopicFilter) Valid() error {
```
sreuland, line 164: should this also inspect the filter for the wildcard validity, so could remove the repetitive checks on filter validity and ensuing `panic` in `TopicFilter.Matches` ?

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -0,0 +1,475 @@
...
+		for operationIndex, operationEvents := range v3.Events {
+			if int32(operationIndex) < operationCursor {
+				continue
+			}
+			for eventIndex, event := range operationEvents.Events {
+				if int32(eventIndex) < eventCursor {
+					continue
+				}
+				if request.Matches(event) {
+					v0 := event.Body.MustV0()
```
sreuland, line 346: does this result in `panic` .. is it acceptable to allow this potentially crash out the overall service or checking for type/null instead and return err?

sreuland (approved):
I left a couple comments for consideration, but nothing critical.

cmd/soroban-rpc/internal/methods/get_events.go
paulbellamy: That's trickier because it captures a bunch of stuff in the lambda (cursor & pagingToken). Instead I refactored out the event -> eventInfo conversion, to make the inner func more readable.

PR #274, opened 2022-11-30
Add a more generic getLedgerEntry rpc method

paulbellamy:
### What

Add a new rpc `getLedgerEntry` method to supersede `getContractData`.

### Why

Because the contract storage has changed with the new xdr, `getContractData` can no longer be used by soroban-cli to fetch the contract code. We could have added the `getContractCode` endpoint from the original spec. But, it's come up several times that actually `getLedgerEntry` is both lower-level and more useful.

The plan will be to deprecate `getContractData`, and remove it in favour of `getLedgerEntry` in the next release.

### Todo

- [x] Rebase this onto main
- [ ] test this
- [x] use the new method from soroban-cli to fetch the contract code
  - Will need two requests, one to get the code address, and a second to get the actual code.
- [x] update the rust dependencies to include the new xdr and conversions
- [x] update the go dependency to include the new xdr
- [ ] Once merged: Update the [api doc](https://docs.google.com/document/d/1TZUDgo_3zPz7TiPMMHVW_mtogjLyPL0plvzGMsxSz6A/edit)

### DIFF
--- a/cmd/soroban-rpc/internal/methods/get_ledger_entry.go
+++ b/cmd/soroban-rpc/internal/methods/get_ledger_entry.go
@@ -23,7 +23,7 @@ type GetLedgerEntryResponse struct {
 	LatestLedger       int64  `json:"latestLedger,string"`
 }
 
-// NewGetLedgerEntryHandler returns a json rpc handler to retrieve a contract data ledger entry from stellar cre
+// NewGetLedgerEntryHandler returns a json rpc handler to retrieve the specified ledger entry from stellar core
 func NewGetLedgerEntryHandler(logger *log.Entry, coreClient *stellarcore.Client) jrpc2.Handler {
 	return handler.New(func(ctx context.Context, request GetLedgerEntryRequest) (GetLedgerEntryResponse, error) {
 		var key xdr.LedgerKey
--- a/cmd/soroban-rpc/internal/methods/get_ledger_entry.go
+++ b/cmd/soroban-rpc/internal/methods/get_ledger_entry.go
@@ -1,0 +1,83 @@
+package methods
+
+import (
+	"context"
+
+	"github.com/creachadair/jrpc2"
+	"github.com/creachadair/jrpc2/code"
+	"github.com/creachadair/jrpc2/handler"
+
+	"github.com/stellar/go/clients/stellarcore"
+	proto "github.com/stellar/go/protocols/stellarcore"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/xdr"
+)
+
+type GetLedgerEntryRequest struct {
+	Key string `json:"key"`
+}
+
+type GetLedgerEntryResponse struct {
+	XDR                string `json:"xdr"`
+	LastModifiedLedger int64  `json:"lastModifiedLedgerSeq,string"`
+	LatestLedger       int64  `json:"latestLedger,string"`
+}
+
+// NewGetLedgerEntryHandler returns a json rpc handler to retrieve a contract data ledger entry from stellar cre
+func NewGetLedgerEntryHandler(logger *log.Entry, coreClient *stellarcore.Client) jrpc2.Handler {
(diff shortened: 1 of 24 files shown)


tomerweller:
Awesome. 
I added a warning in the API doc next to `getContractData`
Is this PR expected to close [this issue](https://github.com/stellar/go/issues/4695) as well?

paulbellamy:
Yes, that is the issue, thanks.

paulbellamy:
Note: Because this updates the rust xdr, but not the go xdr, this will temporarily break the tools repo.

paulbellamy:
FYI: The e2e test on this will fail until quickstart is updated.

cmd/soroban-rpc/internal/test/get_ledger_entry_test.go
```diff
@@ -0,0 +1,150 @@
+package test
```
sreuland, line 1: just wanted to mention, I added this coverage scope as a system test(dapp e2e) also:
https://github.com/stellar/system-test/issues/3, perhaps at some point can consolidate integration tests in there.

cmd/soroban-rpc/internal/methods/get_ledger_entry.go
```diff
@@ -0,0 +1,83 @@
...
+	Key string `json:"key"`
+}
+
+type GetLedgerEntryResponse struct {
+	XDR                string `json:"xdr"`
+	LastModifiedLedger int64  `json:"lastModifiedLedgerSeq,string"`
+	LatestLedger       int64  `json:"latestLedger,string"`
+}
+
+// NewGetLedgerEntryHandler returns a json rpc handler to retrieve a contract data ledger entry from stellar cre
```
sreuland, line 26: nit, comment can simplify to something like 'NewGetLedgerEntryHandler returns a json rpc handler to retrieve the specified ledger entry from stellar core'

cmd/soroban-rpc/internal/test/docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
     # Note: Please keep the image pinned to an immutable tag matching the Captive Core version.
     #       This avoid implicit updates which break compatibility between
     #       the Core container and captive core.
-    image: ${CORE_IMAGE:-sreuland/stellar-core:19.4.1-1097.4e813f20e.focal-soroban}
+    image: ${CORE_IMAGE:-2opremio/stellar-core:19.5.1-1131.f55f88376.focal-soroban}
```
sreuland, line 18: moving target, new core build as of today:

`sreuland/stellar-core:19.5.1-1137.b3a6bc281.focal-soroban`

cmd/soroban-rpc/internal/test/transaction_test.go
```diff
@@ -115,7 +115,7 @@ func TestSendTransactionSucceedsWithResults(t *testing.T) {
 	assert.Nil(t, response.Error)
 	assert.Equal(t,
 		[]methods.SCVal{
-			{XDR: "AAAABAAAAAEAAAAEAAAAIJL0uHwiQNNaPB0brHQ1Wg+WexdpZMjN1H4UDoxE5pQz"},
+			{XDR: "AAAABAAAAAEAAAAGAAAAIOqfy4GuVKKfazvyk4R9P9fpo2n9HICsr+xqvVcTF+DC"},
```
sreuland, line 118: not suggesting change, gets it done, but this pattern could become difficult to triage later, i.e. without a comment saying what the unmarshal'd struct is or maybe just creating the expected xdr struct and then marshaling it to this bas64 string would provide more visibility in the test of the expected value as a baseline for where it went wrong.

sreuland (approved):

cmd/soroban-rpc/internal/test/transaction_test.go
paulbellamy: Yeah. Would be way better to parse it and compare.

sreuland:
looks good, I didn't dig in too deep on the rust though.

PR #275, opened 2022-11-30
feat: Config command

willemneal:
The `soroban config` subcommand, [as proposed](https://docs.google.com/document/d/16lJhyMU6cqRzNtMclr9yXd978MJgbZ0El9UhxbTz_Ts/edit#). Manage Soroban configs (identities, networks, etc)

The `config` subcommand expands on one of the "Other Ideas" in [the ongoing `profile` effort][profile]. It takes its lead from git's excellent and well-known config management system. We suggest following git's config system closely, although we prefer standard [toml](https://toml.io/), rather than git's special format. Envisioned similarities:

* A home-directory `~/.soroban/config.toml` stores user defaults, and can be updated with `soroban config --global`
* A workspace `.soroban/config.toml` can store project-specific settings, which would override any `~/.soroban/config.toml` settings of the same name (unless it is a setting that can be merged).

We suggest subsuming the current `profile` effort into this new `config` system. Further, as [@leighmcculloch suggests](https://github.com/stellar/soroban-tools/pull/254#issuecomment-1318065805), we suspect developers will build more precise mental models if `identity` and `network` are two separate config groups, rather than being grouped together into `profile`.

Altogether, here are the `config` subcommands we suggest starting with:

    USAGE:
        soroban config <SUBCOMMAND>

    SUBCOMMANDS:
        identity    Configure identities (accounts/wallets/keypairs)
        network     Configure named networks

While perhaps long to type, we suggest keeping these nested under `config`. This clarifies for developers that these commands only manipulate local configuration files, rather than make RPC calls or manipulate on-chain resources. It also leaves the first-level subcommands open for other uses; see [soroban-cli-network] below.

### `soroban config identity`

`soroban config identity` will have its own subcommands:

    USAGE:
        soroban config identity <SUBCOMMAND>

    SUBCOMMANDS:
        add        Add a new identity (keypair) to the config file
        rm         Remove an identity from the config file
        default    Set a default identity

`add` takes one positional argument, the name of the identity, and requires either the `--secret-key` or the `--seed-phrase` flag. It then prompts the user to enter either a secret key or seed phrase using an interactive password-style prompt, potentially using [the rpassword crate](https://crates.io/crates/rpassword). The secret key or seed phrase can then be used to generate the public key to store in the config file.

Adding a new identity to the workspace config:

    soroban config identity add ahalabs --secret-key

This will save the identity to a new toml file within the workspace under `.soroban/identities/*`, using [toml tables](https://toml.io/en/v1.0.0#table)We could also consider using a different file format, such as [jsonc](https://code.visualstudio.com/docs/languages/json#_json-with-comments)).

Assuming toml format, the previous command would result in a file saved to `.soroban/identities/ahalabs.toml` with contents:

    secret_key = "S..."

The public key here is not included, because it can be generated on the fly when commands need it. Including it in the config could result in errors, if people hand-edit these files.

To add a similar entry to the user's home `~/.soroban/config.toml`:

    soroban config identity add --global chadoh --seed-phrase

Resulting in:

    seed_phrase = "correct horse battery staple..."

This seed phrase can be used to generate secret public keys as needed.

These could then be used in other commands with either the `--identity` flag or its alias `--as`:

    soroban contract invoke --id ... --fn ... --identity ahalabs
    soroban contract invoke --id ... --fn ... --as chadoh

Importantly, this can also put an end to people putting secret keys in command line arguments. `.bash_history` and other shell history files are readable by other system processes, making the inclusion of keys in arguments a dangerous practice.

It would also be good to support adding identities that work with Ledger hardware wallets and the macOS keystore:

    soroban config identity add --global chadoh --ledger
    soroban config identity add --global chadoh --mac-os-keychain

### `soroban config network`

Adding configurations for networks should follow similar logic to adding identity configs.

    USAGE:
        soroban config network <SUBCOMMAND>

    SUBCOMMANDS:
        add        Add a new named network to interact with
        rm         Remove a named network
        default    Set a default identity

Example:

    soroban config network add --global localnet  https://localhost:8000/soroban/rpc --network-passphrase "Local Sandbox Stellar Network ; September 2022"

The network security passphrase would be saved to the toml file along with the rest of the settings. The network passphrase should never change for a network, it is constant. Any change is probably malicious; a useful error needs to be thrown if making calls to a network with an incorrect passphrase. There doesn't need to be a convenient way to update a network passphrase; in the rare event that a network passphrase needs to change, people can simply remove and re-add the network, as with git remotes.

These could then be used together with identities to cover all use-cases currently considered by [the ongoing `profile` effort][profile].

    soroban invoke --contract ... --fn ... \
      --network localnet \
      --as chadoh

Note that these commands only modify local config files, and do not help users run or connect to these networks. It may be worth displaying a warning to this effect when running `soroban config network` commands, or at least mention it prominently in the help info. Something like "If you want to run network nodes locally, install the [soroban-cli-network] plugin."

### Default network and identity

See comment below about removing this.

Both the `config identity` and `config network` commands have a `default` subcommand that allows setting a default.

    soroban config identity default ahalabs
    soroban config network default --global localnet

The default can be stored as a top-level config in the toml file. The `add` subcommands should also take a `--default` boolean option, which allows adding a new identity/network and setting it as the default at the same time.

    soroban config identity add --default --global chadoh --secret-key

### `soroban config alias`

Give developers easily-typeable alternatives to contract IDs.

    USAGE:
        soroban config alias <SUBCOMMAND>

    SUBCOMMANDS:
        add        Add a new contract alias
        rm         Remove an alias

`add` also accepts a `--global` flag to add the alias to the config in the user's home directory. It takes two positional arguments, the name and value of the config to add.

    soroban alias add --global incrementalist d17953b1...

Resulting in `~/.soroban/config.toml` with:

    [alias]
    incrementalist = "d17953b1..."

`rm` accepts just the name of the alias to remove.

These aliases can then be referenced by other commands.


cmd/soroban-cli/src/config/network/mod.rs
```diff
@@ -0,0 +1,45 @@
+use clap::Parser;
+
+pub mod add;
+pub mod rm;
+pub mod default;
+pub mod ls;
+
+#[derive(Debug, Parser)]
+pub enum Cmd {
+    /// Add a new netowrk
```
chadoh, line 10:
```suggestion
    /// Add a new network
```

willemneal:
@leighmcculloch I changed my mind about a default network since we currently signal that we are using a network by using either --network or the network settings.  So it doesn't make much since.  Though we could add a `--default-network` arg.

Similarly for identity I felt that having a default identity might become a footgun as you could end up more easily sign a transaction with an unintended key.

willemneal:
Currently failing windows with multiple network configs on windows.  I don't have a windows machine does anyone have any suggestions for how to troubleshoot this?

How strongly do we want to support using powershell/windows?

tsachiherman:
@willemneal could you try using a virtual machine for that ?

willemneal:
@tsachiherman It seems to be working now. @leighmcculloch @paulbellamy

.gitignore
```diff
@@ -1,2 +1,8 @@
 target/
-.soroban/
+./.soroban/
+!test.toml
+./cmd/soroban-cli/.soroban/*
+./.soroban/**/*
+.soroban/identities/
+
+cmd/soroban-cli/.soroban/events.json
```
paulbellamy, line 8: Why do we need all these things in the `.gitignore`? Is there something `.soroban/` would miss, or that we want to include in there?

paulbellamy: My concern is that, if using soroban with `.gitignore` is this difficult, no one will understand how to do it.

cmd/soroban-cli/.soroban/identities/test_id.toml
```diff
@@ -0,0 +1 @@
+secret_key = "SBGWSG6BTNCKCOB3DIFBGCVMUPQFYPA2G4O34RMTB343OYPXU5DJDVMN"
```
paulbellamy, line 1: This is probably fine for now, but I'd say tests should generate a random strkey with `rs-stellar-strkey`, and put it into a toml file in a tmpdir.

cmd/soroban-cli/tests/it/config.rs
```diff
@@ -0,0 +1,156 @@
...
+        .arg("network")
+        .arg("ls")
+        .assert()
+        .stdout("local2\n");
+
+    let sub_dir = dir.join("sub_directory");
+    fs::create_dir(&sub_dir).unwrap();
+    add_network(&sub_dir, "local3\n");
+
+    //TODO Investigate why there is an extra newline characeter
```
tsachiherman, line 139: Please investigate or remove comment.

cmd/soroban-cli/src/config/identity/generate.rs
```diff
@@ -0,0 +1,33 @@
...
+            "Coming soon! This will generate a new identity named {} and a corresponding seed phrase from seed {:?}",
+            self.name,
+            self.seed.as_deref().unwrap_or("random")
+        );
+        if false {
+            return Err(Error::Unknown);
+        }
+        Ok(())
+    }
+}
```
paulbellamy, line 33: I'd vote for leaving out stuff that doesn't actually exist yet. The API may change when we implement it.

paulbellamy: For example, maybe we'd prefer to do `soroban config identity add mike --seed 5` instead.

cmd/soroban-cli/.soroban/identities/test_id.toml
tsachiherman, line 1: hmm. Do we really want to have a hard-coded secret key ? if the value isn't important, please dynamically generate it for testing.

tsachiherman (dismissed):
Could you please take a look on the comments I've left ?

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    let data = fs::read(&path).map_err(|_| Error::SecretFileRead { path })?;
+    toml::from_slice::<Secret>(&data).map_err(|_| Error::Deserialization)
+}
+
+pub fn read_network(name: &str) -> Result<Network, Error> {
+    // 1. check workspace config files for `name`
+    let local_network = Args { global: false }.network_path(name);
+    // 2. use if found, else, check global config files for `name`
+    let path = local_network.or_else(|_| Args { global: true }.identity_path(name))?;
```
paulbellamy, line 156: Should be this, no?
```suggestion
    let path = local_network.or_else(|_| Args { global: true }.network_path(name))?;
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    println!("Found network \"{name}\" at {}", path.display());
```
paulbellamy, line 158: Should we maybe use `eprintln!`? or remove the printlns to avoid polluting the output of other commands?

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+    #[clap(flatten)]
+    pub network: network::Args,
+
+    #[clap(flatten)]
+    pub ledger_file: ledger_file::Args,
+
+    #[clap(long, alias = "as")]
+    /// Use specified identity to sign transaction
+    pub identity: Option<String>,
+}
```
paulbellamy, line 72: Why are these args needed here? Shouldn't they be handled by the `identity` and `network` subcommands?

cmd/soroban-cli/src/contract/inspect.rs
```diff
@@ -1,102 +1,24 @@
...
-                    ScSpecEntry::UdtErrorEnumV0(udt) => {
-                        println!(" • Error: {udt:?}");
-                    }
-                }
-            }
-        } else {
-            println!("Contract Spec: None");
-        }
+        println!("File: {}", self.wasm.wasm.to_string_lossy());
+        print!("{}", self.wasm.parse()?);
```
paulbellamy, line 21: De we use `wasm.parse()` anywhere else? If not, we should keep that logic here, closer to where it is used.

cmd/soroban-cli/src/contract/install.rs
```diff
@@ -96,42 +51,28 @@ impl Cmd {
...
         Ok(hex::encode(wasm_hash))
     }
 
     async fn run_against_rpc_server(&self, contract: Vec<u8>) -> Result<String, Error> {
-        let client = Client::new(self.rpc_url.as_ref().unwrap());
-        let key = utils::parse_secret_key(self.secret_key.as_ref().unwrap())
-            .map_err(|_| Error::CannotParseSecretKey)?;
+        let network = self.config.get_network()?;
+        let client = Client::new(&network.rpc_url);
+        let key = self.config.key_pair()?;
```
paulbellamy, line 75: May not have a keypair if we're using, say... a hw signer like a ledger. We will only have the public key. So maybe the methods would be `public_key()`, and `sign(msg)`?

paulbellamy: But that's a way bigger refactor, so let's leave it for now.

cmd/soroban-cli/tests/it/config.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    let _ = Sandbox::new_cmd("config")
+        .current_dir(&dir)
+        .arg("network")
+        .arg("add")
+        .arg("--rpc-url")
+        .arg("https://127.0.0.1")
+        .arg("--network-passphrase")
+        .arg("Local Sandbox Stellar Network ; September 2022")
+        .arg("local")
```
paulbellamy, line 16: Shouldn't the network name be first? `network add local ...` or I guess it doesn't matter?

cmd/soroban-cli/.soroban/identities/test_id.toml
paulbellamy: Oh.. We need this for the *other* tests. And there's no default `identity` so they need a key to run.. 🤔 Still feels like a bit of a kludge...

paulbellamy (changes requested):
Thoughts:
- A lot of usage of `clap(flatten)`, which can make it difficult to track which commands take which args... Overall, probably worth it as it keeps things DRY-er in this case.
- Cleaning up all the ledger loading everywhere is great!

Overall, almost there. A couple questions & fixes.

cmd/soroban-cli/src/config/mod.rs
willemneal: This case is when using the identity that was created by the identity subcommands.  e.g. 

```bash
soroban contract invoke .. --identity mike
```

However, you can still supply the raw key value instead.

cmd/soroban-cli/src/contract/inspect.rs
willemneal: My thought here was that eventually we would also make a library so that this parse could be used outside of the CLI.

cmd/soroban-cli/tests/it/config.rs
willemneal: Yeah it's a positional arg so the flag args can be before or after.  I'll swap to make this clearer.

cmd/soroban-cli/tests/it/config.rs
willemneal: Removed! It's not an issue.

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl Args {
+    pub fn key_pair(&self) -> Result<ed25519_dalek::Keypair, Error> {
+        // TODO remove unwrap and provide error
```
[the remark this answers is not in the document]
willemneal: Nope! Removed.

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+            secret::Secret::SecretKey {
+                secret_key: secret_key.clone(),
+            }
+        } else {
+            return Err(Error::CannotParseSecretKey);
+        };
+        let str_key = match &key {
+            secret::Secret::SecretKey { secret_key } => secret_key,
+            secret::Secret::SeedPhrase { seed_phrase: _ } => {
+                todo!("Still need to implement seedphrase")
```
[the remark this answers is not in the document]
willemneal: Done!

cmd/soroban-cli/.soroban/identities/test_id.toml
willemneal: Just removed it in favor of each test getting its own temporary directory. And added a tests to write this dynamically!

.gitignore
willemneal: Not sure why it evolved that way.  Its now just `.soroban` and works as expected.

cmd/soroban-cli/.soroban/identities/test_id.toml
willemneal: See above commend to @tsachiherman, but now the tests each have their own temp directory and identities are written when a test needs one.

cmd/soroban-cli/src/config/locator.rs
willemneal: Yep good catch!

cmd/soroban-cli/src/contract/install.rs
willemneal: Yep.  The same will be true for mac keychain.  Will address this in a future PR, perhaps should be a separate crate.

cmd/soroban-cli/src/config/mod.rs
paulbellamy: Oh, got it.

cmd/soroban-cli/tests/it/config.rs
paulbellamy: All good, just double checking.

paulbellamy (approved):

Issue #277, opened 2022-12-01
CLI: unable to deploy wrapped token contract


overcat:
### What version are you using?
soroban-cli 0.2.1 (2a065eeebc94f6c0b4fcb03a7963e139ff2a583d)
soroban-env-interface-version: 23

### What did you do?
```
soroban token wrap --asset HELLO:GCKGYECWTNN2QTEKS2D5EUWHTQZCEBKURYOVRQS7LU56TBSKNRLK7PKN \
    --secret-key SAMFEBQ467IEBCTXUPZBC6ZDLUNW7FGSCROVFCVNIWZ4HXHQBYGD5Y2C \
    --rpc-url https://horizon-futurenet.stellar.cash:443/soroban/rpc \
    --network-passphrase 'Test SDF Future Network ; October 2022'
```

### What did you expect to see?
The contract can be deployed and returns the contract ID.

### What did you see instead?

```
error: transaction submission failed
```
see: https://horizon-futurenet.stellar.org/transactions/23274a7dfac1bf17db8c2ff7e4b72135e3a7a309ea9573547b950ab6159c4e48

### Note

I think the `footpoint` is not set correctly.

PR #278, opened 2022-12-01
soroban-cli: fix that the footpoint was not set correctly when deploying the wrapped token contract

overcat:
### What
see #277 
Fix that the footpoint was not set correctly when deploying the wrapped token contract.
### Why

see #277 
### Known limitations

Lack of test cases


silence48 (approved):
I tested it today, it works flawlessly now.

leighmcculloch:
Thanks @overcat !!!

silence48 (approved):

leighmcculloch (approved):

PR #279, opened 2022-12-01
Updating xdr for soroban-rpc

paulbellamy:
Update the soroban-rpc xdr, on top of get-ledger-entry branch


paulbellamy:
Ok, hilariously, I think the remaining e2e tests need an updated quickstart docker image to pass.

PR #282, opened 2022-12-02
Add index of the parsed argument in the error message

vinamogit:
### What

Add index of the parsed argument in the error message

### Why

When invoking a contract with multiple arguments, especially if the types accept the same kind of values.

### Known limitations

The index is not only '--arg' index but the index of all the invoke command options

### DIFF
--- a/cmd/soroban-cli/src/invoke.rs
+++ b/cmd/soroban-cli/src/invoke.rs
@@ -105,8 +105,8 @@ pub struct Cmd {
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error("parsing argument {arg}: {error}")]
-    CannotParseArg { arg: String, error: StrValError },
+    #[error("parsing argument '{arg}' at index {index}: {error}")]
+    CannotParseArg { arg: String, error: StrValError, index: usize },
     #[error("parsing XDR arg {arg}: {error}")]
     CannotParseXdrArg { arg: String, error: XdrError },
     #[error("cannot add contract to ledger entries: {0}")]
@@ -229,6 +229,7 @@ impl Cmd {
                     strval::from_string(s, &input.type_).map_err(|e| Error::CannotParseArg {
                         arg: s.clone(),
                         error: e,
+                        index: arg.0,
                     })
                 }
             })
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/invoke.rs
```diff
@@ -104,8 +104,8 @@ pub struct Cmd {
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error("parsing argument {arg}: {error}")]
-    CannotParseArg { arg: String, error: StrValError },
+    #[error("parsing argument '{arg}' at index {index}: {error}")]
+    CannotParseArg { arg: String, error: StrValError, index: usize },
```
leighmcculloch, line 108: I think this would be really cool to add. With the current implementation in this PR it doesn't print the index of the argument in the arg list to be given to the contract, it prints the index in the full argument list on the command line. This is somewhat confusing.

For example:
```
❯ soroban \
    invoke --id 1 \
    --wasm target/wasm32-unknown-unknown/release/soroban_hello_world_contract.wasm \
    --fn hello \
    --arg 'laksdjfklajsdlfkjasd'
error: parsing argument 'laksdjfklajsdlfkjasd' at index 8: parse error: value is not parseable to Symbol
```

Arguments:
- `0`: `invoke`
- `1`: `--id`
- `2`: `1`
- `3`: `--wasm`
- `4`: `target/wasm32-unknown-unknown/release/soroban_hello_world_contract.wasm`
- `5`: `--fn`
- `6`: `hello`
- `7`: `--arg`
- `8`: `'laksdjfklajsdlfkjasd'`

I think I would expect it to say argument 0 or 1, depending if the index is intended to start at zero or one.

vinamogit: Not sure what is the best way to count the arg. Here is an attempt.

willemneal:
I've begun to wonder if makes sense that the args clap type is a tuples of name and value. The xdr does contain the parameter names.  I'll try that today.

tsachiherman:
@paulbellamy could you please take a look and figure out the next steps needed to have this PR approved ?

paulbellamy:
With the new named-arg style `invoke` arg parsing, I don't think this is needed anymore. For example, you now get:
```
$ soroban contract invoke --id "token...id" --fn balance -- --id foo
error: parsing argument id: value is not parseable to Some(
    Address,
)
```

PR #284, opened 2022-12-02
feat: add `args-file` option to `invoke`

willemneal:
### What

add `--args-file` option so that more complex arguments can be saved in files instead of CLI strings.

### Why

Complex JSON structures are hard to write as part of a command line arg.

### Known limitations

[TODO or N/A]


cmd/soroban-cli/src/invoke.rs
```diff
@@ -49,6 +50,9 @@ pub struct Cmd {
     /// Argument to pass to the function (base64-encoded xdr)
     #[clap(long = "arg-xdr", value_name = "arg-xdr", multiple = true)]
     args_xdr: Vec<String>,
+    /// File containing JSON Argument to pass to the function
+    #[clap(long, multiple = true)]
```
leighmcculloch, line 54: Are args files XDR files? That wasn't obvious to me on first read, I thought it was JSON still. Maybe we should make the option `--arg-xdr-file` to make it clear.
```suggestion
    #[clap(long, value_name = "arg-xdr-file", multiple = true)]
```

Separately we could have `arg-xdr` for the JSON form?

cmd/soroban-cli/src/invoke.rs
```diff
@@ -455,6 +459,15 @@ impl Cmd {
     }
 }
 
+impl Cmd {
+    fn contract_id(&self) -> Result<[u8; 32], Error> {
+        utils::contract_id_from_str(&self.contract_id).map_err(|e| Error::CannotParseContractId {
+            contract_id: self.contract_id.clone(),
+            error: e,
+        })
```
leighmcculloch, line 467: Thanks ❤️

leighmcculloch (commented):
Nice. One suggestion.

cmd/soroban-cli/src/invoke.rs
silence48, line 54: the e2e test fails because it passes --arg-file and the command is not found.  so the test works, maybe change
line 54:
`#[clap(long,` multiple = true)]`
to
`#[clap(long = "args-file", value_name="args-file", multiple = true)]`

willemneal: clap converts snake_case to kebab-case, so the rename isn't necessary. The issue was I passed `--arg-file` instead of `--args-file`. After finally setting up the docker server for testing locally it seems to pass now.

willemneal: Yeah I meant for this one to be the same as `--args`, which is both json and raw strings right?  The one test in this repo used `--args=world`.  I'll go ahead and add a `args-xdr-file` to ensure that there is clarity.

silence48: Ah didn't notice that missing 's'.  It worked locally for me also!

leighmcculloch:
@willemneal I can't update your PR to prepare it for merging. Are you able to check the box that lets me push to the branch that this PR is attached to?

leighmcculloch (approved):

leighmcculloch:
Looks like there's a test failing.

chadoh:
@tyvdh (and others) do you think we still need this, given that #307 has been merged? Both might be nice, but ¯\\\_(ツ)\_/¯

tyvdh:
Hmm tough call. I do like the idea of being able to save common args to a file though. Think an account or contract hash. My vote is probably keep it?

chadoh:
Account or contract hash will be taken care of by #275 

- Accounts will be saved as identities:

       soroban config identity add tyvdh --secret-key

- Contract hashes (or anything else) will be saved as aliases:

       soroban config alias add my-contract G...

chadoh:
Search the [CLI Proposal](https://docs.google.com/document/d/16lJhyMU6cqRzNtMclr9yXd978MJgbZ0El9UhxbTz_Ts/edit#heading=h.4tvfd2wht2ga) for how we imagine making use of aliases in `invoke` and other commands.

I'm feeling like `--args-file` was a half solution dreamt up before the rest was thought through, and would rather see us complete the `config` command and try it together with the new dynamic args for a while, then decide if we still feel a need for `--args-file`. But I might be missing some already-obvious use-cases.

tyvdh:
Ah okay nice! Yeah so maybe we table it and see how it feels and revisit this later if the need arises. I'm fine with that outcome as well

leighmcculloch:
> I'm feeling like `--args-file` was a half solution dreamt up before the rest was thought through, and would rather see us complete the `config` command and try it together with the new dynamic args for a while

+1. I think we should leave this change and return to it if we still find a need down the line. Leaving it out means one less thing to maintain, and for users to grok, and it's unclear to me too if we really need it now.

PR #292, opened 2022-12-05
Upgrade crate-git-revision to 0.0.4

brson:
### What

Upgrade crate-git-revision to 0.0.4

### Why

Keep on the same version as other soroban crates. Once fully upgraded there won't be any duplicates in the lockfiles.

I think this is the last of the stellar repos that have this crate in their manifest.

### Known limitations

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -343,9 +343,9 @@ dependencies = [
 
 [[package]]
 name = "crate-git-revision"
-version = "0.0.2"
+version = "0.0.3"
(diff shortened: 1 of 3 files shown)


leighmcculloch (approved):

brson:
The failure is in the e2e-test worker and seems unlikely to be related to this patch, but I am still trying to understand it.

leighmcculloch:
The failure is unrelated. This repo recently got reorganized, and there a few problems in this repo that need resolving. I'm probably going to remove those tests, we can't rely on them.

Issue #290, opened 2022-12-05
soroban-cli: Re-fetch the account sequence when deploying to avoid race condition


paulbellamy:
### What problem does your feature solve?

Currently when we do deploy.rs, we re-use the account sequence (and add 1) for the two successive transactions. This could cause a race condition if there are other txns submitted simultaneously.

### What would you like to see?

We could re-fetch the account seq after submitting the first txn.

### What alternatives are there?

We could also just document that it doesn't work with simultaneous requests with the same private key. There is always a small race condition based on stellar network's design.

PR #294, opened 2022-12-06
Bugfixes for e2e tests in release/v0.3.0 branch

paulbellamy:
### What

- Fix bug in `soroban deploy --wasm-hash` sequence calculation introduced in https://github.com/stellar/soroban-tools/commit/eadacececc2b1071232db7ab045738d461908481
  - Incidentally fixes https://github.com/stellar/soroban-tools/issues/290
- Fix conflicting contract ids in e2e tests
  - The token contract and the newly installed one were using the same salt, resulting in a conflict between the tests.
- Use the temporary `stellar/quickstart:pr-389-soroban-dev` to run e2e tests so they work for now.
- Remove hardcoded `--salt` params from e2e tests altogether.
  - This means the cli will autogenerate a random salt for each one.
  - The e2e tests are then independent from each other and are less likely to conflict.
  - The e2e do not need a clean network to run anymore.

### Why

To fix the e2e tests for https://github.com/stellar/soroban-tools/pull/291

### Known limitations

[TODO or N/A]

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -81,7 +81,7 @@ jobs:
       - run: rustup target add wasm32-unknown-unknown
       # TODO: we should probably manage the container in the test itself
       - run: make build-test-wasms
-      - run: docker run -d -p 8000:8000 stellar/quickstart:pr-389-soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing --protocol-version 20
+      - run: docker run -d -p 8000:8000 stellar/quickstart:soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing --protocol-version 20
       - run: while ! [ "$(curl -s --fail localhost:8000 | jq '.history_latest_ledger')" -gt 0 ]; do echo waiting; sleep 1; done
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       - run: rustup target add wasm32-unknown-unknown
       # TODO: we should probably manage the container in the test itself
       - run: make build-test-wasms
-      - run: docker run -d -p 8000:8000 stellar/quickstart:soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing --protocol-version 20
+      - run: docker run -d -p 8000:8000 stellar/quickstart:pr-389-soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing --protocol-version 20
```
leighmcculloch, line 84: We don't need to merge this, as e2e tests don't block PRs. Thanks for checking that the updated image works though. 
```suggestion
      - run: docker run -d -p 8000:8000 stellar/quickstart:soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing --protocol-version 20
```

PR #298, opened 2022-12-06
soroban-cli: Add `events` subcommand for local and remote event viewing.

Shaptic:
#### 1: Storing Sandbox Events
This stores all events generated by a contract invocation into a local JSON file (much like `ledger.json`) that encodes the event and ledger information exactly in the way the RPC server would return it.

#### 2: Viewing Events
This is a continuation of #260, adding a new events sub-command with the following help text (click to expand):

<details>

```
Watch the network for contract events

USAGE:
    soroban events [OPTIONS]

OPTIONS:
    -s, --start-ledger <START_LEDGER>
            The first ledger sequence number in the range to pull events (required if not in sandbox
            mode)
            
            [default: 0]

    -e, --end-ledger <END_LEDGER>
            The last (and inclusive) ledger sequence number in the range to pull events (required if
            not in sandbox mode).
            https://developers.stellar.org/docs/encyclopedia/ledger-headers#ledger-sequence
            
            [default: 0]

        --output <OUTPUT>
            Output formatting options for event stream
            
            [default: pretty]
            [possible values: pretty, plain, json]

    -c, --count <COUNT>
            The maximum number of events to display (0 = all)
            
            [default: 10]

    -h, --help
            Print help information

OPTIONS (RPC):
        --rpc-url <RPC_URL>
            RPC server endpoint
            
            [env: SOROBAN_RPC_URL=]

OPTIONS (SANDBOX):
        --events-file <PATH>
            Local event store (likely generated by `invoke`) to pull events from
            
            [env: SOROBAN_EVENTS_FILE=]

FILTERS:
        --id <CONTRACT_IDS>...
            A set of (up to 5) contract IDs to filter events on. This parameter can be passed
            multiple times, e.g. `--id abc --id def`, or passed with multiple parameters, e.g. `--id
            abd def`.
            
            Though the specification supports multiple filter objects (i.e. combinations of type,
            IDs, and topics), only one set can be specified on the command-line today, though that
            set can have multiple IDs/topics.

        --topic <TOPIC_FILTERS>...
            A set of (up to 4) topic filters to filter event topics on. A single topic filter can
            contain 1-4 different segment filters, separated by commas, with an asterisk (*
            character) indicating a wildcard segment.
            
            For example, this is one topic filter with two segments:
            
            --topic "AAAABQAAAAdDT1VOVEVSAA==,*"
            
            This is two topic filters with one and two segments each:
            
            --topic "AAAABQAAAAdDT1VOVEVSAA==" --topic '*,*'
            
            Note that all of these topic filters are combined with the contract IDs into a single
            filter (i.e. combination of type, IDs, and topics).

        --type <EVENT_TYPE>
            Specifies which type of contract events to display
            
            [default: all]
            [possible values: all, contract, system]
```
</details>

The options should be self-explanatory and are based on the `getEvents` PR that was recently merged: #270.

  - specify a ledger range you want events for
  - (optionally) pick some contracts and/or topics to filter on
  - (optionally) pick an output format
  - specify either an RPC server or sandbox event file

It's output is something like this:

image

### Why
We should be able to retrieve smart contract events for certain ledger ranges: #61.

### Known limitations
This also needs to be tested end-to-end, now that there's an actual RPC implementation to test against:
 - [x] remove the json fixture mocking the `getEvents` RPC call
 - [x] test against RPC node

#### Out of scope
The following will come in a future PR:

 - support for `getEvents` when using `soroban serve`
 - support for the `--follow` flag with something like `notify`
 - misc. QoL improvements, e.g. showing the latest N events by default


cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// Local event store (likely generated by `invoke`) to pull events from
+    #[clap(
+        long,
+        parse(from_os_str),
+        value_name = "PATH",
+        env = "SOROBAN_EVENTS_FILE",
+        help_heading = HEADING_SANDBOX,
+        conflicts_with = "rpc-url",
+    )]
+    events_file: Option<std::path::PathBuf>,
```
paulbellamy, line 45: What should we do if the file doesn't exist? error? exit silently? It might not exist because the user put the wrong path. But it might not exist because there are no events yet.. Tricky... 🤔

paulbellamy: (or if the file is empty)

cmd/soroban-cli/src/invoke.rs
```diff
@@ -432,12 +442,20 @@ impl Cmd {
...
                 error: e,
             }
         })?;
+
+        snapshot::commit_events(&events.0, &ledger_info, &self.events_file).map_err(|e| {
+            Error::CannotCommitLedgerFile {
+                filepath: self.events_file.clone(),
+                error: e,
+            }
+        })?;
```
paulbellamy, line 457: This needs to create the file if it doesn't exist, otherwise it will error:
```
error: committing file .soroban/events.json: No such file or directory (os error 2)
```

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,304 @@
...
+                            || self.contract_ids.iter().any(|id| *id == evt.contract_id)
+                    })
+                    .filter(|evt| {
+                        // Like before, no topic means pass everything through.
+                        self.topics.is_empty()
+                            || self
+                                .topics
+                                .iter()
+                                // quadratic but both are <= 5 long
+                                .any(|t| evt.topic.iter().any(|t2| *t == *t2))
```
paulbellamy, line 179: I'm not sure this topic matching logic is right. It's missing wildcards for one, and seems like it's doing character comparison?

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,304 @@
...
+                    })
+                    .cloned()
+                    // FIXME: Bubble up errors rather than ignoring them as soon
+                    // as I understand the Rust-isms necessary... ideally we
+                    // could try_collect() here.
+                    .collect::<Vec<Event>>(),
```
paulbellamy, line 185: Filtering and mapping with Results is always super awkward in rust.
I think you want this line to be:
```suggestion
                    .collect::<Result<Vec<Event>>, std::num::ParseIntError>()?,
```
But idk the iterators to do the filtering above.
@leighmcculloch ?

cmd/soroban-cli/src/invoke.rs
paulbellamy: And needs to mkdir the dir in case the file is different from the ledger.json.

cmd/soroban-cli/src/events.rs
paulbellamy: Might be nice to pull this into a function and do some unit tests. You could have a look at the go unit tests for some example cases.
https://github.com/stellar/soroban-tools/blob/main/cmd/soroban-rpc/internal/methods/get_events_test.go#L21

cmd/soroban-cli/src/snapshot.rs
```diff
@@ -120,10 +127,85 @@ where
...
+            events::HostEvent::Debug(_e) => todo!(),
+        };
+
+        // TODO: Handle decoding errors cleanly; I miss errors.Wrap(err, ...) :(
+        let topics = match &contract_event.body {
+            xdr::ContractEventBody::V0(e) => &e.topics,
+        }
+        .iter()
+        .map(|t| t.to_xdr_base64().unwrap())
+        .collect(); // try_collect()? would be nice here
```
paulbellamy, line 177: I think you could remove the `.unwrap()` above and do:
```suggestion
        .collect::<Result<Vec<_>, _>()?;
```
Map (without the unwrap) returns a: `Iter<Result<String>>`, but `Iter<Result<String, Error>>` has an implicit `From` conversion to go to a `Result<Vec<String>, Error>` (by taking the first error and short-circuiting).
So, by giving the collect an explicit `Result<Vec<String>, Error>` type we can force that conversion. But we can let the compiler fill in the String and Error types here by eliding them with: `Result<Vec<_>, _>`

I think...

https://doc.rust-lang.org/std/result/#collecting-into-result

cmd/soroban-cli/src/snapshot.rs
```diff
@@ -120,10 +127,85 @@ where
...
+
+        let cereal_event = rpc::Event {
+            event_type: "contract".to_string(),
+            id: String::new(),
+            paging_token: String::new(),
```
paulbellamy, line 182: Should be an id, no?

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,304 @@
...
+    events_file: Option<std::path::PathBuf>,
+
+    /// Whether or not we should keep watching the events file and displaying
+    /// them in real time.
+    #[clap(
+        short, long,
+        help_heading = HEADING_SANDBOX,
+        conflicts_with = "rpc-url",
+    )]
+    follow: bool,
```
paulbellamy, line 54: If we're not implementing this now let's remove it. Better to not leave features in a broken state.

cmd/soroban-cli/src/events.rs
Shaptic: I already handle this with `InvalidFile` if it doesn't exist and `CannotReadFile` if reading fails. I don't think we need to worry about the non-existence case since this is meant to be run after `invoke`.

Maybe if `--follow` is specified, we can skip the non-existence error? As it implies, more heavily at least, that it's running _alongside_ rather than _after_ invoke.

cmd/soroban-cli/src/snapshot.rs
Shaptic: Nice, this worked :sunglasses:

cmd/soroban-cli/src/events.rs
leighmcculloch: Yeah that looks right. If you have collect collect into a Result, it'll stop at the first error.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,304 @@
...
+    start_ledger: u32,
+
+    /// The last (and inclusive) ledger sequence number in the range to pull events
+    /// https://developers.stellar.org/docs/encyclopedia/ledger-headers#ledger-sequence
+    #[clap(short, long)]
+    end_ledger: u32,
+
+    /// Formatting options for outputted events
+    #[clap(long, arg_enum, default_value = "pretty")]
+    format: OutputFormat,
```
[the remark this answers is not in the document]
Shaptic: Good call, done.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,304 @@
...
+    contract_ids: Vec<String>,
+
+    /// A set of (up to 5) topic filters to filter events on
+    #[clap(
+        long = "topic",
+        multiple = true,
+        max_values(5),
+        help_heading = "FILTERS"
+    )]
+    topics: Vec<String>,
```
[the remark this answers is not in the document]
Shaptic: Good idea, clarified :+1:

cmd/soroban-cli/src/invoke.rs
Shaptic: This is done by the snapshot, no? e.g. this part of `snapshot::commit_events`:

https://github.com/stellar/soroban-tools/blob/77a2120cf34028cc0084a6be95205a18abcddb63/cmd/soroban-cli/src/snapshot.rs#L156-L165

cmd/soroban-cli/src/events.rs
Shaptic: Nice, great idea - I did this haphazardly to get a "this is what it would look like" out there

leighmcculloch (commented):
I defer to @paulbellamy on everything as looks like he's already reviewing, just sharing some thoughts on events.json:

It's unclear to me that we need the sandbox events.json. If there was any feature to defer to see if we can get away without it, this would probably be it. Why: RPC sandbox is going to disappear soon according to https://github.com/stellar/soroban-tools/issues/197#issuecomment-1331155133. Invoke sandbox will be the only things-are-in-files sandbox experience. Outputting the events to stdout is probably sufficient for that use case. Options like querying for ledger number, etc, seem overkill in the cases you need to invoke and query the events. When I imagine using `soroban invoke` in a human capacity, I just want to 👀  the events. When I imagine using `soroban invoke` in a script capacity, it's easier to not have to deal with additional files and getting stdout is simpler.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+impl Cmd {
+    pub async fn run(&self, _matches: &clap::ArgMatches) -> Result<(), Error> {
+        if self.start_ledger > self.end_ledger {
+            return Err(Error::InvalidLedgerRange {
+                low: self.start_ledger,
+                high: self.end_ledger,
+            });
+        }
```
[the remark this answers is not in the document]
Shaptic: Yeah this is not something belonging to the CLI side, right? The error should bubble from the `get_events()` call, I think.

cmd/soroban-cli/src/events.rs
Shaptic: Eh I tried... I think it's the fact that the `Result` only appears after the first `.filter()` which borks up the rest of the process.

cmd/soroban-cli/src/snapshot.rs
Shaptic: Yep, good catch: added in fe2a78a alongside a toid package.

cmd/soroban-cli/src/rpc/mod.rs
```diff
@@ -188,4 +219,15 @@ impl Client {
             .request("getLedgerEntry", rpc_params![base64_key])
             .await?)
     }
+
+    pub async fn get_events(
```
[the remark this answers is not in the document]
Shaptic: fixed now!

cmd/soroban-cli/src/events.rs
Shaptic: Improved this but untested yet

Shaptic:
@leighmcculloch since I've already taken the time to add the sandbox event support, my vote is to keep it until it gets dropped. Stdout is nice and all, but developing over several hours / days basically makes `events.json` a log file you can go back to and cross-reference. The output format here is much nicer than in `invoke`, and I improved the UX such that you can specify the events file as the only arg and replay everything by default. Anyway, I'm probably justifying its existence after the fact, but I'm happy to gut it when the time comes.

@paulbellamy can you review again? barring any changes that come out of further testing, I think this is RFR again.

cmd/soroban-cli/src/events.rs
paulbellamy: I'd either say we should either always skip non-existence error, or always output, it for consistency.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+                error: err.to_string(),
+            })?
+            .iter()
+            // FIXME: We assume here that events are read off-disk in
+            // chronological order, but we should probably be sorting by ledger
+            // number (and ID, for events within the same ledger), instead,
+            // though it's likely that this logic belongs more in
+            // `snapshot::read_events()`.
+            .rev()
+            // The ledger range is optional in sandbox mode.
```
paulbellamy, line 269: *style*
I'd move this comment down a line so it is next to the relevant check, as I missed it initially.

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+}
+
+// Determines whether or not a particular filter matches a topic based on the
+// same semantics as the RPC server:
+//
+//  - for an exact segment match, the filter is a base64-encoded ScVal
+//  - for a wildcard, single-segment match, the string "*" matches exactly one
+//    segment
+//
+// The expectation is that a `topic_filter` is a comma-separated list of
```
paulbellamy, line 319: *style*
do you mean?
```suggestion
// The expectation is that a `segments` is a comma-separated list of
```

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+// segments that has previously been validated.
+//
+// [API
+// Reference](https://docs.google.com/document/d/1TZUDgo_3zPz7TiPMMHVW_mtogjLyPL0plvzGMsxSz6A/edit#bookmark=id.35t97rnag3tx)
+// [Code
+// Reference](https://github.com/stellar/soroban-tools/blob/bac1be79e8c2590c9c35ad8a0168aab0ae2b4171/cmd/soroban-rpc/internal/methods/get_events.go#L182-L203)
```
paulbellamy, line 325: 🚀

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+pub fn does_topic_match(topics: &Vec<String>, segments: &Vec<&str>) -> bool {
```
paulbellamy, line 326: *style*
Wdyt re param naming?
```suggestion
pub fn does_topic_match(topic: &Vec<String>, filter: &Vec<&str>) -> bool {
```

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+        } else if *segment != topics[idx] {
+            // Exact match the ScVal (decodability is assumed)
+            return false;
+        }
+
+        idx += 1;
+    }
+
+    // Check we had no leftovers
+    idx >= topics.len()
```
paulbellamy, line 346: *bug*
strict equal, no?
```suggestion
    idx == topics.len()
```

paulbellamy: Could even do this before we even bother iterating?

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+pub fn does_topic_match(topics: &Vec<String>, segments: &Vec<&str>) -> bool {
+    let mut idx = 0;
+
+    for segment in segments {
+        if idx >= topics.len() {
+            // Nothing to match, need at least one segment.
+            return false;
+        }
+
+        if *segment != "*" {
```
paulbellamy, line 335: *bug*
Shouldn't this check if the segment *is* a wildcard?
```suggestion
        if *segment == "*" {
```

Could we add a couple unit-tests for this? Could just steal the test cases from the rpc server

cmd/soroban-cli/src/rpc/mod.rs
```diff
@@ -188,4 +226,45 @@ impl Client {
...
+            match event_type {
+                Some(t) => match t {
+                    EventType::All => "",
+                    EventType::Contract => "contract",
+                    EventType::System => "system",
+                },
+                None => "",
+            }
+            .into(),
+        );
```
paulbellamy, line 251: *style*
Inserting a blank `"type": ""` and depending on that not being an error seems a bit dependent on a server implementation quirk, so it would be better to exclude if we could.

There's a few other options for how we could write this here, which would avoid inserting the empty key.

The flattest is probably:
```rust
match event_type {
  Some(EventType::Contract) => filters.insert("type".to_string(), "contract".into()),
  Some(EventType::System) => filters.insert("type".to_string(), "system".into()),
  _ => {},
};
```
But using the `_` matcher means we lose the completeness checking. We could either add the `Some(All)`, and `None` cases, or use a few chaining operators for:
```rust
event_type.and_then(|t| match t {
  EventType::All => None,
  EventType::Contract => Some("contract"),
  EventType::System => Some("system"),
}).map(|t| filters.insert("type".to_string(), t.into()))
```
Not really an issue though, so seems fine as is, if you want tbh.

cmd/soroban-cli/src/snapshot.rs
```diff
@@ -120,10 +127,95 @@ where
...
+            // is a safe assumption
+            1,
+            1,
+        );
+        let cereal_event = rpc::Event {
+            event_type: "contract".to_string(),
+            id: id.to_string(),
+            // stolen from
+            // https://github.com/stellar/soroban-tools/blob/main/cmd/soroban-rpc/internal/methods/get_events.go#L264
+            paging_token: format!("{}-{:010}", id.to_paging_token(), i + 1),
```
paulbellamy, line 192: *bug*
`id` should be the same as this I think?

cmd/soroban-cli/src/snapshot.rs
```diff
@@ -120,10 +127,95 @@ where
...
+            ledger: ledger_info.sequence_number.to_string(),
+            ledger_closed_at: ledger_info.timestamp.to_string(),
```
paulbellamy, line 194: *question*
Unless I'm misreading this, this is a `u64::to_string()`? Which means we're re-using `rpc::Event` here, but serializing stuff differently. In events from the soroban-rpc server this is an ISO8601 string, but here this is a stringified u64. It would be less confusing in future if we are consistent, so can we serialize this as an iso8601?

Edit: In the test fixtures file it is an iso8601, so.. what?

cmd/soroban-cli/src/snapshot.rs
```diff
@@ -120,10 +127,95 @@ where
...
+            },
+        };
+
+        events.push(cereal_event);
+    }
+
+    let mut file = std::fs::OpenOptions::new()
+        .write(true)
+        .truncate(true)
+        .open(output_file)?;
```
paulbellamy, line 216: *question*
Feels a bit awkward to rewrite the entire file every time and truncate. vs append. Probably fine for v1, tho. And avoids race-condition read-inconsistencies, maybe?

cmd/soroban-cli/src/toid/mod.rs
```diff
@@ -0,0 +1,69 @@
+/// A barebones implementation of Total Order IDs (TOIDs) from
```
paulbellamy, line 1: *style*
I think if there is only one file, the best practice would be to do:
`cmd/soroban-cli/src/toid.rs`

paulbellamy (commented):
A couple bugs, but looking good. Almost there.
I wouldn't worry too much about sandbox implementation, given: https://github.com/stellar/soroban-tools/pull/298#pullrequestreview-1214252501

cmd/soroban-cli/src/toid/mod.rs
paulbellamy: Otherwise this is nice. 👍

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+        arg_enum,
+        default_value = "all",
+        help_heading = "FILTERS"
+    )]
+    event_type: rpc::EventType,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("invalid ledger range: low bigger than high ({low} > {high})")]
```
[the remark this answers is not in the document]
Shaptic: nice

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        for raw_contract_id in &self.contract_ids {
+            // We parse the contract IDs to ensure they're the correct format,
+            // but since we'll be passing them as-is to the RPC server anyway,
+            // we disregard the return value.
+            utils::id_from_str::<32>(raw_contract_id).map_err(|e| Error::InvalidContractId {
+                contract_id: raw_contract_id.clone(),
+                error: e,
+            })?;
+        }
```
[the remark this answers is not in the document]
Shaptic: Sorry, I don't think I follow - aren't contract IDs still proper hex hashes in sandbox mode?

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+            }
+        }
+
+        let events = match self.rpc_url.as_ref() {
+            Some(rpc_url) => self.run_against_rpc_server(rpc_url).await?,
+            _ => match self.events_file.as_ref() {
+                Some(path) => self.run_in_sandbox(path)?,
+                _ => return Err(Error::TargetRequired),
+            },
+        };
```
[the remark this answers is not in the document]
Shaptic: sick - didn't know you could do this

cmd/soroban-cli/src/events.rs
```diff
@@ -0,0 +1,437 @@
...
+    /// events (required if not in sandbox mode).
+    /// https://developers.stellar.org/docs/encyclopedia/ledger-headers#ledger-sequence
+    #[clap(short, long, default_value = "0")]
+    end_ledger: u32,
+
+    /// Output formatting options for event stream
+    #[clap(long, arg_enum, default_value = "pretty")]
+    output: OutputFormat,
+
+    /// The maximum number of events to display (0 = all)
```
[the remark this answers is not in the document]
Shaptic: Can't we? The server can enforce the appropriate max if the limit isn't set. I can update the docstring to be more accurate (e.g. `0 = all in sandbox or server limit`)

cmd/soroban-cli/src/events.rs
Shaptic: Strict equal is strictly (heh) more correct, but `>=` is more "defensive"? From the code, there's no way it'd ever actually be `>`, but I like to make my invariants as generic as possible to defend against future code changes going wonky :man_shrugging:

cmd/soroban-cli/src/rpc/mod.rs
Shaptic: Lol I didn't expect something like this to be so convoluted :joy: your second solution is quite neat and clever, but wow that's quite a Rust-ism. What do you think about something a little more "traditional"?

```rust
        if event_type.is_some() && event_type.unwrap() != EventType::All {
            filters.insert(
                "type".to_string(),
                match event_type.unwrap() {
                    EventType::Contract => "contract",
                    EventType::System => "system",
                    EventType::All => "", // never happens, see condition
                }
                .into(),
            );
        }
```

(I wish I could do `if let Some(t) = event_type && t != EventType::All` but rustc doesn't like that.)

cmd/soroban-cli/src/snapshot.rs
Shaptic: oh you're right - I thought we discussed doing paging token and ID as being padded and unpadded, respectively, but I can't find it hah. Same is easier anyway :+1:

cmd/soroban-cli/src/snapshot.rs
Shaptic: yeahhh I thought about this and basically "good enough"d it: on one hand you could dump raw xdr and just append new entries, on the other hand you can have json and readability but then you need full rewrites on appends (unless you want to be clever and inject the serialized json in the correct part of the array in the file, but it seems like we shouldn't cross that bridge if we don't need to). I guess you could also have the file contain json blobs but not be a valid json blob itself

Shaptic: I dunno if race conditions are worth worrying about - who is running multiple sandbox invokes simultaneously?

cmd/soroban-cli/src/toid/mod.rs
Shaptic: cool, sure, I was just following `rpc/mod.rs`, and thought this might evolve into its own crate some day

cmd/soroban-cli/src/rpc/mod.rs
Shaptic: Ok I wrote this and it's kinda ugly - gonna stick with your second approach. Thanks for teaching me about `and_then`!

cmd/soroban-cli/src/toid/mod.rs
paulbellamy: Eh. Let's move it if/when that happens. the go toid package has been around a while without risk of expansion.

cmd/soroban-cli/src/rpc/mod.rs
paulbellamy: fwiw, I couldn't remember if I wanted `map` or what the name of the method that did that was. So I went to https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then, and dug around to find `and_then`, with the right type signature.

paulbellamy:
Note: This might want to be moved under `contract` subcommand with https://github.com/stellar/soroban-tools/pull/319

cmd/soroban-cli/src/events.rs
Shaptic: yeah good idea

cmd/soroban-cli/src/events.rs
Shaptic: Done! Grabbed them all from RPC.

cmd/soroban-cli/src/snapshot.rs
Shaptic: Yeah I bamboozled myself by creating the fixture file manually :yum: Fixed in 39376cfa33e464afd823efb93fec32fc5727d270!

cmd/soroban-cli/src/events.rs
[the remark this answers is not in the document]
Shaptic: Jk I got you - they aren't. Fixed!

cmd/soroban-cli/src/events.rs
Shaptic: Tested now w/ unit tests from RPC

paulbellamy (approved):

cmd/soroban-cli/src/events.rs
Shaptic: Yep, always error.

PR #300, opened 2022-12-07
Remove crate-git-revision 0.0.3 from dep graph

### LINKED ISSUE #296: Must rebuild git deps every time
Currently a rebuild which normally would take a fraction of a second must rebuild all git deps and takes ~6-9 seconds.

This happens before any `cargo test` too:

Screenshot 2022-12-06 at 11 10 55 AM

Is this due to the build script?

leighmcculloch:
### What
Update soroban-env, stellar-xdr, soroban-sdk.

### Why
To remove the crate-git-revision 0.0.3 dep from the dep graph because it causes invalidations of the entire build.

Close #296


paulbellamy (approved):

willemneal:
Can confirm locally that it's fixed! Thanks! It was even making my RA stall after any edit.

PR #301, opened 2022-12-07
Fix apt-get install failures in ci

leighmcculloch:
### What
Run `apt-get update` before `apt-get install` in ci.

### Why
We're seeing failures which are likely to do with the apt-get repositories that are cached being out of date. It's common practice to run update before install.


paulbellamy (approved):
Oh nice. I didn't realize we weren't doing that already!

leighmcculloch:
> I didn't realize we weren't doing that already!

An oversight.

PR #307, opened 2022-12-08
[WIP] create dynamic args and parse normal json

willemneal:
### What

This is still very much a WIP, but currently anything after `--` is treated as a CLI for the contract method.  For example, from the hello world contract

```rust
     pub fn hello(env: Env, to: Symbol) -> Vec<Symbol> {
        vec![&env, symbol!("Hello"), to]
    }
```

would be invoked with:

```
soroban invoke ... --fn hello -- --to world
```

More excitingly you can pass normal JSON for complex types. For example,

```rust
pub struct State {
    pub count: u32,
    pub last_incr: u32,
}


pub fn set_state(env: Env, state: State) -> u32 {
...
}
```

Could be be invoked with

```
soroban invoke ... --fn set_state -- --state '{"count": 0, "last_incr": 0}'
```

### Why

Improve the devx for passing arguments and make it clearer what the arguments are.

### Known limitations

[TODO or N/A]


willemneal:
Oops seems to include #284, which should be merged first anyway.

leighmcculloch:
This is beautiful. 🌈

willemneal:
@leighmcculloch is it possible to include comments in the xdr?

leighmcculloch:
> @leighmcculloch is it possible to include comments in the xdr?

What do you mean? Include comments in the contract specification that is stored in the XDR?

leighmcculloch:
Yup we could definitely do that. We could use regular rust docs for this. Capture them in the `contractimpl` attribute macro and include them in the spec.

tyvdh:
This is nice 👍 
Would love to see examples for each of these custom types https://github.com/tyvdh/soroban-quest/tree/main/quests/5-custom-types#create-your-custom-types. Specifically things like account ids, Accounts and Contracts, just to get a more clear sense of what a collection of cli commands would actually look like.

tyvdh:
This is amazing. Is there any value in not requiring the empty `--` flag? e.g. Yarn just forwards extra flags.
```
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
```
On the one hand it helps make it obvious what's intended to be interpreted as an arg but on the other hand `--fn hello -- --to world` feels _slightly_ weird

willemneal:
@tyvdh Currently the solution is to not parse the last args until after the full args are parsed.  It would be different if I could hook into the help doc generation for the main command, but when parsing args clap and others that I looked at require that the arguments be statically known (that is the lifetime of the names of the arguments is the same as the whole program run). However, you can't know what the arguments are while parsing since you need to look up the xdr.

One idea that I had, as I agree it feels a bit awkward, is that the function name is also on the right side.

```
soroban invoke ... -- hello --world arg
```

The other option is that we treat the args as one json object with the names of the parameters as keys, this is how near does it.

```
soroban invoke ... --fn hello --args '{"world": "arg"}'
```

leighmcculloch:
I agree I prefer the current implementation that uses `--` as the separator. While it isn't pretty on the eyes, it is simple, and is well established way to separate args of a tool from args. The Rust tooling also uses it in some places so developers are likely to come across it.

willemneal:
@leighmcculloch @tyvdh 

What do you think of a `file` input, e.g. `--world-file world.txt`?

tyvdh:
> @leighmcculloch @tyvdh
> 
> What do you think of a `file` input, e.g. `--world-file world.txt`?

I don't think it adds much unless again, it allows for positional replacements.

leighmcculloch:
Unless there's immediate value I'd leave it out, at least for now, nothing stopping adding it later. But this might be easier to improve/maintain/iterate on if there is less code to start out.

tyvdh:
@willemneal is there any reason we couldn't convert `String` into `Bytes`? Would make the fact Soroban doesn't have a string type easier to cope with. 

E.g.
```bash
$ soroban invoke --id 1d06cdae1e71f98427d55a056139a9db401f5368fd2bad5299109bd9bd4d0457 --fn c_user -- --_user '{"name": "Tyler van der Hoeven", "age": 31, "pet": "Dog"}'
```
Vs
```bash
$ soroban invoke --id 1d06cdae1e71f98427d55a056139a9db401f5368fd2bad5299109bd9bd4d0457 --fn c_user -- --_user '{"name": [84,121,108,101,114,32,118,97,110,32,100,101,114,32,72,111,101,118,101,110], "age": 31, "pet": "Dog"}'
```

willemneal:
@tyvdh My current strategy for testing is a round trip for the input. So with Bytes, we have no way of knowing that the type returned contract should convert back into a string. Personally I would like to see a new type just for UTF-8 strings. Also bytes stored as a string are usually encoded in Base64.

I do agree that without proper string support Dev's will continue to have to handle their own string encoding/decodings.

tyvdh:
Makes sense. @leighmcculloch did I see somewhere a proposal for a utf8 String type in Soroban? I would like to throw my vote behind that proposal

leighmcculloch:
We may have a way to signal that a bytes is intended to be rendered as a string. I think this is something we'll explore in the new year.

leighmcculloch:
With bytes I would suggest rendering them as hex strings rather than as arrays. We do that already with XDR, and I think the current CLI does that too. The main benefit is that when things like contract IDs show up in JSON, they look like hashes.

willemneal:
@leighmcculloch  and @tyvdh 

I removed the files and xdr args.

Is it worth adding and xdr suffix like file? Or implementing that in the future?

e.g.
```bash
--world-xdr 189AB...
```

leighmcculloch:
> Is it worth adding and xdr suffix like file? Or implementing that in the future?

I'm fine with implementing it in the future. cc @paulbellamy @tyvdh for alternative viewpoints.

tyvdh:
I don't think so, leave it off until someone starts asking for it

cmd/soroban-cli/src/strval.rs
```diff
@@ -29,76 +40,454 @@ impl From<()> for Error {
...
         }
+    }
 
-        // For all others we just use the json parser
-        _ => serde_json::from_str(s)
-            .map_err(Error::Serde)
-            .and_then(|raw| from_json(&raw, t))?,
-    };
-    Ok(val)
+    fn parse_tuple_strukt(
```
paulbellamy, line 173:
```suggestion
    fn parse_tuple_struct(
```

paulbellamy (approved):

cmd/soroban-cli/src/strval.rs
willemneal: Siem en stuct is a keyword I used strukt throughout. It's what rust analyzer does. But here I guess it doesn't matter as much. Definitely no strong feelings if you want to rename it.

paulbellamy: Ok. if rust analyzer does it, then fine. Thought it was just a typo. :)

paulbellamy (approved):

PR #308, opened 2022-12-09
chore: StrValError --> and implemented using thiserror

willemneal:
### What

Rename StrValError to Error and implement it with thiserror. Also remove Error suffix for `UnknownError` since `Error::Unknown` is clearly an error.

### Why

Make it more idiomatic and follow the same convention for errors used across the crate.

### Known limitations

[TODO or N/A]


leighmcculloch:
Thanks @willemneal!

PR #310, opened 2022-12-14
Remove cyclical reference to quickstart on integration tests

sreuland:
### What

removed integration tests that referenced quickstart, instead the same integration test coverage is established by invoking the [system-test](https://github.com/stellar/system-test) docker image instead from gh workflow.

### Why

The reference from tests to quickstart had potential for backwards reference, it's pinned to specific version of core/rpc and tests will need to refer different versions at different times such during release prep, when new version references are needed that quickstart doesn't have yet. Also, the e2e integration tests are mostly duplicative of same coverage done by system tests, don't need to maintain the same test code base here.

### Known limitations

this has a dependency on [system-test PR #8](https://github.com/stellar/system-test/pull/8) to merge first, as it will provide the updated docker image used by the gh workflow here.


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+
+  integration:
+    name: System tests
+    strategy:
+      matrix:
+        scenario-filter: ["^TestDappDevelop$/^.*$"]
+    runs-on: ubuntu-latest
+    env:
+      SYSTEM_TEST_DOCKER_REPO_TAG: "sreuland/system-test:latest"
```
sreuland, line 17: this is just short term to check that workflow works, will be changing this to `stellar/system-test:latest` shortly

leighmcculloch (approved):

sreuland:
the GH workflow passed when running the new system-tests as e2e action, but, looks like the cargo executions are taking alot of time to run here on GH, to build soroban-cli and doing compilations of example wasms, total time for the all the tests was about 25mins: 
https://github.com/stellar/soroban-tools/actions/runs/3699847241/jobs/6267646367

when I run the tests locally(not in docker though) on an m1/arm64 macbook, it's 4-5 minutes

i've added `--config net.git-fetch-with-cli=true --config build.jobs=6` to the cargo invocations, but I don't think it's getting enough cores from GH compute maybe..

leighmcculloch:
When we build soroban-cli and other Rust projects in GHA, we typically use the `stellar/actions/rust-cache` action to cache build dependencies and artifacts so that it doesn't have to download all the dependencies every time, and only builds the things that have changed.

sreuland:
> When we build soroban-cli and other Rust projects in GHA, we typically use the `stellar/actions/rust-cache` action...

@leighmcculloch , thanks for that pro-tip! cut the time in half at least, so now the e2e tests take 18 minutes. This includes building cli with cargo and compiling sample contracts with cargo, so, given gha is only getting two cores, I think that's pretty good improvement.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,40 @@
...
+    strategy:
+      matrix:
+        scenario-filter: ["^TestDappDevelop$/^.*$"]
+    runs-on: ubuntu-latest
+    env:
+      SYSTEM_TEST_DOCKER_REPO_TAG: "sreuland/system-test:latest" 
+      SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_HASH: "main"
+      SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_REPO: "https://github.com/stellar/soroban-examples.git"
+      SYSTEM_TEST_RUST_TOOLCHAIN_VERSION: "1.65.0"
+      SYSTEM_TEST_CORE_DEBIAN_VERSION: "19.5.1-1137.b3a6bc281.focal~soroban"
```
tsachiherman, line 21: I think that it's good for now, but for the long run, we need to derive the core version from the one in the go mono repository. To do that:
1. open the go.mod file and figure out the go repository hash.
2. checkout whatever file in the go repo that contain the core dependency
3. populate the ENV vars.

tsachiherman (commented):
The change looks reasonable to me, but I'd like @leighmcculloch 's feedback as well.

tsachiherman (approved):

tsachiherman:
@sreuland could you please see if you could merge this one in ? It would help us in the coming release.

sreuland:
> @sreuland could you please see if you could merge this one in ? It would help us in the coming release.

Yes, I need to get this [system-test PR #8](https://github.com/stellar/system-test/pull/8) merged first as it publishes the docker image used by tests here, do you mind taking a look at that? thanks!

.github/workflows/e2e.yml
sreuland: removed, use stellars repo now

.github/workflows/e2e.yml
sreuland: created https://github.com/stellar/soroban-tools/issues/328 to track this

PR #311, opened 2022-12-15
makefile: add support for soroban-rpc packaging

tsachiherman:
### What

add support for soroban-rpc packaging in the makefile.

### Why

avoid versioning setup being deduced both during the packaging as well during the makefile execution.

### Known limitations

N/A


sreuland (approved):
thanks for the adjustment, will be useful in the package building pipeline

PR #314, opened 2022-12-15
Add GitHub linting for GO code

tsachiherman:
### What

Add GitHub linting for GO code

### Why

Better code, less bugs.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,33 @@
+name: Linters
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  pull-requests: read
+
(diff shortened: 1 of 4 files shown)


sreuland:
can some content be added to README.md to mention linting config/requirements, but also, an example of how to run the same linting locally to pre-check before pushing?

sreuland (approved):
nice improvement, left one comment just for consideration of extra docs in readme, thanks!

leighmcculloch:
> can some content be added to README.md to mention linting config/requirements, but also, an example of how to run the same linting locally to pre-check before pushing?

Alternatively we could add a `lint` target to `Makefile` which runs the command.

tsachiherman:
> > can some content be added to README.md to mention linting config/requirements, but also, an example of how to run the same linting locally to pre-check before pushing?
> 
> Alternatively we could add a `lint` target to `Makefile` which runs the command.

exactly.

PR #320, opened 2022-12-16
fix(CI): Move clippy to fmt step so e2e test can proceed regardless

willemneal:
### What

Title

### Why

Prevent e2e tests from being blocked by clippy error

### Known limitations

N/A

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,8 +5,6 @@ on:
     branches: [main, release/**]
   pull_request:
 
-env:
-  RUSTFLAGS: -Dwarnings -Dclippy::all -Dclippy::pedantic
 
 defaults:
   run:
@@ -27,7 +25,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
-    - run: cargo fmt --all --check
+    - run: cargo fmt --check
+    - run: cargo clippy -- -Dwarnings -Dclippy::all -Dclippy::pedantic
 
   rust-analyzer-compat:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


willemneal:
@leighmcculloch Does clippy differ for different targets?

.github/workflows/rust.yml
```diff
@@ -27,7 +25,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
-    - run: cargo fmt --all --check
+    - run: cargo fmt --check
+    - run: cargo clippy -- -Dwarnings -Dclippy::all -Dclippy::pedantic
```
leighmcculloch, line 29: Clippy actually builds all of the code, it's like `cargo check` or `cargo build` plus additional linters, so I don't think it should live in the fmt step. The fmt step is intended to be a fast fail on formatting issues, where as clippy will build the entire application which will take a while.

Also by moving clippy out of the build-and-test step, it's not actually running a build other than the builds that the test command runs.

leighmcculloch:
I have seen clippy fail differently once on a different platform, I think it was a bug though. If you see that happen worth opening an issue over at https://github.com/rust-lang/rust-clippy.

leighmcculloch:
Because of the way clippy is built it is difficult to separate it. We could run it separately from the test run, we used to do that, but it made CI slower because we needed to wait for more instance availability and clippy and test repeat build work. While we do cache builds pretty well, we know the test will reuse the same and it is already loaded on the same instance.

willemneal:
I see, I thought it was on syntactic and didn't need the whole compiler. Well I'll open another PR to just add husky so that we can prevent code being pushed that will fail fmt and clippy.  This will prevent the trigger much more often so this isn't so much an issue.

PR #321, opened 2022-12-17
feat(CI): add git hooks and improve tests to calculate file hash first

willemneal:
### What

`./install_githooks.sh` copies the committed two hooks to `.git/hooks`.

Add `Wasm` type that adds a hash method so tests can calculate the expected hash instead of hard coding it.

### Why

Prevent users from commit code that won't pass `fmt`. Also must pass tests and `clippy` before pushing.

Hard coding hash makes tests fail if the resulting wasm file changes, which can happen for example when upgrading the compiler version or updating deps.

### Known limitations

If the hooks are updated devs need to rerun the script.


willemneal:
@leighmcculloch I couldn't figure out how to make this opt-in.  But I could look into that if you'd prefer not forcing this on users.  But I do think it will save CI actions from failing as much.

willemneal:
Also is push all we need?

leighmcculloch:
I don't think cargo-husky has an opt-in built-in. It has an opt-out though with `CARGO_HUSKY_DONT_INSTALL_HOOKS=true`.

I don't mind about whether we use this or not. I'm personally likely to opt out, but that's not hard.

I'll defer to @tsachiherman @paulbellamy and co.

---

Something I and some others do to avoid discovering clippy warnings late is we enable clippy in VSCode. That way we get immediate in-editor feedback about clippy warnings. If you don't have that enabled that may be another way to get the same result.

If you don't have VSCode configured this way and want to try it out, these are the settings to add:
```js
    "rust-analyzer.checkOnSave.command": "clippy",
    "rust-analyzer.checkOnSave.extraArgs": [
        "--",
        "-Wwarnings",
        "-Wclippy::all",
        "-Wclippy::pedantic",
    ],
```

I use `-W` for everything in VSCode so they show up as warnings and not errors, so I can distinguish compile errors from them in the editor.

willemneal:
I have it too, but when I'm doing quick edits I find myself ignoring the warnings. Or sometimes like it tests they fail to show up sometimes.  I've also been having issues with rust-analyzer since I started a couple months ago and I suspect it's because it's rebuilding each save.

Personally in general I'd prefer to fail before a push then after.

paulbellamy (changes requested):
I'm very on-the-fence about this.

Pros:
- The consistency would be nice, and making it easier would be great.

Cons:
- The husky security model is awful. It sets the precedent for devs to download, trust, and run random bits of code off the internet without even seeing it or knowing it is happening. It's had some issues in the past, where if it can't detect the git repo you are in, it will modify your global ~/.git instead 😱 .
- Pre-commit hooks always work poorly with `git add --interactive`, which I do a lot.

I'd suggest making this opt-in. Maybe have a `./install-hooks.sh` script (or something) that installs the scripts?

willemneal:
@paulbellamy This version of husky, copies the scripts checked into the repo and places it in the `.git/hooks` local folder.  It runs with a `cargo test`  as a `build.rs` so it will always find the root of the repo. This also means that it's not random scripts or code;  It's a published crate and a checked in script.

However, as you point out it could cause issues for some users. So perhaps it's better if we just have the script.

paulbellamy:
> This also means that it's not random scripts or code; It's a published crate and a checked in script.
What happens in the case where a user forks the code and submits a PR. If I check that out and run `cargo test` it will start executing their scripts?

tsachiherman:
> > This also means that it's not random scripts or code; It's a published crate and a checked in script.
> 
> What happens in the case where a user forks the code and submits a PR. If I check that out and run `cargo test` it will start executing their scripts?

I think that this problem might be unavoidable. Before checking out a PR, you would need to review it online and check it out only in case you believe it's not dangerous.

paulbellamy (approved):

tsachiherman:
@willemneal could you please resolve the merge conflict ?

PR #322, opened 2022-12-19
soroban-rpc: Add LedgerEntry SQLite storage

2opremio:
Initially migrated from https://github.com/stellar/go/pull/4698 and then adapted to use an on-disk SQLite DB


2opremio:
CC @tamirms

2opremio:
Depends on https://github.com/stellar/go/pull/4715

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -19,6 +24,53 @@ func Start(cfg config.LocalConfig) (exitCode int) {
...
 
+	httpPortUint := uint(cfg.CaptiveCoreHTTPPort)
+	var captiveCoreTomlParams ledgerbackend.CaptiveCoreTomlParams
+	captiveCoreTomlParams.HTTPPort = &httpPortUint
+	captiveCoreTomlParams.HistoryArchiveURLs = cfg.HistoryArchiveURLs
+	captiveCoreTomlParams.NetworkPassphrase = cfg.NetworkPassphrase
+	captiveCoreTomlParams.Strict = true
+	captiveCoreToml, err := ledgerbackend.NewCaptiveCoreTomlFromFile(cfg.CaptiveCoreConfigPath, captiveCoreTomlParams)
+	if err != nil {
+		logger.WithError(err).Fatal("Invalid captive core toml")
```
tsachiherman, line 35: I think that you might want to exit here; the sooban-rpc wouldn't be operative without a captive core.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,295 @@
...
+		return nil, errors.Wrap(err, "open failed")
+	}
+
+	ret := &sqlDB{
+		db: db,
+	}
+
+	err = runMigrations(ret.db.DB, "sqlite3")
+
+	return ret, err
```
tsachiherman, line 57: I'd suggest that we'll be more atomic; if err != nil, do not return ret. ( and close the db )

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,295 @@
...
+	if err := tx.Select(&results, sqlStr, args...); err != nil {
+		return xdr.LedgerEntry{}, err
+	}
+	if len(results) != 1 {
+		return xdr.LedgerEntry{}, sql.ErrNoRows
+	}
+	ledgerEntryBase64 := results[0]
+	var result xdr.LedgerEntry
+	err = xdr.SafeUnmarshalBase64(ledgerEntryBase64, &result)
+	return result, err
```
tsachiherman, line 80: same comment as before: if err != nil, return the empty `xdr.LedgerEntry`

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func flushLedgerEntryBatch(tx *sqlx.Tx, encodedKeyEntries map[string]*string) error {
```
tsachiherman, line 83: could you modify the encodedKeyEntries to be an array rather then a map ?
while I do like maps, iteration over maps yields unpredictable order. That means that the entries in the database would have different ordering for the same data. ideally, this would not be the case.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,295 @@
...
+			return err
+		}
+	}
+
+	if len(deleteKeys) > 0 {
+		sqlStr, args, err := sq.Delete(ledgerEntriesTableName).Where(sq.Eq{"key": deleteKeys}).ToSql()
+		if err != nil {
+			return err
+		}
+		_, err = tx.Exec(sqlStr, args...)
```
tsachiherman, line 111: duplicated line ?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -19,6 +24,53 @@ func Start(cfg config.LocalConfig) (exitCode int) {
 	logger := supportlog.New()
 	logger.SetLevel(cfg.LogLevel)
 
+	httpPortUint := uint(cfg.CaptiveCoreHTTPPort)
+	var captiveCoreTomlParams ledgerbackend.CaptiveCoreTomlParams
+	captiveCoreTomlParams.HTTPPort = &httpPortUint
+	captiveCoreTomlParams.HistoryArchiveURLs = cfg.HistoryArchiveURLs
+	captiveCoreTomlParams.NetworkPassphrase = cfg.NetworkPassphrase
+	captiveCoreTomlParams.Strict = true
+	captiveCoreToml, err := ledgerbackend.NewCaptiveCoreTomlFromFile(cfg.CaptiveCoreConfigPath, captiveCoreTomlParams)
```
sreuland, line 33: @SirTyson , has [a PR merging soon in go repo](https://github.com/stellar/go/pull/4733/files#diff-c7702ef819e52917ed3868588dff5273774c61a6aa8d36c1644d7ba251fb7756) which will config captive core to use BucketListDB by default, no change needed here, just wanted to mention it, as once it's in, might help in some of the spin up time.

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -0,0 +1,238 @@
+package ledgerentry_storage
```
tsachiherman, line 1: how do you feel about naming the packages `data/ledger` and `data/events` ?

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -0,0 +1,238 @@
...
+	"time"
+
+	"github.com/stellar/go/historyarchive"
+	"github.com/stellar/go/ingest"
+	backends "github.com/stellar/go/ingest/ledgerbackend"
+	"github.com/stellar/go/support/log"
+	"github.com/stellar/go/xdr"
+)
+
+// TODO: Make this configurable?
```
tsachiherman, line 17: yes - good call. no need to expose that to the UI for now, but having this a configurable parameter makes sense.

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -0,0 +1,238 @@
...
+func (ls *ledgerEntryStorage) fillEntriesFromLatestCheckpoint(ctx context.Context, archive historyarchive.ArchiveInterface) (uint32, error) {
+	root, err := archive.GetRootHAS()
+	if err != nil {
+		return 0, err
+	}
+	startCheckpointLedger := root.CurrentLedger
+
+	ls.logger.Infof("Starting processing of checkpoint %d", startCheckpointLedger)
+	// TODO: should we make this configurable?
+	checkpointCtx, cancelCheckpointCtx := context.WithTimeout(ctx, 30*time.Minute)
```
tsachiherman, line 66: move the `30*time.Minute` to be a configurable paramter.

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -0,0 +1,238 @@
...
+			return 0, context.Canceled
+		default:
+		}
+		change, err := reader.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			// TODO: we probably shouldn't panic, at least in case of timeout
+			panic(err)
```
tsachiherman, line 92: no need to panic here. Just return the error and let the caller decide what to do next.

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -0,0 +1,238 @@
...
+		if err != nil {
+			return 0, err
+		}
+		if !relevant {
+			continue
+		}
+		tx.UpsertLedgerEntry(key, *entry)
+		entryCount++
+
+		if entryCount%10000 == 0 {
```
tsachiherman, line 106: define the 10000 as a const at the top of the file.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
2opremio: A map is used to coalesce multiple modifications to the same key (the same key can be modified multiple times or deleted). You cannot do that with a slice.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
2opremio: Deletions and updates are queried separately

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
2opremio: Also, with a slice you cannot batch queries since updates and deletions would need to be done sequentially and in order.

tsachiherman: sure; we can use a data structure like the following, which would allow us to be persistent:
```golang
type encodedKeyEntry struct {
  key string
  val *string
}
type encodedKeyEntries struct {
   keys []encodedKeyEntry
   keysMaps map[string]uint // translation of key -> index in keys array
}
```

tsachiherman: as for the batching : 
for update from archives, there are no deletions, so it's very much consistent.
when we write ledger entries, we always want to batch these on ledger boundary ( which would guarantee the same outcome.. )

2opremio: I think this overcomplicates things just to ensure a fixed traversal ordering (which isn’t important for the queries we use)

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: Ah, fantastic. I will bump the PR and check the spin up time!

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
2opremio: 👍

2opremio:
@tamirms  Thanks! I took a look and it seems more robust but considerably less efficient than what I am using (both in the implementation and in the fact that it seems to treat creation and updates separately, which u avoid using upserts)

2opremio:
> and in the fact that it seems to treat creation and updates separately, which I avoid using upserts

I saw that you can simply treat both creations and updates as upserts.

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: It's a fatal log, it will exit

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
2opremio: I think that the extra level of indirection is not necessary and `ledgerentry_storage` is much more descriptive than `data/ledger`

2opremio:
I didn't dig on where the performance difference lies but I also observed that it breaks the `db.go` module abstraction:

* I _compact_ entires at the `ledgerkey->ledgerentry` level within the db module, which is where the compaction makes sense since it's a DB performance detail
*  However, the change compactor does so at the change level, which is a bit odd.

2opremio:
PTAL

cmd/soroban-rpc/internal/ledgerentry_storage/db_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	var key xdr.LedgerKey
+	err := key.SetContractData(data.ContractId, data.Key)
+	if err != nil {
+		panic(err)
+	}
+	return key, entry
+}
+
+func TestConcurrency(t *testing.T) {
```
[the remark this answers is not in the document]
2opremio: The test already does that (it checks that there is no value in the DB after the Upsert)

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -0,0 +1,226 @@
...
+
+	// First, make sure the DB has a complete ledger entry baseline
+
+	startCheckpointLedger, err := ls.db.GetLatestLedgerSequence()
+	if err == ErrEmptyDB {
+		// DB is empty, let's fill it from a checkpoint
+		ls.logger.Infof("Found an empty database, filling it in from the most recent checkpoint (this can take up to 30 minutes, depending on the network)")
+		startCheckpointLedger, err = ls.fillEntriesFromLatestCheckpoint(ctx, archive)
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
2opremio: Yeah, good point, I need to change GoLand's settubgs so that it doesn't hide the error handling.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,308 @@
...
+	if err = xdr.SafeUnmarshal([]byte(ledgerEntryBin), &result); err != nil {
+		return xdr.LedgerEntry{}, err
+	}
+	return result, nil
+}
+
+func flushLedgerEntryBatch(tx *sqlx.Tx, encodedKeyEntries map[string]*string) error {
+	upsertCount := 0
+	upsertSQL := sq.Replace(ledgerEntriesTableName)
+	var deleteKeys = make([]interface{}, 0, len(encodedKeyEntries))
```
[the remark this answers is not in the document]
2opremio: Good catch, I didn't know it would cast automagically

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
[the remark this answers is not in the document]
2opremio: Ah, actually, I cannot do bubble them up because `run()` is executed in a goroutine.

cmd/soroban-rpc/internal/ledgerentry_storage/db_test.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+		if err := tx.Done(); err != nil {
+			panic(err)
+		}
+	}
+	b.StopTimer()
+}
+
+func NewTestDB() (DB, string) {
+	dbPath := path.Join(os.TempDir(), fmt.Sprintf("%08x.sqlite", rand.Int63()))
```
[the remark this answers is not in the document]
2opremio: Unfortunately I cannot use it in the benchmark and it's not worth having two separate version of `NewTestDB()`

2opremio:
@tamirms PTAL

2opremio:
Ah, I get it now!! Thanks! I will incorporate another test

2opremio:
>     1. Mitigate [checkpoint starvation](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/performance.md#checkpoint-starvation)

I was actually looking into this.

When running against pubnet I did notice that the WAL file grows to the same size as the DB file and never goes down in size.

```
fons@studioso soroban-tools % ls -lh soroban_rpc.*
-rw-r--r--  1 fons  staff   9.3G Jan 20 12:28 soroban_rpc.sqlite
-rw-r--r--  1 fons  staff    19M Jan 20 03:47 soroban_rpc.sqlite-shm
-rw-r--r--  1 fons  staff   9.4G Jan 20 12:28 soroban_rpc.sqlite-wal
```

I did see the `wall_checkpoint` pragma, but I don't see how to supply it to the Go SQLite driver, it's not available as an option to the connection string (see https://github.com/mattn/go-sqlite3#connection-string ).
 

>     2. Investigate how soroban-rpc can be deployed using a persistent volume. I anticipate if we do a rolling deployment (where the new version is up before the old version is stopped) we will encounter some errors because sqlite does not allow multiple concurrent writers to the db.

You mean, when used in the quickstart image? 👍

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,322 @@
...
+		tx:                tx,
+		forLedgerSequence: forLedgerSequence,
+		buffer:            xdr.NewEncodingBuffer(),
+		keyToEntryBatch:   make(map[string]*string, maxBatchSize),
+	}
+	ret.stmtCache = sq.NewStmtCache(tx)
+	return ret, nil
+}
+
+func (l *ledgerUpdaterTx) PrepareContext(ctx context.Context, query string) (*sql.Stmt, error) {
```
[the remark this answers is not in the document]
2opremio: It implements an interface used by at:

```
ret.stmtCache = sq.NewStmtCache(tx)
```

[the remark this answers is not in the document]
2opremio: Ah I get it now. That is true!

2opremio:
I have addressed all the comments and also made sure that the WAL file is checkpointed after every write commit.

Ah, while I was at it, I also parallelized the checkpoint DB pre-filling with captive core's preparation

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,340 @@
...
+	sq "github.com/Masterminds/squirrel"
+	"github.com/jmoiron/sqlx"
+	_ "github.com/mattn/go-sqlite3"
+	migrate "github.com/rubenv/sql-migrate"
+	"github.com/stellar/go/support/errors"
+	"github.com/stellar/go/xdr"
+)
+
+//go:embed migrations/*.sql
+var migrations embed.FS
```
tsachiherman, line 20: I like this one.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	postWriteCommitHook := func() error {
+		_, err := db.Exec("PRAGMA wal_checkpoint(TRUNCATE)")
+		return err
+	}
+
+	ret := &sqlDB{
+		db:                  db,
+		postWriteCommitHook: postWriteCommitHook,
```
tsachiherman, line 64: I think that implementing this method as a sqlDB method would be cleaner; i.e.
`func (db *sqlDB) PostWriteCommit() error`.

I think that the code above is not safe:
The `db` variable is a pointer; and when using anonymous functions, the compiler attempts to keep track of all the external memory location. in this case the `db` escapes the scope of the function and therefore need to be tracked. However, given that the `db` itself was not allocated as a pointer, I'm not sure if it would always work correctly.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -0,0 +1,340 @@
...
+		return err
+	}
+	if err := upsertLatestLedgerSequence(l.tx, l.forLedgerSequence); err != nil {
+		_ = l.tx.Rollback()
+		return err
+	}
+	if err := l.tx.Commit(); err != nil {
+		return err
+	}
+	if l.postWriteCommitHook() != nil {
```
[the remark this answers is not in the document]
2opremio: Good catch!

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
2opremio: The garbage collector will keep track of the pointer just fine.

The idea behind using a hook was to make the implementation DB-independent (all the SQLite-specific stuff is only included in the constructor function)

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
2opremio: 👍 I got it from @tamirms

cmd/soroban-rpc/main.go
```diff
@@ -62,6 +74,37 @@ func main() {
 			},
 			Usage: "minimum log severity (debug, info, warn, error) to log",
 		},
+		&config.ConfigOption{
```
sreuland, line 77: I created separate ticket on quickstart to update it's config on soroban runtime, which will now need to include some of these new settings:
https://github.com/stellar/quickstart/issues/406

quickstart won't be able to run this version of rpc until then.

2opremio: 👍 thanks!

2opremio:
OK, I think everything is finally addressed. Thanks for all the review rounds!

Integration tests are failing due to an unrelated reason https://github.com/stellar/soroban-tools/pull/344 (I will rebase and get it all green before merging)

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -255,6 +249,7 @@ func (l *ledgerUpdaterTx) UpsertLedgerEntry(key xdr.LedgerKey, entry xdr.LedgerE
 	// safe since we cast to string right away
```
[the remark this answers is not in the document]
2opremio: Good idea, I implemented it

PR #324, opened 2022-12-20
soroban-cli, soroban-rpc: Simplify topic matching code.

Shaptic:
### What
The topic matching code was a little convoluted given the simplified invariants it needs to maintain:
 - the topic filter needs to match the **number** of segments exactly (e.g. `transfer/GABC/10` should not match `transfer/GABC`)
 - a filter segment should match the topic segment **exactly** or be a wildcard (`*`)

### Why
Simpler code = easier to read/maintain/debug.

### Known limitations
Both of these have unit tests, so N/A.


paulbellamy (approved):
🚀 🎸 🔥

PR #327, opened 2022-12-22
soroban-rpc: fix linting warnings

tsachiherman:
### What

fix linting warnings in soroban-rpc

### Why

remove warnings

### Known limitations

N/A


.golangci.yml
```diff
@@ -73,20 +66,20 @@ linters:
     - dupl
     - errcheck
     - exportloopref
-    - funlen
+    #- funlen
```
2opremio, line 69: Isn’t it better to just remove the line?

cmd/deptool/analyze.go
```diff
@@ -252,7 +252,8 @@ func findLatestVersion(repo *git.Repository, latestCommitRef *plumbing.Reference
 	return commits[len(commits)-1], versions[len(commits)-1], workspaceVer[len(commits)-1], nil
 }
 
-func findCargoVersionForCommit(repo *git.Repository, pkgName string, commit *object.Commit) (string, bool, error) {
+//lint:ignore funlen gocyclo
```
2opremio, line 255: Isn’t it worth fixing the cyclomatic complexity? The function seems rather complex.

2opremio (approved):

.golangci.yml
tsachiherman: The problem is that if I'll remove that line, I would never know that it's an option..

cmd/deptool/analyze.go
tsachiherman: yes, I agree with that. I was hoping to try and fix as many of the issues, without any substantial code change ( since that would require more testing, etc. ).
I'll refactor that in a subsequent PR, as this isn't really a concrete concern.

PR #331, opened 2023-01-01
Upgrade wasm-opt to 0.111.0

brson:
### What

This upgrades binaryen to version 111.

Binaryen (and wasm-opt) now enables the `SignExt` and `MutableGlobals` features by default, which are also enabled in the LLVM backend. In the future Binaryen will align its default feature selection with the LLVM backend.

This is _probably_ is ok for soroban - wasmi supports both these.

This also points to potential future optimizations by explicitly turning on more features that wasmi supports but that aren't enabled in wasm-opt by default.

The [changelog](https://github.com/brson/wasm-opt-rs/blob/master/CHANGELOG.md) for 0.111.0.

### Why

Just upgrading deps.

### Known limitations

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2620,9 +2620,9 @@ checksum = "1c38c045535d93ec4f0b4defec448e4291638ee608530863b1e2ba115d4fff7f"
 
 [[package]]
 name = "wasm-opt"
-version = "0.110.2"
+version = "0.111.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


tsachiherman (commented):
I'm not opposing to this one, but wondering if we have a similar dependency that might use a different version. @leighmcculloch ?

leighmcculloch (approved):
Thanks! This sounds good to me!

leighmcculloch:
> wondering if we have a similar dependency that might use a different version. @leighmcculloch ?

@tsachiherman There shouldn't be a consistency issue with upgrading here. I don't think we use wasm-opt elsewhere.

PR #335, opened 2023-01-05
Disable optional features in wasm-opt

brson:
### What

Disable optional features when calling wasm-opt.

### Why

In https://github.com/stellar/soroban-tools/pull/331 I mentioned that the new wasm-opt enables signext and mutable globals by default.

After a closer look, soroban disables these features explicitly when setting up the wasmi vm, so it seems to me they should be disabled in wasm-opt too.

### Known limitations

na


leighmcculloch (approved):
Nice, +1.

PR #338, opened 2023-01-10
integration testing: fix "unable to find commit hash" during tests

tsachiherman:
### What

The current code was producing the message `unable to find commit hash` without specifying what commit hash was being looked for. 

### Why

The code for scanning the commits was incorrect. It should have evaluated parents of commits in order to properly find the full hash based on the short hash.

### Known limitations

N/A


sreuland (approved):
looks good!

PR #353, opened 2023-01-23
soroban-rpc: Update getLedgerEntry to use the internal storage

2opremio:
### What

Update the `getLedgerEntry` JSON-rpc endpoint to use the internal ledger-entry storage provided by #322 

I've also had to:
* refactor the `daemon.go` module to make it testeable.
* make the ledger storage stop grafefully when being closed (also to make it testable).
* remove `TestSimulateTransactionDeadlineError` early (it wasn't worth fixing since we are going to remove the core dependency on `simulateTransaction` very soon.

### Why

Core is deprecating its `getLedgerEntry` endpoint

### Known limitations
N/A


2opremio:
Addresses https://github.com/stellar/go/issues/4720

PR #355, opened 2023-01-24
soroban-rpc: Add in-memory events storage

tamirms:
Add in-memory events db with a retention window specified in number of ledgers. For example, the event store can be configured to have a retention window of 17280 ledgers, which corresponds to approximately 24 hours assuming an average ledger close time of 5 seconds.

The event store is implemented using a circular buffer.

Close https://github.com/stellar/go/issues/4718

### DIFF
--- a/cmd/soroban-rpc/internal/events/events.go
+++ b/cmd/soroban-rpc/internal/events/events.go
@@ -1,6 +1,7 @@
 package events
 
 import (
+	"fmt"
 	"io"
 	"math"
 	"sort"
@@ -24,23 +25,6 @@ type Cursor struct {
 	Event uint32
 }
 
-var (
-	// MinCursor is the smallest possible cursor
-	MinCursor = Cursor{}
-	// MaxCursor is the largest possible cursor
-	MaxCursor = Cursor{
-		Ledger: math.MaxUint32,
-		Tx:     math.MaxUint32,
-		Op:     math.MaxUint32,
-		Event:  math.MaxUint32,
-	}
-)
-
-type event struct {
-	contents xdr.ContractEvent
-	cursor   Cursor
-}
-
 func cmp(a, b uint32) int {
 	if a < b {
 		return -1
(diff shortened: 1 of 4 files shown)


cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,219 @@
...
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	m.trim(txReader.GetSequence())
+	m.maybeResize(len(events))
+	m.append(events)
+	return nil
+}
+
+func readEvents(txReader *ingest.LedgerTransactionReader) ([]event, error) {
```
tsachiherman, line 146: this function requires to be executed under write lock; please add a comment about that.
( the implementation is fine; but we need to add a comment so it would be clear )

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,219 @@
...
+						Event:  uint32(eventIndex),
+					},
+				})
+			}
+		}
+	}
+	return events, nil
+}
+
+func (m *MemoryStore) trim(latestSequence uint32) {
```
tsachiherman, line 182: this function requires to be executed under write lock; please add a comment about that.
( the implementation is fine; but we need to add a comment so it would be clear )

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,219 @@
...
+		if c.Tx == other.Tx {
+			return cmp(c.Event, other.Event)
+		}
+		return cmp(c.Tx, other.Tx)
+	}
+	return cmp(c.Ledger, other.Ledger)
+}
+
+// MemoryStore is an in-memory store of soroban events.
+type MemoryStore struct {
```
tsachiherman, line 69: I think that there is some misaligment here between the retention window ( which is measured in ledgers) and the events array. Also, it looks like we would store ( in-memory ), redundant information.

It looks to me like the following would be a more efficient memory storage:
```golang
type MemoryStore struct {
        lock            sync.RWMutex
	eventLedgers          []eventLedger
        start            uint32 // starting index into the eventLedgers circular buffer.
}

type eventLedger struct {
   ledgerSeq int
   contractEvents []contractEvent
}
type contractEvent struct {
   xdr.ContractEvent
   // Tx is the index of the transaction within the ledger which emitted the event.
	Tx uint32
	// Op is the index of the operation within the transaction which emitted the event.
	Op uint32
	// Event is the index of the event within in the operation which emitted the event.
	Event uint32
}
```

when `Scan` is called, we could dynamically construct the needed data.
The MemoryStore.eventLedgers would be pre-initialize with the desired retention window.

The goal in the above schema is that there is no copy/realloc of any type during a new ledger.

cmd/soroban-rpc/internal/ledgerentry_storage/migrations/01_init.sql
```diff
@@ -10,11 +10,12 @@ CREATE TABLE metadata (
     value TEXT NOT NULL
 );
 
-
-
-CREATE INDEX ledger_entries_key ON ledger_entries(key);
-
+CREATE TABLE ledger_close_meta (
```
tsachiherman, line 13: I don't see why both the ledger entries and the events needs to live in the same database; they have quite a distinct an unrelated "characteristics" and not ever accessed together.
I'm concerned that placing both in the same database would result in a database-level locking between the two that isn't needed.

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -171,19 +191,35 @@ func (ls *ledgerEntryStorage) run(ctx context.Context, archive historyarchive.Ar
...
 			panic(err)
 		}
+		if err = tx.TrimLedgers(ls.retentionWindow); err != nil {
+			panic(err)
+		}
+		if err = tx.InsertLedger(ledger); err != nil {
+			panic(err)
+		}
 
+		reader.Rewind()
```
tsachiherman, line 222: seems wrong that we need to rewind and re-process. But maybe I'm not understanding the codeflow here correctly.

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,219 @@
...
+		return entry.cursor.Cmp(start) < 0
+	}) + 1
+
+	for ; i < m.length; i++ {
+		index := (m.start + i) % m.length
+		entry := m.events[index]
+		if entry.cursor.Cmp(end) >= 0 {
+			break
+		}
+		if !f(entry.cursor, entry.contents) {
```
tsachiherman, line 119: I don't see any usage of the `Scan` method, but I do have concerns here - it looks like it would ( potentially ) block the RLock for a long time. I can see how this would result in a deadlock if the caller would be blocking for writing operation.

As long as the caller is a "responsible" enough, this shouldn't be a concern. But maybe we should add a comment on the top of the function to alert the user of this danger ?

tsachiherman (commented):
I've made several comments. I'd be happy to review them together if any questions comes up !

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,219 @@
...
+		txMeta, ok := tx.UnsafeMeta.GetV3()
+		if !ok {
+			continue
+		}
+		for opIndex, op := range txMeta.Events {
+			if len(op.Events) == 0 {
+				continue
+			}
+			for eventIndex, opEvent := range op.Events {
+				events = append(events, event{
```
sreuland, line 167: is it worthwhile to offer an external config to be used here as event topic filters? as this is crunching events for the whole network, maybe rpc hosting use cases will only be interested in events for their contract or known topics, etc?

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,219 @@
...
+// remaining events in the range).
+func (m *MemoryStore) Scan(start, end Cursor, f func(Cursor, xdr.ContractEvent) bool) {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	i := sort.Search(m.length, func(i int) bool {
+		index := (m.start + i) % m.length
+		entry := m.events[index]
+		return entry.cursor.Cmp(start) < 0
+	}) + 1
```
paulbellamy, line 111: If we do @tsachiherman's circular buffer above, do we still need the `Search`? If we assume each index in `m.events` is a ledger, couldn't we directly calculate it something like this:
```go
oldest := m.events[m.start].cursor
if oldest > start {
  return fmt.Errorf("cursor out of range") // too old. Some error message here.
}
i := (start - oldest) % m.length
```

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,219 @@
...
+	return cmp(c.Ledger, other.Ledger)
+}
+
+// MemoryStore is an in-memory store of soroban events.
+type MemoryStore struct {
+	lock            sync.RWMutex
+	events          []event
+	length          int
+	start           int
+	retentionWindow uint32
```
2opremio, line 74: Please add a comment or rename it to indicate the unit in which the retention window is expressed.

cmd/soroban-rpc/internal/events/events.go
2opremio: I agree, this makes is much easier to deallocate the the events in a ledger.

It may make sense to extend this hierarchy and and separate the events for each transaction, operation in subslices

```go
type MemoryStore struct {
        lock            sync.RWMutex
	contractEvent    [][][]xdr.ContractEvent
        start           uint32 // starting index into the eventLedgers circular buffer.
        startLedger uint32 // ledger Sequence for the starting index
}
```

This comes with a few advantages and disadvantages

Advantages:
1. You can preallocate most (if not all) the memory, since the maximum transactions per ledger and maximum operations per transactions are known (I don't know if that's the case for events pero operation, but if that's the case, we can also preallocate those)
2. You can use native indexing and subslicing instead of trimming etc ...

Disadvantages:
We need to watch cache behavior being worse due to iterating through subslices

cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go
```diff
@@ -51,6 +65,8 @@ func NewLedgerEntryStorage(cfg LedgerEntryStorageCfg) (LedgerEntryStorage, error
 type ledgerEntryStorage struct {
 	logger            *log.Entry
 	db                DB
+	events            *events.MemoryStore
+	retentionWindow   uint32
```
2opremio, line 69: If possible, I would separate the event ingestion from `ledgerentry_storage.go`. We can have a common daemon consuming `CloseMeta` entries, which are passed to `ledgerentry_storage/ledgerentry_storage.go` and, say `events_storage/events_storage.go`.

Something similar (but probably more lightweight) than the Horizon processors would do.

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -25,15 +27,20 @@ const (
...
 	LedgerEntryStorage
 	GetLatestLedgerSequence() (uint32, error)
+	GetLedger(sequence uint32) (xdr.LedgerCloseMeta, bool, error)
+	GetAllLedgers() ([]xdr.LedgerCloseMeta, error)
 	NewLedgerEntryUpdaterTx(forLedgerSequence uint32, maxBatchSize int) (LedgerEntryUpdaterTx, error)
 }
 
 type LedgerEntryUpdaterTx interface {
+	TrimLedgers(retentionWindow uint32) error
+	InsertLedger(ledger xdr.LedgerCloseMeta) error
```
2opremio, line 43: I would move this into a separate interface since this is for Ledger entries

2opremio:
I think that mixing up the Ledger Entry and Events logic into the same files makes the code complicated. I would change the file hierarchy to decouple them and do the processing of each separately.

1. I would have a common module for the db stuff (like we have today but making sure we get the naming right)
2. A would have two other modules for the events and ledger_entry stuff which can work as the Horizon processors (that requires some orchestrator for captive-core which passes the ledger information to each processor).

cmd/soroban-rpc/internal/ledgerentry_storage/db.go
```diff
@@ -180,6 +187,70 @@ func (s *sqlDB) GetLedgerEntry(key xdr.LedgerKey) (xdr.LedgerEntry, bool, uint32
...
+	err = s.db.Select(&results, sqlStr, args...)
+	ledgers := make([]xdr.LedgerCloseMeta, len(results))
+	for i, ledger := range results {
+		ledgers[i] = xdr.LedgerCloseMeta(ledger)
+	}
+	return ledgers, err
+}
+
+// GetLedger fetches a single ledger from the db.
+func (s *sqlDB) GetLedger(sequence uint32) (xdr.LedgerCloseMeta, bool, error) {
```
sreuland, line 235: before sql gets more verbose, is it worth considering a type-safe compile time binding like go-jet.

cmd/soroban-rpc/internal/ledgerentry_storage/migrations/01_init.sql
tsachiherman: I do like the fact that we can commit the entire ledger changes in a single commit. In this particular case, I don't believe that it really give us much advantage. The events are only being "added" and not really requiring any synchronization with the rest of the data. You have already noticed that the current ledger data tend to become very large. That's because we have lots of write/delete. In order to avoid that, we need to perform periodic full vacuuming.
Unfortunately, performing the full vacuuming on a large database would take a long time - which is why it's advisable to break out the tables that have a different insert/delete profile into a separate database files.

tsachiherman: That sounds like the right thing to do. It would allow us to release this as a working prototype while we improve on the implementation.

paulbellamy:
Is this closed in favor of https://github.com/stellar/soroban-tools/pull/361, or still relevant?

tsachiherman:
> Is this closed in favor of #361, or still relevant?

yes; @tamirms created #361 in order to break this one into a smaller set of reviewable components.
So, he would need to rebase this one and keep working on completing this one.

PR #356, opened 2023-01-25
include captive-core-use-db command option

sreuland:
### What

include a `captive-core-use-db` command option to be passed to captive core config 

### Why

have more control of ram vs disk usage at deployment/runtime

### DIFF
--- a/cmd/soroban-rpc/main.go
+++ b/cmd/soroban-rpc/main.go
@@ -23,6 +23,7 @@ func main() {
 	var endpoint, horizonURL, stellarCoreURL, binaryPath, configPath, networkPassphrase, dbPath, captivecoreStoragePath string
 	var captiveCoreHTTPPort, ledgerEntryStorageTimeoutMinutes uint
 	var checkpointFrequency uint32
+	var useDb bool
 	var historyArchiveURLs []string
 	var txConcurrency, txQueueSize int
 	var logLevel logrus.Level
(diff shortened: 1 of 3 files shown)


2opremio (approved):

2opremio:
As the linter says, please rename `useDb` to `useDB`. Otherwise, it LGTM.

cmd/soroban-rpc/main.go
```diff
@@ -23,6 +23,7 @@ func main() {
 	var endpoint, horizonURL, stellarCoreURL, binaryPath, configPath, networkPassphrase, dbPath, captivecoreStoragePath string
 	var captiveCoreHTTPPort, ledgerEntryStorageTimeoutMinutes uint
 	var checkpointFrequency uint32
+	var useDb bool
```
tsachiherman, line 26: please change `useDb` -> `useDB`

tsachiherman (commented):
one comment above; looks great otherwise.

tsachiherman (approved):

cmd/soroban-rpc/main.go
sreuland: updated, I ran the `make lint-changes` locally, but it emits nothing in response, it should have shown this right? I probably have local config problem, installed golangci-lint from brew on mac.

tsachiherman: I'm not sure if the `lint-changes` works well after you've committed the changes. however, you should be able to run `make lint`, which should not generate any more warnings.

PR #361, opened 2023-01-27
soroban-rpc: Add in-memory event store backed by circular buffer

tamirms:
### What

Add in-memory event store backed by circular buffer. This code is taken from https://github.com/stellar/soroban-tools/pull/355 . To minimize merge conflicts, the ingestion and DB code changes will be included in a separate PR.

### Why

This is necessary to complete https://github.com/stellar/go/issues/4718

### Known limitations

[N/A]

### DIFF
--- a/cmd/soroban-rpc/internal/events/events.go
+++ b/cmd/soroban-rpc/internal/events/events.go
@@ -1,6 +1,7 @@
 package events
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -77,9 +78,9 @@ type event struct {
 	eventIndex uint32
 }
 
-func (e event) cursor(ledger uint32) Cursor {
+func (e event) cursor(ledgerSeq uint32) Cursor {
 	return Cursor{
-		Ledger: ledger,
+		Ledger: ledgerSeq,
 		Tx:     e.txIndex,
 		Op:     e.opIndex,
 		Event:  e.eventIndex,
@@ -92,9 +93,6 @@ type MemoryStore struct {
 	// buckets is a circular buffer where each cell represents
 	// all events occurring within a specific ledger.
 	buckets []bucket
-	// length is equal to the number of ledgers contained within
-	// the circular buffer.
-	length uint32
 	// start is the index of the head in the circular buffer.
 	start uint32
 }
@@ -108,10 +106,10 @@ type MemoryStore struct {
 // older entries outside the retention window.
 func NewMemoryStore(retentionWindow uint32) (*MemoryStore, error) {
 	if retentionWindow == 0 {
-		return nil, fmt.Errorf("retention window must be positive")
+		return nil, errors.New("retention window must be positive")
 	}
 	return &MemoryStore{
-		buckets: make([]bucket, retentionWindow),
+		buckets: make([]bucket, 0, retentionWindow),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


2opremio:
As [I mentioned private](https://github.com/stellar/soroban-tools/pull/355#discussion_r1086599491), I would consider some modifications to the data structure so that we can (mostly) allocate it in advance 

Also, it would be good to know how much memory could be consumed by this (e.g. assuming full ledgers and certain amount of events per operation)

2opremio (approved):

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+type event struct {
+	contents   xdr.ContractEvent
+	txIndex    uint32
+	opIndex    uint32
+	eventIndex uint32
+}
+
+func (e event) cursor(ledger uint32) Cursor {
```
tsachiherman, line 80: The ledger here is a ledger sequence number, right ? could you rename it to ledgerSeq ?

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,276 @@
...
+			if eventRange.End.Cmp(cur) <= 0 {
+				return nil
+			}
+			if !f(cur, event.contents) {
+				return nil
+			}
+		}
+		i = (i + 1) % uint32(len(m.buckets))
+		curLedger++
+		if m.buckets[i].ledgerSeq != curLedger {
```
tsachiherman, line 162: this for loop isn't incorrect, but it makes me nervous since there is no obvious termination condition.
can you move the `m.buckets[i].ledgerSeq != curLedger` to be the terminating condition for the loop ?
i.e.
```golang
for ;m.buckets[i].ledgerSeq == curLedger; curLedger++ {
...
}
```

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,276 @@
...
+		events = m.buckets[i].events
+	}
+}
+
+// validateRange checks if the range falls within the bounds
+// of the events in the memory store.
+// validateRange should be called with the read lock.
+func (m *MemoryStore) validateRange(eventRange *Range) error {
+	if m.length == 0 {
+		return fmt.Errorf("event store is empty")
```
tsachiherman, line 174: consider replacing with `errors.New`, since you're not really formatting anything in here.

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -0,0 +1,276 @@
...
+// [ latestLedger - retentionWindow, latestLedger ]
+// will be included in the MemoryStore. If the MemoryStore
+// is full, any events from new ledgers will evict
+// older entries outside the retention window.
+func NewMemoryStore(retentionWindow uint32) (*MemoryStore, error) {
+	if retentionWindow == 0 {
+		return nil, fmt.Errorf("retention window must be positive")
+	}
+	return &MemoryStore{
+		buckets: make([]bucket, retentionWindow),
```
tsachiherman, line 114: by having `length` member in `MemoryStore`, you're duplicating the inner member. I can't find any place where this is an issue in the codebase, but in general, it would be safer to use `make([]bucket, 0, retentionWindow)` and use `len(buckets)` to find the length of the buckets and `cap(buckers)` to check the max size.

tsachiherman (commented):
Few small comments, but looks great otherwise.

tsachiherman (approved):
looks great! thanks for the changes.

PR #366, opened 2023-01-30
soroban-rpc: Implement getNetwork method

paulbellamy:
### What

Implement the `getNetwork` method of soroban-rpc

### Why

So clients can fetch the current network passphrase, and friendbot URL.

### Known limitations

Right now, this just asks core directly for the current protocol version. We should probably cache this instead when we ingest ledgers. Currently the ledger ingestion is done by the LedgerEntryStorage system. We could add a method there for this (which feels a bit like a hack), or abstract it into a horizon-style ingestion pipeline (which seems like a faff). Opinions @tamirms ?


leighmcculloch (approved):

cmd/soroban-rpc/main.go
```diff
@@ -134,6 +134,13 @@ func main() {
 			},
 			Usage: "comma-separated list of stellar history archives to connect with",
 		},
+		{
+			Name:      "friendbot-url",
+			Usage:     "URL where friendbot requests should be sent",
```
tsachiherman, line 139: The description here isn't incorrect, but I'm not sure that it's what we want in this context.
i.e. I was thinking something in the lines of "The friendbot URL which would be returned by getNetwork endpoint".

The reason for that is that the soroban-rpc doesn't use this field in any way other then copy-paste it into it's responses.

cmd/soroban-rpc/internal/test/integration.go
```diff
@@ -35,10 +35,11 @@ import (
 
 const (
 	StandaloneNetworkPassphrase = "Standalone Network ; February 2017"
-	stellarCoreProtocolVersion  = 20
+	StellarCoreProtocolVersion  = 20
```
tsachiherman, line 38: any particular reason that we should export this const ?
given that it's being used as part of a test, I think that we should keep it as "local" as possible.

tsachiherman (approved):
Few comments, but looks great otherwise.

cmd/soroban-rpc/main.go
leighmcculloch: I thought it will be used by soroban-rpc for doing account funding for test networks. If it's not going to use it itself in anyway, it seems pretty strange for the soroban-rpc to offer it up as data or know about it at all.

cmd/soroban-rpc/internal/test/integration.go
paulbellamy: Yep, good catch. I was thinking it was a separate test package, but it is not. Will fix.

cmd/soroban-rpc/main.go
paulbellamy: The idea is that the client can call `getNetwork` to determine if there is a friendbot available on this network, and figure out how to request from it. This is replacing the proposed soroban-rpc `requestAirdrop` method, which would do the same thing server-side.

leighmcculloch: 👍

PR #371, opened 2023-01-31
update docker build for latest make target

sreuland:
### What

the docker build file for soroban rpc fails with error during go build:
```
#6 33.81 /usr/bin/ld: cannot find -lpreflight
```

### Why

need the build to work, as quickstart uses it to build soroban-rpc, updated the dockerfile to use the Makefile `build-soroban-rpc` target

### DIFF
--- a/cmd/soroban-rpc/docker/Dockerfile
+++ b/cmd/soroban-rpc/docker/Dockerfile
@@ -1,23 +1,23 @@
 FROM golang:1.19.1 as build
+ARG RUST_TOOLCHAIN_VERSION=stable
 
-ADD . /src/soroban-rpc
-WORKDIR /src/soroban-rpc
-RUN go build -o /bin/soroban-rpc ./cmd/soroban-rpc
+WORKDIR /go/src/github.com/stellar/soroban-tools
 
+ADD . ./
 
-FROM ubuntu:20.04
-ARG STELLAR_CORE_VERSION
-ENV STELLAR_CORE_VERSION=${STELLAR_CORE_VERSION:-*}
-ENV STELLAR_CORE_BINARY_PATH /usr/bin/stellar-core
+RUN git config --global --add safe.directory "/go/src/github.com/stellar/soroban-tools"
+
+ENV CARGO_HOME=/rust/.cargo
+ENV RUSTUP_HOME=/rust/.rust
+ENV PATH="/usr/local/go/bin:$CARGO_HOME/bin:${PATH}"
 ENV DEBIAN_FRONTEND=noninteractive
-
-# ca-certificates are required to make tls connections
-RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates curl wget gnupg apt-utils
-RUN wget -qO - https://apt.stellar.org/SDF.asc | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=true apt-key add -
-RUN echo "deb https://apt.stellar.org focal stable" >/etc/apt/sources.list.d/SDF.list
-RUN echo "deb https://apt.stellar.org focal unstable" >/etc/apt/sources.list.d/SDF-unstable.list
-RUN apt-get update && apt-get install -y stellar-core=${STELLAR_CORE_VERSION}
+RUN apt-get update
+RUN apt-get install -y build-essential
 RUN apt-get clean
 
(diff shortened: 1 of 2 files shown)


cmd/soroban-rpc/docker/Dockerfile
```diff
@@ -1,23 +1,23 @@
...
+ARG RUST_TOOLCHAIN_VERSION=stable
 
-ADD . /src/soroban-rpc
-WORKDIR /src/soroban-rpc
-RUN go build -o /bin/soroban-rpc ./cmd/soroban-rpc
+WORKDIR /go/src/github.com/stellar/soroban-tools
 
+ADD . ./
 
-FROM ubuntu:20.04
```
leighmcculloch, line 8: It looks like the second stage of the multistage is being removed. And there's no longer a stellar-core being installed in the second stage. I suspect that the stellar-core is still supposed to be present because soroban-rpc needs it to run. And the stellar-core should only be installed in the second stage, and the build tos like rust should only be installed in the first stage. 

The context is this keeps the final image smaller because it doesn't have the image layers that has the build tools, and any consumer of this image who only needs the built artifact can instruct docker to only build the first stage. That's what the quickstart build process does.

sreuland: I didn't see this Dockerfile being referenced from within this repo, such as by integration tests, those use docker-compose.yml.  @2opremio, @tamirms , do you know where this cmd/soroban-rpc/docker/Dockerfile is used? It appeared like maybe it was used as test fixture initially and then deprecated in favor of the docker compose. 

I cleaned it out to be single purposed for building soroban-rpc only. Removed the core installation in the docker as it's not needed for build, and was debian pkg based which would have run into arm64 hassles. The end stage of this build remains compatible to the way quickstart builds it:
```
docker build -t <tag> -f cmd/soroban-rpc/docker/Dockerfile --target build https://github.com/stellar/soroban-tools.git#<ref>

```

tsachiherman: I think that we currently don't publish the soroban-rpc docker image; the docker is being used as part of the "construction" of the quickstart image.

leighmcculloch: > I do think if we are going to release a soroban-rpc docker image we should include the core binary as it is now a requirement for captive core ingestion

+1 The design of the dockerfile looks like it is based on the design of the horizon dockerfile. I also assumed the intent was to publish an image  in the future hence why core is included. 

It looks like this pr is both fixing a bug with this image in that it can no longer build successfully, as well as refactoring the image. We could limit the PR to do one thing – fix the bug – and debate whether the refactor is appropriate independently.

tsachiherman (approved):

cmd/soroban-rpc/docker/Dockerfile
sreuland: @leighmcculloch @tsachiherman - I created new ticket to capture this deployment requirement, a new Dockerfile providing deployable soroban-rpc image, one that can be used in containerized environments, aws ecs, k8s/helm, etc. - https://github.com/stellar/soroban-tools/issues/382

leighmcculloch: > For building only, if the debian core pkg install is left in this Dockerfile, I think it may lead to problems

@sreuland I recommend looking at the Dockerfile.dev in the go repo for Horizon. It solves this problem by using a multi-stage image. Importantly, quickstart only builds the first stage, the `build` stage of the image. The build stage works for amd64 or arm64. Only the final stage of the image has a dependency on a deb that's only built for amd64.

leighmcculloch: Fwiw this Dockerfile before this change already addresses the issue of separating build and runtime into two stages where only the runtime stage has a hard dependency on amd64.

sreuland: yes, good point, the `from ubuntu:20.04` started a 2nd stage, which provides the separation as long as builder uses `--target`, so, will keep the debian core part in - https://github.com/stellar/soroban-tools/pull/371/commits/e621f014e64339a50e2d6eaa1a6a51c83b7704ae and I updated https://github.com/stellar/soroban-tools/issues/382 to reference re-use of this Dockerfile

PR #377, opened 2023-01-31
increased the rust toolchain version on e2e test config

sreuland:
### What

bumped toolchain version used on e2e test to be same as code config

### Why

prior e2e config was pinned at 1.66.0, code config has recently bumped to 1.67.0

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -18,7 +18,7 @@ jobs:
       SYSTEM_TEST_DOCKER_REPO_TAG: "2opremio/system-test"
       SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_HASH: "main"
       SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_REPO: "https://github.com/stellar/soroban-examples.git"
-      SYSTEM_TEST_RUST_TOOLCHAIN_VERSION: "1.66.0"
+      SYSTEM_TEST_RUST_TOOLCHAIN_VERSION: "stable"
       SYSTEM_TEST_CORE_DEBIAN_VERSION: "19.6.1-1158.c0ad35aa1.focal~soroban"
       SYSTEM_TEST_HORIZON_DEBIAN_VERSION: "2.22.0~soroban-322"
       SYSTEM_TEST_VERBOSE_OUTPUT: "true"


tsachiherman:
Where do we use this ? I'm asking since I think it would be better to derive this one from the compiler version used in the soroban-tools repo.

sreuland:
> Where do we use this ? I'm asking since I think it would be better to derive this one from the compiler version used in the soroban-tools repo.

it's used on the e2e gha test config here in tools - https://github.com/stellar/soroban-tools/pull/377/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR21

@tomerweller suggested similar in chat earlier to automate how this setting is obtained for tests to use, I think we should create a ticket to capture that, this is to fix it now, as pr's are getting stuck on this test fail.

tsachiherman:
> > Where do we use this ? I'm asking since I think it would be better to derive this one from the compiler version used in the soroban-tools repo.
> 
> it's used on the e2e gha test config here in tools - https://github.com/stellar/soroban-tools/pull/377/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR21
> 
> @tomerweller suggested similar in chat earlier to automate how this setting is obtained for tests to use, I think we should create a ticket to capture that, this is to fix it now, as pr's are getting stuck on this test fail.

Yes - I think that what I'm saying is that we should load this value from the root Cargo.toml. see https://github.com/stellar/soroban-tools/pull/362/files.

If you'd like to make a temp fix, I'm perfectly fine with that as well - just please don't forget to create an issue to track this. ( Otherwise, we'll forget about that until we'll start seeing weird compilation errors )

sreuland:
> > > Where do we use this ? I'm asking since I think it would be better to derive this one from the compiler version used in the soroban-tools repo.
> > 
> > 
> > it's used on the e2e gha test config here in tools - https://github.com/stellar/soroban-tools/pull/377/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR21
> > @tomerweller suggested similar in chat earlier to automate how this setting is obtained for tests to use, I think we should create a ticket to capture that, this is to fix it now, as pr's are getting stuck on this test fail.
> 
> Yes - I think that what I'm saying is that we should load this value from the root Cargo.toml. see https://github.com/stellar/soroban-tools/pull/362/files.
> 
> If you'd like to make a temp fix, I'm perfectly fine with that as well - just please don't forget to create an issue to track this. ( Otherwise, we'll forget about that until we'll start seeing weird compilation errors )

@tsachiherman captured the refactor to dynamic toolchain version value as new ticket - https://github.com/stellar/soroban-tools/issues/381

tsachiherman (approved):

PR #385, opened 2023-02-01
soroban-rpc: Use in-memory event store in getEvents handler

tamirms:
### What

This is the last remaining PR to complete https://github.com/stellar/go/issues/4718 . This PR uses the in-memory event store in the getEvents rpc handler instead of relying on Horizon.

### Why

Close https://github.com/stellar/go/issues/4718

### Known limitations

Still need to fix the tests in get_events_test.go , I'm currently working on that and will push a commit to the PR once I'm done

### DIFF
--- a/cmd/soroban-rpc/internal/events/events.go
+++ b/cmd/soroban-rpc/internal/events/events.go
@@ -192,6 +192,7 @@ func readEvents(networkPassphrase string, ledgerCloseMeta xdr.LedgerCloseMeta) (
 		var tx ingest.LedgerTransaction
 		tx, err = txReader.Read()
 		if err == io.EOF {
+			err = nil
 			break
 		}
 		if err != nil {


cmd/soroban-rpc/internal/events/events.go
```diff
@@ -3,72 +3,18 @@ package events
 import (
```
paulbellamy, line 3: Seems to stil be some references to `Cursor` here which is undefined?

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -42,19 +37,18 @@ type EventInfoValue struct {
...
-		return errors.New("endLedger must be after or the same as startLedger")
+	// Validate start
+	// Validate the paging limit (if it exists)
+	if g.StartLedger <= 0 {
+		return errors.New("startLedger must be positive")
 	}
-	if g.EndLedger-g.StartLedger > maxLedgerRange {
-		return fmt.Errorf("endLedger must be less than %d ledgers after startLedger", maxLedgerRange)
+	if g.Pagination != nil && g.Pagination.Limit > maxLimit {
+		return fmt.Errorf("limit must not exceed %d", maxLimit)
```
paulbellamy, line 51: Should make sure we update the docs for this as well, as currently the `limit` part is 200 (iirc)

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -259,223 +252,86 @@ type PaginationOptions struct {
...
-			return err
+			return nil, errors.Wrap(err, "invalid pagination cursor")
 		}
-
-		v3, ok := meta.GetV3()
-		if !ok {
-			return nil
+		// increment event index because, when paginating,
+		// we start with the item right after the cursor
+		start.Event++
```
paulbellamy, line 274: Again, should make sure we document this, as I think it might currently say it is inclusive.

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -259,223 +252,86 @@ type PaginationOptions struct {
...
-
-	// TODO: Use a more efficient backend here. For now, we stream all ledgers in
-	// the range from horizon, and filter them. This sucks.
-	cursor := EventID{
-		ID:         toid.New(request.StartLedger, int32(0), 0),
-		EventOrder: 0,
-	}
+	start := events.Cursor{Ledger: uint32(request.StartLedger)}
+	limit := maxLimit
 	if request.Pagination != nil && request.Pagination.Cursor != "" {
```
paulbellamy, line 266: You should be able to set the limit without a cursor.

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -259,223 +252,86 @@ type PaginationOptions struct {
...
 	}
 
-	eventType := EventTypeContract
-	if event.Type == xdr.ContractEventTypeSystem {
-		eventType = "system"
+	var eventType string
+	switch event.Type {
+	case xdr.ContractEventTypeSystem:
+		eventType = EventTypeSystem
+	case xdr.MaskAccountFlags:
```
paulbellamy, line 329: eh?

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -261,14 +217,16 @@ func (m *MemoryStore) append(sequence uint32, events []event) error {
 
 	if length < uint32(cap(m.buckets)) {
 		m.buckets = append(m.buckets, bucket{
-			ledgerSeq: sequence,
-			events:    events,
+			ledgerCloseTimestamp: ledgerCloseTimestamp,
```
tsachiherman, line 220: nice to have: move the bucket creation ahead of the if statement, so that you won't need to repeat the bucket creation on either of the cases.

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -207,16 +155,21 @@ func seek(events []event, cursor Cursor) []event {
 // IngestEvents adds new events from the given ledger into the store.
 // As a side effect, events which fall outside the retention window are
 // removed from the store.
-func (m *MemoryStore) IngestEvents(txReader *ingest.LedgerTransactionReader) error {
-	events, err := readEvents(txReader)
+func (m *MemoryStore) IngestEvents(networkPassphrase string, ledgerCloseMeta xdr.LedgerCloseMeta) error {
```
tsachiherman, line 158: The `networkPassphrase` here would never change with every ledgerClose event. We should be setting it up once.
Separately, the usage of that ( for cross referencing txHash ) is a bit.. questionable.
( i.e. it's not so obvious to me that the soroban-rpc need to perform validation on the incoming data )

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -243,11 +196,14 @@ func readEvents(txReader *ingest.LedgerTransactionReader) ([]event, error) {
 			}
 		}
 	}
+	if err := txReader.Close(); err != nil {
```
tsachiherman, line 199: could you defer that to ensure that we call Close in the case of an error ?
( the returned error in this case is a bit tricky )

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -5,25 +5,20 @@ import (
...
-	"github.com/stellar/go/toid"
 	"github.com/stellar/go/xdr"
+
+	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/events"
 )
 
 // maxLedgerRange is the maximum allowed value of endLedger-startLedger
 // Just guessed 4320 as it is ~6hrs
-const maxLedgerRange = 4320
+const maxLimit = 10000
```
tsachiherman, line 21: move this const to be configurable value; I'm not particular about the value itself, but would rather all the configuration options to reside in a single place.

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -259,223 +252,88 @@ type PaginationOptions struct {
...
-			return nil
+			return nil, errors.Wrap(err, "could not parse event")
 		}
+		results = append(results, info)
 	}
+	return results, nil
 }
 
-func eventInfoForEvent(opID *toid.ID, eventIndex int32, ledgerClosedAt string, event xdr.ContractEvent) (EventInfo, error) {
+func eventInfoForEvent(cursor events.Cursor, ledgerClosedAt string, event xdr.ContractEvent) (EventInfo, error) {
```
tsachiherman, line 321: could you reorder the arguments 
event xdr.ContractEvent, cursor events.Cursor, ledgerClosedAt string
?
right now, the first two are only being used when constructing the output EventInfo.
given that the entire function is about the "event", it should be the first argument.

cmd/soroban-rpc/internal/events/cursor_test.go
```diff
@@ -0,0 +1,84 @@
+package events
+
+import (
+	"github.com/stretchr/testify/assert"
```
tsachiherman, line 4: As you can see, github actions generated the following warning:
`File is not `goimports`-ed with -local github.com/golangci/golangci-lint (goimports)`

to work around it, just move the `github.com/stretchr/testify/assert` to the bottom.

tsachiherman (changes requested):
looks promising.. few small requests.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -134,13 +142,14 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
 		AccountStore:      methods.AccountStore{Client: hc},
-		EventStore:        methods.EventStore{Client: hc},
+		EventStore:        eventStore,
 		FriendbotURL:      cfg.FriendbotURL,
 		NetworkPassphrase: cfg.NetworkPassphrase,
 		Logger:            logger,
 		TransactionProxy:  transactionProxy,
 		CoreClient:        &stellarcore.Client{URL: cfg.StellarCoreURL},
 		LedgerEntryReader: db.NewLedgerEntryReader(dbConn),
+		MaxEventsLimit:    cfg.MaxEventsLimit,
```
tsachiherman, line 152: hmm. we might want to pass the entire `*cfg` instead of specifying specific fields. That would save us from adding fields over an over. but, that's for another PR.

cmd/soroban-rpc/internal/events/events.go
```diff
@@ -207,31 +163,47 @@ func seek(events []event, cursor Cursor) []event {
...
-func readEvents(txReader *ingest.LedgerTransactionReader) ([]event, error) {
-	var events []event
+func readEvents(networkPassphrase string, ledgerCloseMeta xdr.LedgerCloseMeta) (events []event, err error) {
+	var txReader *ingest.LedgerTransactionReader
+	txReader, err = ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(networkPassphrase, ledgerCloseMeta)
+	if err != nil {
+		return
+	}
+	defer func() {
+		err = txReader.Close()
```
tsachiherman, line 185: this is the "tricky" portion I mentioned in my comment...
I think that what you want in the following:
```golang
defer func() {
		closeErr := txReader.Close()
		if err == nil {
			err = closeErr
		}
}
```

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -494,28 +355,27 @@ func eventInfoForEvent(opID *toid.ID, eventIndex int32, ledgerClosedAt string, e
...
-		PagingToken:    id,
+		ID:             cursor.String(),
+		PagingToken:    cursor.String(),
 		Topic:          topic,
 		Value:          EventInfoValue{XDR: data},
 	}, nil
 }
 
+//
+//// maxLedgerRange is the maximum allowed value of endLedger-startLedger
```
tsachiherman, line 369: these need to be cleaned up, right ?

tsachiherman (approved):
I left few more comments, but don't want to block merging this whenever done.

PR #388, opened 2023-02-02
soroban-rpc: Auth Next

paulbellamy:
Staging area for soroban-rpc auth-next changes

### DIFF
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -98,8 +98,9 @@ type SimulateTransactionCost struct {
 }
 
 type InvokeHostFunctionResult struct {
-	XDR       string `json:"xdr"`
-	Footprint string `json:"footprint"`
+	XDR       string   `json:"xdr"`
+	Footprint string   `json:"footprint"`
+	Auth      []string `json:"auth"`
 }
 
 type SimulateTransactionResponse struct {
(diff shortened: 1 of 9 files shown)


tsachiherman (commented):
The change here looks very reasonable. One thing that I think we should work on is decoupling the C calls from the handler.
i.e. all the C calls should be placed in it's own package, exposing only GO datatypes. That would allow us to decouple the handler logic from the C wrapper code.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -171,10 +172,11 @@ func NewSimulateTransactionHandler(logger *log.Entry, networkPassphrase string,
 			}
 		}
 		li := C.CLedgerInfo{
-			network_passphrase: C.CString(networkPassphrase),
-			sequence_number:    C.uint(latestLedger),
-			protocol_version:   20,
-			timestamp:          C.uint64_t(time.Now().Unix()),
+			// TODO: Read network_id, 32 bytes
```
sreuland, line 175: can this TODO be resolved now, not sure what the blocker is?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -98,13 +98,14 @@ type SimulateTransactionCost struct {
 }
 
 type InvokeHostFunctionResult struct {
-	XDR string `json:"xdr"`
+	XDR       string   `json:"xdr"`
```
tsachiherman, line 101: Could we rename this field to something a bit more descriptive ?
i.e. "output", "returnValue", "outcome"

2opremio:
@paulbellamy should we close this?

paulbellamy:
Closed in favour of `auth-next2` branch

PR #389, opened 2023-02-02
soroban-rpc: Configure default limit, update cursor / startLedger validation, and include latest ledger for getEvents

tamirms:
### What

The max limit for `getEvents` is configured to be 10000. This commit sets the default limit to 100 (previously it was equal to the max limit).

This commit also updates the `getEvents` validation logic to accept either `startLedger` or `cursor` (meaning `startLedger` must be omitted if `cursor` is present) and adds a `latestLedger` field to the response.

### Why

The default should be lower than the max limit. By having a lower default we save some CPU cycles by avoiding lengthy requests which aren't necessary. This is also consistent with Horizon which configures a default lower than the maximum allowed limit.

The `latestLedger` field is necessary so that the client can understand if there are potentially additional events remaining beyond their query.
### Known limitations

[N/A]

### DIFF
--- a/cmd/soroban-cli/src/events.rs
+++ b/cmd/soroban-cli/src/events.rs
@@ -252,7 +252,7 @@ impl Cmd {
                 Some(self.count),
             )
             .await
-            .map_err(|e| Error::Rpc(e))
+            .map_err(Error::Rpc)
     }
 
     fn run_in_sandbox(


tsachiherman (commented):
few requests, but looks great otherwise.

tsachiherman (approved):

cmd/soroban-cli/src/events.rs
```diff
@@ -167,12 +167,11 @@ pub enum OutputFormat {
...
-        if self.start_ledger > self.end_ledger {
-            return Err(Error::InvalidLedgerRange {
-                low: self.start_ledger,
-                high: self.end_ledger,
-            });
-        }
+        let start = match (self.start_ledger, self.cursor.clone()) {
+            (Some(start), _) => rpc::EventStart::Ledger(start),
+            (_, Some(c)) => rpc::EventStart::Cursor(c),
+            _ => return Err(Error::MissingStart),
```
2opremio, line 173: If you use [required_unless_present](https://docs.rs/clap/latest/clap/struct.Arg.html#method.required_unless_present), we won't need this error (clap will make sure one of them is present)

cmd/soroban-cli/src/events.rs
```diff
@@ -196,8 +195,8 @@ impl Cmd {
         }
 
         let events = match (self.rpc_url.as_ref(), self.events_file.as_ref()) {
-            (Some(rpc_url), _) => self.run_against_rpc_server(rpc_url).await,
-            (_, Some(path)) => self.run_in_sandbox(path),
+            (Some(rpc_url), _) => self.run_against_rpc_server(rpc_url, start).await,
+            (_, Some(path)) => self.run_in_sandbox(path, start),
```
2opremio, line 199: [required_unless_present](https://docs.rs/clap/latest/clap/struct.Arg.html#method.required_unless_present) could also be used here (but it's unrelated to this PR)

cmd/soroban-cli/src/events.rs
```diff
@@ -275,30 +282,17 @@ impl Cmd {
...
-                // The ledger range is optional in sandbox mode.
-                if self.start_ledger == 0 && self.end_ledger == 0 {
-                    return true;
-                }
+            .filter(|evt| match parse_cursor(&evt.id) {
+                Ok(event_cursor) => event_cursor > start_cursor,
+                Err(e) => {
+                    eprintln!("error parsing key 'ledger': {e:?}");
+                    eprintln!(
+                        "your sandbox events file ('{}') may be corrupt",
```
2opremio, line 290: maybe we could suggest removing the file?

cmd/soroban-cli/src/rpc/mod.rs
```diff
@@ -278,8 +283,10 @@ impl Client {
         // TODO: cursor
```
2opremio, line 283: This can be removed, can't it?

cmd/soroban-rpc/main.go
```diff
@@ -21,7 +21,7 @@ import (
 
 func main() {
 	var endpoint, horizonURL, stellarCoreURL, binaryPath, configPath, friendbotURL, networkPassphrase, dbPath, captivecoreStoragePath string
-	var captiveCoreHTTPPort, ledgerEntryStorageTimeoutMinutes, maxEventsLimit uint
+	var captiveCoreHTTPPort, ledgerEntryStorageTimeoutMinutes, maxEventsLimit, defaultEventsLimit uint
```
2opremio, line 24: We may want to declare a `LocalConfig` variable instead of declaring every variable independently and then  initializing a local config, but maybe that's out of scope for this PR.

2opremio (approved):
only minor comments

PR #391, opened 2023-02-03
Update soroban-cli to support Auth next

tamirms:
### What

Update soroban-cli to support Auth next.

### Known limitations

* the strval changes only support unions with 1 type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1912,6 +1912,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde-aux"
+version = "4.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c599b3fd89a75e0c18d6d2be693ddb12cccaf771db4ff9e39097104808a014c0"
+dependencies = [
+ "chrono",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "serde_derive"
 version = "1.0.152"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -10,32 +10,33 @@ default-members = ["cmd/soroban-cli"]
...
 version = "0.4.1"
 git = "https://github.com/stellar/rs-soroban-sdk"
-rev = "e1c3de33942f0e997680645941787102ebf61e85"
+rev = "f2f4bf0e8197ee43fd6242c4a31943f466815bed"
 
 [workspace.dependencies.soroban-sdk]
 version = "0.4.1"
 git = "https://github.com/stellar/rs-soroban-sdk"
-rev = "e1c3de33942f0e997680645941787102ebf61e85"
+rev = "f2f4bf0e8197ee43fd6242c4a31943f466815bed"
```
sreuland, line 28: @tamirms , these sdk refs will bump to v0.5.0 tag correct?
https://github.com/stellar/rs-soroban-sdk/releases/tag/v0.5.0

paulbellamy (approved):
Since this incorporates the changes in `auth-next2` as well (looks like?), we should merge that into main after this.

The auth changes lgtm. But I'm not familiar with the tuple/void changes, so you may want to have @leighmcculloch check them.

paulbellamy (approved):

PR #392, opened 2023-02-03
update makefile

tsachiherman:
### What

The makefile was failing when building from scratch.

### Why

The makefile targets were not updated correctly when the preflightlib was added.

### Known limitations

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ Cargo.lock: Cargo.toml
 
 install_rust: Cargo.lock
 	cargo install --path ./cmd/soroban-cli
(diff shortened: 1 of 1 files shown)


sreuland:
looks good, do you by chance test the make through Dockerfile build also, I only ask b/c there's several remote systems(quickstart/jenkins) that run Make indirectly via that Dockerfile build:

```
 docker build -f cmd/soroban-rpc/docker/Dockerfile -t stellar/soroban-rpc:test --target build --no-cache .
```

sreuland (approved):

PR #394, opened 2023-02-03
feat!: contract functions are subcommands and docs are included

willemneal:
### What

Use new docs fields to improve the help generated by the CLI.

Instead of using the `--fn` argument the contract function is a subcommand. E.g.

```
soroban contract invoke ... -- hello
```

`-- --help` now displays a list of the contract methods with their description from the rust doc.

### Why

Provide a better way for users to understand what a function expects.  It acts like interactive documentation.

More intuitive for functions to be subcommands.

### Known limitations

Currently doesn't use all documentation provided, notably the parameters.


willemneal:
@leighmcculloch it seems to be working well.  I'm not using any docs from the fields here.

Another thought is now we could make each function a subcommand. 


```
soroban contract invoke --contract-id ... --fn contract_method -- --help
```

```
soroban contract invoke --contract-id ... -- contract_method --help
```

I made an MVP that if you don't use `--fn` it will print the help where the contract's methods are a list of subcommands with docs.

leighmcculloch:
The most recent SDK has docs on functions, types, fields of types, variants of enums. The XDR also has docs for function parameters, but the SDK doesn't put anything in them.

Dropping the `--fn` sounds good. I really like the idea that the list of functions will be displayed as a list of subcommands. It's easy to inspect.

paulbellamy:
If we're dropping the `--fn`, I'd vote for dropping the `--id` as well, and getting rid of the `slop` params. e.g. just having:
```
soroban contract invoke <contract_id> <fn_name> --arg1 val1
```

willemneal:
@paulbellamy It's hard to remove the slop.  Currently the solution is to create a new command on the fly. This differs from the derive solution we are currently using for the top level.  I had difficulty mixing a derived command with one that was to be parsed partially and then needs to look up the next args to parse.  This is why the slop helped because it let the extra args be arbitrary and be parsed later.

I would need to dig into clap to figure out how to handle the partial parsing, but it's definitely doable.

Is your objection to it because you think it's ugly?

How about

```bash
soroban contract invoke <contract_id> -- <fn_name> --arg1 val1
```

leighmcculloch:
The two character slop seems like a pretty minor detail, with a significantly large cliff to climb to eliminate it. It might be worth doing at some point, but maybe not at this stage. We still need to get feedback over time on how people use the new CLI UX.

paulbellamy:
> ```shell
> soroban contract invoke <contract_id> -- <fn_name> --arg1 val1
> ```

That seems like a perfectly fine compromise for now.

paulbellamy (approved):

Issue #348, opened 2023-01-23
soroban-rpc: Update SimulateTransaction code to support authNext


tsachiherman:
### What

Update SimulateTransaction code to support authNext

paulbellamy:
- Update the rust dependency on rs-soroban-env
- Return the new authorizations result from simulate_transaction back to the client.

paulbellamy:
Doc updates for this are in https://github.com/stellar/soroban-docs/pull/291

paulbellamy:
https://github.com/stellar/rs-soroban-env/pull/645/ is in review now, so we could use that git revision to start work before it is merged.

Issue #360, opened 2023-01-26
CLI: Auth Next Support


chadoh:
Soroban's [Auth Next](https://docs.google.com/document/d/1J-J3ClTUkrsLiJag906OH4hmNkZI3Jk6_Y9ZYt_psAI/edit#heading=h.mne05sjh63jf) feature is shaping up. The current idea is to supersede the `invoker` pattern, `soroban-auth`, and custom auth with a new, more-flexible approach.

As an example, let's consider two contract methods:

```rs
pub fn set_state(env: Env, acc: Account, state: State) {}

pub fn swap(env: Env, acc1: Account, acc2: Account, ...) {}
```

The idea here is that there is no implicit "invoker" added to `env`, but instead, anytime a smart contract author wants to require a signature from a given account, they use the Account type for one or more arguments. So `set_state` requires one signing account, but `swap` requires two. We could even imagine a function that requires an arbitrary number of signatures:

```rs
pub fn foo(env: Env, accs: Vec<Account>) {}
```

Each of these Account args have a complex structure, and would need to be passed as a JSON object looking something like:

```json
{"identity": "my_id", "invocations": [{ "contract_id": "C...", "function": "set_state", "args": [{"state": {...}]}
```

# Making this usable from the CLI

We have the prerequisites we need to add developer-friendly support for Auth Next to the CLI:

- #275
- #307

To start with, the CLI can allow passing all of this explicitly. Using the `set_state` example from above:

```bash
soroban invoke --contract-id $ID --fn set_state -- --acc '{ "identity": "my_id", "invocations": ... }' --state '{ ... }'
```

But paired with the `soroban config identity` feature (#275), it can also support passing Accounts implicitly. Perhaps this could look like:

```bash
soroban invoke --contract-id $ID --fn set_state -- --acc identity:chadoh --state '{ ... }'
```

Here, the "implicit CLI" built by `invoke` would check if the argument is a string prepended with "identity", and would perform the logic & preflight calls necessary to expand that argument into the full JSON structure expected by the smart contract.

Since we have full control over the implicit invoke CLI, we can build this in whatever best optimizes both maintenance and UX. Some other examples of how this might look, using the `swap` example from above:

```bash
# prefixed with `identity:`, as above
soroban invoke --contract-id $ID --fn swap -- --acc1 identity:chadoh --acc2 identity:ahalabs

# prefixed with just `id:`
[...]

PR #395, opened 2023-02-06
soroban-rpc: Add auth-next information to simulateTransaction response

2opremio:
Addresses #348 

TODO:
- [x] Fix tests


cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -196,11 +198,56 @@ fn preflight_host_function_or_maybe_panic(
...
+            nonce: n,
+        }),
+        // TODO: can the address and the nonce really be present independently?
+        _ => None,
+    };
+    ContractAuth {
+        address_with_nonce: address_with_nonce,
+        root_invocation: payload.invocation.clone(),
+        // TODO: ???
+        signature_args: Default::default(),
```
2opremio, line 247: I am not sure how to fill this info.

Any ideas @paulbellamy ?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -199,9 +200,29 @@ func NewSimulateTransactionHandler(logger *log.Entry, networkPassphrase string,
...
 		}
 
+		// Get the auth data
+		var auth []string
+		if res.auth != nil {
+			// CGo doesn't have an easy way to do pointer arithmetic so,
+			// we are better off transforming the memory buffer into a large slice
+			// and finding the NULL termination after that
+			for _, a := range unsafe.Slice(res.auth, 1<<20) {
+				if a == nil {
```
paulbellamy, line 210: If auth is a null-terminated array of null-terminated strings, won't this hit the end of the first string then stop?

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -196,11 +198,56 @@ fn preflight_host_function_or_maybe_panic(
         error: null_mut(),
         result: result_cstr.into_raw(),
         footprint: fp_cstr.into_raw(),
+        // TODO
```
paulbellamy, line 201: What is todo here?

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -196,11 +198,56 @@ fn preflight_host_function_or_maybe_panic(
...
+        auth: null_mut(),
         cpu_instructions: budget.get_cpu_insns_count(),
         memory_bytes: budget.get_mem_bytes_count(),
     })
 }
 
+fn recoded_auth_payloads_to_c(
```
paulbellamy, line 208:
```suggestion
fn recorded_auth_payloads_to_c(
```

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -196,11 +198,56 @@ fn preflight_host_function_or_maybe_panic(
...
+    let address_with_nonce = match (payload.address.clone(), payload.nonce) {
+        (Some(a), Some(n)) => Some(AddressWithNonce {
+            address: a,
+            nonce: n,
+        }),
+        // TODO: can the address and the nonce really be present independently?
+        _ => None,
+    };
+    ContractAuth {
+        address_with_nonce: address_with_nonce,
```
paulbellamy, line 244:
```suggestion
        address_with_nonce,
```

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -196,11 +198,56 @@ fn preflight_host_function_or_maybe_panic(
...
+    // TODO: replace by `out_vec.into_raw_parts()` once the API stabilizes
+    let ptr = out_vec.as_mut_ptr();
+    mem::forget(out_vec);
+
+    Ok(ptr)
+}
+
+fn recoded_auth_payload_to_xdr(payload: &RecordedAuthPayload) -> ContractAuth {
+    let address_with_nonce = match (payload.address.clone(), payload.nonce) {
+        (Some(a), Some(n)) => Some(AddressWithNonce {
```
paulbellamy, line 236:
```suggestion
        (Some(address), Some(nonce)) => Some(AddressWithNonce { address, nonce }),
```

cmd/soroban-rpc/internal/methods/simulate_transaction.go
paulbellamy: (same question about the deallocator as well)

cmd/soroban-rpc/lib/preflight/src/lib.rs
paulbellamy: Good question. I'd expect the args passed to `auth_required`, but those should already be captured as part of the `root_invocation`, so I'm not sure what this is. @dmkozh ?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: This is Checking the top-level array, the null or each string is checked by `C.GoString()`

paulbellamy: i.e. doesn't the `unsafe.Slice(...)` return a slice like:
```
hi<null>bye<null><null>
```

2opremio: > i.e. I expect that the `unsafe.Slice(...)` of `["hi", "bye"]` would return a slice like:
> 
> ```
> hi<null>bye<null><null>
> ```

That’s not the format Rust (or rather, C) will return.

`["hi", "bye"]` will be returned from Rust as:

```
pointer_to("hi\0"), pointer_to("bye\0"),NULL
```

paulbellamy: Got it. Thanks.

cmd/soroban-rpc/lib/preflight/src/lib.rs
paulbellamy: From @dmkozh We should leave this blank (or return a different type to exclude it). This is for the client to fill in when submitting the txn. Updated the comment.

2opremio:
The soroban rpc tests are passing. The system test and soroban-cli tests are failing (as expected since that’s dealt with at #360 )

After adding a simulateTransaction test which exercises auth next (the current tests don’t) we should be ready to go.

cmd/soroban-rpc/lib/preflight.h
```diff
@@ -15,6 +15,7 @@ typedef struct CPreflightResult {
     char *error; // Error string in case of error, otherwise null
     char *result; // SCVal XDR in base64
     char *preflight; // LedgerFootprint XDR in base64
+    char **auth; // NULL terminated array of XDR ContractAuths in base64
```
tsachiherman, line 18: given that you're packing on the auths in a single contagious array, I think that the "correct" datatype here is `char*` rather then `char**`.

2opremio: I am not packing the auths in a contiguous array.

tsachiherman: sorry - my bad; I wasn't reading this correctly.

2opremio: See https://github.com/stellar/soroban-tools/pull/395#discussion_r1097685983 in case it helps.

2opremio: It follows the same format as `**char argv` in C

paulbellamy: aiui it now, they're not actually packed in a single contiguous array. This main array is an array of pointers to the individual arrays of each string.

2opremio:
This should be ready to go. Note that the soroban-cli tests (including the Rust tests) and system tests are broken.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -97,16 +97,17 @@ type SimulateTransactionCost struct {
...
 }
 
 type SimulateTransactionResponse struct {
-	Error        string                     `json:"error,omitempty"`
-	Results      []InvokeHostFunctionResult `json:"results,omitempty"`
-	Footprint    string                     `json:"footprint"`
-	Cost         SimulateTransactionCost    `json:"cost"`
-	LatestLedger int64                      `json:"latestLedger,string"`
+	Error        string                      `json:"error,omitempty"`
+	Results      []SimulateTransactionResult `json:"results,omitempty"`
```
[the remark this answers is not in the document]
2opremio: Right now there can only be one because Core only accepts a host function invocation per transaction.

However, that may change in the future and a slice is used to future-proof it.

Cargo.toml
```diff
@@ -10,32 +10,32 @@ default-members = ["cmd/soroban-cli"]
 [workspace.dependencies.soroban-env-host]
 version = "0.0.12"
 git = "https://github.com/stellar/rs-soroban-env"
-rev = "993c527abce72748405d71467498bffd63e061c1"
+rev = "d65b04f1793c48471995bf67850fffee53b90bd2"
```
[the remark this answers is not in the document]
2opremio: I bumped it to what Dima suggested last night https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1675726203937539?thread_ts=1675725615.816279&cid=C030Z9EHVQE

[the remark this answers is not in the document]
2opremio: I just bumped it

paulbellamy (approved):

PR #397, opened 2023-02-06
soroban-rpc: Fix rollback errors in logs

tamirms:
### What

During ingestion we can see the following logs:

WARN[2023-02-06T21:16:22.733+01:00] could not rollback ingest write transactions  error="sql: transaction has already been committed or rolled back" pid=78232

This commit will ignore sql.ErrTxDone errors when rolling back a transaction.


2opremio:
Is it too hard to simply make sure we rollback/commit once?

I think that would be much cleaner.

tsachiherman (approved):

PR #399, opened 2023-02-09
Add option for running contract with unlimited budget

graydon:
Does what it says on the title


cmd/soroban-cli/src/contract/invoke.rs
```diff
@@ -46,6 +46,9 @@ pub struct Cmd {
     /// Output the cost execution to stderr
     #[clap(long = "cost")]
     cost: bool,
+    /// Run with an unlimited budget
+    #[clap(long = "unlimited-budget")]
```
paulbellamy, line 50:
```suggestion
    #[clap(long = "unlimited-budget", conflicts_with = "rpc-url")]
```
This only applies for local sandbox mode at the moment, so we can reject if doing a remote invoke.

paulbellamy (approved):
We probably want to have a more general `--max-budget 1000` option as well at some point, but this could live alongside. And to support for remote we need to add that to the soroban-rpc api anyway.

so lgtm

PR #401, opened 2023-02-09
fix: improve error messages and remove unused default args

### LINKED ISSUE #400: soroban-cli: config subcommand usability improvements
I'd suggest we should:
- Throw a more informative error if we are trying to use a network with no identity configured. Currently it just says something like, "failed to parse secret key".
- Add a pair of `soroban config identity|network default` commands to show/set the default for each. Currently you can only set them as default when creating the network or identity.
- Actually handle the `--default` flags to do something. Currently afaict it does nothing.

willemneal:
fixes #400


paulbellamy (approved):

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -17,9 +17,13 @@ pub enum Error {
     CurrentDirNotFound,
     #[error("Failed to create directory: {path:?}")]
     DirCreationFailed { path: PathBuf },
-    #[error("Failed to read secret's file: {path}")]
+    #[error(
+        "Failed to read secret's file: {path}.\nPropably need to use `soroban config identity add`"
```
willemneal, line 21:
```suggestion
        "Failed to read secret's file: {path}.\nProbably need to use `soroban config identity add`"
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -17,9 +17,13 @@ pub enum Error {
...
+    )]
     SecretFileRead { path: PathBuf },
-    #[error("Failed to read network file: {path}")]
+    #[error(
+        "Failed to read network file: {path};\nPropably need to use `soroban config network add`"
```
willemneal, line 25:
```suggestion
        "Failed to read network file: {path};\nProbably need to use `soroban config network add`"
```

PR #402, opened 2023-02-09
feat(invoke): support i128 & u128 arguments

chadoh:
### What

Support functions that use `u128` and `i128` arguments

### Why

The [advanced auth](https://soroban.stellar.org/docs/how-to%20guides/auth-advanced#run-the-contract) example uses `i128`. The [updated docs](https://github.com/stellar/soroban-docs/pull/301/files#diff-560b3b1cd6262a68eb4a4c2a45cbb8b836e106e004cf958e27de3d57c0d6bffaL657-R656) break without this CLI change.

### Known limitations

- [x] Support negative `i128` values (currently gets parsed by Clap as a flag)
- [ ] Support quoted values


willemneal:
Was able to get hyphens to work for negative int types. I don't think we need to handle the quoted case as it won't be needed if it's the top level arg.  If a i128 is used inside a more complex type then it will be quoted (or should be quoted and is an error otherwise).

PR #410, opened 2023-02-13
feat(cli): address command to read addresses of stored keys

willemneal:
### What

Add `address` sub command for `identity`.

For example,

```bash
soroban config identity address test-identity
```

For identities that are seed phrases there is an hd-path option. E.g.

```bash
soroban config identity address test-identity --hd-path 1
```

The default is zero.

### Why

Tests can now deterministically generate the same addresses from a single seed phrase. As shown in the tests provided.

### Known limitations

Might be confusing for new users. Though I found it very useful for writing tests for the examples.


paulbellamy:
> Though I found it very useful for writing tests for the examples.

Yeah, the idea came up because in the example-dapp we invoke the contract with `initialize(owner: Address, ...)`, and it's nice to use an identity for that owner address.

paulbellamy (approved):
LGTM. Maybe we should throw an error if they pass `--hd-path`, but the key is not a seed phrase? wdyt? Could prevent confusion, but could also be annoying in scripting...

PR #416, opened 2023-02-14
update cargo.toml to use the latest xdr versions.

tsachiherman:
### What

update cargo.toml to use the latest xdr versions.

### Why

new XDR, new dependencies.

### Known limitations

N/A


cmd/deptool/scanner.go
```diff
@@ -102,6 +102,9 @@ func addTomlDependencies(dependencies *projectDependencies, tomlDeps map[string]
 			fmt.Printf("Conflicting entries in Cargo.toml file :\n%v\nvs.\n%v\n", existing, current)
 			exitErr()
 		}
+		if current.githubPath == "" {
+			continue
```
sreuland, line 106: could guard check earlier after line 92 for `crateGit.Git = ""` instead ?

tsachiherman: good point. updated.

sreuland (approved):

PR #421, opened 2023-02-15
Bump Go version to 1.20.1

tamirms:
### What

Bump Go version to 1.20.1.

related: https://github.com/stellar/go/pull/4770

### Why

Go version 1.20.1 contains some important security updates that we need to include in our builds.

### Known limitations

[N/A]


mollykarcher (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/stellar/soroban-tools
 
-go 1.18
+go 1.20
```
2opremio, line 3: What's the reason for this? This enforces a minimum Go version of 1.20 to compile the project. I think it should be 1.18 if we want to support the latest two releases.

2opremio: (I am sure that there will be patch releases for 1.19 and 1.18 containing the important security fixes)

PR #430, opened 2023-02-16
feat: add `show` command and allow generating a secret key

### LINKED ISSUE #428: Add `identity generate foo --secret-key` flag
For the example dapp, it would be nice to use `soroban config identity generate token-admin` to generate the admin account. But `generate` outputs a seed phrase, whereas we need the secret key.

We could either:
- Add a `--secret-key` flag to `generate` that outputs an `S` key instead of a seed phrase.
- Add a subcommand similar to `identity address` which outputs the secret key
- Replace `identity address` with a more general command like `identity show`?

Thoughts, @willemneal?

_Originally posted by @paulbellamy in https://github.com/stellar/soroban-example-dapp/pull/87#discussion_r1108806403_

willemneal:
`config identity show` returns the identities corresponding private key.
fixes #428 

@paulbellamy As for renaming `address` any ideas?


cmd/soroban-cli/src/config/identity/show.rs
```diff
@@ -0,0 +1,38 @@
...
+        println!("{}", self.private_key()?.to_string());
+        Ok(())
+    }
+
+    pub fn private_key(&self) -> Result<stellar_strkey::ed25519::PrivateKey, Error> {
+        let res = locator::read_identity(&self.name)?;
+        let key = res.key_pair(self.hd_path)?;
+        Ok(stellar_strkey::ed25519::PrivateKey::from_payload(
+            key.public.as_bytes(),
+        )?)
```
Shaptic, line 36: Does this work even if it was generated with a seed phrase?

cmd/soroban-cli/src/config/identity/generate.rs
```diff
@@ -15,18 +15,35 @@ pub enum Error {
 pub struct Cmd {
     /// Name of identity
     pub name: String,
+
     /// Optional seed to use when generating seed phrase.
     /// Random otherwise.
     #[clap(long, short = 's')]
```
Shaptic, line 21: just a thought - what about `p` for `phrase` instead of `s` so that it doesn't conflict with `S` (then `S` can be `s`)?

cmd/soroban-cli/src/config/identity/generate.rs
```diff
@@ -15,18 +15,35 @@ pub enum Error {
...
     pub seed: Option<String>,
 
+    /// Output the generated identity as a secret
+    #[clap(long, short = 'S')]
+    pub as_secret: bool,
+
     #[clap(flatten)]
     pub config_locator: locator::Args,
+
+    /// When generating a seceret key, which hd_path should be used from the original seed_phrase.
```
Shaptic, line 31: typo:
```suggestion
    /// When generating a secret key, which hd_path should be used from the original seed_phrase.
```

Shaptic (approved):
> As for renaming `address` any ideas?

maybe `resolve`? or why not just have `show` display both?

cmd/soroban-cli/src/config/identity/show.rs
willemneal: Yep! key_pair uses the hd_path if the identity was a seed phrase. This is also why the default is a seed phrase.

willemneal:
How should they be displayed? As JSON?

Could show have two subcommands (private/public) or should they be conflicting flags?

cmd/soroban-cli/src/config/identity/generate.rs
willemneal: This seed to generate the seed phrase and must be 16 characters long. Seed seems short enough in my opinion so perhaps just the other one gets 's'

Shaptic:
It'd be nice to have it easily CLI-parseable e.g. with `awk -F' '` or `head -n1` or something rather than JSON since it stays human readable. Two plaintext lines to stdout seems fine to me :man_shrugging: 

As for the flags, maybe showing the secret by default is a "security issue" (or more like unexpected given its secret nature)? So you could have a `--both` lol. I don't have a strong opinion either way, though.

cmd/soroban-cli/src/config/identity/generate.rs
Shaptic: Fair enough, no strong opinion :)

cmd/soroban-cli/src/config/identity/generate.rs
```diff
@@ -15,18 +15,45 @@ pub enum Error {
...
+
     #[clap(flatten)]
     pub config_locator: locator::Args,
+
+    /// When generating a secret key, which hd_path should be used from the original seed_phrase.
+    #[clap(long)]
+    pub hd_path: Option<usize>,
+
+    /// Generate the default seed phrase. Useful for testing.
+    /// Equilivant to --seed 0000000000000000
```
Shaptic, line 36: nit: typo 
```suggestion
    /// Equivalent to --seed 0000000000000000
```

paulbellamy:
Bit late here, and not to use `gpg` as an example of good UX, but `gpg` uses `gpg --list-keys` and `gpg --list-secret-keys` (along with `--export-key` and `--export-secret-key`).

Issue #373, opened 2023-01-31
soroban-rpc: Deprecate and remove `getAccount` method


paulbellamy:
### What problem does your feature solve?

`getAccount` is just a ledger entry lookup. We can use `getLedgerEntry` in the client to implement this without specific server-side support. This is already done in `js-soroban-client`.

### What would you like to see?

Deprecate `getAccount` in soroban-docs, and remove it from the server next release.

PR #439, opened 2023-02-20
Remove soroban-rpc getAccount method

paulbellamy:
Fixes https://github.com/stellar/soroban-tools/issues/373

### What

Remove soroban-rpc's getAccount method.

### Why

The jsonrpc method for this was marked as deprecated in the latest release, and the same functionality is implemented in the client using the `getLedgerEntry` jsonrpc method.

### Known limitations

[TODO or N/A]


cmd/soroban-rpc/internal/test/transaction_test.go
```diff
@@ -11,12 +11,54 @@ import (
...
+}
+
+func getAccount(client *jrpc2.Client, address string) (xdr.AccountEntry, error) {
+	decoded, err := strkey.Decode(strkey.VersionByteAccountID, address)
+	if err != nil {
+		return xdr.AccountEntry{}, err
+	}
+	var key xdr.Uint256
+	copy(key[:], decoded)
+	keyXdr, err := xdr.LedgerKey{
```
Shaptic, line 33: nit: this `err` isn't checked (thanks golangci)

cmd/soroban-cli/src/rpc/mod.rs
```diff
@@ -181,11 +186,20 @@ impl Client {
...
+                stellar_strkey::ed25519::PublicKey::from_string(address)?.0,
+            ))),
+        });
+        let response = self.get_ledger_entry(key).await?;
+        if let LedgerEntryData::Account(entry) =
+            LedgerEntryData::read_xdr_base64(&mut response.xdr.as_bytes())?
+        {
+            Ok(entry)
+        } else {
+            panic!("Invalid soroban-rpc response")
```
Shaptic, line 201: why not return error?

Shaptic (approved):
small nits, otherwise :shipit:

cmd/soroban-rpc/internal/test/transaction_test.go
```diff
@@ -11,12 +11,57 @@ import (
...
 
 	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/methods"
 )
 
+type AccountInfo struct {
+	ID       string
+	Sequence string
+}
+
+func getAccount(client *jrpc2.Client, address string) (xdr.AccountEntry, error) {
```
2opremio, line 26: @paulbellamy this doesn't seem to be used? I think its uses (which make sense) got lost in the last rebase.

I removed the old uses of `getAccount()` at #438 because it broke the tests but the same checks still make sense when using `getLedger()`

paulbellamy: Yeah good catch. I think I resolved the merge wrong.

2opremio: I don't think this got solved (`getAccount()` is still unused)

PR #441, opened 2023-02-21
Ingest feebump transactions

tamirms:
### What

Ingest feebump transactions so that we can lookup a transaction by both its inner hash and outer hash. I also added a `FeeBump` field to the `Transaction` struct so that the client can determine how to parse the result.


2opremio (approved):
Good catch!

PR #443, opened 2023-02-21
feat: reorg into a proper library

willemneal:
### What

Currently the crate is both a library and a binary. However, it's a weird mix of the two with the library only exposes a small number of modules.

This PR reorganizes it into a library first and then makes the binary import the needed functionality for its main function.

Also adds a CommandParser trait that is implemented for all commands allowing them to be constructed with a string.

```rs
let cmd = contract::invoke::Cmd::parse("--contract .. -- method --arg .").unwrap();
```

### Why

This is the first step to allowing tests and other crates to use the CLI's functionality directly without needed to use an external command.

Currently the tests are beginning to take longer and this is due to needed to execute a command for each interaction and lots of file i/o.

### Known limitations

The current design of the commands need to be revamped so that they return a string or buffer, which then get's printed by the make root.run method.  This will allow users to construct the needed commands, execute them and get the results without needed to go through `assert_cmd` or the normal `Command`.


cmd/soroban-cli/src/commands/completion.rs
```diff
@@ -26,6 +26,6 @@ pub struct Cmd {
 impl Cmd {
     pub fn run(&self) {
         let cmd = &mut Root::command();
-        generate(self.shell, cmd, env!("CARGO_BIN_NAME"), &mut io::stdout());
+        generate(self.shell, cmd, "soroban", &mut io::stdout());
```
willemneal, line 29: This env var wasn't present for a library build so this was failing.  If this is important I can make the 'main.rs` pass this value.

paulbellamy (approved):

PR #460, opened 2023-02-24
fix: allow using ENV to add identity and pass secret

willemneal:
### What
```
soroban config identity add <NAME>
```

Now accepts SOROBAN_SECRET_KEY to set the secret and --secret-key also accepts it.
### Why

Needed for SQ

### Known limitations

[TODO or N/A]


paulbellamy (approved):
Conflicts with https://github.com/stellar/soroban-tools/pull/459, but that needs more work, so merging this first.

PR #465, opened 2023-03-02
feat: add better docs for UDT and add examples for each arg

willemneal:
### What

Each contract method argument provides a full type for UDT and add an example input.

Examples:

```
ssome_types 


USAGE:
    some_types [OPTIONS]

OPTIONS:
        --address <Address>                          Example:
                                                       --address GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4
        --bool                                       Example:
                                                       --bool 
    -h, --help                                       Print help information
        --i32 <i32>                                  Example:
                                                       --i32 -1
        --option <Option<u32>>                       Example:
                                                       --option 1
        --set <Set<Address>>                         Example:
                                                       --set '[ "GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4" ]'
        --strukt <{ a: u32, b: bool, c: Symbol }>    This is from the rust doc above the struct Test
                                                     Example:
                                                       --strukt '{ "a": 1, "b": true, "c": "hello" }'
        --symbol <Symbol>                            Example:
                                                       --symbol hello
        --tuple <Tuple<Symbol, Address>>             Example:
                                                       --tuple '["hello", "GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4"]'
        --vec <Array<u32>>                           Example:
                                                       --vec '[ 1 ]'
```
```
othertypes 


USAGE:
    othertypes [OPTIONS]

OPTIONS:
        --bytes <hex_bytes>                       Example:
                                                    --bytes beefface123
        --bytes_n <9_hex_bytes>                   Example:
                                                    --bytes_n efefefefefefefefe
        --const_enum <11 | 12 | 13>               Example:
                                                    --const_enum 11
    -h, --help                                    Print help information
        --map <Map<Symbol, u128>>                 Example:
                                                    --map '{"hello": "1000" }'
        --simple_enum <First | Second | Third>    Example:
                                                    --simple_enum First
```

### Why

Currently it is difficult to understand how an XDR argument should be encoded into JSON.  By providing the type info and an example it should make it easier for users.

### Known limitations

Struct and map examples should include two different options for different shells.  For *sh's you can do the following `'{"a": 1}'` vs `"{\"a\": 1}"`.

### DIFF
--- a/cmd/soroban-cli/src/contract/invoke.rs
+++ b/cmd/soroban-cli/src/contract/invoke.rs
@@ -27,7 +27,7 @@ use soroban_spec::read::FromWasmError;
 
 use crate::config;
 use crate::rpc::Client;
-use crate::strval::{arg_value_name, Spec};
+use crate::strval::Spec;
 use crate::utils::{create_ledger_footprint, default_account_ledger_entry};
 use crate::HEADING_SANDBOX;
 use crate::{events, rpc, strval, utils};
@@ -585,7 +585,7 @@ fn build_custom_cmd<'a>(name: &'a str, spec: &Spec) -> Result<clap::App<'a>, Err
             .value_parser(clap::builder::NonEmptyStringValueParser::new())
             .long_help(spec.doc(name, type_).unwrap());
 
-        if let Some(value_name) = arg_value_name(spec, type_) {
+        if let Some(value_name) = spec.arg_value_name(type_, 0) {
             arg = arg.value_name(Box::leak(value_name.into_boxed_str()));
         }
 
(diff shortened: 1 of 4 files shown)


willemneal:
Another issue is that in the examples any type that's a JSON string doesn't have "s around it when it's the arg type. However, if used inside another type e.g. struct, it requires the quotes.

I didn't provide the "s initially because I felt it looked bad and would signal to the user that they are required.

However, I think it will cause confusion down the line.

paulbellamy (approved):

paulbellamy (approved):

Issue #382, opened 2023-01-31
Deployment Image for Soroban RPC


sreuland:
### What problem does your feature solve?

1. Add multi-arch(linux/amd64 and linux/arm64) support for building the image from cmd/soroban-rpc/Dockerfile.

2. Document the image in cmd/soroban-rpc/docker/README:
*  describe current limitation on linux/amd64 builds only 
* how to build the image from command line using provided cmd/soroban-rpc/docker/Makefile
  ```
  STELLAR_CORE_VERSION="19.10.1-1310.6649f5173.focal~soroban" SOROBAN_RPC_VERSION=0.8.0-21 make docker-build
  ```
* define the release process for pushing new versions of the image for a soroban release.
  * implement the necessary GH CI workflow to be triggered automatically to build/push image to dockerhub when a GH release is created on soroban-tools.

sreuland:
@satyamz , when the new docker image is published from this effort, `stellar/soroban-rpc`, we should update [soroban-rpc helm chart](https://github.com/stellar/helm-charts/blob/main/charts/soroban-rpc/values.yaml#L22) to refer to it. Also, soroban rpc runtime config parameters have progressed, there are new settings now for captive core configurations needed. I think can use the[ latest quickstart environment config for rpc](https://github.com/stellar/quickstart/blob/master/futurenet/soroban-rpc/etc/soroban-rpc.env) as reference to latest settings include in helm chart. Noticed the helm chart spins up a core container on futurenet already, so, the captive core config settings would just need to be tailored to point to that instance for peer'ing.

satyamz:
> when the new docker image is published from this effort, stellar/soroban-rpc, we should update [soroban-rpc helm chart](https://github.com/stellar/helm-charts/blob/main/charts/soroban-rpc/values.yaml#L22) to refer to it.

You can open a PR for that whenever docker image is ready. Also you would need to update the image `tag` [here](https://github.com/stellar/helm-charts/blob/main/charts/soroban-rpc/values.yaml#L24)


>  Noticed the helm chart spins up a core container on futurenet already, so, the captive core config settings would just need to be tailored to point to that instance for peer'ing.


Captive core settings needs to be updated on the futurenet horizon which is not included in the soroban-rpc helm chart.

jacekn:
[...]

PR #470, opened 2023-03-03
soroban rpc deployment manifest

### LINKED ISSUE #418: soroban-rpc: deploy public instance on futurenet

We'd like to facilitate more engagement/usage of RPC in the ecosystem prior to launching on testnet. We eventually plan to host a public instance on testnet as well, but not pubnet.

A deployed public instance of soroban-rpc on futurenet on stellar domain,`rpc-futurenet.stellar.org`. Deployment should be integrated into the soroban release process so that it receives updates appropriately.

Initial deployment to use k8s and the quickstart soroban-dev image for soroban rpc, and ingress route for dns.

This initial deployment will later be refactored to use new rpc service deployment image from #382 and updated helm chart to use/config that image.

- #382 as a pre-requisite/blocker
- quickstart in its current form (potentially faster/easier, but to be followed up prior to testnet release by #382)

sreuland:
### What

added a reference manifest to deploy soroban rpc to a k8s cluster. This manifest uses quickstart dockerhub image for the soroban-rpc runtime as a first iteration of deployment.

### Why

provide a running instance of soroban rpc on futurenet, for convenience to explore and test with as a reference example, this instance is meant for learning/discovery, not meant to be used for real/prod scenarios.

Closes #418 

### Known limitations

this relies on quickstart image reference for soroban-rpc runtime, is meant as first iteration, longer term k8s deployments should use a dedicated deployment image for rpc


2opremio:
Shouldn’t we just use https://github.com/stellar/helm-charts/tree/main/charts/soroban-rpc ?

cmd/soroban-rpc/docker/k8s_deployment.yml
```diff
@@ -0,0 +1,128 @@
...
+      - name: clean-volume
+        image: busybox:1.28
+        command: ["rm", "-rf", "/opt/stellar/soroban-rpc/captive-core"]
+        volumeMounts:
+        - mountPath: /opt/stellar/soroban-rpc
+          name: captive-core-data
+      containers:    
+      - name: rpc-server
+        image: stellar/quickstart:soroban-dev@sha256:81c23da078c90d0ba220f8fc93414d0ea44608adc616988930529c58df278739
+        args: ["--futurenet", "--enable-soroban-rpc", "--logs"]
```
jacekn, line 47: This is meant to be separate from the soroban-rpc helm chart yes?

cmd/soroban-rpc/docker/k8s_deployment.yml
```diff
@@ -0,0 +1,128 @@
...
+                - /bin/sh
+                - -c
+                - | 
+                  curl --location --request POST 'http://localhost:8000/soroban/rpc' \
+                    --header 'Content-Type: application/json' \
+                    --data-raw '{
+                      "jsonrpc": "2.0",
+                      "id": 10235,
+                      "method": "getHealth"
+                    }' | jq --exit-status 'has("jsonrpc")'
```
jacekn, line 65: Nice. Is this recommended way to check health of the service? If so I'd like to incorporate it in the helm chart too

cmd/soroban-rpc/docker/k8s_deployment.yml
```diff
@@ -0,0 +1,128 @@
...
+metadata:
+  annotations:
+    cert-manager.io/cluster-issuer: default
+    ingress.kubernetes.io/ssl-redirect: "true"
+    kubernetes.io/ingress.class: public
+    nginx.ingress.kubernetes.io/rewrite-target: /soroban/rpc
+  name: soroban-rpc-futurenet
+spec:
+  rules:
+  - host: rpc-futurenet.prototypes.kube001.services.stellar-ops.com
```
jacekn, line 115: Maybe we can keep the hostname generic?

cmd/soroban-rpc/docker/k8s_deployment.yml
sreuland: for readiness probe concerns, this should be adequate, it validates json-rpc service is responsive, it doesn't check further functional readiness, i.e. the response of `getHealth` , which could be `result:healthy` or `error` depending on the underlying ingestion state which rpc has ongoing with it's captive core, but in that case we don't want to allow k8s to restart the service due to probe failures as the error response means ingestion is in-process of catch up.

cmd/soroban-rpc/docker/k8s_deployment.yml
sreuland: yes, this manifest is intended as short-term gap to provide deployment up for now. Next up is to create formal deployment artifact for rpc on https://github.com/stellar/soroban-tools/issues/382, that ticket addresses creation of a dedicated dockerfile definition for rpc runtime image(includes captive core config) and integrate that to the helm chart.

sreuland:
> Shouldn’t we just use https://github.com/stellar/helm-charts/tree/main/charts/soroban-rpc ?

this manifest was done to provide stop gap deployment now using quickstart image which has the captive core config for latest rpc version wired up.

the helm chart is using older image/verrsion of rpc, doesn't have captive core yet. The next step https://github.com/stellar/soroban-tools/issues/382, it will shore up this rpc deployment, updating it's dockerfile for multi-stage, build, and runtime with captive core, and pushing as a dockerhub image, then use that from the helm.

jacekn:
> the helm chart is using older image/verrsion of rpc, doesn't have captive core yet. The next step #382, it will shore up this rpc deployment, updating it's dockerfile for multi-stage, build, and runtime with captive core, and pushing as a dockerhub image, then use that from the helm.

I update the linked issue but maybe we should aim to not have captivecore built into the image? Docker best practice is one process per container and bundling captive core with soroban-rpc goes against that. Maybe the design should be discussed a bit more before we commit to it?

cmd/soroban-rpc/docker/k8s_deployment.yml
sreuland: I added alias for `rpc-futurenet.stellar.org` on ingress, https://github.com/stellar/soroban-tools/pull/470/commits/7635d42a8cffe0ad1aa56fef5dfe778bd16655cd , does this address that concern?

sreuland:
> > the helm chart is using older image/verrsion of rpc, doesn't have captive core yet. The next step #382, it will shore up this rpc deployment, updating it's dockerfile for multi-stage, build, and runtime with captive core, and pushing as a dockerhub image, then use that from the helm.
> 
> I updated the linked issue but maybe we should aim to not have captivecore built into the image? Docker best practice is one process per container and bundling captive core with soroban-rpc goes against that. Maybe the design should be discussed a bit more before we commit to it?

Yes, see your [comment on #382](https://github.com/stellar/soroban-tools/issues/382#issuecomment-1456624880), we can continue that design discussion as part of scope on that effort, sounds good, I think is one reason why have taken the approach here with standalone manifest referring to existing known working config in quickstart/rpc to enable rapid initial deployment in short term, but with acknowledgement that [#382](https://github.com/stellar/soroban-tools/issues/382) is needed as follow-on iteration for formalized single-purpose(rpc) deployment.

jacekn (approved):

sreuland:
@mollykarcher , this is effectively done, this k8s manifest has been used to provide the short term deployment on k8s cluster, I don't think it's worth merging this as the manifest is only for short term usage to support ops, can refer to the file here on pr as needed, I will close the pr for completeness.

PR #484, opened 2023-03-06
Add support for soroban-cli --version flag

paulbellamy:
### What

Use `clap` flags to enable support for:
```sh
$ soroban -V
soroban 0.6.0 (v0.6.1-19-g1b1557497b33f15f114d9b278385c20ec898bf0a)

$ soroban --version
soroban 0.6.0 (v0.6.1-19-g1b1557497b33f15f114d9b278385c20ec898bf0a)
soroban-env 0.0.14 (d06aaddca61f011cc64ec098b464233423197c3a)
soroban-env interface version 29
stellar-xdr 0.0.14 (55f47d302a3bbcd34cf32bfcd28abccfaeffc5e0)
xdr next (df18148747e807618acf4639db41c4fd6f0be9fc)
```

### Why

Because I keep typing `soroban --version` and expecting it to work, so other people probably do too.

### Known limitations

This does not remove `soroban version` subcommand. However the `-V/--version` flags override any subcommands. So:
```sh
$ soroban -V contract # if we specify both -V and a subcommand:
soroban 0.6.0 (v0.6.1-19-g1b1557497b33f15f114d9b278385c20ec898bf0a)
```

### DIFF
--- a/cmd/soroban-cli/src/main.rs
+++ b/cmd/soroban-cli/src/main.rs
@@ -19,10 +19,10 @@ const HEADING_RPC: &str = "OPTIONS (RPC)";
 #[derive(Parser, Debug)]
 #[clap(
     name = "soroban",
-    version,
+    version = Box::leak(Box::new(version::short())).as_str(),
+    long_version = Box::leak(Box::new(version::long())).as_str(),
     about = "https://soroban.stellar.org",
     disable_help_subcommand = true,
-    disable_version_flag = true
 )]
 #[clap(global_setting(AppSettings::DeriveDisplayOrder))]
 struct Root {
--- a/cmd/soroban-cli/src/version.rs
+++ b/cmd/soroban-cli/src/version.rs
@@ -10,19 +10,27 @@ pub struct Cmd;
 impl Cmd {
     #[allow(clippy::unused_self)]
     pub fn run(&self) {
-        println!("soroban {} ({GIT_REVISION})", env!("CARGO_PKG_VERSION"));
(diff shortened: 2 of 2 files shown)


cmd/soroban-cli/src/main.rs
```diff
@@ -19,10 +19,10 @@ const HEADING_RPC: &str = "OPTIONS (RPC)";
 #[derive(Parser, Debug)]
 #[clap(
     name = "soroban",
-    version,
+    version = Box::leak(Box::new(version::short())).as_str(),
+    long_version = Box::leak(Box::new(version::long())).as_str(),
```
paulbellamy, line 23: Annoyingly these have to be leaked so they are `'static` lifetime for clap.

tsachiherman (approved):

cmd/soroban-cli/src/main.rs
Shaptic: they can't be callables? :/

Shaptic (approved):

2opremio (approved):

cmd/soroban-cli/src/main.rs
paulbellamy: Unfortunately, clap wants a `&'help str` :(

Well.. it actually wants an `IntoResettable<Str>`  (https://docs.rs/clap/latest/clap/builder/struct.Command.html#method.version), so we could build a struct that implements that, or is convert-able to Str or something, but... I figured it wasn't worth it for now.

PR #487, opened 2023-03-08
Remove unused `config` cargo crate

paulbellamy:
### What

Remove the `config` cargo crate dep.

### Why

`cargo +nightly udeps --all-targets`, says it is unused. And afaict it has never been used. I think it was added accidentally in https://github.com/stellar/soroban-tools/pull/275, can you confirm @willemneal ?

### Known limitations

[TODO or N/A]


willemneal:
We had added it with plans to use it, but indeed never did. Good catch!

tsachiherman (approved):

PR #488, opened 2023-03-08
soroban-rpc: Decouple simulateTransaction method and preflight computation

2opremio:
### What

Decouple the `simulateTransaction` method handling and the preflight compuation

### Why

In preparation for limiting the concurrency of preflight computation to the number of processors (although it will be fully configurable)

Also, the code is much cleaner this way

### Known limitations

N/A


cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if res.error != nil {
+		return Preflight{}, errors.New(C.GoString(res.error))
+	}
+
+	// Get the auth data
+	var auth []string
+	if res.auth != nil {
+
+		// CGo doesn't have an easy way to do pointer arithmetic so,
```
tsachiherman, line 158: could you fix the above "unnecessary leading newline (whitespace)"?

tsachiherman (approved):
looks good - could you fix the above comment ?

PR #489, opened 2023-03-09
soroban-rpc: Limit preflight-computation concurrency through a worker pool

2opremio:
### What

Limit preflight-computation concurrency through a worker pool

The number of workers can be configured through `--preflight-worker-count` and defaults to the number of CPUs in the machine.

### Why

preflight-obtention is CPU-bound, according to @tsachiherman it is a good idea to limit it to the number of course. 

### Known limitations

I am on the fence about this though (the OS will be self-limiting in this regard). Also, although the worker pool code is not overly complicated it does add some complexity.

### DIFF
--- a/cmd/soroban-rpc/internal/preflight/pool.go
+++ b/cmd/soroban-rpc/internal/preflight/pool.go
@@ -2,6 +2,7 @@ package preflight
 
 import (
 	"context"
+	"errors"
 	"sync"
 
 	"github.com/stellar/go/support/log"
@@ -25,6 +26,7 @@ type PreflightWorkerPool struct {
 	ledgerEntryReader db.LedgerEntryReader
 	networkPassphrase string
 	logger            *log.Entry
+	done              chan struct{}
 	requestChan       chan workerRequest
 	wg                sync.WaitGroup
 }
@@ -34,6 +36,7 @@ func NewPreflightWorkerPool(workerCount uint, ledgerEntryReader db.LedgerEntryRe
 		ledgerEntryReader: ledgerEntryReader,
 		networkPassphrase: networkPassphrase,
 		logger:            logger,
+		done:              make(chan struct{}),
 		requestChan:       make(chan workerRequest),
 	}
 	for i := uint(0); i < workerCount; i++ {
@@ -45,14 +48,19 @@ func NewPreflightWorkerPool(workerCount uint, ledgerEntryReader db.LedgerEntryRe
 
 func (pwp *PreflightWorkerPool) work() {
 	defer pwp.wg.Done()
-	for request := range pwp.requestChan {
-		preflight, err := GetPreflight(request.ctx, request.params)
-		request.resultChan <- workerResult{preflight, err}
+	for {
+		select {
+		case <-pwp.done:
+			return
+		case request := <-pwp.requestChan:
+			preflight, err := GetPreflight(request.ctx, request.params)
+			request.resultChan <- workerResult{preflight, err}
+		}
 	}
 }
 
 func (pwp *PreflightWorkerPool) Close() {
-	close(pwp.requestChan)
+	close(pwp.done)
 	pwp.wg.Wait()
 }
 
@@ -66,6 +74,8 @@ func (pwp *PreflightWorkerPool) GetPreflight(ctx context.Context, sourceAccount 
 	}
 	resultC := make(chan workerResult)
 	select {
+	case <-pwp.done:
+		return Preflight{}, errors.New("preflight worker pool is closed")
 	case pwp.requestChan <- workerRequest{ctx, params, resultC}:
 		result := <-resultC
 		return result.preflight, result.err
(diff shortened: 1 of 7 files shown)


cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -0,0 +1,72 @@
...
+	params := PreflightParameters{
+		Logger:             pwp.logger,
+		SourceAccount:      sourceAccount,
+		InvokeHostFunction: op,
+		NetworkPassphrase:  pwp.networkPassphrase,
+		LedgerEntryReader:  pwp.ledgerEntryReader,
+	}
+	resultC := make(chan workerResult)
+	defer close(resultC)
+	pwp.requestChan <- workerRequest{ctx, params, resultC}
```
paulbellamy, line 69: Probably needs a timeout/cancel in case the pool is full for too long.

cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -0,0 +1,72 @@
...
+	result := <-resultC
```
tsachiherman, line 70: due to the above issue, this line might be hunged up forever..

tsachiherman (changes requested):
I believe there is a corner case issue here that need to be addressed..

cmd/soroban-rpc/internal/preflight/pool.go
2opremio: closed channels return inmediately

tsachiherman: I don't think that resultC can be a closed channel here. ( nor does the `defer close(resultC)` really needed ).
But you're correct that in the current code, it can't hang:
- if the write to pwp.requestChan was successful, it means that the channel is still open, and there is a processing goroutine.
- if there is a processing go routine, then resultC should get called, eventually.

tsachiherman (changes requested):

cmd/soroban-rpc/internal/preflight/pool.go
2opremio: I guess we can read from the `ctx`'s  channel at the same time

tsachiherman: yes, selecting between `pwp.requestChan <-` and `ctx.Done()` sounds like the right thing to do.
Also, if `ctx.Done()` is the one selected, return ctx.Err

cmd/soroban-rpc/main.go
```diff
@@ -246,6 +255,10 @@ func main() {
 				)
 				os.Exit(-1)
 			}
+			if serviceConfig.PreflightWorkerCount < 1 {
+				fmt.Print("preflight-worker-count must be > 0")
```
tsachiherman, line 259: eol ?

cmd/soroban-rpc/internal/preflight/pool.go
2opremio: Yep, exactly my thought

cmd/soroban-rpc/internal/preflight/pool.go
2opremio: When saying _closed channels return immediately_ I was referring to `pwp.requestChan`

And, you are right, I forgot unused (and unreachable) channels are garbage-collected. I removed the close.

cmd/soroban-rpc/main.go
2opremio: Good catch

2opremio:
@tsachiherman @paulbellamy PTAL

tsachiherman (approved):
changes looks great.

thanks!

tsachiherman:
thinking about that again - I think that we'll also need to return 429 if the pool is full. Otherwise, we're just enqueueing infinite amount of blocking go routines.

tsachiherman:
it can easily be implemented by sizing the requestChan and adding a default case to the select.

2opremio:
Wouldn't it be better to just set a timeout?

tsachiherman:
> Wouldn't it be better to just set a timeout in the http handle?

I think that timeout should be additive. passing a context would a deadline would do the trick, right ?

2opremio:
> I think that timeout should be additive. passing a context would a deadline would do the trick, right ?

I do think that the client should be the one deciding how much to wait when the workers are busy. What's the downside of doing that?

tsachiherman:
> > I think that timeout should be additive. passing a context would a deadline would do the trick, right ?
> 
> I do think that the client should be the one deciding how much to wait when the workers are busy. What's the downside of doing that?

two reasons for having local timeout ( regardless of client-side timeout ):
- control recovery timeout in case of full queue.
- limit to reasonable execution time. A processing time of 10 second is something we don't want to allow by definition.

2opremio:
I think we should start by measuring how long it takes to compute the preflights

2opremio:
Do you mind if I merge as is and we have an offline conversation about the rest?

PR #491, opened 2023-03-09
Update to Clap v4

### LINKED ISSUE #180: CLI: Update clap-rs to v4

chadoh:
Fixes #180 

### What

Update [clap](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md) and [clap_complete](https://docs.rs/clap_complete) to version 4

### Why

- it's good to be up-to-date 
- clap v4 was released months ago and there's lots of good stuff we're missing out on
- it will allow us to [generate man pages](https://lib.rs/crates/clap_mangen) based off of the `--help` docs we already generate. We will also be able to publish these man pages as web pages.

### Known limitations

- needed to change all `conflicts_with` arguments to snake_case rather than kebab-case; see bug report:
   - https://github.com/clap-rs/clap/pull/4759


cmd/soroban-cli/src/commands/events.rs
```diff
@@ -67,7 +66,11 @@ pub struct Cmd {
     /// Though the specification supports multiple filter objects (i.e.
     /// combinations of type, IDs, and topics), only one set can be specified on
     /// the command-line today, though that set can have multiple IDs/topics.
```
paulbellamy, line 68: Seems like this change means we support multiple now, so we could update the comment.

chadoh: I think the behavior remains unchanged.

Prev: `multiple = true, max_values(5)`
New: `num_args = 1..=5`

Looking at the comment on `topic` below, I think this comment is trying to tell us that if you pass a combination of `--id`, `--topic`, and `--type`, they will all be combined with AND, rather than OR. I did not change the filtering logic, so I believe this should still be true.

Changing the logic or even the comment seems outside the scope of this PR.

chadoh:
@paulbellamy I'm having trouble figuring out why system-test is failing. I get no error on the specific step (`cargo build` of hello world example) locally. When I try actually running system-tests locally, I get a different error, but I might be running it incorrectly. What I tried:

- in one terminal, I run:
   ```
    docker run --rm -it -p 8000:8000 --name stellar stellar/quickstart:soroban-dev --standalone --enable-soroban-rpc --enable-core-artificially-accelerate-time-for-testing
   ```
- in `system-test` repo, I run:
   ```
   SorobanExamplesGitHash="main" SorobanExamplesRepoURL="https://github.com/stellar/soroban-examples.git" TargetNetworkPassPhrase="Standalone Network ; February 2017" TargetNetworkSecretKey="SC5O7VZUXDJ6JBDSZ74DSERXL7W3Y5LTOAMRF7RQRL3TAGAPS7LUVG3L"  TargetNetworkPublicKey="GBZXN7PIRZGNMHGA7MUUUF4GWPY5AYPV6LY4UV2GL6VJGIQRXFDNMADI" TargetNetworkRPCURL="http://localhost:8000/soroban/rpc" VerboseOutput=false go test -v --run "^TestDappDevelop$/^DApp developer compiles, deploys and invokes a contract.*$" ./features/dapp_develop/...
   ```

cmd/soroban-cli/src/commands/events.rs
paulbellamy: Oh yeah. You're right. That comment is somewhat confusingly worded...

paulbellamy (approved):

willemneal:
@chadoh I don't see your commit that fixed the other kebab issue we found yesterday.

chadoh:
I rebased it in; it was [this `key_xdr` line](https://github.com/AhaLabs/soroban-tools/blob/build/update-clap/cmd/soroban-cli/src/commands/contract/read.rs#L24)

cmd/soroban-cli/src/commands/config/identity/add.rs
```diff
@@ -10,14 +11,15 @@ pub enum Error {
 }
 
 #[derive(Debug, clap::Args)]
+#[group(skip)]
```
chadoh, line 14: pasting this here for visibility and indexing 

looks like the named ArgGroup stuff is there mostly to enable future features; We use the same struct names ALL OVER THE PLACE so it would be quite a pain to give them all unique names; that's why the group(skip) stuff is needed

PR #492, opened 2023-03-09
soroban-rpc: Miscelaneous fixes

2opremio:
* Print cmdline parsing errors to stderr (and uniformly)
* Do not use negative exit error codes (we should be using positive exit codes for errors, e.g. in Unix they go from 0 to 255)
* Make sure we deallocate resources after the webserver is stopped and not while it's happening
* Remove spurious exit code return value from the daemon


cmd/soroban-rpc/main.go
```diff
@@ -296,12 +294,10 @@ func main() {
 	cmd.AddCommand(versionCmd)
 
 	if err := configOpts.Init(cmd); err != nil {
-		supportlog.New().WithError(err).Fatal("could not parse config options")
-		os.Exit(-1)
+		fmt.Fprintf(os.Stderr, "could not parse config options: %v\n")
```
tsachiherman, line 297: fmt.Fprintf format %v reads arg #1, but call has 0 args

cmd/soroban-rpc/main.go
```diff
@@ -267,8 +266,7 @@ func main() {
 			serviceConfig.LedgerEntryStorageTimeout = time.Duration(ledgerEntryStorageTimeoutMinutes) * time.Minute
 			serviceConfig.CoreRequestTimeout = time.Duration(coreTimeoutSeconds) * time.Second
 			serviceConfig.MaxHealthyLedgerLatency = time.Duration(maxHealthyLedgerLatencySeconds) * time.Second
-			exitCode := daemon.Run(serviceConfig, endpoint)
-			os.Exit(exitCode)
+			daemon.Run(serviceConfig, endpoint)
```
tsachiherman, line 269: not sure why you'd ignoring the deamon.Run return.. it could be useful to let the Deamon return an error code.

cmd/soroban-rpc/main.go
```diff
@@ -296,12 +294,10 @@ func main() {
...
 	if err := configOpts.Init(cmd); err != nil {
-		supportlog.New().WithError(err).Fatal("could not parse config options")
-		os.Exit(-1)
+		fmt.Fprintf(os.Stderr, "could not parse config options: %v\n")
 	}
 
 	if err := cmd.Execute(); err != nil {
-		supportlog.New().WithError(err).Fatal("could not run")
-		os.Exit(-1)
+		fmt.Fprintf(os.Stderr, "could not run: %v\n", err)
```
tsachiherman, line 301: If we have an error which makes us exit the process, we should probably exit with a non-zero. This is particulary important for automation scripts.

cmd/soroban-rpc/main.go
2opremio: We can add it when/if needed, right now the daemon just starts  the webserver, which returns nothing

cmd/soroban-rpc/main.go
2opremio: Yep, it was an oversight. Thanks

2opremio:
@tsachiherman PTAL

cmd/soroban-rpc/main.go
tsachiherman: yes, I just reviewed `support.http.Run`; admittedly, I'm not a big fan of that - I'd rather having a single location ( i.e. main function ) have all the os.Exit calls.
But I agree that at this point it's not needed.

tsachiherman (approved):

PR #493, opened 2023-03-09
Upgrade wasm-opt to 0.112.0

brson:
### What

Just keeping the crate up to date.

There don't appear to be significant changes in binaryen 112:

https://github.com/WebAssembly/binaryen/blob/main/CHANGELOG.md#v112

### Known limitations

n/a


paulbellamy (approved):
Thanks!

PR #517, opened 2023-03-15
soroban-rpc: Add simulateTransaction job queue and error when it fills up

2opremio:
Also, add cache for `LedgerEntryReadTx`'s `GetLatestLedgerSequence()` because:
1. It doesn't change within the transaction
2. We were awkwardly passing the latest ledger around to avoid extra calls


cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -70,13 +72,34 @@ func NewSimulateTransactionHandler(logger *log.Entry, getter PreflightGetter) jr
 			}
 		}
 
-		result, err := getter.GetPreflight(ctx, sourceAccount, xdrOp)
+		jobBufferLen, jobBufferCapacity := pwp.GetJobQueueLenAndCapacity()
+		if jobBufferLen != 0 && jobBufferLen == jobBufferCapacity {
+			return SimulateTransactionResponse{
+				Error: "All workers are busy, try again later",
+			}
+		}
```
2opremio, line 80: We discussed returning a 429 HTTP error here but, unfortunately that's not really how JSON RPC works but:

1. I don't think that's how you are supposed to use JSON RPC (you should have a response object for every request). Note that you get a stream of JSON request objects and the client expects a matching stream of responses. See https://www.jsonrpc.org/specification .
2. The Go library we are using doesn't allow overriding the HTTP error we are getting.

We may want to reserve and return one of the `-32000 to -32099` server errors for this though.

I am aware this will make it more difficult for loadbalancers to detect and I am open to alternatives (like having a global approach for all endpoints).

cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -48,34 +48,38 @@ func NewPreflightWorkerPool(workerCount uint, ledgerEntryReader db.LedgerEntryRe
...
-		LedgerEntryReader:  pwp.ledgerEntryReader,
+		LedgerEntryReadTx:  readTx,
 	}
 	resultC := make(chan workerResult)
 	select {
-	case <-pwp.done:
-		return Preflight{}, errors.New("preflight worker pool is closed")
 	case pwp.requestChan <- workerRequest{ctx, params, resultC}:
 		result := <-resultC
 		return result.preflight, result.err
```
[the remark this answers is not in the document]
2opremio: That was my first approach, but I thought it was better to let the pool user choose. In hindsight I think it's cleaner like you suggest. I will implement it

cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -26,18 +27,17 @@ type PreflightWorkerPool struct {
...
 	networkPassphrase string
 	logger            *log.Entry
-	done              chan struct{}
+	isClosed          atomic.Bool
 	requestChan       chan workerRequest
 	wg                sync.WaitGroup
 }
 
-func NewPreflightWorkerPool(workerCount uint, ledgerEntryReader db.LedgerEntryReader, networkPassphrase string, logger *log.Entry) *PreflightWorkerPool {
+func NewPreflightWorkerPool(workerCount uint, jobQueueCapacity uint, ledgerEntryReader db.LedgerEntryReader, networkPassphrase string, logger *log.Entry) *PreflightWorkerPool {
```
paulbellamy, line 35: There's probably a refactor we could do (at some point), to:
```go
func NewJsonRPCMethodWorkerPool(
  workerCount uint,
  jobQueueuCapacity uint,
  handler jrpc2.Handler,
) *WorkerPool
```
Or even using generics for the handler 🙈

2opremio: I had actually started some work work in those lines at https://github.com/2opremio/soroban-cli/commit/afa1c5daff32c19516026071b363f6b5fa6945e2 .

cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -48,38 +48,42 @@ func NewPreflightWorkerPool(workerCount uint, ledgerEntryReader db.LedgerEntryRe
...
+	}
+	close(pwp.requestChan)
 	pwp.wg.Wait()
 }
 
-func (pwp *PreflightWorkerPool) GetPreflight(ctx context.Context, sourceAccount xdr.AccountId, op xdr.InvokeHostFunctionOp) (Preflight, error) {
+var PreflightQueueFullErr = errors.New("preflight queue full")
+
+func (pwp *PreflightWorkerPool) GetPreflight(ctx context.Context, readTx db.LedgerEntryReadTx, sourceAccount xdr.AccountId, op xdr.InvokeHostFunctionOp) (Preflight, error) {
+	if pwp.isClosed.Load() {
```
tsachiherman, line 69: this would not, unfortunately, provide adequate protection against writing to closed channels.
my suggestion is that you would use a bool channel instead, and select on that one.

PR #518, opened 2023-03-15
feat: auto-generate comprehensive CLI docs

chadoh:
### What

- Add a separate `doc-gen` binary to the CLI project; needs to be run with a feature flag so that we can avoid adding its dependencies (currently only [clap-markdown](https://docs.rs/clap-markdown)) to the main project's dependencies.
- Run new bin with `cargo run --bin doc-gen --features clap-markdown` to generate a markdown documentation page in the `docs/` folder, so people can read about the entire CLI all together
- Link to this from `soroban --help` output
- Start strategically adding both long help (available with `--help`; gets included in this comprehensive doc) and short help (available with just `-h`) for root `soroban` help docs and some subcommands (only `soroban contract invoke --help`, for now)

### Why

This fixes the frustration of the current CLI, which requires typing `--help` for every specific subcommand. Especially since soroban-cli was reorganized into more nested subcommands, it has become difficult to understand all possible functionality quickly.

### Known limitations

- I originally wanted these docs to be available with `man soroban`, similar to `man git`. However, not even `cargo` ships with a man page, so maybe this is not yet a common pattern with CLIs authored in Rust.
- The `--help` info now links to `https://github.com/stellar/soroban-tools/tree/main/docs/soroban-cli-full-docs.md`. We may want to create a short link?
- The long help (`soroban --help`) now mostly duplicates the "CLI Reference" [in the docs](https://soroban.stellar.org/docs/reference/command-line). I will create a follow-up PR to the docs repo that shrinks that page down to a stub that links people to the doc in GitHub.
- The easiest thing to do with this markdown doc is to put it here in this same repo, and commit it to source control so that it can be viewed as HTML. However, this means that we will now be committing a built file. As always when committing built files, this has two risks: 1. People may manually edit the file. 2. People may forget to commit an updated version. As discussed in [the review below](https://github.com/stellar/soroban-tools/pull/518#pullrequestreview-1343872669), I've attempted to prevent these problems two ways:
   1. Add a README file to `docs/README.md` explaining that this directory contains auto-generated files, and that they should not be manually edited.
   2. Added the task to the pre-commit hook we encourage people to install.
   3. Added a GitHub Action that checks if the doc file is up-to-date, and explains what to do if it is not (see [example failure here](https://github.com/AhaLabs/soroban-tools/actions/runs/4470915330/jobs/7855105646))

### DIFF
--- a/.github/workflows/full-help-docs.yml
+++ b/.github/workflows/full-help-docs.yml
@@ -1,0 +1,32 @@
+name: CLI Help Doc
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  pull-requests: read
+
+jobs:
+  doc_check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: stellar/actions/rust-cache@main
+
+    - name: Generate help doc
+      # this looks goofy to get GITHUB_OUTPUT to work with multi-line return values;
+      # see https://stackoverflow.com/a/74266196/249801
+      run: |
+        cargo run --bin doc-gen --features clap-markdown
+        raw_diff=$(git diff docs/soroban-cli-full-docs.md)
+        if [ "$raw_diff" != "" ]; then echo ""; echo "Unexpected docs change:"; echo "======================="; echo ""; echo "$raw_diff"; echo ""; echo "======================="; echo ""; fi
+        echo diff=$raw_diff >> $GITHUB_OUTPUT
+      id: doc-gen
+
+    - name: Check diff
+      if: steps.doc-gen.outputs.diff != ''
+      uses: actions/github-script@v3
+      with:
+        script: |
+          core.setFailed('Expected `doc-gen` to generate no diffs, but got diff shown in previous step.\n\nUpdate the full help docs:\n\n    cargo run --bin doc-gen --features clap-markdown\n\nDo this automatically on every commit by installing the pre-commit hook as explained in the README.')
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -4,3 +4,6 @@ set -e
 
 echo '+cargo fmt --check'
 cargo fmt --check || (cargo fmt && exit 1)
+
+echo '+cargo run --bin doc-gen --features clap-markdown'
+cargo run --bin doc-gen --features clap-markdown
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -253,6 +253,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap-markdown"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "325f50228f76921784b6d9f2d62de6778d834483248eefecd27279174797e579"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_complete"
 version = "4.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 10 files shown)


docs/soroban-cli-help.md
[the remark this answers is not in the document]
chadoh: @paulbellamy I agree we would ideally not commit it, or we would have it created by CI. But I'm not aware of a way to actually accomplish that. Having it in the `/docs` folder makes it easy to read as HTML, and gives us a low-maintenance pipeline for publishing these docs. We can then link to them from `soroban --help`, which I've done in the most recent commits here.

> Add a comment at the top noting how it was autogenerated, and when.

Since the whole file is auto-generated, adding a comment at the top seems difficult. I've added a README to the `/docs` folder explaining that the files within are auto-generated. Do you think that's a good enough substitute?

> Add a ci job to check that the committed matches what is generated.

Great idea, I'll look into this.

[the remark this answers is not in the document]
chadoh: @paulbellamy I added a GH Action to check the generated file; see the failure message here https://github.com/AhaLabs/soroban-tools/actions/runs/4470915330/jobs/7855105646

If you think the `docs/README.md` + GH Action approach works for now, then I think this is ready to merge!

.github/workflows/full-help-docs.yml
```diff
@@ -0,0 +1,26 @@
...
+      run: |
+        echo "xtask=$(cargo xtask markdown && git status -s docs/soroban-cli-full-docs.md)" >> $GITHUB_OUTPUT
+      id: run_generate
+
+    - name: Check diff
+      if: ${{ steps.run_generate.outputs.xtask }} != ""
+      uses: actions/github-script@v3
+      with:
+        script: |
+            core.setFailed('You need to run `cargo xtask markdown` to update the full help docs. You can also install the pre-commit hook as explained in the README to do this automatically on every commit.')
```
paulbellamy, line 26:
```suggestion
            core.setFailed('You need to run `cargo xtask markdown` to update the full help docs. You can also install the pre-commit hook as explained in the README to do this automatically on every commit, or run: cargo xtask markdown')
```

paulbellamy (approved):

.github/workflows/full-help-docs.yml
chadoh: This made it a bit redundant 😄

chadoh: I more-or-less changed this back; I hope you don't mind. Having the `cargo xtask markdown` in there twice seems unnecessary to me.

chadoh:
@paulbellamy 🙈 

(working on it; I know no good way to troubleshoot GH Actions)

willemneal:
@chadoh I just pushed a fix to the githook since you reference it in the error message.

chadoh:
@willemneal the error message mentions the pre- _commit_ hook, which I did modify

chadoh:
My thought: we don't want to do this pre-push since it creates a dirty working tree and then the rest of the push fails; we want to include the doc change in every commit where it's relevant

chadoh:
Right now it's quite slow, but if it ran fast I think it would be less intrusive to run it every commit. Maybe we can find a way to make it run faster, though?

chadoh:
@paulbellamy Ohhh Kaaayyy. After MUCH troubleshooting, I finally got this new build step, job, and pre-commit hook to a place that feels reasonable.

- ditched xtask; cargo's caching didn't seem to kick in with it fully, so it always took ~9 seconds (on my (v fast) computer)
- instead now using a separate cargo bin; command is `cargo run --bin doc-gen --features clap-markdown` – the `--features` is necessary to avoid including `clap-markdown` dependency in main `dependencies`
- command now caches properly so it takes ~1s if it's already been run, which feels reasonable-enough to run in a pre-commit hook (I think?)
- GH Action now looks like this:
   image

paulbellamy (approved):
Approved. One stray file to remove.

PR #521, opened 2023-03-17
soroban-rpc: integration cleanup

tsachiherman:
### What

This PR provides some cleanup to the common execution path:
- breaks up the initialization of the captive core into it's own function.
- eliminate the need to error handling on the events.NewMemoryStore/transactions.NewMemoryStore.
- move the default for ledgerbucketwindow where the implementation is.
- add a check in the case where no history archives were specified.
- eliminate the non-existing error from internal.NewJSONRPCHandler


cmd/soroban-rpc/internal/ledgerbucketwindow/ledgerbucketwindow.go
```diff
@@ -21,14 +20,18 @@ type LedgerBucket[T any] struct {
 	BucketContent        T
 }
 
+// DefaultEventLedgerRetentionWindow represents the max number of ledgers we would like to keep
+// an incoming event in memory
+const DefaultEventLedgerRetentionWindow = 17280
```
Shaptic, line 25: I know you added this to the help text, but it'd be nice to note in the comment here that this is 24hr worth of ledgers (assuming 5s/ledger close times)

Shaptic (approved):
nice refactor :+1:

cmd/soroban-rpc/internal/ledgerbucketwindow/ledgerbucketwindow.go
tsachiherman: done.

PR #524, opened 2023-03-21
soroban-rpc: eliminate mac compilation warnings

tsachiherman:
### What

Compiling on mac, I ran into the following warnings:
```
ld: warning: object file (/var/folders/fs/dzj3b9r14jv5x6qf0w60lc5h0000gn/T/go-link-3656807857/000002.o) was built for newer macOS version (13.0) than being linked (12.3)
ld: warning: object file (/var/folders/fs/dzj3b9r14jv5x6qf0w60lc5h0000gn/T/go-link-3656807857/000004.o) was built for newer macOS version (13.0) than being linked (12.3)
ld: warning: object file (/var/folders/fs/dzj3b9r14jv5x6qf0w60lc5h0000gn/T/go-link-3656807857/000003.o) was built for newer macOS version (13.0) than being linked (12.3)
ld: warning: object file (/var/folders/fs/dzj3b9r14jv5x6qf0w60lc5h0000gn/T/go-link-3656807857/000005.o) was built for newer macOS version (13.0) than being linked (12.3)
...
```

### Why

It seems that the rust compiler, which is used to create the preflight library, is compiling the rust code with a min-version of 13.0, whereas the go compiler is compiling it with a min-version of 12.3.

To align these on Mac, I've set the go compiler to use 13.0 as the target min-version.

### Known limitations

N/A


sreuland:
it looks like the system tests in e2e.yml are using older version of core(git commit ref), seeing tests fail with:

`TransactionResultExt has invalid V (int32) switch value '6'`

can that be bumped to the newer core version that has same xdr version that tools repo is on

https://github.com/stellar/soroban-tools/blob/tsachi/macos-min-ver/.github/workflows/e2e.yml#L25

or, if there are many changes that need to settle, it may be worth to just disable the e2e workflow action for now and re-enable with update to the right version when ready.

Makefile
```diff
@@ -31,13 +41,13 @@ install_rust: Cargo.lock
 	cargo install --path ./cmd/soroban-cli
 
 install: install_rust build-libpreflight
-	go install -ldflags="${GOLDFLAGS}" ./...
+	go install -ldflags="${GOLDFLAGS}" -gcflags='CXXFLAGS="-mmacosx-version-min=12.3"' ${MACOS_MIN_VER} ./...
```
paulbellamy, line 44: Shouldn't the 12.3 here align with the 13.0 above?

PR #523, opened 2023-03-21
soroban-rpc: adapt to value overhaul changes

2opremio:
### What

Adapt soroban-rpc to the new value overhaul XDR changes (that is https://github.com/stellar/go/pull/4814 , which in turn comes from https://github.com/stellar/stellar-xdr/pull/70 )

### Why

It's part of https://github.com/stellar/soroban-tools/issues/471

It closes https://github.com/stellar/soroban-tools/issues/477

### Known limitations

TODO:
- [ ] fix integration tests (it requires bumping the core build)

PR #527, opened 2023-03-22
Bump all dependencies to support structured debug events

2opremio:
The Golang dependencies already had XDR support for structured debug events whilst Core and the Rust dependencies didn't.

This caused the soroban-rpc integration tests to break with XDR decoding errors.

This change bumps Core and the Rust dependencies to also support structured debug events.


2opremio:
Interestingly, the OOM panic from #523 went away.

paulbellamy (approved):

PR #528, opened 2023-03-22
feat: add new soroban-test crate

willemneal:
### What

Take advantage of the cli being a library and create a new `TestEnv` type which helps ensure that tests are run within their own directory.

### Why

Invoking tests with `assert_cmd` is costly and as our tests were growing so was the time it took.  Furthermore, it's API is not as friendly to new devs. Since we can parse commands directly it allows executing them directly and the cost savings have already be shown `9`s down to `<2`

### Known limitations

Currently there is not a standard way of setting the pwd of a test.

Using the `std` to set the `pwd` and using a Environment variable, both suffer from being shared mutable variables, which causes tests to become flaky.

The current approach was adding a `--pwd` command to the config locator and thus any method that requires the config.  This allows the `TestEnv` to prepend `--pwd={}`, however, this breaks any command that doesn't have a `--pwd` flag.  This commands can be invoke directly by parsing them with methods provided by the `CommandParser` trait.


cmd/crates/soroban-test/src/lib.rs
```diff
@@ -0,0 +1,205 @@
+//! **Soroban Test** - Test framework for invoking Soroban externally.
+//!
+//! Currently soroban provides a mock test environment for writing unit tets.
+//!
+//! However, it does not provide a way to run tests aganist a local sandbox or rpc endpoint.
+//!
+//! ## Overview
+//!
+//! - `TestEnv` is a test environment for running runing tests isolated from each other.
```
paulbellamy, line 9:
```suggestion
//! - `TestEnv` is a test environment for running tests isolated from each other.
```

cmd/crates/soroban-test/tests/it/config.rs
```diff
@@ -0,0 +1,267 @@
...
+        .new_assert_cmd("config")
+        .env("XDG_CONFIG_HOME", dir.to_str().unwrap())
+        .arg("network")
+        .arg("ls")
+        .assert()
+        .stdout("\n");
+}
+
+#[test]
+fn mulitple_networks() {
```
paulbellamy, line 110:
```suggestion
fn multiple_networks() {
```

cmd/soroban-cli/src/commands/config/identity/show.rs
```diff
@@ -13,15 +13,17 @@ pub enum Error {
     StrKey(#[from] stellar_strkey::DecodeError),
 }
 
-#[derive(Debug, clap::Args)]
-#[group(skip)]
```
paulbellamy, line 17: So, this will add an arggroup for these?

docs/soroban-cli-full-docs.md
```diff
@@ -189,15 +185,11 @@ soroban contract invoke ... -- --help
 * `--unlimited-budget` — Run with an unlimited budget
 * `--footprint` — Output the footprint to stderr
 * `--auth` — Output the contract auth for the transaction to stderr
-* `--events-file <EVENTS_FILE>` — File to persist event output
-
-  Default value: `.soroban/events.json`
+* `--events-file <EVENTS_FILE>` — File to persist event output, default is .soroban/events.json
```
paulbellamy, line 188:
```suggestion
* `--events-file <EVENTS_FILE>` — File to persist event output, default is `.soroban/events.json`
```

docs/soroban-cli-full-docs.md
```diff
@@ -189,15 +185,11 @@ soroban contract invoke ... -- --help
...
 * `--rpc-url <RPC_URL>` — RPC server endpoint
 * `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
 * `--network <NETWORK>` — Name of network to use from config
-* `--ledger-file <LEDGER_FILE>` — File to persist ledger state
-
-  Default value: `.soroban/ledger.json`
+* `--ledger-file <LEDGER_FILE>` — File to persist ledger state, default is `.soroban/ledger.json`
 * `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. S...          a seceret key alice         an identity 'kite urban.  a seed phrase DEFAULT       Is the key generated with `identity generate --seed 0000000000000000
```
paulbellamy, line 193:
```suggestion
* `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. S...          a secret key alice         an identity 'kite urban.  a seed phrase DEFAULT       Is the key generated with `identity generate --seed 0000000000000000
```

paulbellamy (commented):
A couple of the rust tests seem to be failing, like mulitple_networks (sp)

cmd/soroban-cli/src/commands/config/identity/show.rs
willemneal: Oops added them back.

willemneal:
@paulbellamy The windows failures seem to be addressed.  Turns out you should not use `.display` for paths.

The other issue left is naming.  Should we merge as is and then address the API and how things are named?

paulbellamy (approved):

Issue #473, opened 2023-03-06
soroban-rpc: update getEvents query to support optional diagnostic events


tsachiherman:
## What

soroban-rpc: update getEvents query to support optional debug events

We should include any emitted diagnostic events alongside the normal contract events (if they are in storage). We'll need to add a new event type for this to filter.

Note: They will already have been ingested & stored alongside normal events in https://github.com/stellar/soroban-tools/issues/472

Question: Should we return a value indicating that diagnostic events are enabled?

## Epic

https://github.com/stellar/soroban-tools/issues/471

PR #530, opened 2023-03-23
soroban-rpc: Add filtering support for diagnostic events

2opremio:
### What

Add support for diagnostic events filtering in the  `getEvents` method

### Why

Closes https://github.com/stellar/soroban-tools/issues/473

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/methods/get_events.go
+++ b/cmd/soroban-rpc/internal/methods/get_events.go
@@ -81,6 +81,12 @@ const EventTypeSystem = "system"
 const EventTypeContract = "contract"
 const EventTypeDiagnostic = "diagnostic"
 
+var eventTypeFromXDR = map[xdr.ContractEventType]string{
+	xdr.ContractEventTypeSystem:     EventTypeSystem,
+	xdr.ContractEventTypeContract:   EventTypeContract,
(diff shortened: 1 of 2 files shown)


2opremio:
Done! See https://github.com/stellar/soroban-docs/pull/349

cmd/soroban-rpc/internal/methods/get_events.go
```diff
@@ -335,20 +330,27 @@ func (h eventsRPCHandler) getEvents(request GetEventsRequest) (GetEventsResponse
...
+	switch xdrType {
+	case xdr.ContractEventTypeSystem:
+		return EventTypeSystem
+	case xdr.ContractEventTypeContract:
+		return EventTypeContract
+	case xdr.ContractEventTypeDiagnostic:
+		return EventTypeDiagnostic
+	}
+	return ""
+}
```
Shaptic, line 343: can/should this just be a static map rather than a runtime lookup?

Shaptic (approved):

cmd/soroban-rpc/internal/methods/get_events.go
2opremio: That’s a good idea, I will do that

PR #494, opened 2023-03-09
feat: initial work with new XDR

willemneal:
### What

Use newest XDR.

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]

PR #541, opened 2023-03-29
fix: parse XDR Result to JSON; encode RawVal return types

chadoh:
### What

Two new tests show some errors that revealed themselves while [adding CLI tests](https://github.com/stellar/soroban-examples/pull/216):

- [the errors example](https://github.com/stellar/soroban-examples/blob/main/errors/src/lib.rs) showed that valid `Ok` results were not being correctly parsed to JSON
- [the deployer example](https://github.com/AhaLabs/soroban-examples/blob/feat/tests/deployer/deployer/src/lib.rs#L18) showed that `RawVal` return types were causing errors

### Why

These tests expose real problems that have existed in soroban-cli's `main` branch since [updating to the latest XDR definitions](https://github.com/stellar/soroban-tools/pull/494)

### Known limitations

N/A


cmd/soroban-cli/src/strval.rs
```diff
@@ -438,7 +438,8 @@ impl Spec {
     /// May panic
     pub fn xdr_to_json(&self, val: &ScVal, output: &ScType) -> Result<Value, Error> {
         Ok(match (val, output) {
-            (ScVal::Map(None) | ScVal::Vec(None), ScType::Option(_)) => Value::Null,
+            (ScVal::Void, ScType::Val)
```
paulbellamy, line 441: Is there anything sensible we can do if `val != ScVal::Void && output == ScType::Val`? And wouldn't this still fail in that case?

Now that we have the value overhaul in, the below matchers seem to mostly be pretty trivial ScVal -> ScType "equality" checks... 🤔

willemneal: Most of them were before too.  The tricky part is still Enums, Unions, Maps, & Sets since they don't exist in ScVal.

willemneal: In the case of `val != ScVal::Void && output == ScType::Val` really it just means that we should just ignore the `ScType` and look only at the `ScVal`?  If so we should add back the `ScVal` --> `JSON` for this case.

paulbellamy: I'm saying, that really for `xdr_to_json`, the `output: &ScType` is sort of irrelevant. So yeah, for that case we should ignore the type, and just do `ScVal -> JSON`.

But more generally than that, I'm saying, maybe we should *always* ignore the type. The only cases where we do anything non-trivial are nulls, addresses, and results. 🤔

paulbellamy (approved):
Approved, as this is strictly better than what we have, and we can do a follow-up PR to address `val != ScVal::Void && output == ScType::Val` case.

willemneal:
The reason that we we used both was for the UDT cases. That is `JSON -> ScVal` for Enums and Structs differs than if we use `ScVal  -> JSON`

PR #543, opened 2023-03-29
cmd/soroban-rpc: Retry transient startup errors in soroban-rpc with exponential backoff

### LINKED ISSUE #358: Startup, resilient to transient in-complete network/core states
soroban-rpc process will exit at startup if the archive urls greatest ledger is presently at 0.
you will see this in log:
```
time="2023-01-26T00:18:51.358Z" level=info msg="Starting Soroban JSON RPC server on localhost:8003" pid=1099
time="2023-01-26T00:18:51.358Z" level=info msg="Found an empty database, filling it in from the most recent checkpoint (this can take up to 30 minutes, depending on the network)" pid=1099
time="2023-01-26T00:18:51.363Z" level=info msg="Starting processing of checkpoint 0" pid=1099
panic: 0 is not a checkpoint ledger, try 63 or 63 (in general, try n where n+1 mod 64 == 0)

goroutine 66 [running]:
github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/ledgerentry_storage.(*ledgerEntryStorage).run.func1()
	/src/soroban-rpc/cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go:150 +0x114
created by github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/ledgerentry_storage.(*ledgerEntryStorage).run
	/src/soroban-rpc/cmd/soroban-rpc/internal/ledgerentry_storage/ledgerentry_storage.go:143 +0x2e8

```

the startup process performs retry mechanism to heal from possible transient core/network state issues that may be overcome such as to re-attempt archives at a backoff and optionally log each failed attempt for status/visibility.

a bonus, would be to move the init routine onto async, and let the http service come up as fast as possible, but then report 'not healthy' on the rpc `getHealth` endpoint until init routine finishes successfully. This enables remote systems interop, using `getHealth` for an accurate heartbeat signal for rpc readiness.

do a brute force retry loop from the parent process that is trying to start soroban rpc, that is the current work around being used in quickstart to launch rpc when it's on standalone network, which takes about 5 minutes before the first archive of 64 ledgers is published.

stellarsaur:
### What

This PR retries potential transient errors during startup of soroban-rpc using exponential back-off with a max duration of 5 minutes. All retries are additionally logged.

- Connecting to history archive
- Opening SQLite DB
- Getting txmeta cache from DB
- Creating new ingest service

Tested manually locally by returning errors from each affected operation individually.

Closes #358 

### Why

See #358 for full context.

soroban-rpc should be more resilient during startup when encountering core/network state issues that may resolve on their own by retrying.


cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -143,19 +180,27 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
-		EventStore:        eventStore,
-		TransactionStore:  transactionStore,
-		NetworkPassPhrase: cfg.NetworkPassphrase,
-		Archive:           historyArchive,
-		LedgerBackend:     core,
-		Timeout:           cfg.LedgerEntryStorageTimeout,
-	})
+	ingestService, err := backoff.RetryNotifyWithData(
+		func() (*ingest.Service, error) {
+			return ingest.NewService(ingest.Config{
```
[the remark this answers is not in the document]
stellarsaur: Good catch. I changed the signature so that it no longer returns an error, and added constant backoff (retry every second for 5 seconds) to running ingestion.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -100,20 +104,44 @@ func MustNew(cfg config.LocalConfig) *Daemon {
 	if len(cfg.HistoryArchiveURLs) == 0 {
 		logger.Fatalf("no history archives url were provided")
 	}
-	historyArchive, err := historyarchive.Connect(
-		cfg.HistoryArchiveURLs[0],
-		historyarchive.ConnectOptions{
-			CheckpointFrequency: cfg.CheckpointFrequency,
+	// Defaults: https://pkg.go.dev/github.com/cenkalti/backoff/v4@v4.2.0#pkg-constants
```
sreuland, line 107: I think for db and archive resources, the backoff may not be adding much value and not worth the xtra code to wrap it as such.  both of those resources are either on/off, they don't have transition states, I think fast failing on both of these is fine still as that means it's likely a configuration error, @tamirms do you agree?

tsachiherman: I think that as a service, this back off initialization isn't appropriate.
We should startup fast ( i.e. less than 2 sec ), and fail fast.
Any network activity should follow later and keep trying to recover forever

cmd/soroban-rpc/internal/ingest/service.go
```diff
@@ -46,12 +47,21 @@ func NewService(cfg Config) (*Service, error) {
 	}
 	o.wg.Add(1)
 	go func() {
-		err := o.run(ctx, cfg.Archive)
+		// Retry running ingestion every second for 5 seconds.
+		constantBackoff := backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 5)
```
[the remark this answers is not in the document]
sreuland: Have you looked into unit test coverage against this, such as with a new service_test.go, might be able to exercise the backoff err handler, could mock an instance of `db.ReadWriter` that you pass as `Config.DB`  to `NewService()`, and then wire up `mockReadWriter.GetLatestLedgerSequence()` to return an error.

cmd/soroban-rpc/internal/daemon/daemon.go
tsachiherman: I.e. it's a server. There is no one behind the console. If it can start up, it should endlessly try to take care of itself.
If it's not completely initialized, it should report that to the various endpoints.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -100,20 +104,44 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
+		func(err error, dur time.Duration) {
+			logger.Errorf("Failed connecting to history archive with error: %v. Retrying after %v", err, dur)
+		})
 	if err != nil {
 		logger.Fatalf("could not connect to history archive: %v", err)
 	}
+	expBackoff.Reset()
 
-	dbConn, err := db.OpenSQLiteDB(cfg.SQLiteDBPath)
+	dbConn, err := backoff.RetryNotifyWithData(
```
2opremio, line 131: Considering we are using SQLite (which is mapped to a file) I am not sure retries would help if we fail to open the file.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -3,14 +3,18 @@ package daemon
 import (
 	"context"
 	"net/http"
+	"time"
 
 	"github.com/jmoiron/sqlx"
```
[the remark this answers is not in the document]
stellarsaur: Good catch. Fixed!

cmd/soroban-rpc/internal/daemon/daemon.go
stellarsaur: Makes sense. I removed the retry here.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -127,10 +155,19 @@ func MustNew(cfg config.LocalConfig) *Daemon {
 
 	// initialize the stores using what was on the DB
 	// TODO: add a timeout?
-	txmetas, err := db.NewLedgerReader(dbConn).GetAllLedgers(context.Background())
+	txmetas, err := backoff.RetryNotifyWithData(
```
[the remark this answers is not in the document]
stellarsaur: This makes sense to me the way you put it. I removed the retrying like as suggested.

cmd/soroban-rpc/internal/daemon/daemon.go
stellarsaur: The reason I added retrying with back-off to the archive here is because #358 suggested it as a possible solution. From my understanding, connecting to the history archive may fail for some network related reason, and retrying would help with this. However, based on the comments in this PR, it seems like it is only maybe worth retrying running ingestion.

This makes me wonder: **what exactly are the possible transient core/network state issues we want to make more resilient?** It seems I'm misunderstanding something important here, or there are very little/no transient core/network state issues to begin with. Would appreciate any insight here.

cmd/soroban-rpc/internal/ingest/service.go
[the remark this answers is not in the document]
stellarsaur: You're right. I changed this to use "WithContext" instead.

In terms of unit test coverage, I'll update this PR with all tests necessary once we align on the functionality.

cmd/soroban-rpc/internal/daemon/daemon.go
sreuland: I think published archives are one transient state to retry, the bug report on #358 , referred to a stack trace related to rpc exiting early when it tries to do ingestion system startup due to the ingest library throwing error when not able to handle empty archive service, it expected at least one archive be published by the service, however in case of starting a new networks like standalone, the archives http service will be up, but no archives are published yet until the first checkpoint occurs on the network(once every 64 ledgers, or 8 ledgers if using accelerated mode). The error from ingest library worked up stack into rpc service initialization routine and it panic'd, causing rpc service to exit, however, In 5 minutes(or sooner on accelerated network) that archive state will change as the first 64 ledgers complete and the checkpoint archive will be published at which point the ingest routine can proceed, so a retry from rpc over that initialization routine and timeframe should prevent a panic, allowing the same rpc service to continue and recover from that type of evolving network state.

The retry applied on https://github.com/stellar/soroban-tools/pull/543/files#diff-b8bd4f66b43a1d0998b64e915e915f36752c2d8d8a50f46ed74f27f92f92eb4aR56 covers the rpc starting the ingestion system, so, seems like it would be effective in this regard.

There were some recent changes in the ingestion library, [go#4773](https://github.com/stellar/go/pull/4773) to be more robust in handling some aspects of ledger ranges based on archives, but I'm not sure those changes prevent the error due to no-archives present case. It would be good to replicate the bug on present day source, since #358 was reported several weeks ago, by running a standalone network from `stellar/quickstart:soroban-dev`, and check the supervisor rpc logs in the running container to see if rpc process is still exiting early and supervisor is restarting it repeatedly.

stellarsaur: I learned a lot from these comments. Thank you for the detailed responses! I removed retrying connecting to the history archives here since it's unnecessary.

I'll figure out unit testing retries for running ingestion in order to finalize this PR.

cmd/soroban-rpc/internal/ingest/service.go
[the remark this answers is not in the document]
stellarsaur: Ok, I added a test by passing in a callback on retrying ingestion. I'm aware that "Sleep" in tests isn't ideal but ingestion is run in a co-routine so this is a little tricky. I ran the test locally with -count=1000 to ensure it's not flaky.

PR #547, opened 2023-04-03
We should actually run the rust unit tests in CI

paulbellamy:
### What

Run the rust unit tests in CI

### Why

Because they are failing and we haven't noticed.

### Known limitations

[TODO or N/A]


.github/workflows/rust.yml
```diff
@@ -63,6 +63,7 @@ jobs:
       run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
     - run: cargo clippy --all-targets --target ${{ matrix.target }}
     - run: make build-test-wasms
+    - run: make test
```
willemneal, line 66:
```suggestion
    - run: cargo install --path ./cmd/soroban-cli --debug
    - run: make test
```

willemneal (changes requested):
Binary was missing for tests.   Also good find!

paulbellamy:
Oh, duh. I put the `make test` in the wrong order when pushing this change. Locally it was still failing with:
```
#0: debug: previous invocation is missing - no payloads recorded
thread 'invoke_sandbox::invoke_with_seed' panicked at 'called `Result::unwrap()` on an `Err` value: Events(Generic("debug events unsupported: DebugEvent {\n    msg: Some(\n        \"previous invocation is missing - no payloads recorded\",\n    ),\n    args: [\n        ,\n    ],\n}"))', cmd/crates/soroban-test/tests/it/invoke_sandbox.rs:218:20
```
Which it is on CI now, as well.

tsachiherman (approved):

paulbellamy:
Hmmm.. Seems to be failing on windows for some reason...

willemneal:
I finally dual booted an old computer so I'll try testing it locally, though this seems to be some environment issue.

paulbellamy:
Yeah, clearly the actual build works when we release, so maybe it's an issue with tests, or the actions worker...

willemneal:
It's passing for me locally on windows.

paulbellamy:
Ok, so it's a github actions issue of some sort... 🤔

.github/workflows/rust.yml
```diff
@@ -64,6 +64,7 @@ jobs:
     - run: cargo clippy --all-targets --target ${{ matrix.target }}
     - run: make build-test-wasms
     - run: cargo install --path ./cmd/soroban-cli --debug
+    - run: cargo test
```
leighmcculloch, line 67: The tests for the cli are run a few lines down, in the for loop, because that includes the cli. So I don't think we need this. Please let me know if I'm missing something.

PR #566, opened 2023-04-06
cmd/soroban-rpc: Shutdown HTTP server gracefully using server.Shutdown instead of supporthttp Package

### LINKED ISSUE #499: soroban-rpc: Graceful shutdown for HTTP server

When you kill soroban-rpc it may drop ongoing HTTP requests.

We should use the HTTP server graceful shutdown mechanism to finish pending HTTP requests before shutdown.

stellarsaur:
### What

Currently, we rely on the supporthttp package to finish pending HTTP requests before shutting down. This PR removes that dependency and uses the server.Shutdown method directly, alongside the appropriate OS signal handling.

This PR still uses the same values as before for relevant timeouts:
- default read timeout still 5s
- shutdown grace period still 10s (supporthttp default)

Open to suggestions on testing this, or if it's even necessary.

Closes #499 

### Why

Without graceful shutdown, when you kill soroban-rpc it may drop ongoing HTTP requests. We would like to wait a grace period while finishing any outgoing requests and closing any open connections. Additionally, we don't want to rely on external packages for this (like how supporthttp relies on the [Graceful](https://github.com/tylerstillwater/graceful) package).


cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -189,14 +194,37 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
+	}()
+
+	// Shutdown gracefully when we receive an interrupt signal.
+	// First server.Shutdown closes all open listeners, then closes all idle connections.
+	// Finally, it waits a grace period (10s here) for connections to return to idle and then shut down.
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+	<-signals
+
+	// Default supporthttp Shutdown grace period.
```
2opremio, line 221:
```suggestion
	// Default Shutdown grace period.
```

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -189,14 +194,37 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
+		ReadTimeout: defaultReadTimeout,
+	}
+
+	d.logger.Infof("Starting Soroban JSON RPC server on %v", endpoint)
+
+	go func() {
+		if err := server.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+			// Error starting or closing listener:
+			d.logger.Fatalf("Soroban JSON RPC server encountered fatal error: %v", err)
+			os.Exit(1)
```
2opremio, line 210: Fatalf already exists, this is a noop.

```suggestion
```

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -189,14 +194,37 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+	<-signals
+
+	// Default supporthttp Shutdown grace period.
+	shutdownCtx, shutdownRelease := context.WithTimeout(context.Background(), defaultShutdownGracePeriod)
+	defer shutdownRelease()
+
+	if err := server.Shutdown(shutdownCtx); err != nil {
+		// Error from closing listeners, or context timeout:
+		d.logger.Errorf("Error during Soroban JSON RPC server Shutdown: %v", err)
```
tsachiherman, line 227: I'm not sure if a special use case here would be ContextTimedOut

tsachiherman (approved):
I like this implementation. I think that it could be further improve shutting down all the other services by passing the context to d.Close().

What I had in mind is a "hard limit" of terminating the service within 5 second or letting it shut down on its own faster. Keep in mind that if we won't persist data or disconnect existing connections - that's perfectly fine.

In case the shutdown takes more than 5 seconds, we should be generating a log entry telling us why this is the case ( i.e. one of the services is hanged.. )

Issue #496, opened 2023-03-10
soroban-rpc: Add /metrics endpoint


paulbellamy:
### What problem does your feature solve?

As an operator, I would like to have access to system health metrics for my RPC instance, so that I can understand system resource process consumption and available capacity.

### What would you like to see?

Metrics should include:
- General
  - Version running
  - Uptime (comes out of the box w prometheus, I think)
  - Golang stuff like goroutines etc
  - Question: Should/can we include disk/cpu/io/ram? These would normally come from the environment monitoring (vm/k8s)
- Ingestion
  - Same as horizon
- txsub
  - Same as horizon
  - Note: We might not need as much as horizon here, given sendTransaction is simpler. But tx type broken down by envelope would be nice.
- JsonRPC
  - RPS (with route and status)
  - Request duration (with route and status)
- Logging
  - Lines/entries logged (with severity)
- DB (similar to horizon)
  - connections
    - open
    - in-use
    - waited-for
    - wait-duration
  - queries per-method
    - mean duration
    - count
  - disk size
    - total
    - just events
    - just ledger entries

We should also provide a document with example alerts and their expected response, so operators know how to configure their own alerts.

PR #573, opened 2023-04-10
soroban-rpc: Add admin endpoint including pprof debug endpoints

### LINKED ISSUE #498: soroban-rpc: Add configurable pprof support

As a developer I would like to acquire runtime profiling for my RPC sercvice.

Add a `--pprof` flag (and matching env var) to soroban-rpc, which should enable the golang pprof web interface. It should be disabled by default.
See: https://pkg.go.dev/net/http/pprof

Note: Make sure that when this is disabled it doesn't impact performance.

2opremio:
### What

Add admin http endpoint containing pprof debug endpoints

It adds an `--admin-endpoint` flag (matching the normal `--endpoint` flag), which indicates where to listen.

For instance, when providing `--admin-endpoint http://localhost:9000` the endpoints like the following will be exposed:

```
http://localhost:9000/debug/pprof/
http://localhost:9000/debug/pprof/cmdline
http://localhost:9000/debug/pprof/profile
http://localhost:9000/debug/pprof/symbol
http://localhost:9000/debug/pprof/trace
...
```

Depends on #566

### Why

Closes #498 and paces the way for https://github.com/stellar/soroban-tools/issues/496 which also needs an admin base endpoint

### Known limitations

[TODO or N/A]


cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -187,16 +192,54 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
+	shutdownCtx, shutdownRelease := context.WithTimeout(context.Background(), defaultShutdownGracePeriod)
+	defer shutdownRelease()
+
+	if err := server.Shutdown(shutdownCtx); err != nil {
+		// Error from closing listeners, or context timeout:
+		d.logger.Errorf("Error during Soroban JSON RPC server Shutdown: %v", err)
+	}
+	d.Close()
+
+	if err := adminServer.Shutdown(shutdownCtx); err != nil {
```
tsachiherman, line 241: this would panic if `adminServer` is nil. I suggest that you'll add a defer function right after `go func() {` in line 217.

tsachiherman (changes requested):

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -187,16 +192,54 @@ func MustNew(cfg config.LocalConfig) *Daemon {
...
+		if err := server.ListenAndServe(); !errors.Is(err, http.ErrServerClosed) {
+			// Error starting or closing listener:
+			d.logger.Fatalf("Soroban JSON RPC server encountered fatal error: %v", err)
+			os.Exit(1)
+		}
+	}()
+	var adminServer *http.Server
+	if adminEndpoint != "" {
+		// after importing net/http/pprof, debug endpoints are implicitly registered in the default serve mux
+		adminServer = &http.Server{Addr: adminEndpoint, Handler: http.DefaultServeMux}
```
tsachiherman, line 216: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)

tsachiherman, line 216: Should we support adminEndpoint == endpoint ? ( it's a bit more complicated then it sounds )

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: It should, I actually prepared it for it, but forgot to check for nil. I don't think there is a need for defer

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: The admin port should never be exposed publicly. Way worse things can happen if it does.

2opremio: No, we shouldn't. There will be a binding error if both endpoints are the same.

tsachiherman: it wasn't me.. it's just to appease the linter.

2opremio: Ah, I see

2opremio: If you meant we should be adding a `//nolint:gosec` problem, I think it's better not to add noise to the file based on the tooling we are currently using.

2opremio:
@tsachiherman PTAL, can't merge without your confirmation.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	_ "net/http/pprof"
```
paulbellamy, line 7: It would be nice if we could avoid the automatic http handler installed by this... :/

cmd/soroban-rpc/internal/daemon/daemon.go
tsachiherman: yes, that's why I said it's a little bit more complicated. ( i.e. in this case, we don't need another listener )

tsachiherman: but I'm not really picky about this one. Just figured it's the "complementary" to ensure we support all end user permutations. in particular, avoid generating binding error if the two endpoints are the same.

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: It won't change until Go 2 at least, https://github.com/golang/go/issues/42834

tsachiherman (approved):

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: Problem is that the default mux code isn't exposed so I would need to replicate it privately

tsachiherman: since we're using a custom server anyway ( i.e. http.Server ), couldn't we use a non-default mux ?

2opremio: We could use a non-default mux but the default mux would still be filled automatically. See https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/net/http/pprof/pprof.go;l=85-91

tsachiherman:
> @tsachiherman PTAL, can't merge without your confirmation.

I blocked this one only for the nil issue I pointed out ( and you fixed ).

PR #574, opened 2023-04-10
soroban-rpc: libpreflight: Avoid panicking when address and nonce are both None

2opremio:
Followup to  #553 

(Found out by @sreuland at https://github.com/stellar/soroban-tools/commit/9be3f2c5036ad2db237df98516b9ae872f9fd997#r108326767 )


sreuland (approved):
looks good, do you expect any unit/integration tests local in tools repo that could have coverage on this path eventually?

2opremio:
We should probably add some unit tests to libpreflight at some point, yeah

PR #576, opened 2023-04-10
Infer 80/443 ports for http/https rpc-url

leighmcculloch:
### What
Infer 80/443 ports for http/https rpc-url

### Why
When not including a port in the rpc-url an error message is displayed. This is not particularly ergonomic. Although it has no otherwise functional impact and causes no other problem. Folks are just not accustomed to thinking about ports for http and https.


cmd/soroban-cli/Cargo.toml
```diff
@@ -53,6 +53,7 @@ csv = "1.1.6"
 ed25519-dalek = "1.0.1"
 jsonrpsee-http-client = "0.15.1"
 jsonrpsee-core = "0.15.1"
+http = "0.2.9"
```
leighmcculloch, line 56: This is not really a new dependency. The `http` dependency is the same dependency that the jsonrpsee package uses for parsing URLs.

tomerweller (approved):

paulbellamy (approved):

PR #579, opened 2023-04-12
soroban-rpc: Add /metrics endpoint

tamirms:
### What

Add `/metrics` prometheus endpoint with the following set of metrics:

- DB (similar to horizon)
  - connections
    - open
    - in-use
    - waited-for
    - wait-duration
  - queries per-method
    - mean duration
    - count

- General
  - Version running
  - Uptime (comes out of the box w prometheus, I think)
  - Golang stuff like goroutines etc

- Logging
  - Lines/entries logged (with severity)

### Why

This PR does not provide all the metrics specified in the issue above. The remaining metrics be included in separate upcoming PRs.

### Known limitations

[N/A]

### DIFF
--- a/.github/workflows/update-completed-sprint-on-issue-closed.yml
+++ b/.github/workflows/update-completed-sprint-on-issue-closed.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   update_completed_sprint:
-    uses: stellar/actions/.github/workflows/update-completed-sprint-on-issue-closed.yml@main
+    uses: paulbellamy/stellar-actions/.github/workflows/update-completed-sprint-on-issue-closed.yml@update-completed-sprint-on-issue-closed
     with:
       project_name: "Platform Scrum"
       field_name: "CompletedSprint"


tsachiherman (changes requested):
Tamir, I don't believe that the PrometheusRegistry should be part of the local configuration.
( unless you can convince me otherwise ).
I think that it should be a member of the daemon.

The local config is for entries that we would load from a configuration file. I don't believe the registry is part of that..

cmd/soroban-rpc/internal/test/metrics_test.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMetrics(t *testing.T) {
+	test := NewTest(t)
+	metrics, err := test.daemon.PrometheusRegistry().Gather()
+	assert.NoError(t, err)
```
tsachiherman, line 12: I think that *require* here would work better, since the rest of the test could easily stop here.

tsachiherman (approved):
Tamir, my intent in asking for a unit test was to have a http-client unit test that would verify that the metrics endpoint works as expected.

I don't have any reason to believe that it wouldn't, and I don't want to block you from merging this one in - but I do want to have ( eventually ) a concrete unit test to cover all the http endpoints.

cmd/soroban-rpc/internal/ingest/service.go
```diff
@@ -36,7 +36,7 @@ type Config struct {
 
 func NewService(cfg Config) *Service {
 	ctx, done := context.WithCancel(context.Background())
-	o := Service{
+	service := Service{
```
tsachiherman, line 39: I'd suggest making this into a `&Service` here so that subsequent calls would avoid copying the data structure.

cmd/soroban-rpc/main.go
```diff
@@ -283,7 +283,7 @@ func main() {
 			serviceConfig.IngestionTimeout = time.Duration(ingestionTimeoutMinutes) * time.Minute
 			serviceConfig.CoreRequestTimeout = time.Duration(coreTimeoutSeconds) * time.Second
 			serviceConfig.MaxHealthyLedgerLatency = time.Duration(maxHealthyLedgerLatencySeconds) * time.Second
-			daemon.Run(serviceConfig, endpoint, adminEndpoint)
+			daemon.MustNew(serviceConfig, endpoint, adminEndpoint).Run()
```
tsachiherman, line 286: I honestly don't feel comfertable using the function output without testing it.
( yes, I'm aware of the content.. )
But I'm ok if we'll tackle this another time.

tsachiherman (approved):

PR #583, opened 2023-04-13
Try using the update thing as an action not a workflow

paulbellamy:
### What

Try using https://github.com/stellar/actions/pull/52 as an action, not a workflow.

### Why

This is what the github action docs say for doing a typescript action.
https://github.com/actions/typescript-action

### Known limitations

[TODO or N/A]

### DIFF
--- a/.github/workflows/update-completed-sprint-on-issue-closed.yml
+++ b/.github/workflows/update-completed-sprint-on-issue-closed.yml
@@ -4,13 +4,15 @@ on:
   issues:
     types: [closed]
   pull_request:
-    types: [closed]
+    types: [closed, comment]
 
 jobs:
   update-completed-sprint:
     runs-on: ubuntu-latest
     steps:
       - uses: paulbellamy/stellar-actions/update-completed-sprint-on-issue-closed@update-completed-sprint-on-issue-closed
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           project_name: "Platform Scrum"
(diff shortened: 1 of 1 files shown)


.github/workflows/update-completed-sprint-on-issue-closed.yml
```diff
@@ -7,8 +7,19 @@ on:
...
-    uses: paulbellamy/stellar-actions/.github/workflows/update-completed-sprint-on-issue-closed.yml@update-completed-sprint-on-issue-closed
-    with:
-      project_name: "Platform Scrum"
-      field_name: "CompletedSprint"
+  update-completed-sprint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate token
+        id: generate_token
+        uses: tibdex/github-app-token@v1
```
jacekn, line 15: Is it worth pinning to a specific sha to avoid code changing without us knowing?

tsachiherman (approved):

.github/workflows/update-completed-sprint-on-issue-closed.yml
paulbellamy: Let's do that if it breaks

jacekn: The risk I was thinking of was compromised action or developer's account. With that somebody could publish malicious action that steals creds or repo data.
It's not very frequent problem and in this particular case our exposure would be limited to access to projects so it maybe acceptable.

But getting into habit of using sha everywhere might be good from security perspective.

Issue #497, opened 2023-03-10
soroban-rpc: Audit and standardize logging


paulbellamy:
### What problem does your feature solve?

Currently the logging in soroban-rpc is a bit ad-hoc. We should audit what we have, and standardize it for better ingestion by people & other tools.

### What would you like to see?

- Standardize the logging output format
- Standardize the logging severity we use for different info/warnings/errors
- Add a `--log-level` flag so operators can use what level of logging output they want.
- Add clear logging during startup so we have tools in-place to debug any potential issues there, and can understand where the service is in the startup process.
- Log all incoming jsonrpc method requests and their resolution.
  - Note, this should be two log entries, one when the request is received, and another one when it is completed, tied together with a request id. (Paul's opinion)
- Log all incoming http requests and their response with duration, etc.
- Add support for json log formatting (should this be a flag? default?)
- Audit to make sure the log outputs do not include any avoidable identifying information, or secrets.

2opremio:
Closing, I think I've covered everything in this ticket. @paulbellamy please reopen if you think otherwise.

PR #584, opened 2023-04-13
soroban-rpc: Add logging for HTTP requests and JSONRPC requests

2opremio:
### What

Each HTTP request is assigned an ID logged using the the stellar/go/support/http middleware

Within each HTTP request, each JSONRPC request is logged using the global HTTP ID and the per-request JSONRPC ID.

Additionally, I added the version and commit of soroban-rpc to the startup log.

### Why

Done as part of #497 

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
-	_ "net/http/pprof"
+	"net/http/pprof" //nolint:gosec
 	"os"
 	"os/signal"
 	"sync"
@@ -242,8 +242,11 @@ func MustNew(cfg config.LocalConfig, endpoint string, adminEndpoint string) *Dae
 	}
 	if adminEndpoint != "" {
 		adminMux := supporthttp.NewMux(logger)
-		// after importing net/http/pprof, debug endpoints are implicitly registered in the default serve mux
-		adminMux.Handle("/", http.DefaultServeMux)
+		adminMux.HandleFunc("/debug/pprof/", pprof.Index)
+		adminMux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+		adminMux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+		adminMux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+		adminMux.HandleFunc("/debug/pprof/trace", pprof.Trace)
 		adminMux.Handle("/metrics", promhttp.HandlerFor(d.prometheusRegistry, promhttp.HandlerOpts{}))
 		d.adminServer = &http.Server{Addr: adminEndpoint, Handler: adminMux}
 	}
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
-	"net/http/pprof"
+	_ "net/http/pprof"
 	"os"
 	"os/signal"
 	"sync"
@@ -17,6 +17,7 @@ import (
 	"github.com/stellar/go/historyarchive"
 	"github.com/stellar/go/ingest/ledgerbackend"
 	dbsession "github.com/stellar/go/support/db"
+	supporthttp "github.com/stellar/go/support/http"
 	supportlog "github.com/stellar/go/support/log"
 
 	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal"
@@ -39,7 +40,8 @@ type Daemon struct {
 	core                *ledgerbackend.CaptiveStellarCore
 	ingestService       *ingest.Service
 	db                  dbsession.SessionInterface
-	handler             *internal.Handler
+	jsonRPCHandler      *internal.Handler
+	httpHandler         http.Handler
 	logger              *supportlog.Entry
 	preflightWorkerPool *preflight.PreflightWorkerPool
 	prometheusRegistry  *prometheus.Registry
@@ -55,7 +57,7 @@ func (d *Daemon) PrometheusRegistry() *prometheus.Registry {
 }
 
 func (d *Daemon) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
-	d.handler.ServeHTTP(writer, request)
+	d.httpHandler.ServeHTTP(writer, request)
 }
 
 func (d *Daemon) GetDB() dbsession.SessionInterface {
@@ -63,19 +65,16 @@ func (d *Daemon) GetDB() dbsession.SessionInterface {
 }
 
 func (d *Daemon) close() {
-	// Default Shutdown grace period.
 	shutdownCtx, shutdownRelease := context.WithTimeout(context.Background(), defaultShutdownGracePeriod)
 	defer shutdownRelease()
 	var closeErrors []error
 
 	if err := d.server.Shutdown(shutdownCtx); err != nil {
-		// Error from closing listeners, or context timeout:
 		d.logger.Errorf("Error during Soroban JSON RPC server Shutdown: %v", err)
 		closeErrors = append(closeErrors, err)
 	}
 	if d.adminServer != nil {
 		if err := d.adminServer.Shutdown(shutdownCtx); err != nil {
-			// Error from closing listeners, or context timeout:
 			d.logger.Errorf("Error during Soroban JSON admin server Shutdown: %v", err)
 			closeErrors = append(closeErrors, err)
 		}
@@ -89,7 +88,7 @@ func (d *Daemon) close() {
 		d.logger.WithError(err).Error("Error closing captive core")
 		closeErrors = append(closeErrors, err)
 	}
-	d.handler.Close()
+	d.jsonRPCHandler.Close()
 	if err := d.db.Close(); err != nil {
 		d.logger.WithError(err).Error("Error closing db")
 		closeErrors = append(closeErrors, err)
@@ -104,7 +103,7 @@ func (d *Daemon) Close() error {
 	return d.closeError
 }
 
-// newCaptiveCore create a new captive core backend instance and returns it.
+// newCaptiveCore creates a new captive core backend instance and returns it.
 func newCaptiveCore(cfg *config.LocalConfig, logger *supportlog.Entry) (*ledgerbackend.CaptiveStellarCore, error) {
 	httpPortUint := uint(cfg.CaptiveCoreHTTPPort)
 	var captiveCoreTomlParams ledgerbackend.CaptiveCoreTomlParams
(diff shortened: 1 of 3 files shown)


cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -53,16 +60,42 @@ func NewJSONRPCHandler(cfg *config.LocalConfig, params HandlerParams) Handler {
...
+}
+
+func (r *rpcLogger) LogRequest(ctx context.Context, req *jrpc2.Request) {
+	r.logger.WithFields(log.F{
+		"subsys":   "jsonrpc",
+		"req":      middleware.GetReqID(ctx),
+		"json_req": req.ID(),
+		"method":   req.Method(),
+		// TODO: is this overkill?
+		"params": req.ParamString(),
```
paulbellamy, line 83: It means we would be logging transactionenvelopes and everything.. Probably there might be PII in here. Let's not for now. We could hash the string and log that? That would tell us uniqueness of the method+params

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -53,16 +60,42 @@ func NewJSONRPCHandler(cfg *config.LocalConfig, params HandlerParams) Handler {
...
+	}).Info("starting JSONRPC request")
+}
+
+func (r *rpcLogger) LogResponse(ctx context.Context, rsp *jrpc2.Response) {
+	// TODO: Print the elapsed time (there doesn't seem to be a way to it with with jrpc2, since
+	//       LogRequest cannot modify the context)
```
paulbellamy, line 89: Could you check the request receive time from the context? That wouldn't be super accurate if you have multiple methods in a single http connection, though...

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -53,16 +60,42 @@ func NewJSONRPCHandler(cfg *config.LocalConfig, params HandlerParams) Handler {
...
+	logger := r.logger.WithFields(log.F{
+		"subsys":   "jsonrpc",
+		"req":      middleware.GetReqID(ctx),
+		"json_req": rsp.ID(),
+		// TODO: is this overkill?
+		"result": rsp.ResultString(),
```
paulbellamy, line 95: Same comment as the request params.

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: Considering we will have a global log entry for the full HTTP request I am not sure it's worth it.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -239,10 +241,10 @@ func MustNew(cfg config.LocalConfig, endpoint string, adminEndpoint string) *Dae
 		ReadTimeout: defaultReadTimeout,
 	}
 	if adminEndpoint != "" {
-		adminMux := http.NewServeMux()
+		adminMux := supporthttp.NewMux(logger)
+		// after importing net/http/pprof, debug endpoints are implicitly registered in the default serve mux
+		adminMux.Handle("/", http.DefaultServeMux)
 		adminMux.Handle("/metrics", promhttp.HandlerFor(d.prometheusRegistry, promhttp.HandlerOpts{}))
-		adminMux.HandleFunc("/debug/pprof/heap", pprof.Index)
-		adminMux.HandleFunc("/debug/pprof/profile", pprof.Profile)
```
2opremio, line 245: @tamirms FYI, we were missing some pprof endpoints, see https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/net/http/pprof/pprof.go;l=85-91

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: I think that having the parameters will be very helpful when debugging problems.

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: Here we can also cap it, but I think it's less important than with the parameteres

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: It would be cool that fields could individually have a log level (we could add params with a debug level), but that's not possible

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -223,11 +222,14 @@ func MustNew(cfg config.LocalConfig, endpoint string, adminEndpoint string) *Dae
...
 
+	httpHandler := supporthttp.NewAPIMux(logger)
+	httpHandler.Handle("/", jsonRPCHandler)
 	d := &Daemon{
 		logger:              logger,
 		core:                core,
 		ingestService:       ingestService,
-		handler:             &handler,
+		jsonRPCHandler:      &jsonRPCHandler,
+		httpHandler:         httpHandler,
```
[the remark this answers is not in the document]
2opremio: good point, I will remove it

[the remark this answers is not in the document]
2opremio: Actually, we need it, for `ServeHTTP()` (which we should probably remove, but that should happen in a separate PR).

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -53,16 +60,42 @@ func NewJSONRPCHandler(cfg *config.LocalConfig, params HandlerParams) Handler {
...
+}
+
+type rpcLogger struct {
+	logger *log.Entry
+}
+
+func (r *rpcLogger) LogRequest(ctx context.Context, req *jrpc2.Request) {
+	r.logger.WithFields(log.F{
+		"subsys":   "jsonrpc",
+		"req":      middleware.GetReqID(ctx),
```
2opremio, line 79: This unfortunately doesn't work, because the HTTP request context is not propagated to the JSONRPC handlers :S . I will see if I can fix it.

2opremio: I don't see a way to do it, I created an upstream ticket https://github.com/creachadair/jrpc2/issues/99

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -53,16 +60,42 @@ func NewJSONRPCHandler(cfg *config.LocalConfig, params HandlerParams) Handler {
...
+		"json_req": rsp.ID(),
+		// TODO: is this overkill?
+		"result": rsp.ResultString(),
+	})
+	if err := rsp.Error(); err != nil {
+		logger = logger.WithField("error", err.Error())
 	}
+	logger.Info("finished JSONRPC request")
```
tsachiherman, line 100: this one would go away, right ?

2opremio: what do you mean? why would it go away?

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: Or ... we can add a debug-level log for those so that they only show when the log level is set to debug

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: The maintainer confirms there is no way to do it and discourages mixing HTTP with JSONRPC. I think he has a point (CC @paulbellamy )

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -53,16 +60,48 @@ func NewJSONRPCHandler(cfg *config.LocalConfig, params HandlerParams) Handler {
...
+		"req":      middleware.GetReqID(ctx),
+		"json_req": req.ID(),
+		"method":   req.Method(),
+	})
+	logger.Info("starting JSONRPC request")
+
+	// Params are useful but can be really verbose, let's only print them in debug
+	logger = logger.WithField("params", req.ParamString())
+	logger.Debug("starting JSONRPC request params")
+
```
tsachiherman, line 88: unnecessary trailing newline (whitespace)

cmd/soroban-rpc/internal/jsonrpc.go
tsachiherman: I think that you've added quite many log entries here. I think that it would work well for debugging, but might be overkill for production.
i.e. the amount of logging might be such that would end up impacting the server throughput.

I think that for now it's probably just fine, but for the long run, we'll need to update the logging to limit it to 1 log line per incoming request.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
-	"net/http/pprof"
+	_ "net/http/pprof"
```
tsachiherman, line 7: could you postfix //nolint:gosec on this line to avoid the linter warning ?

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -239,10 +241,10 @@ func MustNew(cfg config.LocalConfig, endpoint string, adminEndpoint string) *Dae
 		ReadTimeout: defaultReadTimeout,
 	}
 	if adminEndpoint != "" {
-		adminMux := http.NewServeMux()
+		adminMux := supporthttp.NewMux(logger)
+		// after importing net/http/pprof, debug endpoints are implicitly registered in the default serve mux
+		adminMux.Handle("/", http.DefaultServeMux)
```
tsachiherman, line 246: Using explicit handlers like you had before is a bit safer since you can be sure that all the handlers are being defined here. On it's own, I'm sure that it's fine - but it might trigger undesired exposed endpoint in the future.

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: Also, the JSONRPC context is different from the HTTP context, see https://github.com/stellar/soroban-tools/pull/584#discussion_r1166710936

cmd/soroban-rpc/internal/daemon/daemon.go
2opremio: Problems is there are quite a few handlers. I can do that though

2opremio: Another problem is that new handles are added with time and we will miss them.

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -53,16 +60,49 @@ func NewJSONRPCHandler(cfg *config.LocalConfig, params HandlerParams) Handler {
...
+type rpcLogger struct {
+	logger *log.Entry
+}
+
+func (r *rpcLogger) LogRequest(ctx context.Context, req *jrpc2.Request) {
+	logger := r.logger.WithFields(log.F{
+		"subsys": "jsonrpc",
+		// FIXME: the HTTP request context is independent from the JSONRPC context, and thuse the code below doesn't work
+		// "req":      middleware.GetReqID(ctx),
```
paulbellamy, line 80: Thought: in opentracing terminology, it seems like the jsonrpc handler would be a subspan of the http handler. 🤔

paulbellamy (approved):

cmd/soroban-rpc/internal/jsonrpc.go
2opremio: Exactly what I thought. The problem is we don’t have a way to join the spans

Issue #374, opened 2023-01-31
soroban-rpc: Replace `getLedgerEntry` with `getLedgerEntries`


paulbellamy:
### What problem does your feature solve?

Currently, to fetch multiple ledger entries you have to issue multiple `getLedgerEntry` calls, and you might get them as of multiple ledger states.

### What would you like to see?

Replacing:
```ts
getLedgerEntry(key: xdr.LedgerKey): {
  xdr: xdr.LedgerEntryData,
  lastModifiedLedgerSeq: number,
  latestLedger: number,
}
```
with:
```ts
getLedgerEntries(keys: xdr.LedgerKey[]): {
  entries: (null | {
    xdr: xdr.LedgerEntryData,
    lastModifiedLedgerSeq: number
  })[],
  latestLedger: number,
}
```
If any of the keys are not found they should be put as `null` into the output fields. This is so the client can pair up the requested keys with their results.

We will need to:
- Add the new method, and deprecate the old one in the first release.
- Remove the old method in the next release.

stellarsaur:
Have we considered adding `getLedgerEntries`, but keeping `getLedgerEntry` as-is? Saves users the effort to update how they're calling the RPC server, and makes more sense semantically to use `getLedgerEntry` when you only want to get a single entry.

Just a thought.

tomerweller:
Instead of modifying `getLedgerEntry` to work on multiple keys, can we have our json-rpc handler allow for a batch of requests to operate on a consistent state? Consistent state across a batch of requests can potentially have benefits beyond just batching ledger entry reads.

paulbellamy:
+1 for batching support in future, but that's a bigger lift, and there's no reason not to do both `getLedgerEntries` now, and batching later, imo.

PR #587, opened 2023-04-14
cmd/soroban-rpc: Add getLedgerEntries JSON RPC Method and Use it in Soroban CLI

stellarsaur:
### What

This PR accomplishes the following:
* Adds new `getLedgerEntries` JSON RPC handler to soroban-rpc
* Deprecates `getLedgerEntry` JSON RPC method
* Replaces instances of `get_ledger_entry` with the matching call to `get_ledger_entries` in soroban-cli

Part of [#374](https://github.com/stellar/soroban-tools/issues/374)

### Why

To fetch multiple ledger entries you have to issue multiple `getLedgerEntry` calls, and you might get them as of multiple ledger states. Now you will be able to get multiple ledgers for the same ledger state.


cmd/soroban-rpc/internal/methods/get_ledger_entries.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/db"
+)
+
+type GetLedgerEntriesRequest struct {
+	Keys []string `json:"keys"`
+}
+
+type LedgerEntryResult struct {
+	Found              bool   `json:"found"`
```
stellarsaur, line 22: Wasn't sure how to cleanly have "null" as a result (not sure if changing `Entries` in `GetLedgerEntriesResponse` to type `[]*LedgerEntryResult` is the way to go), so I went with adding this `Found` field which clients should check is True before processing an entry. Open to suggestions on how to handle this.

sreuland: could keep it this way, or another way to consider from a restful resource pattern when query is by list of ids, it's more a search then a getter and response can just contain entries that matched, ranging from an empty list to whatever. it won't indicate 'not found' per id, but client can derive that on their own if it needed. I think if you want that signal in the response model, another way to accomplish it over json-rpc would be:
```
GetLedgerEntryRequest struct {
   Key string `json:"key"`
}
```
And client submits multiple of these in a json-rpc batch request, that way client gets a response for each requested id in a batch result.

stellarsaur: Both good points. I went with replacing the `Found` field in `LedgerEntryResult` with a `Key` field, and only including present ledger entries in the final `GetLedgerEntriesResponse`

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -567,39 +567,45 @@ async fn get_remote_contract_spec_entries(
...
-                    let code_vec: Vec<u8> = code.into();
-                    soroban_spec::read::from_wasm(&code_vec)
-                        .map_err(Error::CannotParseContractSpec)?
+    let contract_key = LedgerKey::ContractData(LedgerKeyContractData {
+        contract_id: xdr::Hash(*contract_id),
+        key: ScVal::LedgerKeyContractExecutable,
+    });
+    let contract_ref = client.get_ledger_entries(Vec::from([contract_key])).await?;
+    if contract_ref.entries.is_empty() {
+        return Err(Error::MissingOperationResult);
```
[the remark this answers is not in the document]
stellarsaur: I couldn't find a better existing error code here so I added a "MissingResult" error similar to what `mod.rs` uses.

cmd/soroban-rpc/internal/methods/get_ledger_entries.go
```diff
@@ -0,0 +1,112 @@
...
+		if err != nil {
+			return GetLedgerEntriesResponse{}, &jrpc2.Error{
+				Code:    code.InternalError,
+				Message: "could not get latest ledger",
+			}
+		}
+
+		var ledgerEntryResults []LedgerEntryResult
+		for i, ledgerKey := range ledgerKeys {
+			present, ledgerEntry, err := tx.GetLedgerEntry(ledgerKey)
```
2opremio, line 75: @stellarsaur Sorry this comes late, I was reading this code and I think:

1. we should probably query all entries at once
2. we should probably set a limit in the amount of entries requested

PR #594, opened 2023-04-17
cmd/soroban-rpc: Add metrics for ingestion, event store, and transaction store

tamirms:
### What

Add metrics for ingestion, event store, and transaction store.

### Why

### Known limitations

[N/A]

### DIFF
--- a/cmd/soroban-rpc/internal/daemon/daemon.go
+++ b/cmd/soroban-rpc/internal/daemon/daemon.go
@@ -42,7 +42,6 @@ type Daemon struct {
 	ingestService       *ingest.Service
 	db                  dbsession.SessionInterface
 	jsonRPCHandler      *internal.Handler
-	httpHandler         http.Handler
 	logger              *supportlog.Entry
 	preflightWorkerPool *preflight.PreflightWorkerPool
 	server              *http.Server
@@ -51,10 +50,6 @@ type Daemon struct {
 	closeError          error
 	done                chan struct{}
 	metricsRegistry     *prometheus.Registry
-}
-
-func (d *Daemon) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
-	d.httpHandler.ServeHTTP(writer, request)
 }
 
 func (d *Daemon) GetDB() dbsession.SessionInterface {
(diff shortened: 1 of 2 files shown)


cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -162,8 +162,18 @@ func MustNew(cfg config.LocalConfig, endpoint string, adminEndpoint string) *Dae
 	}
 	dbConn := dbsession.RegisterMetrics(session, "soroban_rpc", "db", prometheusRegistry)
 
-	eventStore := events.NewMemoryStore(cfg.NetworkPassphrase, cfg.EventLedgerRetentionWindow)
-	transactionStore := transactions.NewMemoryStore(cfg.NetworkPassphrase, cfg.TransactionLedgerRetentionWindow)
+	eventStore := events.NewMemoryStore(
+		prometheusRegistry,
+		"soroban_rpc",
```
tsachiherman, line 167: I think there is no added value in passing "soroban_rpc" all around the place.
Instead, in /internal/metrics, create a file named `metrics.go` that would have a single const
```golang
const (
const PrometheusNamespacesSorobanRPC = "soroban_rpc"
)
```
and just use it everywhere instead of passing it on the stack.

tsachiherman (changes requested):
Could you take a look on the `Unchanged files with check annotations` in https://github.com/stellar/soroban-tools/pull/594/files ? it shows quite many testing bugs.

I think that if you could have used the singleton pattern here, you would reduce ( by much ) the amount of files you had to change to support this. it would also allow you to have the dependency on `github.com/prometheus/client_golang/prometheus` get used only in a single place ( i.e. /internal/metrics ) rather than all around the place.

cmd/soroban-rpc/internal/daemon/daemon.go
```diff
@@ -3,6 +3,7 @@ package daemon
 import (
 	"context"
 	"errors"
+	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/metrics"
```
tsachiherman, line 6: could you move this line to the group on the bottom ?

tsachiherman (approved):

PR #598, opened 2023-04-18
soroban-rpc: standarize logging output format

2opremio:
### What

Standarize logging output format. (e.g. always start logs in lowercase, use an error field)

### Why

#497 

### Known limitations

N/A


Shaptic (approved):
nice :clap:

PR #600, opened 2023-04-20
cmd/soroban-rpc: Add txsub metrics

tamirms:
### What

Add metrics to:

-  count transaction submission counts / duration broken down by status.
- count number of operations per envelope broken down by tx sub status.

### Why

### Known limitations

[N/A]

### DIFF
--- a/cmd/soroban-rpc/internal/daemon/metrics.go
+++ b/cmd/soroban-rpc/internal/daemon/metrics.go
@@ -1,11 +1,18 @@
 package daemon
 
 import (
+	"context"
 	"runtime"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stellar/go/clients/stellarcore"
+	proto "github.com/stellar/go/protocols/stellarcore"
 	"github.com/stellar/go/support/logmetrics"
+	"github.com/stellar/go/xdr"
+
 	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/config"
+	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/daemon/interfaces"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/soroban-rpc/internal/daemon/metrics.go
```diff
@@ -40,3 +47,56 @@ func (d *Daemon) MetricsRegistry() *prometheus.Registry {
...
 }
+
+type CoreClientWithMetrics struct {
+	stellarcore.Client
+	submitMetric  *prometheus.SummaryVec
+	opCountMetric *prometheus.SummaryVec
+}
+
+func newCoreClientWithMetrics(client stellarcore.Client, registry *prometheus.Registry) *CoreClientWithMetrics {
+	submitMetric := prometheus.NewSummaryVec(prometheus.SummaryOpts{
```
paulbellamy, line 58: Should we aim to match the metrics we emit from horizon txsub? wdyt?
https://github.com/stellar/go/blob/master/services/horizon/internal/txsub/system.go#L410

paulbellamy: No, the concern is that it seems like horizon tracks these a bit more in-depth. open/buffered/failed, etc. I like using the response status, but we don't currently get visibility into the actual txsub queueing. But I guess we don't have any queueing here, so txsub is a bit simpler... 🤔

paulbellamy: (And that it would be nice to match naming and metrics conventions for consistency)

paulbellamy (approved):

tsachiherman (approved):
Looks great. I'm not sure if these metrics are overkill as Paul mentioned.

PR #602, opened 2023-04-20
cmd/soroban-rpc: Add metrics for preflight worker pool

tamirms:
### What

Add metrics for preflight worker pool.

### Why

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/preflight/pool.go
+++ b/cmd/soroban-rpc/internal/preflight/pool.go
@@ -68,7 +68,7 @@ func NewPreflightWorkerPool(daemon interfaces.Daemon, workerCount uint, jobQueue
 	preflightWP.durationMetric = prometheus.NewSummaryVec(prometheus.SummaryOpts{
 		Namespace: daemon.MetricsNamespace(),
 		Subsystem: "preflight_pool",
-		Name:      "request_duration",
+		Name:      "request_ledger_get_duration_seconds",
 		Help:      "preflight request duration broken down by status",
 	}, []string{"status", "type"})
 	daemon.MetricsRegistry().MustRegister(
@@ -150,7 +150,7 @@ func (pwp *PreflightWorkerPool) GetPreflight(ctx context.Context, readTx db.Ledg
 		}
 		pwp.durationMetric.With(
 			prometheus.Labels{"type": "db", "status": status},
-		).Observe(float64(wrappedTx.totalDurationMs * 1000))
+		).Observe(float64(wrappedTx.totalDurationMs) / 1000.0)
 		return result.preflight, result.err
 	case <-ctx.Done():
 		return Preflight{}, ctx.Err()


paulbellamy (approved):

cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -49,7 +76,9 @@ func NewPreflightWorkerPool(workerCount uint, jobQueueCapacity uint, ledgerEntry
 func (pwp *PreflightWorkerPool) work() {
 	defer pwp.wg.Done()
 	for request := range pwp.requestChan {
+		pwp.concurrentRequestsMetric.Inc()
```
tsachiherman, line 79: We could do that inside GetPreflight as well, but it might be cleaner to do it here : 
useful metrics for the execution of GetPreflight would be:
1. the time it takes to execute
2. the amount of time spent on DB access for a single GetPreflight
3. successfull vs. non-successful preflights invocations.

2opremio (approved):

cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -65,25 +113,49 @@ func (pwp *PreflightWorkerPool) Close() {
...
 	select {
 	case pwp.requestChan <- workerRequest{ctx, params, resultC}:
 		result := <-resultC
+		status := "ok"
+		if result.err != nil {
+			status = "error"
+		}
+		pwp.durationMetric.With(
+			prometheus.Labels{"type": "db", "status": status},
+		).Observe(float64(wrappedTx.totalDurationMs * 1000))
```
tsachiherman, line 153: Which unit did you intend to meter here ? It's not seconds here but rather microsecond. Was that your intent ?

cmd/soroban-rpc/internal/preflight/pool.go
```diff
@@ -24,21 +27,56 @@ type workerRequest struct {
...
+	preflightWP.errorFullCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: daemon.MetricsNamespace(),
+		Subsystem: "preflight_pool",
+		Name:      "queue_full_errors",
+		Help:      "number of preflight full queue errors",
+	})
+	preflightWP.durationMetric = prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		Namespace: daemon.MetricsNamespace(),
+		Subsystem: "preflight_pool",
+		Name:      "request_duration",
```
tsachiherman, line 71: I think that you want to call it "request_ledger_get_duration_seconds" for consistency with the rest of the duration-like metrics. This isn't just the request duration, but rather measuring the total of the database operations.

cmd/soroban-rpc/internal/preflight/pool.go
tsachiherman: assuming you'll call the metric `request_ledger_get_duration_seconds`, then you'd need to divide by 1000. obviously,  do so outside of the float64 conversion

tsachiherman (approved):

PR #603, opened 2023-04-20
soroban-rpc: Add metrics for JSON RPC handlers

tamirms:
### What

Add metrics for JSON RPC handlers

### Why

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/test/metrics_test.go
+++ b/cmd/soroban-rpc/internal/test/metrics_test.go
@@ -2,13 +2,15 @@ package test
 
 import (
 	"fmt"
-	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/config"
 	"io"
 	"net/http"
+	"net/url"
 	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/config"
 )
 
 func TestMetrics(t *testing.T) {
@@ -26,7 +28,9 @@ func TestMetrics(t *testing.T) {
 }
 
 func getMetrics(test *Test) string {
-	response, err := http.Get(test.adminURL() + "/metrics")
+	metricsURL, err := url.JoinPath(test.adminURL(), "/metrics")
+	require.NoError(test.t, err)
+	response, err := http.Get(metricsURL)
 	require.NoError(test.t, err)
 	responseBytes, err := io.ReadAll(response.Body)
 	require.NoError(test.t, err)


tsachiherman (changes requested):
Please add a unit test that verify that this work.

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -42,6 +44,44 @@ type HandlerParams struct {
 	Daemon            interfaces.Daemon
 }
 
+type handlerWithMetrics struct {
+	handler       jrpc2.Handler
+	endpoint      string
```
2opremio, line 49: I don't think we need this. Can't we use `Request.Method()`?

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -42,6 +44,44 @@ type HandlerParams struct {
...
+
+func (h *handlerWithMetrics) Handle(ctx context.Context, r *jrpc2.Request) (interface{}, error) {
+	startTime := time.Now()
+	result, err := h.handler.Handle(ctx, r)
+	duration := time.Since(startTime).Seconds()
+	label := prometheus.Labels{"status": "ok"}
+	simulateTransactionResponse, ok := result.(methods.SimulateTransactionResponse)
+	if err != nil || ok && simulateTransactionResponse.Error != "" {
+		label = prometheus.Labels{"status": "error"}
+	}
```
2opremio, line 61: Shouldn't the same be done for the generic JSON-RPC `Response.Error()` field?

https://pkg.go.dev/github.com/creachadair/jrpc2#Response.Error

2opremio: We may actually want to record the error codes separately

cmd/soroban-rpc/internal/test/metrics_test.go
```diff
@@ -28,5 +22,14 @@ func TestMetrics(t *testing.T) {
 		runtime.Version(),
 		config.Version,
 	)
-	require.Contains(t, responseString, buildMetric)
+	require.Contains(t, metrics, buildMetric)
+}
+
+func getMetrics(test *Test) string {
+	response, err := http.Get(test.adminURL() + "/metrics")
```
tsachiherman, line 29: this is probably ok here, but just fyi, this is usually being done using "path.Join", which would eliminate double `/`, etc.

tsachiherman (approved):

cmd/soroban-rpc/internal/test/metrics_test.go
tsachiherman: sorry - I should have written [`url.JoinPath`](https://pkg.go.dev/net/url@go1.19beta1#JoinPath)

Issue #403, opened 2023-02-09
soroban-cli: Replace all hex-encoded contract ids with strkeys


paulbellamy:
### What problem does your feature solve?

Currently there are places we input/output contract ids as hex-encoded strings `abcd1234...`. But there are others, like contract `Address` params, where we input/output them as `C...` strkeys.

### What would you like to see?

We should stop encoding contract ids as hex bytes, and always use `C...` strkeys for contract IDs. This would include when you deploy them, invoke them, etc...

We should do it sooner the better, so that tooling and contracts in the ecosystem can standardize on this.

### What alternatives are there?

- Support both? This is tricky, as we would need to try parsing as a strkey, then parse as hex. And people might use `BytesN` for their contract params. And we would need to pick an output format anyway, so it'll be as confusing as it is now.
- Use hex everywhere? But then it'll clash with `Address` type in the contracts, so there is a weird exception.

tyvdh:
Discord conversation context
https://discord.com/channels/897514728459468821/1111673776317399103

PR #604, opened 2023-04-20
feat: Allow passing an identity for an address 

willemneal:
### What

Reuse the `identity::address` command to look up identities when parsing an `Address`.

### Why

Allows aliases instead of having to pass the public key.

### Known limitations

[TODO or N/A]


.cargo/config.toml
paulbellamy, line 1: Accidental commit?

.github/workflows/full-help-docs.yml
paulbellamy, line 1: Also seems unrelated?

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -189,7 +189,17 @@ impl Cmd {
             .map(|i| {
                 let name = i.name.to_string().unwrap();
                 if let Some(mut raw_val) = matches_.get_raw(&name) {
-                    let s = raw_val.next().unwrap().to_string_lossy().to_string();
+                    let mut s = raw_val.next().unwrap().to_string_lossy().to_string();
+                    if matches!(i.type_, ScSpecTypeDef::Address) {
+                        let cmd = crate::commands::config::identity::address::Cmd {
+                            name: Some(s.clone()),
+                            hd_path: Some(0),
```
paulbellamy, line 196: Unrelated side note: Been thinking about this hd-path thing more, and it feels like a sort of leaky abstraction in identities. What if it was rolled into a named identity thing? So, instead of passing `--hd-path` on the CLI you'd first have to do:
```
# Explicitly generate a derived subpath key for usage, something like
$ soroban config identity generate --parent existing_identity --hd-path 1 child_name

# and then you'd use it like a normal identity
$ soroban contract invoke --source child_name ...
```
Thoughts?

cmd/soroban-cli/src/strval.rs
```diff
@@ -82,14 +82,16 @@ impl Spec {
...
-            | ScType::Address
             | ScType::Void
             | ScType::Timepoint
             | ScType::Duration
             | ScType::U256
             | ScType::I256
             | ScType::String
             | ScType::Bool => String::new(),
+            ScType::Address => String::from(
+                "Can be public key (G13..), a contract hash (6c45307) or an identity (alice), ",
```
paulbellamy, line 93: Or an `S...` key?

paulbellamy: Also, we should get rid of contract hash support here when we do https://github.com/stellar/soroban-tools/issues/403

willemneal: Yeah we could, but it's a bad practice to put a secret in plaintext in a CLI command.   But for scripts I guess it's okay.

.github/workflows/full-help-docs.yml
willemneal: Yeah.  I'll remove these files.

cmd/soroban-cli/src/commands/contract/invoke.rs
willemneal: I was thinking that we could also do something like `existing_id:<hd_path>` then that id file can just reference the other id.

.cargo/config.toml
willemneal: Ah yeah this was meant for another PR. But it ensures that the clippy warnings happen when building and adds an alias for markdown

paulbellamy (approved):

PR #625, opened 2023-05-03
Bump Rust dependencies to support fees

### LINKED ISSUE #621: soroban-tools: Update Cargo.toml to meet core's fees changes
### What ?

The XDR was modified to support fees changes. This issue is about the update of the rust components in the soroban-tools repository to compile with these changes.
Two "big" changes that this XDR includes are:
- InvokeHostFunction now includes a list of functions rather then a single function
- The Footprint doesn't go on the operations anymore. Instead, it's a Transaction level field.

## Epic

2opremio:
### What

Bump dependencies and add support fee support to soroban-tools

### Why

Closes #621 

### Known limitations

This PR breaks Soroban-RPCs API (the `xdr` field is renamed to `xdrs` which is now a list containing multiple results).

This PR just handles proper complication of the Rust/Go code following XDR changes.
It still has a few TODOs and is complementary to https://github.com/stellar/soroban-tools/pull/627 .

This will *break* CI when merged. The remaining tasks for this to work are:

* Updating the Go code https://github.com/stellar/soroban-tools/pull/627
* Producing and filling in the new fee fields. This will require API changes
* Updating the downstream SDKs to accomodate the new API.


cmd/soroban-cli/src/commands/contract/install.rs
```diff
@@ -125,7 +120,23 @@ pub(crate) fn build_install_contract_code_tx(
...
+                    read_write: vec![ContractCode(LedgerKeyContractCode { hash: hash.clone() })]
+                        .try_into()?,
+                },
+                // TODO: what values should be used here?
+                instructions: 0,
+                read_bytes: 0,
+                write_bytes: 0,
+                extended_meta_data_size_bytes: 0,
+            },
+            // TODO: what value to use here?
```
tsachiherman, line 136: I discussed that with @dmkozh and @sisuresh. Leave it like this for now, and once the simulateTransaction endpoint is ready, we'll need to update this code to use the endpoint to figure out these values.

2opremio: Yes, I was planning to do that :) 

However, the question is, what values to use for cases in which we were not preflight-ing  the transaction (e.g. when uploading the contract).

Will we now need to start preflight-ing  those transactions or can the values be inferred (like we used to do before introducing fees)

tsachiherman (approved):

Issue #519, opened 2023-03-16
[epic] Platform May Release 


tsachiherman:
This epic contains all the upcoming core features that needs to be supported by the platform team:
* Network Config Ledger Entries ( a.k.a. Network Settings )
* Token Interface Changes
* Fees
* Metering / Limits
* Instance Metadata
* Contract Upgrades
* Soroban RPC Productization
 
### Network Config Ledger Entries ( a.k.a. Network Settings ) (1 day)
* https://github.com/stellar/soroban-tools/issues/513 (2 days)
* https://github.com/stellar/go/issues/4788 (1 day)

### Token Interface Changes (9 story points)
* https://github.com/stellar/soroban-tools/issues/590 (1 sp)
* https://github.com/stellar/soroban-tools/issues/591 (2 sp)
* https://github.com/stellar/soroban-example-dapp/issues/103 (2 sp)
* https://github.com/stellar/soroban-docs/issues/387 (1 sp)
* https://github.com/stellar/soroban-examples/issues/230 (1 sp)
* https://github.com/stellar/go/issues/4850 (2 sp)

### Fees
* https://github.com/stellar/soroban-tools/issues/588
* https://github.com/stellar/go/issues/4863
* https://github.com/stellar/soroban-tools/issues/621
* https://github.com/stellar/soroban-tools/issues/622
* https://github.com/stellar/soroban-tools/issues/623
  - https://github.com/stellar/soroban-tools/pull/627
  - https://github.com/stellar/soroban-tools/pull/628
* https://github.com/stellar/soroban-tools/issues/624
* https://github.com/stellar/js-stellar-base/issues/596
* https://github.com/stellar/js-soroban-client/issues/76
* https://github.com/stellar/soroban-example-dapp/issues/107
* https://github.com/stellar/soroban-docs/pull/439

### Instance Metadata
* https://github.com/stellar/soroban-tools/issues/593 (2 sp)
* https://github.com/stellar/soroban-example-dapp/issues/104

### Soroban RPC Productization
( the list below is a subset of the tasks https://github.com/stellar/soroban-tools/issues/483 )
* https://github.com/stellar/soroban-tools/issues/496
* https://github.com/stellar/soroban-tools/issues/507
* https://github.com/stellar/soroban-tools/issues/499
* https://github.com/stellar/soroban-tools/pull/449
* https://github.com/stellar/soroban-tools/issues/498
* https://github.com/stellar/soroban-tools/issues/497
[...]

PR #627, opened 2023-05-04
soroban-rpc: update xdr

tsachiherman:
### What

Update the go monorepo dependency

### Why

Updating the go monorepo dependency also brings the XDR library updates.

### Known limitations

This PR just handles proper complication of the code following the massive XDR changes.

### Epic


cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -123,7 +123,7 @@ func GoNullTerminatedStringSlice(str **C.char) []string {
 }
 
 func GetPreflight(ctx context.Context, params PreflightParameters) (Preflight, error) {
-	hfB64, err := xdr.MarshalBase64(params.InvokeHostFunction.Function)
+	hfB64, err := xdr.MarshalBase64(params.InvokeHostFunction.Functions)
```
2opremio, line 126: I guess this implies we should change the `simulateTransaction` method to support multiple functions

2opremio: Actually, if we do it like that in Rust, it complicate the Rust<->C<->Go interface quite a bit. So, at least in that layer, we will still do it one by one.

2opremio: Also, I think that we should pass the whole operation to preflight

2opremio: i.e. 

```suggestion
	hfB64, err := xdr.MarshalBase64(params.InvokeHostFunction)
```

tsachiherman: That sounds like a good idea. I wasn't going to complete the implementation here, since I need your PR ready first.

tsachiherman: I.e. the preflightlib dependencies needs to be updated before any of these changes are meaningful.

2opremio (approved):

PR #630, opened 2023-05-04
cli: Unveil broken Rust tests

2opremio:
### What

Unveil tests which were broken and we didn't notice because they were not run by CI :S

### Why

We want all tests to pass

### Known limitations

N/A


2opremio:
Related: https://github.com/stellar/soroban-tools/pull/547

2opremio:
After further thought .... I think that tests are running successfully in CI (x86), but are failing in my machine (Apple Silicon).

For instance, it seems like they are all run successfully at  https://github.com/stellar/soroban-tools/actions/runs/4863433526/jobs/8671137386

willemneal:
@2opremio What errors are you getting?

PR #631, opened 2023-05-04
Try new docker compose plugin

paulbellamy:
### What

Apparently the docker issue in https://github.com/stellar/soroban-tools/pull/628, was *not* fixed, but is just intermittent. :(

### Why

All the docker support issues say to use docker-compose v2, so let's try installing that.

### Known limitations

[TODO or N/A]


sreuland (approved):
nice work, it's substantial amount of scaffolding to get it working, too bad there isn't a solid  GHA action for docker-compose it seems.

PR #636, opened 2023-05-05
soroban-rpc: Calculate soroban transaction data in libpreflight

### LINKED ISSUE #622: preflightlib : add support for fee calculations
### What ?

The rs-soroban-env now contains library functions to calculate the fees needed by a transaction, based on it's resources.
Our goal is to integrate that into simulateTransaction endpoint in soroban-rpc. This issue is about the first step in the process:
We need to expose the newly added function via the preflightlib so that soroban-rpc would be able to call it.
This issue is focused around adding support for calling the new method, plus ensuring that soroban-rpc is capable of importing and calling. It doesn't cover any plumbing around the simulateTransaction endpoint itself ( that would be handled in a separate ticket ).

### Epic

2opremio:
### What

Calculate soroban transaction data in libpreflight

This changes the API of `simulateTransaction`. `SimulateTransactionResponse` changes as follows:

* Instead of returning a footprint, we return the full soroban transaction data.
* The list of events is returned globally instead of per host function
* A new field, `minResourceFee`, is added to tell the user what fee should be used with respect to the resources consumed by the host function invocations.

### Why

Closes #622 

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -100,14 +100,11 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 
 		hostFunctionResults := make([]SimulateHostFunctionResult, len(result.Results))
 		for i := 0; i < len(hostFunctionResults); i++ {
-			hostFunctionResults[i].XDR = result.Results[i]
+			hostFunctionResults[i] = SimulateHostFunctionResult{
+				Auth: result.Results[i].Auth,
+				XDR:  result.Results[i].Result,
+			}
 		}
-
-		// For now, attribute the full auth and and events to the first function
-		//
-		// FIXME: this is wrong! we should be able to get the auth and events for each separate function
-		//        but needs to be implemented in libpreflight first
-		hostFunctionResults[0].Auth = result.Auth
 
 		return SimulateTransactionResponse{
 			Results:         hostFunctionResults,
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -33,8 +33,8 @@ type SimulateTransactionResponse struct {
 	TransactionData string                       `json:"transactionData"` // SorobanTransactionData XDR in base64
 	Events          []string                     `json:"events"`          // DiagnosticEvent XDR in base64
 	MinResourceFee  int64                        `json:"minResourceFee,string"`
-	Results         []SimulateHostFunctionResult `json:"results,omitempty"`
-	Cost            SimulateTransactionCost      `json:"cost"`
+	Results         []SimulateHostFunctionResult `json:"results,omitempty"` // an array of the individual host function call results
+	Cost            SimulateTransactionCost      `json:"cost"`              // the effective cpu and memory cost of the invoked transaction execution.
 	LatestLedger    int64                        `json:"latestLedger,string"`
 }
 
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -30,8 +30,8 @@ type SimulateHostFunctionResult struct {
 type SimulateTransactionResponse struct {
 	Error string `json:"error,omitempty"`
 	// TODO: update documentation and review field names
-	TransactionData string                       `json:"transaction_data"` // SorobanTransactionData XDR in base64
-	Events          []string                     `json:"events"`           // DiagnosticEvent XDR in base64
+	TransactionData string                       `json:"transactionData"` // SorobanTransactionData XDR in base64
+	Events          []string                     `json:"events"`          // DiagnosticEvent XDR in base64
 	MinResourceFee  int64                        `json:"minResourceFee,string"`
 	Results         []SimulateHostFunctionResult `json:"results,omitempty"`
 	Cost            SimulateTransactionCost      `json:"cost"`
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -21,19 +21,23 @@ type SimulateTransactionCost struct {
 	MemoryBytes     uint64 `json:"memBytes,string"`
 }
 
+// TODO: should this refer to each InvokeHostFunctionOp within the transaction
+//
+//	or to each HostFunction within each InvokeHostFunctionOp
 type SimulateTransactionResult struct {
-	Auth      []string `json:"auth"`
-	Events    []string `json:"events"`
-	Footprint string   `json:"footprint"`
+	Auth   []string `json:"auth"`
+	Events []string `json:"events"`
 	// TODO: update documentation and review field name
-	XDRs []string `json:"xdrs"`
+	XDRs []string `json:"xdrs"` // SCVal XDRs in base64
 }
 
 type SimulateTransactionResponse struct {
-	Error        string                      `json:"error,omitempty"`
-	Results      []SimulateTransactionResult `json:"results,omitempty"`
-	Cost         SimulateTransactionCost     `json:"cost"`
-	LatestLedger int64                       `json:"latestLedger,string"`
+	Error string `json:"error,omitempty"`
+	// TODO: update documentation and review field name
+	TransactionData string                      `json:"transaction_data"` // SorobanTransactionData XDR in base64
+	Results         []SimulateTransactionResult `json:"results,omitempty"`
+	Cost            SimulateTransactionCost     `json:"cost"`
+	LatestLedger    int64                       `json:"latestLedger,string"`
 }
 
 type PreflightGetter interface {
@@ -99,12 +103,12 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 		return SimulateTransactionResponse{
 			Results: []SimulateTransactionResult{
 				{
-					Events:    result.Events,
-					Auth:      result.Auth,
-					Footprint: result.Footprint,
-					XDRs:      result.Results,
+					Events: result.Events,
+					Auth:   result.Auth,
+					XDRs:   result.Results,
 				},
 			},
+			TransactionData: result.TransactionData,
 			Cost: SimulateTransactionCost{
 				CPUInstructions: result.CPUInstructions,
 				MemoryBytes:     result.MemoryBytes,
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -23,15 +23,15 @@ type SimulateTransactionCost struct {
 
 // SimulateHostFunctionResult contains the simulation result of each HostFunction within each InvokeHostFunctionOp
 type SimulateHostFunctionResult struct {
-	Auth   []string `json:"auth"`
-	Events []string `json:"events"`
-	XDR    string   `json:"xdr"`
+	Auth []string `json:"auth"`
+	XDR  string   `json:"xdr"`
 }
 
 type SimulateTransactionResponse struct {
 	Error string `json:"error,omitempty"`
-	// TODO: update documentation and review field name
+	// TODO: update documentation and review field names
 	TransactionData string                       `json:"transaction_data"` // SorobanTransactionData XDR in base64
+	Events          []string                     `json:"events"`           // DiagnosticEvent XDR in base64
 	MinResourceFee  int64                        `json:"minResourceFee,string"`
 	Results         []SimulateHostFunctionResult `json:"results,omitempty"`
 	Cost            SimulateTransactionCost      `json:"cost"`
@@ -107,11 +107,11 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 		//
 		// FIXME: this is wrong! we should be able to get the auth and events for each separate function
 		//        but needs to be implemented in libpreflight first
-		hostFunctionResults[0].Events = result.Events
 		hostFunctionResults[0].Auth = result.Auth
 
 		return SimulateTransactionResponse{
 			Results:         hostFunctionResults,
+			Events:          result.Events,
 			TransactionData: result.TransactionData,
 			MinResourceFee:  result.MinFee,
 			Cost: SimulateTransactionCost{
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -113,9 +113,7 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 		return SimulateTransactionResponse{
 			Results:         hostFunctionResults,
 			TransactionData: result.TransactionData,
-			// FIXME: this cast shouldn't be necessary
-			//       find out why it doesn't work without it
-			MinFee: int64(result.MinFee),
+			MinFee:          result.MinFee,
 			Cost: SimulateTransactionCost{
 				CPUInstructions: result.CPUInstructions,
 				MemoryBytes:     result.MemoryBytes,
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -21,7 +21,7 @@ type SimulateTransactionCost struct {
 	MemoryBytes     uint64 `json:"memBytes,string"`
 }
 
-// SimulateHostFunctionResult contains the simulation result of each HostFunction within each InvokeHostFunctionOp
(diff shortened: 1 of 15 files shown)


cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -21,19 +21,23 @@ type SimulateTransactionCost struct {
...
 
 type SimulateTransactionResponse struct {
-	Error        string                      `json:"error,omitempty"`
-	Results      []SimulateTransactionResult `json:"results,omitempty"`
-	Cost         SimulateTransactionCost     `json:"cost"`
-	LatestLedger int64                       `json:"latestLedger,string"`
+	Error string `json:"error,omitempty"`
+	// TODO: update documentation and review field name
+	TransactionData string                      `json:"transaction_data"` // SorobanTransactionData XDR in base64
+	Results         []SimulateTransactionResult `json:"results,omitempty"`
```
sreuland, line 38: is there an implied association in `results` where it will have one item for each HostFunction that was present in `InvokeHostFunctionOp.functions`?

2opremio: I clarified it in the new commits I pushed.

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -185,25 +191,233 @@ fn preflight_host_function_or_maybe_panic(
...
+        write_bytes: soroban_resources.write_bytes,
+        metadata_size_bytes: soroban_resources.extended_meta_data_size_bytes,
+        // Note: we could get a better transaction size if the full transaction was passed down to libpreflight
+        transaction_size_bytes: estimate_max_transaction_size(
+            invoke_hf_op,
+            &soroban_resources.footprint,
+        )?,
+    };
+    let (min_fee, ref_fee) =
+        compute_transaction_resource_fee(&transaction_resources, &fee_configuration);
```
2opremio, line 245: @dmkozh I am interpreting `min_fee` as the fee required to access resources, on top of which the transaction submitted needs to add the _Stellar classic_ fee. Is that correct?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -96,15 +98,22 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
...
+		hostFunctionResults := make([]SimulateHostFunctionResult, len(result.Results))
+		for i := 0; i < len(hostFunctionResults); i++ {
+			hostFunctionResults[i].XDR = result.Results[i]
+		}
+
+		// For now, attribute the full auth and and events to the first function
+		//
+		// FIXME: this is wrong! we should be able to get the auth and events for each separate function
+		//        but needs to be implemented in libpreflight first
+		hostFunctionResults[0].Auth = result.Auth
```
sreuland, line 110: for consideration, maybe set each `hostFunctionResults[n].Auth = result.Auth` for now to just set in place the expectations on model, upstream clients can code to that, later when libpreflight can support Auth per host function invocation then upstream doesn't need to roll new version for parsing any different.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -21,19 +21,21 @@ type SimulateTransactionCost struct {
...
-	Results      []SimulateTransactionResult `json:"results,omitempty"`
-	Cost         SimulateTransactionCost     `json:"cost"`
-	LatestLedger int64                       `json:"latestLedger,string"`
+	Error string `json:"error,omitempty"`
+	// TODO: update documentation and review field names
+	TransactionData string                       `json:"transaction_data"` // SorobanTransactionData XDR in base64
+	Events          []string                     `json:"events"`           // DiagnosticEvent XDR in base64
+	MinResourceFee  int64                        `json:"minResourceFee,string"`
+	Results         []SimulateHostFunctionResult `json:"results,omitempty"`
+	Cost            SimulateTransactionCost      `json:"cost"`
```
sreuland, line 37: is `Cost` providing anything unique, it appeared like `TransactionData Resources` has the same attributes with further detail?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -21,19 +21,21 @@ type SimulateTransactionCost struct {
...
-type SimulateTransactionResult struct {
-	Auth      []string `json:"auth"`
-	Events    []string `json:"events"`
-	Footprint string   `json:"footprint"`
-	// TODO: update documentation and review field name
-	XDRs []string `json:"xdrs"`
+// SimulateHostFunctionResult contains the simulation result of each HostFunction within each InvokeHostFunctionOp
+type SimulateHostFunctionResult struct {
+	Auth []string `json:"auth"`
+	XDR  string   `json:"xdr"`
```
sreuland, line 27: I think this may need to include `hostFunctionIndex` and `operationIndex` to provide association, looking at it from upstream usage, it will need to match up each sim result instance here to a InvokeHostFunctionOp::HostFunction instance in the user's tx so that it can set the `auth` from sim result. or I may be mis-understanding intent of usage.

sreuland: oh, I think understand further, the association back to source/user's tx can be derived from the serialized order of these in `SimulateTransactionResponse.Results`, the first `SimulateHostFunctionResult` belonging to first InvokeHostFunctionOp and first HostFunction within source/user's tx and so on.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -21,19 +21,21 @@ type SimulateTransactionCost struct {
...
 type SimulateTransactionResponse struct {
-	Error        string                      `json:"error,omitempty"`
-	Results      []SimulateTransactionResult `json:"results,omitempty"`
-	Cost         SimulateTransactionCost     `json:"cost"`
-	LatestLedger int64                       `json:"latestLedger,string"`
+	Error string `json:"error,omitempty"`
+	// TODO: update documentation and review field names
+	TransactionData string                       `json:"transaction_data"` // SorobanTransactionData XDR in base64
+	Events          []string                     `json:"events"`           // DiagnosticEvent XDR in base64
+	MinResourceFee  int64                        `json:"minResourceFee,string"`
```
sreuland, line 35: is `minResourceFee` the extent of suggested inclusion fees from preflight for now?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: There can only be a single InvokeHostFunctionOp, so each result refers to each HostFunction

sreuland: got it, thanks, I read into the docs stated on line 24 too literally, and thought the single-op restriction had been lifted, just for consideration, could state that a bit more directly   `SimulateHostFunctionResult contains the simulation result of each HostFunction within the single InvokeHostFunctionOp allowed on a Tx`

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -21,19 +21,21 @@ type SimulateTransactionCost struct {
...
+	Results         []SimulateHostFunctionResult `json:"results,omitempty"`
```
sreuland, line 36: for consideration, remove the `omitempty`, to include results array in response, consistent with other collections in the struct like the events. makes a little easier to use from client side, doesn't have to check for absence first, etc.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -21,19 +21,21 @@ type SimulateTransactionCost struct {
...
 }
 
-type SimulateTransactionResult struct {
-	Auth      []string `json:"auth"`
-	Events    []string `json:"events"`
-	Footprint string   `json:"footprint"`
-	// TODO: update documentation and review field name
-	XDRs []string `json:"xdrs"`
+// SimulateHostFunctionResult contains the simulation result of each HostFunction within each InvokeHostFunctionOp
+type SimulateHostFunctionResult struct {
```
tsachiherman, line 25: It seems that we have lots of "Simulate" prefix here - which is correct, but might not place the focus where it should be. How do you feel about the following naming for this struct:
```golang
type ExecutedFunction {
	Auth []string `json:"auth"`
	Results  string   `json:"results"` // base64 xdr encoded...
}
```
The field name that I feel is really the issue is the `XDR`, since this is a type rather then a data element descriptor.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: That is true ... maybe we can remove them. @tsachiherman thoughts?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: Good point. I've updated the comment

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: The XDR name predates this release iteration. Also, `results` is redundant with the higher level name, plus it's a single result of a single function, so either way I think it should be singular.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: Yeah, that's the idea. Also, see https://github.com/stellar/soroban-tools/pull/636/files#r1186931388

cmd/soroban-rpc/internal/methods/simulate_transaction.go
tsachiherman: I'll look into that tomorrow morning, but I think that we shouldnt have any duplicate info in here.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
tsachiherman: How would it behave with classic transaction ?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
tsachiherman: I think that you agree with me that's it's undesirable, but don't have any concrete solution here. I'm not very opinionated here, since there is nothing structurally wrong, but want to try and figure out a better "names" for the constructs that we're creating.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
sreuland: the `results` array would be empty, clients figure out what do when no results, but at least they can depend on presence and don't have to null check first then reference the length, in cases where you expect this to be absent instead, should those be conveyed more as an `error`?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -21,19 +21,21 @@ type SimulateTransactionCost struct {
...
 }
 
 type SimulateTransactionResponse struct {
-	Error        string                      `json:"error,omitempty"`
-	Results      []SimulateTransactionResult `json:"results,omitempty"`
-	Cost         SimulateTransactionCost     `json:"cost"`
-	LatestLedger int64                       `json:"latestLedger,string"`
+	Error string `json:"error,omitempty"`
+	// TODO: update documentation and review field names
+	TransactionData string                       `json:"transaction_data"` // SorobanTransactionData XDR in base64
```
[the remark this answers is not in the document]
2opremio: Yes, my bad!

dmkozh (approved):
The fee calculation looks good to me.

tsachiherman:
@sreuland look at @dmkozh 's response above.

2opremio:
Thanks for the review @dmkozh !!

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -185,25 +191,235 @@ fn preflight_host_function_or_maybe_panic(
...
+    let mut signatures: Vec<DecoratedSignature> = vec![];
+    let mut signatures_left = 20;
+    while signatures_left > 0 {
+        signatures.push(DecoratedSignature {
+            hint: SignatureHint([0; 4]),
+            signature: Default::default(),
+        });
+        signatures_left -= 1;
+    }
+    let envelope = TransactionV1Envelope {
```
dmkozh, line 277: I'm actually looking only at the internal `Transaction` size, specifically to avoid the signatures coupling (not sure if that's 100% correct, but that's how it behaves now).

cmd/soroban-rpc/lib/preflight/src/lib.rs
sreuland: @tsachiherman , @2opremio , based on this statement and [response on the js-soroban-client pr for handling inclusion fee](https://github.com/stellar/js-soroban-client/pull/77#discussion_r1190451805) , I've tentatively [added 'suggestedInclusionFee' to expected rpc simulation response in js-soroban-client ](https://github.com/stellar/js-soroban-client/pull/77/files#diff-718319220445fb0cf5b6fe5c0d0c88085f8f32ee0a90d41b496037b00836b8c3R124), if you agree, would need to be added here on `SimulateTransactionResponse`, let me know if this is right direction, thanks!

cmd/soroban-rpc/internal/methods/simulate_transaction.go
tsachiherman: yes, absolutely. I don't see much value keeping this if that information can be obtained via the SorobanTransactionData.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -213,6 +217,47 @@ func getContractID(t *testing.T, sourceAccount string, salt [32]byte, networkPas
...
+
+	request := methods.SimulateTransactionRequest{Transaction: txB64}
+	var response methods.SimulateTransactionResponse
+	err = client.CallResult(context.Background(), "simulateTransaction", request, &response)
+	assert.NoError(t, err)
+	if !assert.Empty(t, response.Error) {
+		fmt.Println(response.Error)
+	}
+	var transactionData xdr.SorobanTransactionData
+	err = xdr.SafeUnmarshalBase64(response.TransactionData, &transactionData)
```
tsachiherman, line 240: add
```golang
assert.NoError(t, err)
```
to avoid `ineffectual assignment to err (ineffassign)`

cmd/soroban-rpc/lib/preflight/src/lib.rs
2opremio: I had already added `MinResourceFee` to the `SimulateTransactionResponse`, isn't that enough?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: Alright, I will remove them

cmd/soroban-rpc/lib/preflight/src/lib.rs
sreuland: yes, if `MinResourceFee` sums in the 'inclusion fee' suggested by @dmkozh also, I wasn't sure if that was the case as he was mentioning this 'inclusion fee' as something that should be added on top by 'they' which I understood to be remote clients after getting simulation response?

2opremio: `MinResourceFee` needs to be added on top of the classic Stellar fee (typically the minimum base fee of 100).

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: Actually, the soroban transaction data doesn't contain memory consumption and the instructions it contains are the resources it allows in for execution (which contain a leeway), so I think I will leave it.

sreuland: ok, will retain reference to `cost` on the soroban-client simulation response model.

cmd/soroban-rpc/lib/preflight/src/lib.rs
sreuland: ok, I'm trying to confirm this rule on handling fees to complete [js-soroban-client](https://github.com/stellar/js-soroban-client/pull/77) when it prepares a users tx with preflight results, @tsachiherman [you mentioned a different approach of taking max](https://github.com/stellar/js-soroban-client/pull/77#discussion_r1189248525) of classic fee(already set on the tx by user) or MinResourceFee. Which approach is correct for applying MinResourceFee on a user's tx, Max comparison or Summation? And this means there will be no other suggested inclusion fees provided by simulation response?

sreuland: received clarification, the rule is to use summation with existing tx classic fee:
>given that it's called MinResourceFee, it would imply that the user would need to pay a minimum of MinResourceFee + classic fees. so, if the user provided an amount that is anywhere less then (MinResourceFee + classic fees), then we should bump that up to this number.  ( since we know it won't work otherwise )

PR #640, opened 2023-05-09
cmd/soroban-rpc: Improve error message when starting before first checkpoint is published

tamirms:
### What

Improve error message when starting before first checkpoint is published.

### Why

soroban RPC cannot start captive core before the first checkpoint is published because captive core requires the history archive snapshot to be present. If someone tries to startup soroban RPC before the first checkpoint is published we emit the following log messages:

```
time="2023-05-08T23:43:13.994Z" level=info msg="Found an empty database, creating ledger-entry baseline from the most recent checkpoint (0). This can take up to 30 minutes, depending on the network" pid=1021
time="2023-05-08T23:43:13.995Z" level=fatal msg="could not run ingestion" error="error starting prepare range: opening subprocess: error calculating ledger and hash for stellar-core run: CaptiveCore is unable to start from ledger 1, start from ledger 2" pid=1021
```

These error messages are misleading because it seems like there is a bug in soroban rpc where we are starting from the incorrect ledger. In actuality, the problem is caused by the fact that the history archive snapshot is missing for the first checkpoint ledger.

### Known limitations

[N/A]

### DIFF
--- a/cmd/soroban-rpc/internal/ingest/service.go
+++ b/cmd/soroban-rpc/internal/ingest/service.go
@@ -3,6 +3,7 @@ package ingest
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jcx120:
Aside from the error message issue, there another issue that needs to be addressed with this:
> In actuality, the problem is caused by the fact that the history archive snapshot is missing for the first checkpoint ledger.

PR #643, opened 2023-05-17
soroban-cli: update to support fees

tsachiherman:
### What

Update soroban-cli to support fees.

### Why

Fees are kind-of important

### Known limitations

n/a


cmd/soroban-cli/src/rpc/transaction.rs
```diff
@@ -0,0 +1,357 @@
...
+    // operation in the txn, so we need to enforce that here. I (Paul) think that is a bug
+    // in soroban-rpc.simulateTransaction design, and we should fix it there.
+    // TODO: We should to better handling so non-soroban txns can be a passthrough here.
+    if tx.operations.len() != 1 {
+        return Err(Error::UnexpectedOperationCount {
+            count: tx.operations.len(),
+        });
+    }
+
+    // TODO: Should we keep this?
```
paulbellamy, line 27: The `log_events` piece here is slightly awkward (passing a log fn into this fn). We could instead return the events somehow and do the logging externally where it is needed (only invoke for now). Or we could leave this as-is for now...

tsachiherman: Given that it has no functional downside, I'd suggest leave it as is for this release and create a ticket to address this for next release.

cmd/soroban-cli/src/commands/contract/deploy.rs
```diff
@@ -220,31 +222,10 @@ fn build_create_contract_tx(
...
-                instructions: 0,
-                read_bytes: 0,
-                write_bytes: 0,
-                extended_meta_data_size_bytes: 0,
-            },
-            // TODO: what value to use here?
-            refundable_fee: 0,
-            ext: ExtensionPoint::V0,
-        }),
+        ext: TransactionExt::V0,
```
2opremio, line 225: woot? I am surprised this works at all

paulbellamy: We replace this with the real one returned from simulateTransaction before we submit the transaction. So this v0 bit doesn't actually matter.

paulbellamy: I had debated adding a comment about that, but it felt like 3 different places to add it, so...

cmd/soroban-cli/src/rpc/transaction.rs
```diff
@@ -0,0 +1,357 @@
...
+    raw: &Transaction,
+    simulation: &SimulateTransactionResponse,
+    log_events: Option<LogEvents>,
+) -> Result<Transaction, Error> {
+    let mut tx = raw.clone();
+
+    // Right now simulate.results is one-result-per-function, and assumes there is only one
+    // operation in the txn, so we need to enforce that here. I (Paul) think that is a bug
+    // in soroban-rpc.simulateTransaction design, and we should fix it there.
+    // TODO: We should to better handling so non-soroban txns can be a passthrough here.
```
2opremio, line 20: I am curious as to what's the usecase for multiple `InvokeHostFunction` operations in the same transaction (unless they can, maybe, be interspersed with operations of other types).

paulbellamy: One thing you can do is upload the wasm, create the contract, and invoke an initialization method in a single txn. You can also invoke functions on multiple contracts atomically (e.g. approve a token, do a deposit, disapprove the token).

2opremio (approved):

PR #644, opened 2023-05-18
feat!: convert bindings to subcommands and add a ts one

willemneal:
### What

Refactor the bindings command into subcommands and adds a TS project.

### Why

This way they can each have their own options.

### Known limitations

Requires: https://github.com/stellar/rs-soroban-sdk/pull/871

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ rev = "0c0cae6fa22b751e7fd95d7ce7556ac6d7b7010e"
 [workspace.dependencies.soroban-spec]
 version = "0.8.0"
 git = "https://github.com/ahalabs/rs-soroban-sdk"
-rev = "4d8dee5934ea1f075632e342ebdf0328fe475551"
+rev = "de9ac823a0a0d3d6ea9c1f5f98024fd6afb4b779"
 # path = "../rs-soroban-sdk/soroban-spec"
 
 [workspace.dependencies.soroban-sdk]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,8 @@ rev = "0c0cae6fa22b751e7fd95d7ce7556ac6d7b7010e"
 [workspace.dependencies.soroban-spec]
 version = "0.8.0"
 git = "https://github.com/ahalabs/rs-soroban-sdk"
-rev = "379d818a134239a5dda057a159d3c342509b5b5c"
-# path = "../rs-soroban-sdk/soroban-spec"
+rev = "b983d06ba03f4066886735f7840c9c38d2d71ca9"
+# path = "../sdk/soroban-spec"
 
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -17,8 +17,8 @@ rev = "0c0cae6fa22b751e7fd95d7ce7556ac6d7b7010e"
 
 [workspace.dependencies.soroban-spec]
 version = "0.8.0"
-git = "https://github.com/stellar/rs-soroban-sdk"
-rev = "090f5c88e8f8b67f4d02010665aedfea6ee43db3"
+git = "https://github.com/ahalabs/rs-soroban-sdk"
+rev = "14ce186845311c75ee9daab023837bbfd5f382a5"
```
paulbellamy, line 21: Does this fork need merging in? Or are we waiting for a new release here? or what?

cmd/soroban-cli/src/commands/contract/bindings/ts.rs
paulbellamy, line 1: I'm assuming this is why we are on a fork of soroban_spec here. Maybe we should leave this out until it is ready and spec is merged. We could just do the json+rust in this PR, then a separate PR once the ts bindings are ready (I'm assuming they are still under development?)

cmd/soroban-cli/src/commands/contract/bindings/ts.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// where to place generated project
+    #[arg(long)]
+    root_dir: PathBuf,
+
+    #[arg(long)]
+    contract_name: String,
+
+    #[arg(long, alias = "id")]
+    contract_id: String,
```
chadoh, line 22: Can we also allow specifying `rpc-url` and `network-passphrase` in the no-network version? These are networked params, so I'm not sure it does. But it also feels weird to [hardcode them](https://github.com/stellar/rs-soroban-sdk/pull/871/files#diff-3cb416aff49f093298d5ed678a7d07d70b591905101ace4c5af2a80c0da2bd07R17-R29) in the generated template, even if they're overridable with environment variables.

Maybe this doesn't matter, since we'll have the networked version soon enough.

Cargo.toml
chadoh: Yes, this is waiting on this PR https://github.com/stellar/rs-soroban-sdk/pull/871

cmd/soroban-cli/src/commands/contract/bindings/ts.rs
chadoh: You can mark this comment as resolved

paulbellamy (commented):
Added a naming suggestion, but overall I think this is ready to go.
I'd like to get the sdk stuff merged/released first, then we can use that and merge this.

PR #652, opened 2023-05-23
soroban-rpc: Fix bugs in config.toml and cli flag parsing

paulbellamy:
### What

Improve parsing of cli and toml flags

### Why

https://github.com/stellar/soroban-tools/pull/609 introduced a couple bugs in setValue. It didn't handle all the types that the toml parsing fed into it, so it would crash.

This touched some of the OptType stuff, so I took the opportunity to also clean that up and auto-infer it where possible.

### Known limitations

- More complex types are still kind of janky.
- Hopefully this is all the types we need. If not, at least it's easy to add new tests for them.
- My kingdom for rust's `match` block :( Would make this so much cleaner.
- I built a local quickstart image with this and quickstart master branch, and it seems to launch fine.
- Could definitely use a few more tests for 100% coverage


cmd/soroban-rpc/internal/config/config.go
```diff
@@ -52,22 +52,29 @@ func (cfg *Config) Init(cmd *cobra.Command) error {
...
 		// Merge in the config file flags
 		if err := cfg.loadConfigPath(); err != nil {
 			return err
 		}
+
+		// Load from cli flags and environment variables again, to overwrite what we
+		// got from the config file
+		if err := cfg.loadFlags(); err != nil {
+			return err
+		}
```
leighmcculloch, line 77: Would it be equivalent for us to `loadFlags` once after the `if cfg.ConfigPath != ""` block instead of twice, once before and once after during?

paulbellamy: We could. We specifically need to load the configPath first, in order to load the config file.

paulbellamy:
Seems like `viper.IsSet` just always returns true... :/

paulbellamy:
So it seems like reintroducting the second `loadFlags` here has reintroduced @sreuland's bug from [here](https://github.com/stellar/soroban-tools/pull/609#issuecomment-1555154370). Because `viper.IsSet` always returns true, it just overwrites all the config file options. :(

paulbellamy:
Ok, I fixed this by removing viper and `go/support/config`, and just doing it ourselves here.

Viper is fine so long as you want to do exactly what it does. Otherwise it gets very hacky.

`go/support/config` is also fine, but would need changes to do what we want, and I don't want to jeopardise the other code that uses it as I don't feel that I fully understand the subtle implementation points that might be used in other codebases.

cmd/soroban-rpc/internal/config/config.go
```diff
@@ -43,31 +40,41 @@ type Config struct {
...
+func (cfg *Config) SetValues(lookupEnv func(string) (string, bool)) error {
+	// We start with the defaults
 	if err := cfg.loadDefaults(); err != nil {
 		return err
 	}
 
-	cfg.Bind()
+	// Then we load from the environment variables and cli flags, to try to find
+	// the config file path
+	if err := cfg.loadEnv(lookupEnv); err != nil {
```
sreuland, line 56: what if add ability to load flag from env by a key name such as `config-path` key, then the overlay sequence simplifies down to:

1. load flags
2. if config-path flag present, load config-path
3. load env

rather than doing a second global apply of env flags in the config-path conditional?

paulbellamy: Yeah, I considered that. Would need to make the config path option extractable or something. Couldn't find a way to do it cleanly with DRY, etc... There's no real downside to loading everything twice since loading it is non-destructive.

PR #658, opened 2023-05-25
fix(CLI): contract id parsing

willemneal:
### What

Allow parsing Contract Addresses

### Why

Part of the move from Bytes<32> to Address anywhere a contract id is used.

### Known limitations

[TODO or N/A]


paulbellamy (approved):
We should remember to remove the hex-parsing once it is converted everywhere.

paulbellamy:
Part of https://github.com/stellar/soroban-tools/issues/403

Issue #572, opened 2023-04-10
Add automatic dependency sanity checker


2opremio:
We have at least 4 dependencies with their own XDR version which can get out of sync.
The XDR dependencies in the go monorepo (needed by soroban-rpc)
[rs-soroban-env](https://github.com/stellar/rs-soroban-env) (used by both soroban-rpc and the cli)
[rs-soroban-sdk](https://github.com/stellar/rs-soroban-sdk) (used by the cli)
Core

We need a way to safely ensure that everything is in sync with respect to XDR definitions. There are too many dependencies to do this manually and we have already fallen into problems with this several times in the past. Last problem required several hours of debugging and several engineers (we suspected a bug in Core, see https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1679397527481429?thread_ts=1679388817.352239&cid=C030Z9EHVQE ) 

Additionally, [rs-soroban-env](https://github.com/stellar/rs-soroban-env)  and [rs-soroban-sdk](https://github.com/stellar/rs-soroban-sdk) are intertwined  (you need to use specific version of both for things to work properly) and sometimes out of sync (e.g. their XDR dependencies are not bumped atomically)

We could add a script that makes sure the dependency versions are sane. We should be able to run it both locally and in CI.

PR #659, opened 2023-05-26
Add dependency checker script

2opremio:
### What

Add dependency checker script and make it run in CI ~(this last part is still TODO)~.

### Why

Close https://github.com/stellar/soroban-tools/issues/572

### Known limitations

Parsing of Core/Golang versions doesn’t work when they don’t contain a commit hash.

I left TODOs for them and we can fix it later if needed.

### DIFF
--- a/scripts/check-dependencies.bash
+++ b/scripts/check-dependencies.bash
@@ -26,7 +26,7 @@ RS_STELLAR_XDR_REVISION=""
 # revision of https://github.com/stellar/stellar-xdr/ used by the Rust code
 STELLAR_XDR_REVISION_FROM_RUST=""
 
-if CARGO_OUTPUT=$(cargo tree --depth 1 -p stellar-xdr 2>&1); then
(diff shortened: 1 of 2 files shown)


scripts/check-dependencies.bash
```diff
@@ -0,0 +1,88 @@
...
+fi
+
+
+# revision of the https://github.com/stellar/rs-stellar-xdr library used by the Rust code
+RS_STELLAR_XDR_REVISION=""
+
+# revision of https://github.com/stellar/stellar-xdr/ used by the Rust code
+STELLAR_XDR_REVISION_FROM_RUST=""
+
+if CARGO_OUTPUT=$(cargo tree --depth 1 -p stellar-xdr 2>&1); then
```
paulbellamy, line 29:
```suggestion
if CARGO_OUTPUT=$(cargo tree --depth 0 -p stellar-xdr 2>&1); then
```
`--depth=0` might make the parsing below a bit simpler

paulbellamy (approved):
one suggestion

PR #667, opened 2023-05-29
soroban-rpc: Obtain fee configuration from the ledger

### LINKED ISSUE #665: soroban-rpc
### What ? 
Read the config settings ledger keys when calculating fees

2opremio:
### What

Obtain the configuration for the fee estimation of `simulateTransaction` from the ledger instead of hardcoding it.

### Why

Close #665 

### Known limitations

Obtaining 4 ledger entries for each `simulateTransaction` call will reduce performance. We need a benchmark to measure the impact.

### DIFF
--- a/cmd/soroban-rpc/lib/preflight/src/fees.rs
+++ b/cmd/soroban-rpc/lib/preflight/src/fees.rs
@@ -164,8 +164,11 @@ fn get_configuration_setting(
 fn get_fee_configuration(
     ledger_storage: &ledger_storage::LedgerStorage,
 ) -> Result<FeeConfiguration, Box<dyn error::Error>> {
-    // TODO: consider caching these entries. In theory, they can change at any given ledger
-    //       (making caching ineffective) but we can maybe relax that requirement?
+    // TODO: to improve the performance of this function (which is invoked every single preflight call) we can
+    //       1. modify ledger_storage.get() so that it can gather multiple entries at once
+    //       2. implement a write-through cache for the configuration ledger entries (i.e. the cache is written to at the
+    //          same time as the DB, ensuring they are always in memory).
+    //
 
     let ConfigSettingEntry::ComputeV0(compute) = get_configuration_setting(ledger_storage, ConfigSettingId::ComputeV0)? else {
             return Err(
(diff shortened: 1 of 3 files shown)


cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -140,23 +141,74 @@ fn calculate_soroban_resources(
...
-    //       Specifically, we need to derive it from ConfigSettingContractComputeV0 which can
-    //       be retrieved using the ConfigSetting/CONFIG_SETTING_CONTRACT_COMPUTE_V0.
+fn get_configuration_setting(
+    ledger_storage: &ledger_storage::LedgerStorage,
+    setting_id: ConfigSettingId,
+) -> Result<ConfigSettingEntry, Box<dyn error::Error>> {
+    let key = LedgerKey::ConfigSetting(LedgerKeyConfigSetting {
+        config_setting_id: setting_id,
+    });
+    match ledger_storage.get(&key)? {
```
paulbellamy, line 151: We could refactor ledger_storage so `get` can fetch multiple keys at the same time. Might reduce the go<->rust round-trips. But we should probably only do that if it proves to be an issue.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -140,23 +141,74 @@ fn calculate_soroban_resources(
...
+        )
+        .into()),
+    }
+}
+
+fn get_fee_configuration(
+    ledger_storage: &ledger_storage::LedgerStorage,
+) -> Result<FeeConfiguration, Box<dyn error::Error>> {
+    // TODO: consider caching these entries. In theory, they can change at any given ledger
+    //       (making caching ineffective) but we can maybe relax that requirement?
```
paulbellamy, line 168: If they don't change frequently, but we need immediate results when they *do* change, a write-through cache might be the right choice here.

cmd/soroban-rpc/lib/preflight/src/fees.rs
paulbellamy: Do we have a good way to benchmark this given the rust/go boundary stuff? would a go benchmark be reliable?

paulbellamy (approved):

cmd/soroban-rpc/lib/preflight/src/fees.rs
2opremio: That's a really good idea

2opremio: I will add a comment about it

PR #663, opened 2023-05-26
fix: return contract Addresses and allow passing hex when used as an contract arg

willemneal:
### What

fixes #661 
fixes #662

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]

PR #668, opened 2023-05-30
fix: return contract Addresses and allow passing hex when used as an contract arg

### LINKED ISSUE #403: soroban-cli: Replace all hex-encoded contract ids with strkeys

### LINKED ISSUE #661: Contract invocation arguments should accept `BytesN<32>` for contract hashes
They used to, they don't anymore, despite what the help text says to the contrary.

```bash
% soroban contract invoke --id 17bd...232f --source Default --network Futurenet -- new_liqpool -h

Usage: new_liqpool [OPTIONS]

Options:
      --salt <32_hex_bytes>  Example:
                               --salt efefefefefefefefefefefefefefefefefefefefefefefefefefefefefefefef
      --token_b <Address>    Can be public key (G13..), a contract hash (6c45307) or an identity (alice), 
                             Example:
                               --token_b GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4
      --token_a <Address>    Can be public key (G13..), a contract hash (6c45307) or an identity (alice), 
                             Example:
                               --token_a GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4
  -h, --help                 Print help (see more with '--help')
% soroban contract invoke --id 17bd...232f --source Default --network Futurenet -- new_liqpool --salt deadbeef --token_a 17c0...357e8 --token_b dab0...772b
```
```bash
error: parsing argument token_a: value is not parseable to Some(
    Address,
)
```

### LINKED ISSUE #662: Contract deployments should spit out the `StrKey` `Address` vs the `BytesN<32>`
We've switched to using `StrKey` `Address` vs the `BytesN<32>` so we should stop spitting out less usable `BytesN<32>` hashes.

The two places I'm aware of would be `contract deploy` and `lab token wrap`. There may be others?

Dependent on:
https://github.com/stellar/soroban-tools/issues/660

Related to:
https://github.com/stellar/soroban-tools/issues/661

paulbellamy:
### What

fixes #661
fixes #662
fixes #403
replaces #663

### Why

https://github.com/stellar/soroban-tools/pull/663 missed a few other commands were we take contract IDs as args. Also, we cannot generically accept hex for `strval::scaddress_from_json`, because with just 32-byte hex, we cannot tell if it is a user address or a contract address.

Incidentally, making this more consistent also fixed a bug in `soroban events --id 1`, where the given contract id was not padded out to the full 32 bytes. Now, these are all equivalent:
```sh
soroban events --id 1
soroban events --id 0000000000000000000000000000000000000000000000000000000000000001
soroban events --id CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM
```
Those ids all also work for `invoke`, `deploy`, and `read`

### Known limitations

- We should also update soroban-rpc to take a contract strkey in it's params, wherever it currently accepts a contract hex (e.g. getEvents). Then we can update this to pass that in.

### DIFF
--- a/cmd/crates/soroban-test/tests/it/invoke_sandbox.rs
+++ b/cmd/crates/soroban-test/tests/it/invoke_sandbox.rs
@@ -1,3 +1,5 @@
+use std::println;
+
 use soroban_cli::commands::{config::identity, contract};
 use soroban_test::TestEnv;
 
@@ -55,6 +57,7 @@ fn invoke_hello_world_with_deploy_first() {
         .success();
     let stdout = String::from_utf8(res.get_output().stdout.clone()).unwrap();
     let id = stdout.trim_end();
+    println!("{id}");
     sandbox
         .new_assert_cmd("contract")
         .arg("invoke")
--- a/cmd/crates/soroban-test/tests/it/invoke_sandbox.rs
+++ b/cmd/crates/soroban-test/tests/it/invoke_sandbox.rs
@@ -29,7 +29,7 @@ fn install_wasm_then_deploy_contract() {
         .arg("--id=1")
         .assert()
         .success()
-        .stdout("0000000000000000000000000000000000000000000000000000000000000001\n");
+        .stdout("CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM\n");
 }
 
 #[test]
@@ -42,7 +42,7 @@ fn deploy_contract_with_wasm_file() {
         .arg("--id=1")
         .assert()
         .success()
-        .stdout("0000000000000000000000000000000000000000000000000000000000000001\n");
+        .stdout("CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM\n");
 }
 
 #[test]
--- a/cmd/soroban-cli/src/utils.rs
+++ b/cmd/soroban-cli/src/utils.rs
@@ -221,21 +221,6 @@ pub fn get_contract_spec_from_storage(
     }
 }
 
-/// # Errors
-///
-/// Might return an error
-pub fn vec_to_hash(res: &ScVal) -> Result<String, XdrError> {
-    if let ScVal::Bytes(res_hash) = &res {
-        let mut hash_bytes: [u8; 32] = [0; 32];
-        for (i, b) in res_hash.iter().enumerate() {
-            hash_bytes[i] = *b;
-        }
-        Ok(hex::encode(hash_bytes))
-    } else {
-        Err(XdrError::Invalid)
-    }
-}
-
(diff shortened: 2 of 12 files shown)


cmd/soroban-cli/src/utils.rs
```diff
@@ -156,10 +159,17 @@ pub fn sign_transaction(
...
 /// Might return an error
-pub fn id_from_str<const N: usize>(contract_id: &String) -> Result<[u8; N], FromHexError> {
-    padded_hex_from_str(contract_id, N)?
-        .try_into()
-        .map_err(|_| FromHexError::InvalidStringLength)
+pub fn contract_id_from_str(contract_id: &str) -> Result<[u8; 32], stellar_strkey::DecodeError> {
+    stellar_strkey::Contract::from_string(contract_id)
+        .map(|strkey| strkey.0)
+        .or_else(|_| {
+            // strkey failed, try to parse it as a hex string, for backwards compatibility.
```
leighmcculloch, line 166: 👏🏻 Very nice keeping the backwards compatibility.

leighmcculloch (approved):

leighmcculloch (approved):

cmd/soroban-cli/src/utils.rs
Shaptic: Should we make a note to drop the deprecated way of doing things prior to v1.0?

cmd/crates/soroban-test/tests/it/invoke_sandbox.rs
```diff
@@ -27,7 +29,7 @@ fn install_wasm_then_deploy_contract() {
         .arg("--id=1")
         .assert()
         .success()
-        .stdout("0000000000000000000000000000000000000000000000000000000000000001\n");
+        .stdout("CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM\n");
```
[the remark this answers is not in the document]
paulbellamy: 🐦 CAAAA CAAAA CAAA 🐦

Issue #436, opened 2023-02-17
Cannot use Bytes type propperly


aristidesstaffieri:
### What version are you using?
0.6.0

### What did you do?
I have deployed an instance of the token interface on futurenet, and I'm trying to call initialize with [this fn sig](https://github.com/stellar/soroban-examples/blob/main/token/src/contract.rs#L14). The last two args are Bytes that I believe are supposed to be the hex string representation of the value. I am trying the following and am wondering why it wouldn't be able to parse the --symbol arg but had no problem with the --name arg, both are hex strings of the value that I want to pass for name and symbol.
```
$ soroban contract invoke \
    --id 6885ed6b904a6cead909c99cdc482ffa4a74c181803f678cacbfbe831e9efac6 \
    --secret-key S... \
    --rpc-url http://localhost:8000/soroban/rpc \
    --network-passphrase 'Test SDF Future Network ; October 2022' \
    --fn initialize \
    -- \
    --admin GCGORBD5DB4JDIKVIA536CJE3EWMWZ6KBUBWZWRQM7Y3NHFRCLOKYVAL \
    --decimal 7 \
    --name 467265696768746572205465737420546f6b656e \
    --symbol 465454
error: parsing argument symbol: invalid type: integer `465454`, expected string or map
```

### What did you expect to see?
A successful transaction for initializing this contract

### What did you see instead?
cli cannot parse the hex string provided for the bytes arg called symbol, it seems to be fine with the hex string for bytes for the arg called name.

willemneal:
Hmm interesting.  Does it still happen if you change symbol to `symbol_b`?  We do have a test for bytes. But I'll add one with an arg named `symbol`.

aristidesstaffieri:
> Hmm interesting. Does it still happen if you change symbol to `symbol_b`? We do have a test for bytes. But I'll add one with an arg named `symbol`.

Not sure I understand, do you mean change the arg name to `symbol_b`, just to clarify this is the symbol used to identify the token in this instance, not a Symbol type but a Bytes type.

bjorn:
Just ran into this on RPCiege day 3, which required sending in a symbol of which the hash included a certain pattern. I was unable to pass any symbols on the CLI that consisted of just numbers, likely because they were being parsed as integers and then failed a type check later on when invoking the contract's function:
```
[...]

PR #671, opened 2023-05-31
Fix a bug parsing bytes args where the hex are all numbers

paulbellamy:
### What

Fix a bug parsing "bytes" type args, where the value is all numbers.

### Why

Parsing a bytes array of the text "DT" would fail (bytes 0x45, 0x54). The CLI json arg parser would count it as a json number, not a string, so parsing would fail.

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-cli/src/strval.rs
+++ b/cmd/soroban-cli/src/strval.rs
@@ -798,7 +798,7 @@ pub fn from_json_primitives(v: &Value, t: &ScType) -> Result<ScVal, Error> {
         // Bytes parsing
         (bytes @ ScType::BytesN(_), Value::Number(n)) => from_json_primitives(
             &Value::String(format!("{n}")),
-            &ScType::BytesN(bytes.clone()),
+            &bytes,
         )?,
         (ScType::BytesN(bytes), Value::String(s)) => ScVal::Bytes(ScBytes({
             if let Ok(key) = stellar_strkey::ed25519::PublicKey::from_string(s) {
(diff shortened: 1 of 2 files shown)


Shaptic (approved):
nice :clap: 

this could have been worked around by quoting the parameter, correct? e.g. `'"4545"'` (single quotes for the shell, double for the arg itself)

paulbellamy:
Yes, that would force the CLI to treat it as a string.

cmd/soroban-cli/src/strval.rs
```diff
@@ -796,6 +796,10 @@ pub fn from_json_primitives(v: &Value, t: &ScType) -> Result<ScVal, Error> {
         (ScType::Address, Value::String(s)) => sc_address_from_json(s, t)?,
 
         // Bytes parsing
+        (ScType::BytesN(bytes), Value::Number(n)) => from_json_primitives(
```
willemneal, line 799:
```suggestion
        (bytes @ ScType::BytesN(_), Value::Number(n)) => from_json_primitives(
```

cmd/soroban-cli/src/strval.rs
```diff
@@ -796,6 +796,10 @@ pub fn from_json_primitives(v: &Value, t: &ScType) -> Result<ScVal, Error> {
...
+            &Value::String(format!("{n}")),
+            &ScType::BytesN(bytes.clone()),
```
willemneal, line 801:
```suggestion
            &bytes,
```

bjorn:
Does this fix #436, and would it also address the same problem for `Symbol` arguments that I reported at https://github.com/stellar/soroban-tools/issues/436#issuecomment-1570238192?

PR #672, opened 2023-05-31
feat!(CLI): make `contract optimize` optional

willemneal:
### What

Remove wasm-opt by default

### Why

Some users had issues with building wasm-opt and this also makes installing it from source faster.

### Known limitations

[TODO or N/A]


cmd/soroban-cli/src/commands/contract/optimize.rs
```diff
@@ -18,11 +19,18 @@ pub struct Cmd {
...
     Wasm(#[from] wasm::Error),
+    #[cfg(feature = "opt")]
     #[error("optimization error: {0}")]
     OptimizationError(OptimizationError),
 }
 
 impl Cmd {
+    #[cfg(not(feature = "opt"))]
+    pub fn run(&self) -> Result<(), Error> {
+        todo!("Must install with \"opt\" feature");
```
leighmcculloch, line 30: Can we return an error here? The `todo!` macro doesn't feel very appropriate, because there's not really anything for us to implement.

cmd/soroban-cli/Cargo.toml
```diff
@@ -26,6 +26,10 @@ name = "soroban_cli"
 path = "src/lib.rs"
 doctest = false
 
+[features]
+default = []
+opt = ["wasm-opt"]
```
leighmcculloch, line 31: This will cause the binary releases that we package and release on GH to not have the opt feature. Could we either make `opt` the default, which allows folks to opt-out of defaults, and change the github build for release to include this option?

leighmcculloch, line 31: Adding features means we need to start testing feature combinations in this repository. If we make `opt` the default, I think we could get away without testing these two features though. If we don't make it the default, or if we advertise building without the feature, we should test it.

leighmcculloch (commented):
I think this makes sense. I've seen folks struggle to install it from source too because of this.

There's some logistical stuff we need to do to support this. See inline.

willemneal:
@leighmcculloch Ideally this would be a plugin. This prevents the feature issues altogether.

However, currently we can only do `soroban-x` for plugins, but we should be able to add it to any subcommand e.g. `soroban-contract-optimize`.  Then we have a stub that prompts the user to install the plugin.

The correctness of the rest of the CLI is independent from the optimize command, which is well tested underneath. Thanks @brson!

For now can we feature gate it and then return to pull it out before the next release so we can publish the plugin.

cmd/soroban-cli/Cargo.toml
willemneal: I definitely think it should be default off because we are trying to minimize the average install. But I do agree that we should have it in the released binaries.

leighmcculloch:
> Ideally this would be a plugin. This prevents the feature issues altogether.
> 
> However, currently we can only do `soroban-x` for plugins, but we should be able to add it to any subcommand e.g. `soroban-contract-optimize`. Then we have a stub that prompts the user to install the plugin.

I don't think we need too be too precious on location. We can make it a plugin and put that plugin at the top level. We can move it later if we think plugins should be multi level, but multi level plugins may represent a small security concern, so we should consider them carefully.

willemneal:
@leighmcculloch fair enough. Is this good enough for now and we move to a top level plugin later?

leighmcculloch:
If this is a short term gap, 👍🏻. Making this a plugin seems like the right move since the CLI has that capability. It's the perfect use case. Rust features are painful on a few fronts, and we've started trying to eliminate features from other repos. I think we'd want to eliminate this feature as soon as possible.

cmd/soroban-cli/Cargo.toml
leighmcculloch, line 31: This change as it is, is actually adding two new features to the CLI: `opt` and `wasm-opt`. We can make it one new feature by referencing the `wasm-opt` create as a `dep:` feature, which unintuitively tells cargo we don't want the implicitly `wasm-opt` feature to exist.
```suggestion
opt = ["dep:wasm-opt"]
```

leighmcculloch (approved):
I pushed a couple tweaks updating docs, etc. Looks good!

leighmcculloch:
To be followed up by:
- #676

willemneal:
Looks like preflight is failing to link on windows:
```
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/soroban-tools/soroban-tools/cmd/soroban-rpc/internal/preflight/../../../../target/x86_64-pc-windows-gnu/release-with-panic-unwind/\libpreflight.a(std-ca5208825e97b4ba.std.687851ba-cgu.0.rcgu.o): in function `std::sys::windows::fs::open_link_no_reparse':
/rustc/90c541806f23a127002de5b4038be731ba1458ca/library\std\src\sys\windows/fs.rs:800: undefined reference to `NtCreateFile'
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: /rustc/90c541806f23a127002de5b4038be731ba1458ca/library\std\src\sys\windows/fs.rs:829: undefined reference to `RtlNtStatusToDosError'
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/soroban-tools/soroban-tools/cmd/soroban-rpc/internal/preflight/../../../../target/x86_64-pc-windows-gnu/release-with-panic-unwind/\libpreflight.a(std-ca5208825e97b4ba.std.687851ba-cgu.0.rcgu.o): in function `std::sys::windows::handle::Handle::synchronous_read':
/rustc/90c541806f23a127002de5b4038be731ba1458ca/library\std\src\sys\windows/handle.rs:241: undefined reference to `NtReadFile'
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: /rustc/90c541806f23a127002de5b4038be731ba1458ca/library\std\src\sys\windows/handle.rs:272: undefined reference to `RtlNtStatusToDosError'
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/soroban-tools/soroban-tools/cmd/soroban-rpc/internal/preflight/../../../../target/x86_64-pc-windows-gnu/release-with-panic-unwind/\libpreflight.a(std-ca5208825e97b4ba.std.687851ba-cgu.0.rcgu.o): in function `std::sys::windows::handle::Handle::synchronous_write':
/rustc/90c541806f23a127002de5b4038be731ba1458ca/library\std\src\sys\windows/handle.rs:290: undefined reference to `NtWriteFile'
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: /rustc/90c541806f23a127002de5b4038be731ba1458ca/library\std\src\sys\windows/handle.rs:318: undefined reference to `RtlNtStatusToDosError'
collect2.exe: error: ld returned 1 exit status
```

paulbellamy:
Windows build fail seems due to github updating the windows image.

PR #673, opened 2023-06-01
Fixed HTTP Request Smuggling in golang-golang-x-net

wica-sufatmawati:
A request smuggling attack is possible when using MaxBytesHandler. When using MaxBytesHandler, the body of an HTTP request is not fully consumed. When the server attempts to read HTTP2 frames from the connection, it will instead be reading the body of the HTTP request, which could be attacker-manipulated to represent arbitrary HTTP2 requests.

CWE-444
CVE-2022-41721
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,9 +67,9 @@ require (
 	github.com/xanzy/ssh-agent v0.3.0 // indirect
 	golang.org/x/crypto v0.1.0 // indirect
 	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e // indirect
-	golang.org/x/net v0.1.0 // indirect
+	golang.org/x/net v0.7.0 // indirect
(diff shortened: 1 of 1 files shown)


tsachiherman:
@wica-sufatmawati I'm pretty sure that we won't be susceptible to this attack since we're not using the `MaxBytesHandler`.
It is a good idea to update our dependencies, and in particular logrus and github.com/rs/cors; but unless these are being updated with the latests dependencies on their own, changing it in the `go.mod` file won't work.

tsachiherman:
I'm closing this PR as it won't work as-is. Please re-open with a mergable fix if you find this to be a meaningful issue.

noyon3rh:
" onclick=alert(1)//<button ' onclick=alert(1)//> */ alert(1)//

PR #675, opened 2023-06-01
feat(CLI): contract fetch

willemneal:
### What

Downloads the bytes of a contract. Print to stdout or write to a file

### Why

Other commands already do this so it would best to implement it once and it's handy for build tools.

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/fetch.rs
+++ b/cmd/soroban-cli/src/commands/contract/fetch.rs
@@ -6,7 +6,6 @@ use std::str::FromStr;
 use std::{fmt::Debug, fs, io, rc::Rc};
 
 use clap::{arg, command, Parser};
-use hex::FromHexError;
 use soroban_env_host::{
     budget::Budget,
     storage::Storage,
@@ -18,6 +17,7 @@ use soroban_env_host::{
 
 use soroban_ledger_snapshot::LedgerSnapshot;
 use soroban_spec::read::FromWasmError;
+use stellar_strkey::DecodeError;
 
 use super::super::config::{self, locator};
 use crate::commands::config::ledger_file;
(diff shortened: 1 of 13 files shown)


cmd/soroban-cli/src/commands/contract/fetch.rs
```diff
@@ -0,0 +1,235 @@
...
+        Ok(get_contract_wasm_from_storage(&mut storage, contract_id)?)
+    }
+
+    fn contract_id(&self) -> Result<[u8; 32], Error> {
+        utils::contract_id_from_str(&self.contract_id)
+            .map_err(|e| Error::CannotParseContractId(self.contract_id.clone(), e))
+    }
+}
+
+async fn get_remote_wasm(client: &Client, contract_id: &[u8; 32]) -> Result<Vec<u8>, Error> {
```
paulbellamy, line 161: Can we make this a method on the `Client`? Then we could re-use it for `invoke` subcommand where we do: `get_remote_contract_spec_entries` as well.

cmd/soroban-cli/src/commands/contract/fetch.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+            scval => return Err(Error::UnexpectedContractCodeDataType(scval)),
+        },
+    )
+}
+
+pub fn get_contract_wasm_from_storage(
+    storage: &mut Storage,
+    contract_id: [u8; 32],
+) -> Result<Vec<u8>, FromWasmError> {
```
paulbellamy, line 197: I'm assuming this (and `get_remote_wasm` above) are public so they can be re-used by other subcommands (like `inspect`)? If so, we should put them somewhere else. Having subcommands depending on pieces from other subcommands is going to turn this into a mess quickly.

It would be nice if we had a way to put this method onto the `Storage`, but barring that it could go into the `utils` package, where we already have a `get_contract_spec_from_storage` function.

willemneal: I'll move it to util.rs for now, but I agree that the presence of a util suggests we should move the needed functions upstream.

paulbellamy (approved):

PR #690, opened 2023-06-09
preflight: Resolve compilation issue rust 1.70 on windows

tsachiherman:
### What

Fix compilation issue with soroban-rpc & updated rust compiler.

### Why

We previously pinned the rust compiler to 1.69, since the transition to 1.70 included an upgrade to llvm 16.
That upgrade broke binary compatibility between the preflight library and the soroban-rpc.

To overcome these limitations, this PR switches to use the msys2 version of mingw's gcc compiler. This is the same compiler that the gnu implementation of rust is using, and therefore produces the same library symbols.

With this change in place, we can continue upgrading the rust compiler version as needed.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/soroban-rpc.yml
+++ b/.github/workflows/soroban-rpc.yml
@@ -68,6 +68,7 @@ jobs:
           cd cmd/soroban-rpc/lib/preflight
           rustup default stable-${{ matrix.rust_target }}
           cd ../../../..
+          echo "CC=c:\msys2\mingw64\bin\gcc.exe" >> $GITHUB_ENV
 
       # Use cross-compiler for linux aarch64
       - if: matrix.rust_target == 'aarch64-unknown-linux-gnu'
--- a/.github/workflows/soroban-rpc.yml
+++ b/.github/workflows/soroban-rpc.yml
@@ -84,7 +84,7 @@ jobs:
 
       - name: Build Soroban RPC reproducible build
         run: |
-          CGO_ENABLED=1 GOARCH=${{ matrix.go_arch }} go build -trimpath -buildvcs=false ./cmd/soroban-rpc
+          CGO_ENABLED=1 GOARCH=${{ matrix.go_arch }} go build -x -trimpath -buildvcs=false ./cmd/soroban-rpc
           ls -lh soroban-rpc
           file soroban-rpc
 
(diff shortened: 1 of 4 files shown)


.github/workflows/soroban-rpc.yml
```diff
@@ -61,21 +61,16 @@ jobs:
...
       - if: matrix.os == 'windows-latest'
-        name: Configure preflight library to use explicit rust version
+        name: Install the same mingw gcc compiler used by rust
         run: |
-          cd cmd/soroban-rpc/lib/preflight
-          rustup override set $PREFLIGHT_RUST_VERSION-${{ matrix.rust_target }}
-          cd ../../../..
+          C:\\msys64\\usr\\bin\\pacman.exe -S mingw-w64-x86_64-gcc --noconfirm
+          echo "CC=C:\\msys64\\mingw64\\bin\\gcc.exe" >> $GITHUB_ENV
+          echo "C:\\msys64\\mingw64\\bin" >> $GITHUB_PATH
```
Shaptic, line 73: would be cleaner to just use the envvar you set :)

```suggestion
          echo "C:\\msys64\\mingw64\\bin" >> $GITHUB_PATH
          $GITHUB_PATH\\pacman.exe -S mingw-w64-x86_64-gcc --noconfirm
          echo "CC=$GITHUB_PATH\\gcc.exe" >> $GITHUB_ENV
```

but just a nitpick

Shaptic: (also I'm almost positive Windows allows `/` over `\\` but again, just a nit)

Shaptic (approved):

cmd/soroban-rpc/lib/preflight/Cargo.toml
```diff
@@ -2,7 +2,6 @@
 name = "preflight"
 version = "0.8.0"
 publish = false
-rust-version = "1.69"
```
leighmcculloch, line 5: Fyi this `rust-version` field is a minimum version, not a pinned version. The preflight crate will still be compiled with 1.70 if you have 1.70 installed.

tsachiherman: @leighmcculloch In this PR I have disabled the mechanism that treated this as pinned version. Yes, I'm aware that from rust perspective, this is a min compiler version.

PR #693, opened 2023-06-12
Add lab token id command

vinamogit:
### What

Add a new subcommand to "lab token" that displays the id and key of the wrapped asset

```
soroban lab token id --asset native --network futurenet
CDMT6XD3WDV4JKOI64T4LTV4JZARSTJYEV7B2DMRANLLIO74KKEBHYNJ
```

### Why

It is not obvious to derive the contract id of a wrapped asset.
Using the wrap command will return an error when the token is already wrapped, displaying the key at this moment does not seem the best option since your intention might not be to wrap it.

### Known limitations

- ~~The error handling should be rewritten to push the common errors used in the id command instead of using the lab::token::wrap::Error.~~
- ~~The common functions parse_asset and get_contract_id would be better placed in the module.~~

~~For now I failed to do that.~~


cmd/soroban-cli/src/commands/lab/token/wrap.rs
```diff
@@ -154,7 +154,7 @@ pub fn vec_to_hash(res: &ScVal) -> Result<Hash, XdrError> {
     }
 }
 
-fn get_contract_id(asset: &Asset, network_passphrase: &str) -> Result<Hash, Error> {
+pub fn get_contract_id(asset: &Asset, network_passphrase: &str) -> Result<Hash, Error> {
```
willemneal, line 157: Could you move this function to the utils file?

cmd/soroban-cli/src/commands/lab/token/wrap.rs
```diff
@@ -222,7 +222,7 @@ fn build_wrap_token_tx(
     })
 }
 
-fn parse_asset(str: &str) -> Result<Asset, Error> {
+pub fn parse_asset(str: &str) -> Result<Asset, Error> {
```
willemneal, line 225: Same here

cmd/soroban-cli/src/commands/lab/token/id.rs
```diff
@@ -0,0 +1,28 @@
...
+
+impl Cmd {
+    pub fn run(&self) -> Result<(), lab::token::wrap::Error> {
+        let asset = parse_asset(&self.asset)?;
+        let network = self.config.get_network()?;
+        let contract_id = get_contract_id(&asset, &network.network_passphrase)?;
+        let contract_key = stellar_strkey::Contract(contract_id.0).to_string();
+        println!("Id: {contract_id}");
+        println!("Key: {contract_key}");
+        Ok(())
```
willemneal, line 26: I find this distinction confusion. There are both the contract id, one encoded in hex and strkey. We are also trying to move away from hex altogether since strkey encodes the fact that it is a contract. So perhaps it's best to just print the `contract_id`.

willemneal (changes requested):
Great work! I'm sure people will be happy. Glad to see how easy it is to add new commands. Just a few comments.

cmd/soroban-cli/src/commands/lab/token/id.rs
leighmcculloch: I think we should print only the strkey, the `contract_key` in this code. But let's refer to that as the I'd rather than a key.

You're right at launch there should be no hex values for IDs.

paulbellamy (approved):

paulbellamy:
This is awesome @vinamogit. Thanks so much!

PR #695, opened 2023-06-14
soroban-rpc: State Expiration

### LINKED ISSUE #664: soroban-rpc: implement state-compaction support
### What ?
The goal of this task is to implement a state compaction support in the soroban-rpc so that it would behave equal to the core.

### Entry Creation

- [x] Soroban-RPC should store the `expirationLedgerSeq` field for ContractDataEntry and ContractCodeEntry type LedgerEntries.
- [x] Nice-to-have: The `getLedgerEntries` should include the entry's expirationLedgerSeq value if set. This might already happen automatically, but double-check it.

Normal ledger entry creation, as before, except ContractDataEntry and ContractCodeEntry have a new xdr field, `expirationLedgerSeq`.

### Entry Extension

- [x] When Soroban-RPC sees an EXPIRATION_EXTENSION, it should update the stored entry with the new expirationLedgerSeq.

Users can extend the expirationLedgerSeq via several ways (manual/automatic/host-fn). But whichever way they do, a LedgerEntry will be emitted.

The ContractDataEntry and ContractCodeEntry have a new subtype called: [EXPIRATION_EXTENSION](https://github.com/stellar/stellar-xdr/blob/956da0a0d6020efd539f3adbc6f7a0d4876478e4/Stellar-ledger-entries.x#L522-L523). That just includes the ledger entry key, and the new expirationLedgerSeq, but doesn’t re-emit the full entry value.

### Entry Expiration

- [x] v1: Deny access to ledger entries where `expirationLedgerSeq` is in the past.
- [ ] v2: Allow access to expired entries in `simulateTransaction`, but record them and return that to the client.

Entries at this stage should *not* be deleted from Soroban-RPC. Users can still extend the entry’s expiration to make it accessible again.

Nothing is emitted by Core at this stage. It is up to downstream systems to track the `expirationLedgerSeq`, and make sure that entries where the expiration is in the past are not made available to contracts.

### Entry Deletion

- [ ] Soroban-RPC does not do long-term storage, so all entries which the new LedgerCloseMetaV2 lists as “evicted” should be deleted.

The new xdr type [LedgerCloseMetaV2](https://github.com/stellar/stellar-xdr/blob/956da0a0d6020efd539f3adbc6f7a0d4876478e4/Stellar-ledger.x#L493-L524) now includes 3 new fields to support state expiration:
```
    // Size in bytes of BucketList, to support downstream
    // systems calculating storage fees correctly.
    uint64 totalByteSizeOfBucketList;

    // Expired temp keys that are being evicted at this ledger.
    LedgerKey evictedTemporaryLedgerKeys<>;

    // Expired restorable ledger entries that are being
    // evicted at this ledger.
    LedgerEntry evictedRestorableLedgerEntries<>;
```

Specifically, when an entry is deleted from the core bucketlist it will be included in either the `evictedTemporaryLedgerKeys`, or the `evictedRestorableLedgerEntries`.

The temporary keys are simply dropped.

The Restorable entries should be picked up by the EvictedStateArchiveNodes for long-term storage.

### Horizon

Update Horizon's ContractData ingest processor to make sure that balances are maintained correctly when they are expired/evicted/restored.

I expect this is a no-op, (aside from the xdr updates), because when it is restored the full balance is emitted as the new ledger-entry. But confirm with @SirTyson when he is back.

paulbellamy:
### What

Add support to soroban-rpc for state expiration in preview 10.

### Why

Fixes #664

### TODO

- [x] Update the xdr in go (`preview-10` branch)
	- [x] Fix the xdr enough so it builds
- [x] Update the xdr in soroban-tools (`664/rpc-state-expiration` branch)
- [x] store the expiration
- [x] Update the expiration seq when extensions are ingested
	- Figure out how to find out the existing expiration
- [x] Deny access to expired entries
	- [x] Check if it is < or <=. Asked garand on slack
	- [x] check if env is going to enforce this, or if rpc needs to
		- From grepping, can't see anywhere that does this in env or core
		- Asked on slack...
- [x] Delete evicted entries
- [x] write unit tests
	- [x] ExtendEntry basics
	- [x] ExtendEntry entry not found
	- [x] expired entries not returned
	- [x] multiple extensions in the same txn
- [x] wait for @SirTyson's new xdr changes
- [x] Add a unit test for LedgerChangeReader
- [x] Check assumptions still match how state expiration works now
- [x] PR/merge/update the changes in the `go` repo:
  - [x] https://github.com/stellar/go/pull/4941
  - [x] https://github.com/stellar/go/pull/4942
  - [x] update the go dep here

### Next Release

- [ ] Improve expired entry handling. Allow simulateTransaction access, and return a recording to the wallet
- [x] decide if getLedgerEntries should include expired entries
	- Decision is to add a flag to `getLedgerEntries(keys, { includeExpired })`
	- [ ] Add the flag to `getLedgerEntries`


cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -148,6 +196,29 @@ func (l *ledgerEntryReadTx) GetLedgerEntry(key xdr.LedgerKey) (bool, xdr.LedgerE
 	if err = xdr.SafeUnmarshal([]byte(ledgerEntryBin), &result); err != nil {
 		return false, xdr.LedgerEntry{}, err
 	}
+
+	// Disallow access to entries that have expired. Expiration includes the
+	// "current" ledger, which we are building.
+	// TODO: Support allowing access, but recording for simulateTransaction
+	switch result.Data.Type {
+	case xdr.LedgerEntryTypeContractData:
```
tsachiherman, line 204: fallthrough ?

paulbellamy: Are you suggesting a fallthrough? or?

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -40,6 +42,53 @@ type ledgerEntryWriter struct {
...
+			return err
+		}
+	}
+
+	// Unmarshal the existing entry
+	if err := xdr.SafeUnmarshal([]byte(existing), &entry); err != nil {
+		return err
+	}
+
+	// Update the expiration
```
tsachiherman, line 80: should we add validation here that `entry.Data.Type == key.Type` ?

cmd/soroban-rpc/internal/db/ledgerentry.go
tsachiherman: The codepath here seems to be identical other then the referenced field ContractData / ContractCode.
maybe something like:
```golang
expirationLedger := -1
switch result.Data.Type {
	case xdr.LedgerEntryTypeContractData:
		expirationLedger = result.Data.ContractData.ExpirationLedgerSeq
        case xdr.LedgerEntryTypeContractCode:
		expirationLedger = result.Data.ContractCode.ExpirationLedgerSeq
}
if expirationLedger > -1 {
       latestLedger, err := l.GetLatestLedgerSequence()
	if err != nil {
		return false, xdr.LedgerEntry{}, err
	}
        if expirationLedger <= xdr.Uint32(latestLedger) {
		return false, xdr.LedgerEntry{}, nil
	}
}
```

tsachiherman (commented):
Ledger entries should be removed at some point, as a result of the ingestion. That code wasn't implemented here yet, right ?

cmd/soroban-rpc/internal/db/ledgerentry.go
paulbellamy: No? The key and data are different things?

paulbellamy:
It actually *does* remove ledger entries when they are evicted from core. Neatly, the LedgerChangeReader support [in the `go` repo](https://github.com/stellar/go/pull/4941), means that evictions get converted into a `Change{Pre: entry, Post: nil}` change in ingest, and the code [here](https://github.com/stellar/soroban-tools/blob/145416d3d19819bb3e19f806ba4598ed4c91196f/cmd/soroban-rpc/internal/ingest/ledgerentry.go#L50-L55), will pick that up like a normal deletion.

tsachiherman:
> It actually _does_ remove ledger entries when they are evicted from core. Neatly, the LedgerChangeReader support [in the `go` repo](https://github.com/stellar/go/pull/4941), means that evictions which core sends out in the LedgerCloseMeta get converted into a `Change{Pre: entry, Post: nil}` change in ingest, and the code [here](https://github.com/stellar/soroban-tools/blob/145416d3d19819bb3e19f806ba4598ed4c91196f/cmd/soroban-rpc/internal/ingest/ledgerentry.go#L50-L55), will pick that up like a normal deletion.

just saw that PR..

cmd/soroban-rpc/internal/db/ledgerentry.go
tsachiherman: my understanding ( which could be wrong ) is that the entry contains multiple fields and that the key is one of them. and if that key is supposed to be identical to the "main" ledger key, then it would make sense to perform some sort of sanity validation.

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -148,6 +197,20 @@ func (l *ledgerEntryReadTx) GetLedgerEntry(key xdr.LedgerKey) (bool, xdr.LedgerE
...
+
+	// Disallow access to entries that have expired. Expiration excludes the
+	// "current" ledger, which we are building.
+	// TODO: Support allowing access, but recording for simulateTransaction
+	if expirationLedgerSeq, ok := result.Data.ExpirationLedgerSeq(); ok {
+		latestLedger, err := l.GetLatestLedgerSequence()
+		if err != nil {
+			return false, xdr.LedgerEntry{}, err
+		}
+		if expirationLedgerSeq <= xdr.Uint32(latestLedger) {
```
tsachiherman, line 209: @dmkozh how does the core treat expiration ledger ? is it inclusive or exclusive ?
i.e. should it be like the above or just `<` ?
does it mean "the last ledger at which the data is available" or "the first ledger where the data is no longer availble" ?

cmd/soroban-rpc/internal/db/ledgerentry_test.go
```diff
@@ -196,6 +437,23 @@ func getContractDataLedgerEntry(data xdr.ContractDataEntry) (xdr.LedgerKey, xdr.
 	return key, entry
 }
 
+func getContractCodeLedgerEntry(code xdr.ContractCodeEntry) (xdr.LedgerKey, xdr.LedgerEntry) {
```
tsachiherman, line 440: given that this is a testing code, it would be cleaner to add the `t *testing.T` to its input parameters and `require.NotNil` the error below instead of panicing.

cmd/soroban-rpc/internal/db/ledgerentry.go
dmkozh: it's the former, i.e. the entry is still available on expiration ledger. So this should be `<`.

paulbellamy: `latestLedger` here refers to the latest closed ledger. So this will only hide the entry once the expirationLedgerSeq ledger number is closed.

cmd/soroban-rpc/internal/db/ledgerentry.go
paulbellamy: Ah, right. Yes, we should do that in the `Upsert` method though, so it is done pre-emptively, not just on extension. Doing that cleanly means a refactoring PR to the `go` repo, which is in [this new PR.](https://github.com/stellar/go/pull/4942)

paulbellamy: Actually, we can simplify the UpsertLedgerEntry method, since the key is derived from the entry, we don't need to pass in both. Then there is nothing to validate.

tsachiherman (approved):

PR #696, opened 2023-06-14
fix: make opt default; to opt-out `cargo install --no-default-features`

willemneal:
### What

When installing users must do `--no-default-features` to opt out of using wasm-opt.

### Why

Hopefully will fix https://github.com/stellar/soroban-tools/actions/runs/5260321197/jobs/9507078865?pr=694

### Known limitations

[TODO or N/A]

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,7 +49,7 @@ jobs:
         cd target/package
         tar xvfz soroban-cli-$VERSION.crate
         cd soroban-cli-$VERSION
-        cargo build --target-dir=../.. --features opt --release --target ${{ matrix.target }}
+        cargo build --target-dir=../.. --release --target ${{ matrix.target }}
     - name: Compress 
       run: |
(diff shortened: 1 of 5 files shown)


leighmcculloch:
The error in that run looks like it's environmental:
```
Caused by:
  failed to remove file `D:\a\soroban-tools\soroban-tools\target\package\soroban-cli-0.8.5\target\x86_64-pc-windows-msvc\debug\build\wasm-opt-cxx-sys-7458c11355075ff3\out\cxxbridge\crate\wasm-opt-cxx-sys`
```

It's unclear how changing the default feature set would fix this.

leighmcculloch:
I pushed a commit temporarily enabling the publish-dry-run job so that we can see if this fixes the issue.

willemneal:
>  Access is denied. (os error 5)

It seems windows doesn't have the right permissions

leighmcculloch:
Problem is happening with this change too, so I'm going to close this PR.

PR #697, opened 2023-06-14
fix(ts-gen): update 'spec' version; fix i128

chadoh:
Use version of `soroban-spec` from https://github.com/stellar/rs-soroban-sdk/pull/976

### What

Fix `i128` & `u128` encodings for the JS libraries generated with
`soroban contract bindings typescript`

### Why

So that projects using `i128` and `u128` have functional JS libraries.

### Known limitations

Other common types may not yet be functional.


Cargo.toml
```diff
@@ -17,8 +17,8 @@ rev = "0c0cae6fa22b751e7fd95d7ce7556ac6d7b7010e"
 
 [workspace.dependencies.soroban-spec]
 version = "0.8.5"
```
leighmcculloch, line 19: This needs changing to 0.8.6.

willemneal: Do we still need all the rev's and git in all these deps?

tsachiherman: not sure, but let's defer that discussion so we can release it ?

Cargo.toml
```diff
@@ -17,8 +17,8 @@ rev = "0c0cae6fa22b751e7fd95d7ce7556ac6d7b7010e"
...
-git = "https://github.com/stellar/rs-soroban-sdk"
-rev = "7d143e425b690f2981ae804ce249462d98a68b0d"
+git = "https://github.com/ahalabs/rs-soroban-sdk"
```
leighmcculloch, line 20: This should be stellar.
```suggestion
git = "https://github.com/stellar/rs-soroban-sdk"
```

paulbellamy:
Done in the other repo after code move.

PR #701, opened 2023-06-15
build: update slash direction for windows RPC builds

tsachiherman:
### What

In the Github testing, I used \\ ( double slash ) to specify path on windows platform.
In a previous code review, it was suggested to replace these with /, since windows understands both.

### Why

It's sightly more aligned with the rest of the yml file content.

### Known limitations

N/A. If it works, it works.


Shaptic (approved):
nice :ok_hand:

PR #702, opened 2023-06-15
feat(CLI): move soroban-spec-json/typescipt; add soroban-spec-tools

willemneal:
### What

Move the crates `soroban-spec-json` and `soroban-spec-typescript` to this repo and move `strval.rs`  a new crate called sorobon-spec-tools.

### Why

Remove dependency on sdk and allow for using the JSON parsing outside the CLI.

### Known limitations

[TODO or N/A]

PR #708, opened 2023-06-16
feat: upgrade freighter-api to 1.5.0 & soroban-client to 0.8.1

chadoh:
### TODO

- [x] wait for #702 to merge, then rebase on `main` to remove all but the last commit here

### What

For JS libraries generated with `soroban contract bindings typescript`:

- upgrade to freighter-api 1.5.0

- use the new methods from `freighter-api@1.5.0` to avoid popping
  pop-ups over and over every time a view call is invoked

- if the transaction only needs to be simulated, then the user doesn't
  need to have Freighter connected. This now uses a default/dummy
  account for such calls, but throws an error if the app author tries to
  `signAndSend` such calls.

- upgrade `soroban-client` to 0.8.1, which allows removing the pinned version of `stellar-base`

### Why

This greatly improves the devx & ux of apps that use these generated libraries.

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-spec-typescript/fixtures/ts/src/invoke.ts
+++ b/cmd/crates/soroban-spec-typescript/fixtures/ts/src/invoke.ts
@@ -2,7 +2,8 @@ import freighter from "@stellar/freighter-api";
 // working around ESM compatibility issues
 const {
   isConnected,
-  getPublicKey,
+  isAllowed,
+  getUserInfo,
   signTransaction,
 } = freighter;
 import * as SorobanClient from 'soroban-client'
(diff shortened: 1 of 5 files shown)


tomerweller (approved):

cmd/crates/soroban-spec-typescript/fixtures/ts/src/invoke.ts
```diff
@@ -54,7 +53,10 @@ export class NotImplementedError extends Error { }
  * @returns The transaction response, or the simulation result if signing isn't required.
  */
 export async function invoke({ method, args = [], fee = 100, signAndSend = false }: InvokeArgs): Promise<(TxResponse & { xdr: string }) | Simulation> {
-  const account = await getAccount()
+  const freighterAccount = await getAccount()
+
+  // use a placeholder account if not yet connected to Freighter so that view calls can still work
+  const account = freighterAccount ?? new SorobanClient.Account('GBZXP4PWQLOTBL3P6OE6DQ7QXNYDAZMWQG27V7ATM7P3TKSRDLQS4V7Q', '0')
```
leighmcculloch, line 59: Non urgent followup: Not a big deal since the seq num is zero this tx will be unusable and I assume no one has the priv key, but to communicate even clearer that this account is meaningless I'd replace it with the zero account.

willemneal: This was just meant for simulation only calls. Does the sequence number matter?

chadoh: Yes, I'd love to replace it with the zero account. I haven't seen that around yet and don't see it documented at soroban.stellar.org/docs. Is this correct?

```
G0000000000000000000000000000000000000000000000000000000
```

leighmcculloch:
```rust
use stellar_strkey::ed25519::PublicKey;
fn main() {
    let zero = PublicKey([0; 32]);
    println!("{zero}");
}
```
```
GAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAWHF
```

leighmcculloch: strkeys are base32 encoded. Lots of zeros encode into AAAAAAAAAA...

PR #709, opened 2023-06-17
Fix publish with new crates

leighmcculloch:
### What
Fix publish with new crates, soroban-spec-tools, soroban-spec-typescript, soroban-spec-json.

### Why
See the description of https://github.com/stellar/actions/pull/54 for why publish-dry-run-v2 is required for this repository, which contains binaries.


graydon (approved):

.github/workflows/rust.yml
```diff
@@ -86,7 +86,8 @@ jobs:
           target: aarch64-apple-darwin
 #       - os: windows-latest
 #         target: x86_64-pc-windows-msvc
-    uses: stellar/actions/.github/workflows/rust-publish-dry-run.yml@main
+    uses: stellar/actions/.github/workflows/rust-publish-dry-run-v2.yml@main
     with:
+      crates: soroban-spec-tools soroban-spec-json soroban-spec-typescript soroban-test soroban-cli
```
leighmcculloch, line 91: There's one big downside of having to specify the crates here. If we add a new crate, and forget to specify it here, it'll still be published by the publish job, but it won't be verified here. The main downside of this is that if the crate fails to publish it'll break the release. It's not the end of the world.

We should do some improvements, by probably making a corresponding publish v2 that requires an explicit list too. I'll think about it some more.

PR #710, opened 2023-06-20
Add package-based soroban-rpc Dockerfile

jacekn:
### What

This PR adds new soroban-rpc Dockerfile that uses official SDF deb packages as source of stellar-soroban-rpc and stellar-core

### Why

This Dockerfile will allow us to reuse release packages. It will also allow us to run more lighweight soroban-rpc deployments than those using the stellar/quickstart image


cmd/soroban-rpc/docker/Makefile
```diff
@@ -0,0 +1,18 @@
+SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
+
+# https://github.com/opencontainers/image-spec/blob/master/annotations.md
+BUILD_DATE := $(shell date -u +%FT%TZ)
+
+TAG ?= stellar/stellar-soroban-rpc:$(SOROBAN_RPC_VERSION)
+
+STELLAR_CORE_VERSION ?= 19.10.1-1310.6649f5173.focal~soroban
+SOROBAN_RPC_VERSION ?= 0.8.0-21
```
tsachiherman, line 9: Hardcoding these here would force us to update these, or to have us invoke this makefile with the "correct" values.
I think that optionally providing a command-line override is perfectly fine, but these should be derived from the repositories themselves. 
in concrete actions, I'm thinking about the following:
1. find the latest soroban-rpc release tag in the soroban-tools repository. extract the version from there.
2. Use the above tag ( i.e. commit hash ), to extract the go.mod file. This would tell you the go monorepo hash. From there you could get the stellar core version and use it here.

If you need help figuring out how to extract that information, we've done a similar thing in the quickstart repository ( scripts dir )

jacekn: This could work but there are some challenges we'll need to address I think. Currently core builds are "promoted" manually. We can also have multiple builds (and packages) use the same commit and version, they can use different build flags for example. With that we'd need a way to pick which exact build to use.

I was also thinking about getting build tags from the docs but that's also not trivial - what if we want to build images before releasing them?

cmd/soroban-rpc/docker/Makefile
```diff
@@ -0,0 +1,18 @@
...
+
+docker-build:
+	$(SUDO) docker build --pull $(DOCKER_OPTS) \
```
sreuland, line 12: are the deb pkgs published for both amd64 and arm64? If not arm64 yet, should probably set `--platform linux/amd64' , otherwise docker won't be able to build from any other non-amd hosts, like on dev workstations that are apple m1, etc.

cmd/soroban-rpc/docker/Makefile
tsachiherman: while it's true that there are multiple builds for the same hash, these would eventually narrow down to a single build that you'd like to pick. 
Adding another reference is a guaranteed way for us to "miss" the release date.. I'd like to avoid doing that.

leighmcculloch:
> It will also allow us to run more lighweight soroban-rpc deployments than those using the stellar/quickstart image

It's possible for us to create a docker container that isn't quickstart, without introducing `.deb` installers into the mix. There have been some challenges with building our docker contains this way (how horizon's was built):
- There's no support for arm64.
- We can't ship docker images on PRs, because of the needed coordination with deb building on Jenkins.

Do we need the debs, or do we just need a docker image?

cmd/soroban-rpc/docker/Makefile
leighmcculloch: > are the deb pkgs published for both amd64 and arm64?

arm64 isn't published yet:
- https://github.com/stellar/packages/issues/154.

tsachiherman:
> 

I agree with @leighmcculloch 's point here. The fact that we're tying our docker image generation here to a prebuilt debian release ( from jenkins ) is a serious limitation. When building from source, we need *always* build from... source.

sreuland:
> > 
> 
> I agree with @leighmcculloch 's point here. The fact that we're tying our docker image generation here to a prebuilt debian release ( from jenkins ) is a serious limitation. When building from source, we need _always_ build from... source.

yes, 'build from source' has become an approach used in other places to enable multi-arch situations, quickstart does it and system-test followed suit, but it has introduced an unintended side effect, it requires compute resources and time to compile core from source. one other idea getting some discussion was pre-built multi-arch docker images for core, which has caveats also, like which repos can push them and which are most common permutation of build flags to build images on, but, if that existed, then docker's like this rpc one can safely copy the core bin from that image as a build stage.

cmd/soroban-rpc/docker/Makefile
jacekn: > If not arm64 yet, should probably set `--platform linux/amd64' 
Good point, currently we only support amd64 builds, I added the flag

jacekn:
> Do we need the debs, or do we just need a docker image?

We wanted to build images that use supported, tested version of core for production use. Currently we only test and validate amd64 builds of core. amd64 debs are also the only builds that can be cryptographically verified to be coming from SDF. This is a well established release process that we've been using for many years so moving to custom compiled core binaries for soroban-rpc images is a fairly big deal IMO.
I appreciate that the build process is more limited but that's, unfortunately, the price of having higher security guarantees and thorough testing that those builds go through. Maybe there will be ways to integrate things more for easier testing in GHA?

Having said that it looks like we really need to bump priority of arm64 support. It likely won't matter much for production use cases, where most systems run amd64, but being able to run the images on arm laptops would benefit developers. This can be a significant project of course, depends on how much testing infrastructure we need to duplicate on arm64 architecture.

tsachiherman (approved):

PR #712, opened 2023-06-21
rust: fix rust-analyzer testing

tsachiherman:
### What

Following a recent dependency update, the `rust-analyzer` started failing when executing the following:
```sh
rustup run nightly rust-analyzer analysis-stats . 2>&1 | (! grep ERROR)
```

### Why

One of the dependencies was including the text "ERROR", which caused this test to fail. This is a false-positive,
since there is no actual failure here. The solution here is to omit the stdout and use the stderr only when scanning for the "ERROR" text.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,7 +36,7 @@ jobs:
     - run: rustup update
     - run: rustup +nightly component add rust-analyzer
     - name: Check if rust-analyzer encounters any errors parsing project
-      run: rustup run nightly rust-analyzer analysis-stats . 2>&1 | (! grep ERROR)
+      run: rustup run nightly rust-analyzer analysis-stats . 2>&1 >/dev/null | (! grep ERROR)
 
   build-and-test:
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     - run: rustup update
     - run: rustup +nightly component add rust-analyzer
     - name: Check if rust-analyzer encounters any errors parsing project
-      run: rustup run nightly rust-analyzer analysis-stats . 2>&1 | (! grep ERROR)
+      run: rustup run nightly rust-analyzer analysis-stats . 2>&1 >/dev/null | (! grep ERROR)
```
leighmcculloch, line 39: We just fixed this in another repo with the combination of these changes:
- https://github.com/stellar/rs-soroban-sdk/pull/994
- https://github.com/stellar/rs-soroban-sdk/pull/995

The output is important, we do not want to send it to dev null. Instead we need to improve the match:

```suggestion
      run: rustup run nightly rust-analyzer analysis-stats . 2>&1 | (! grep '^\[ERROR')
```

leighmcculloch (dismissed):

.github/workflows/rust.yml
tsachiherman: I wasn't aware of that fix on the other repos. Let me update it here.

leighmcculloch (approved):

PR #713, opened 2023-06-22
Add command for building contracts

leighmcculloch:
### What
Add command for building contracts:

```
soroban contract build
```

```
❯ soroban contract build -h
Build a contract from source

Usage: soroban contract build [OPTIONS]

Options:
      --manifest-path <MANIFEST_PATH>  Path to Cargo.toml [default: Cargo.toml]
      --package <PACKAGE>              Package to build
      --profile <PROFILE>              Build with the specified profile [default: release]
      --out-dir <OUT_DIR>              Directory to copy wasm files to
  -h, --help                           Print help (see more with '--help')

Features:
      --features <FEATURES>  Build with the list of features activated, space or comma separated
      --all-features         Build with the all features activated
      --no-default-features  Build with the default feature not activated

Other:
      --print-commands-only  Print commands to build without executing them
```

```
❯ soroban contract build --help
Build a contract from source

Builds all crates that are referenced by the cargo manifest (Cargo.toml) that have cdylib as their crate-type. Crates are built for the wasm32 target. Unless configured otherwise, crates are built with their default features and with their release profile.

To view the commands that will be executed, without executing them, use the --print-commands-only option.

Usage: soroban contract build [OPTIONS]

Options:
      --manifest-path <MANIFEST_PATH>
          Path to Cargo.toml
          
          [default: Cargo.toml]

      --package <PACKAGE>
          Package to build
          
          If omitted, all packages that build for crate-type cdylib are built.

      --profile <PROFILE>
          Build with the specified profile
          
          [default: release]

      --out-dir <OUT_DIR>
          Directory to copy wasm files to
          
          If provided, wasm files can be found in the cargo target directory, and the specified directory.
          
          If ommitted, wasm files are written only to the cargo target directory.

  -h, --help

Features:
      --features <FEATURES>
          Build with the list of features activated, space or comma separated

      --all-features
          Build with the all features activated

      --no-default-features
          Build with the default feature not activated

Other:
      --print-commands-only
          Print commands to build without executing them
```

### Why

The commands for building contracts are getting increasingly complex, e.g.:
```
cargo rustc --crate-type cdylib --target wasm32-unknown-unknown --release
```

A build command on soroban-cli has been requested many times. Most recently at: https://discord.com/channels/897514728459468821/1121218101489447054/1121219305133064284

Related to https://github.com/stellar/soroban-docs/pull/476

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -1,0 +1,118 @@
+use clap::Parser;
+use itertools::Itertools;
+use std::{
+    collections::HashSet,
+    ffi::OsStr,
+    fmt::Debug,
+    process::{Command, Stdio},
+};
+
+use cargo_metadata::{Metadata, MetadataCommand, Package};
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// Path to Cargo.toml
+    #[arg(long)]
+    pub manifest_path: Option<std::path::PathBuf>,
+    /// Package to build
+    #[arg(long)]
+    pub package: Option<String>,
+    /// Build with the specified profile
+    #[arg(long, default_value = "release")]
+    pub profile: String,
+    /// Build with the list of features activated, space or comma separated
+    #[arg(
+        long,
+        conflicts_with = "all_features",
+        conflicts_with = "no_default_features"
+    )]
+    pub features: Option<String>,
+    /// Build with the all features activated, space or comma separated
+    #[arg(
+        long,
+        conflicts_with = "features",
+        conflicts_with = "no_default_features"
+    )]
+    pub all_features: bool,
+    /// Build with the default feature not activated
+    #[arg(long, conflicts_with = "features", conflicts_with = "all_features")]
+    pub no_default_features: bool,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Metadata(#[from] cargo_metadata::Error),
+}
+
+impl Cmd {
+    pub fn run(&self) -> Result<(), Error> {
+        let packages = self.packages()?;
+
+        for p in packages {
+            let mut cmd = Command::new("cargo");
+            cmd.stdout(Stdio::piped());
+            cmd.arg("rustc");
+            cmd.arg(format!("--manifest-path={}", p.manifest_path));
+            cmd.arg("--crate-type=cdylib");
+            cmd.arg("--target=wasm32-unknown-unknown");
+            cmd.arg(format!("--package={}", p.name));
+            cmd.arg(format!("--profile={}", self.profile));
+            if self.all_features {
+                cmd.arg("--all-features");
+            }
+            if self.no_default_features {
+                cmd.arg("--no-default-features");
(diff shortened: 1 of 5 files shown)


cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -0,0 +1,189 @@
...
+    ///
+    /// If provided, wasm files can be found in the cargo target directory, and
+    /// the specified directory.
+    ///
+    /// If ommitted, wasm files are written only to the cargo target directory.
+    #[arg(long)]
+    pub out_dir: Option<std::path::PathBuf>,
+    /// Print commands to build without executing them
+    #[arg(long)]
+    pub print_commands_only: bool,
```
tsachiherman, line 59: dry run ?

tsachiherman (commented):
Should we check the rustc version to make sure the output is what we expect it to be ?

As long as we left the end user with the responsibility to generate the wasm, it was their responsibility to provide a valid wasm to the rest of the SDK. But with this change, we'll be asked about inconsistent wasm output. I do like this change, and think it's the right move - but I wish we could have "embed" that functionality into the cli.

cmd/soroban-cli/src/commands/contract/build.rs
leighmcculloch: I'm contemplating removing the feature, to keep the feature set a little slimmer. Also, because not everything this command does is captured in cargo commands as output. Let me know if you think we should keep it.

leighmcculloch: Actually, the option is really helpful for testing and development.

leighmcculloch:
> Should we check the rustc version to make sure the output is what we expect it to be ?

I don't think so because we're doing a few things to make sure that a change in rustc would be unlikely to break soroban-cli:

1. The `cargo_metadata` crate we're using passes `--format-version 1` to the `cargo metadata` command, to ensure if the output is changed we continue to get the version 1 output.

2. The subcommand only uses stable cargo and rustc options, meaning that these options are highly unlikely to change.

3. There are no dependencies on the stdout or stderr output of the rustc command, other than assuming a non-zero error code is a failure.

4. The location of .wasm files is unlikely to change because there are many tools that assume the location of artifacts in cargo's target directory structure. Even if it was surprisingly changed on us (unlikely), it would only break the `--out-dir` option, which we could patch quickly. To my knowledge there's no way we could detect this ahead of time. We read the target directory location from the metadata, but nothing in the cargo metadata could tell us the location of the artifact.

cmd/soroban-cli/src/commands/contract/build.rs
tsachiherman: rehashing /w myself, we're not the first ones to invoke a compiler under the hood. Both rustc and go are invoking the gnu compiler, and both have options to allow the enduser to pass in arbitrary parameters and/or extra printing.

leighmcculloch: I named it `--print-commands-only` because that's explicitly describing what it does. The name of the option is similar to the same option that's provided in the `cargo-hack` tool.

I can rename it to dry run if that is preferred / makes more sense.

tsachiherman (approved):

PR #721, opened 2023-06-23
Enable diagnostic debug level in sandbox

leighmcculloch:
### What
Enable diagnostic debug level in sandbox

### Why
They used to be enabled via a crate feature, but that functionality was converted to a function that gets called.

This needs enabling so that when running in sandbox mode debug events are outputted by a contract that is built with debug assertions enabled.


leighmcculloch:
I don't understand why, but this change doesn't appear to be necessary. Contract logs are emitted without turning on diagnostics. cc @graydon

PR #722, opened 2023-06-26
preview10: soroban-cli xdr updates

paulbellamy:
### What

Update the xdr in soroban-cli

### Why

for preview 10

### Known limitations

[TODO or N/A]


cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1185,7 +1193,7 @@ impl Spec {
             ScType::I32 => Some("-1".to_string()),
             ScType::Bool => Some("true".to_string()),
             ScType::Symbol => Some("\"hello\"".to_string()),
-            ScType::Status => Some("Status".to_string()),
+            ScType::Error => Some("Error".to_string()),
```
willemneal, line 1196: These are example inputs, so perhaps an Err is better. Though these are only seen when used and an Error as input is likely never to happen.

tsachiherman (approved):

PR #726, opened 2023-06-27
soroban-rpc: bug workaround and test fixes

2opremio:


cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -468,9 +467,9 @@ func TestSimulateInvokeContractTransactionSucceeds(t *testing.T) {
 	assert.Nil(t, obtainedAuth.RootInvocation.SubInvocations)
 
 	// check the events
-	assert.Len(t, response.Events, 1)
+	assert.Len(t, response.Events, 3)
```
sreuland, line 470: for consideration, maybe add comment or assert message, statying why 3 events are expected?

sreuland (approved):

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -430,7 +429,7 @@ func TestSimulateInvokeContractTransactionSucceeds(t *testing.T) {
 	ro1 := obtainedFootprint.ReadOnly[1]
 	assert.Equal(t, xdr.LedgerEntryTypeContractData, ro1.Type)
 	assert.Equal(t, xdr.ScAddressTypeScAddressTypeContract, ro1.ContractData.Contract.Type)
-	assert.Equal(t, xdr.Hash(contractID), ro1.ContractData.Contract.ContractId)
+	assert.Equal(t, xdr.Hash(contractID), *ro1.ContractData.Contract.ContractId)
```
tsachiherman, line 432: if testing a value in a *uint, I think that adding
```golang
require.NotNil(t, ro1.ContractData.Contract.ContractId)
```
could be a good idea - i.e. we don't want the test to panic due to segmentation fault. We want a "clean" failure.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -468,9 +467,9 @@ func TestSimulateInvokeContractTransactionSucceeds(t *testing.T) {
 	assert.Nil(t, obtainedAuth.RootInvocation.SubInvocations)
 
 	// check the events
-	assert.Len(t, response.Events, 1)
+	assert.Len(t, response.Events, 3)
 	var event xdr.DiagnosticEvent
-	err = xdr.SafeUnmarshalBase64(response.Events[0], &event)
+	err = xdr.SafeUnmarshalBase64(response.Events[1], &event)
```
tsachiherman, line 472: a followup on @sreuland remarks - if the contract now returns 3 events, shouldn't we test all of them ?
note that it shouldn't be a blocker for this PR; but we should note to ourselves to address that.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -486,7 +485,7 @@ func TestSimulateInvokeContractTransactionSucceeds(t *testing.T) {
 	require.Contains(t, metrics, "soroban_rpc_json_rpc_request_duration_seconds_count{endpoint=\"simulateTransaction\",status=\"ok\"} 3")
 	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_get_duration_seconds_count{status=\"ok\",type=\"db\"} 3")
 	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_get_duration_seconds_count{status=\"ok\",type=\"all\"} 3")
-	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_entries_fetched_sum 29")
+	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_entries_fetched_sum 33")
```
tsachiherman, line 488: I'm not sure if we should really test that particular number.. since it looks like it would change over time and cause quite many false positives.. maybe test that the number is between 5 - 500 ?

tsachiherman (commented):
few improvement suggestions, but looks good otherwise.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: This predates the current PR. Please create a separate ticket if you think it’s important

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: The other events are debug events, we should test for the event emitted explicitly by the contract implementation

PR #727, opened 2023-06-27
soroban-rpc integration tests: derive core docker container from go monorepo

tsachiherman:
### What

soroban-rpc integration tests: derive core docker container from go monorepo

### Why

Make sure that both the go monorepo and the soroban are working against the same XDR and the same core binaries.
In essense, instead of specifying the docker container and the debian package in the soroban-tools repository, this change allow us to dynamically retrieve that from the horizon.yml file in the go monorepo.
That reduces the number of locations where the core debian/docker need to be updated and ensures that the repositories dependencies aligns with the CI testing.

### Known limitations

N/A


2opremio:
I am not sure this is worth automating. I would much rather see the all core images being published publicly so that we don't need to do it manually.

2opremio:
@paulbellamy thoughts?

tsachiherman:
> I am not sure this is worth automating. I would much rather see the all core images being published publicly so that we don't need to do it manually.

I think that automation is good for "tip of tree" like branch. But for all archival/historical branches we always want to have explicit dependencies.

I can't, honestly, see us getting to a complete automation here in the near future.. however, I can very well see us "missing" updates and having undesired downstream implications.

tsachiherman:
> This may be a sign that we need to start thinking about ABIs or at least version syncing between components during development.

There is no sign here; we do need to think about XDR compatibility when developing. The goal of this PR is to force us to align the XDR with the binaries being used.

> We have soroban-rpc, horizon and core, all have to use the same XDR. At the same time there can be many versions of each of the components that use a given XDR revision. It may also be possible for components to move independently and bump XDR version they use.

There could be many versions, but that would never present itself as a problem : the "link" between the repositories is always explicit, so there is no confusion on which version is being used.
You're correct that if we had "grab the latest of that component", then it would probably be an issue. But we don't have that anywhere.

> Is it feasible to always test against latest "unstable" core?
It would be a nice to have, but I would caution against it. The reason for that is that it would be only as reliable and complete as our tests are. Given that we have a lot to improve on the testing front, I wouldn't prioritize that..
Separately, when we make releases, we typically want to make a claim that we tested a given "set" of components together. No one would care of any other set other then the one we officially release.

PR #730, opened 2023-06-29
soroban-rpc: Implement preflighting of BumpFootprintExpiration and RestoreFootprint operations

2opremio:


cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -287,3 +288,78 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+    let fee_configuration = get_fee_configuration(snapshot_source)?;
+    let (min_fee, ref_fee) =
+        compute_transaction_resource_fee(&transaction_resources, &fee_configuration);
+    let transaction_data = SorobanTransactionData {
+        resources: soroban_resources,
+        refundable_fee: ref_fee,
+        ext: ExtensionPoint::V0,
+    };
+    Ok((transaction_data, min_fee))
+}
```
2opremio, line 365: @sisuresh @dmkozh could you review the two functions above to confirm the fee computation is correct?

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -67,10 +67,19 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
...
 
-		xdrOp, ok := op.Body.GetInvokeHostFunctionOp()
-		if !ok {
+		footprint := xdr.LedgerFootprint{}
+		switch op.Body.Type {
+		case xdr.OperationTypeInvokeHostFunction:
+		case xdr.OperationTypeBumpFootprintExpiration, xdr.OperationTypeRestoreFootprint:
+			if txEnvelope.Type != xdr.EnvelopeTypeEnvelopeTypeTx && txEnvelope.V1.Tx.Ext.V != 1 {
+				return SimulateTransactionResponse{
+					Error: "Footprint operation provided without SorobanTransactionData",
```
tsachiherman, line 76: It would be better to have more explicit error message:
```To perform a SimulateTransaction for BumpFootprintExpiration or RestoreFootprint operations, SorobanTransactionData must be provided```

File: cmd/soroban-rpc/lib/preflight/src/fees.rs [hunk 3]
```diff
@@ -287,3 +288,78 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+pub(crate) fn compute_bump_footprint_exp_transaction_data_and_min_fee(
+    footprint: LedgerFootprint,
+    _ledgers_to_expire: u32,
+    snapshot_source: &ledger_storage::LedgerStorage,
+) -> Result<(SorobanTransactionData, i64), Box<dyn error::Error>> {
+    let read_bytes =
+        calculate_unmodified_ledger_entry_bytes(footprint.read_only.as_vec(), snapshot_source)?;
+    let soroban_resources = SorobanResources {
+        // TODO: is this correct or should we use the footprint?
+        footprint: LedgerFootprint {
```
dmkozh, line 301: The request should contain the list of keys to bump and `ledgers_to_expire` and this should just copy the key list to RO footprint.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -287,3 +288,78 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+
+pub(crate) fn compute_restore_footprint_transaction_data_and_min_fee(
+    footprint: LedgerFootprint,
+    snapshot_source: &ledger_storage::LedgerStorage,
+) -> Result<(SorobanTransactionData, i64), Box<dyn error::Error>> {
+    let write_bytes =
+        calculate_unmodified_ledger_entry_bytes(footprint.read_write.as_vec(), snapshot_source)?;
+    let soroban_resources = SorobanResources {
+        // TODO: is this correct or should we use the footprint?
+        footprint: LedgerFootprint {
```
dmkozh, line 338: Similar to above, the request should contain the list of keys to restore and here we should copy them to the  RW footprint.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -287,3 +288,78 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+        },
+        instructions: 0,
+        read_bytes: 0,
+        write_bytes,
+        extended_meta_data_size_bytes: 2 * write_bytes,
+    };
+    let transaction_resources = TransactionResources {
+        instructions: 0,
+        read_entries: 0,
+        write_entries: 0,
```
dmkozh, line 350: This should be updated to match the footprint.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -287,3 +288,78 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+            read_write: Default::default(),
+        },
+        instructions: 0,
+        read_bytes,
+        write_bytes: 0,
+        extended_meta_data_size_bytes: 2 * read_bytes,
+    };
+    let transaction_resources = TransactionResources {
+        instructions: 0,
+        read_entries: 0,
```
dmkozh, line 312: This should match the footprint

cmd/soroban-rpc/lib/preflight/src/fees.rs
2opremio: I will just pass on the footprint then, let me know if that's what you meant

dmkozh: This would work, but I would suggest to accept a list of the ledger keys from the user, so that they don't need to worry about RO vs RW.

tsachiherman (changes requested):
Just one request.

cmd/soroban-rpc/lib/preflight/src/fees.rs
tsachiherman: @dmkozh I don't see the user really being concerned about that. In the context of `simulateTransaction`, I think that this is the correct approach.
The user is likely to be using the sdks to use these features, and that functionality would be implemented there.
Depending on the usage, we might want to create a dedicated API for that sole purpose, but other then cleaning up some of the REST interface, the won't be much gains there.

cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -124,7 +126,46 @@ func GoNullTerminatedStringSlice(str **C.char) []string {
...
-	invokeHostFunctionB64, err := xdr.MarshalBase64(params.InvokeHostFunction)
+	handle := cgo.NewHandle(snapshotSourceHandle{params.LedgerEntryReadTx, params.Logger})
+	defer handle.Delete()
+	switch params.OpBody.Type {
+	case xdr.OperationTypeInvokeHostFunction:
+		return getInvokeHostFunctionPreflight(params)
+	case xdr.OperationTypeBumpFootprintExpiration, xdr.OperationTypeRestoreFootprint:
+		return getFootprintExpirationPreflight(params)
+	default:
+		return Preflight{}, fmt.Errorf("unsupported operation type: %s", params.OpBody.Type.String())
```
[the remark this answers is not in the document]
Shaptic: only if you assume the stack to `GetPreflight` goes through `NewSimulateTransactionHandler`

cmd/soroban-rpc/lib/preflight/src/fees.rs
dmkozh: Well, the design is really up to you. Semantically these operations just bump/restore a set of keys and ideally their interface would be to just take a list of keys. I just want to make it clear that the way they're implemented in XDR is indeed confusing, but that seemed like a lesser evil out of the alternatives. Here you have more freedom, so you could provide a more semantically correct interface. It also would be more consistent with invoke fn (where user passes the input data and gets the footprint back).

cmd/soroban-rpc/internal/preflight/preflight.go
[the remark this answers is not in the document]
2opremio: I don't think the extra check hurts. Particularly since we are validating user data input.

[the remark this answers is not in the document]
tsachiherman: no - it doesn't hurt.

dmkozh (approved):
The resource computation looks fine to me.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -67,10 +67,19 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
...
+		footprint := xdr.LedgerFootprint{}
+		switch op.Body.Type {
+		case xdr.OperationTypeInvokeHostFunction:
+		case xdr.OperationTypeBumpFootprintExpiration, xdr.OperationTypeRestoreFootprint:
+			if txEnvelope.Type != xdr.EnvelopeTypeEnvelopeTypeTx && txEnvelope.V1.Tx.Ext.V != 1 {
+				return SimulateTransactionResponse{
+					Error: "To perform a SimulateTransaction for BumpFootprintExpiration or RestoreFootprint operations, SorobanTransactionData must be provided",
+				}
+			}
+			footprint = txEnvelope.V1.Tx.Ext.SorobanData.Resources.Footprint
```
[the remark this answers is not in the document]
2opremio: That's ensured by de-serialization (provided that `txEnvelope.V1.Tx.Ext.V` == 1), which we check above.

See https://github.com/stellar/go/blob/531797f01226d661197cae708f19f2d782c1af6d/xdr/xdr_generated.go#L30287

PR #731, opened 2023-06-30
Fix: error, enum, union types

willemneal:
### What

Upgrade to new error type and fix other types.

### Why

### Known limitations

Currently invoke.ts is failing because of xdr change.


willemneal:
@paulbellamy It's failing to find the xdr in the test wasms. Could there be a version mismatch? I'm very confused how this could happen.

Makefile
```diff
@@ -101,5 +101,14 @@ lint-changes:
 lint:
 	golangci-lint run ./...
 
+temp-bindings:
+	cargo run -- contract bindings typescript \
+					--wasm ./target/wasm32-unknown-unknown/test-wasms/test_custom_types.wasm \
```
paulbellamy, line 106: probably depends on this being built

Makefile
```diff
@@ -101,5 +101,14 @@ lint-changes:
 lint:
 	golangci-lint run ./...
 
+temp-bindings:
```
paulbellamy, line 104: maybe `typescript-bindings-fixtures`? Not sure why these are 'temporary' if we are comitting them.

cmd/crates/soroban-test/tests/it/custom_types.rs
```diff
@@ -211,7 +211,7 @@ fn number_arg_return_err() {
 
 #[test]
 fn void() {
-    invoke(&TestEnv::default(), "void")
+    invoke(&TestEnv::default(), "voidd")
```
paulbellamy, line 214: not `woid` to match above?

paulbellamy (approved):
Couple small suggestions, but looks fine.

willemneal:
The invoke.ts is failing on this line:
```
 const auth = SorobanClient.xdr.SorobanAuthorizationEntry.fromXDR(auths![0]!, 'base64')
    if (auth.addressWithNonce() !== undefined) {
      throw new NotImplementedError(
        `This transaction needs to be signed by ${auth.addressWithNonce()
        }; Not yet supported`
      )
    }

**Comment by paulbellamy** [comment]
You probably need https://github.com/stellar/js-soroban-client/pull/103, because that updates the XDR

**Comment by willemneal** [comment]
Should we pause this PR until then? Or merge and make one just for updating that?

**Comment by tyvdh** [comment]
Is there a clean way to test this before preview 10 is live? I'd love for this typescript bindings thing to work well but I'm currently blocked on errors and apparently this PR may help me

**Comment by tyvdh** [comment]
Alright I did 
```bash
cargo install --git https://github.com/AhaLabs/soroban-tools --rev 3630239df49f4f0cce2efb22f7ed5e9e4e23ef89 soroban-cli
```
which installed that commit for this PR however there are still a couple types errors I'm seeing:

## Error 1
```ts
export async function colors_mine({ miner, to, colors }: { miner: Address, to: Option<Address>, colors: Array<[u32, u32]> }, { signAndSend, fee }: { signAndSend?: boolean, fee?: number } = { signAndSend: false, fee: 100 }): Promise<void> {
    let invokeArgs: InvokeArgs = {
        signAndSend,
        fee,
        method: 'colors_mine',
        args: [((i) => addressToScVal(i))(miner),
        ((i) => (!i) ? xdr.ScVal.scvVoid() : addressToScVal(i))(to),
        ((i) => xdr.ScVal.scvVec(i.map((i) => [
            ((i) => xdr.ScVal.scvU32(i))(i[0]),
            ((i) => xdr.ScVal.scvU32(i))(i[1])
        ])))(colors)],
    };

    // @ts-ignore Type does exist
    const response = await invoke(invokeArgs);
    return;
}
```
Screenshot 2023-06-30 at 2 38 53 PM
```rust
pub fn colors_mine(env: &Env, miner: Address, to: Option<Address>, colors: Vec<(u32, u32)>)
```

## Error 2
```ts
export async function glyph_mint({ minter, to, colors, width }: { minter: Address, to: Option<Address>, colors: Array<[Address, Array<[u32, Array<u32>]>]>, width: u32 }, { signAndSend, fee }: { signAndSend?: boolean, fee?: number } = { signAndSend: false, fee: 100 }): Promise<Buffer> {
    let invokeArgs: InvokeArgs = {
        signAndSend,
        fee,
        method: 'glyph_mint',
        args: [((i) => addressToScVal(i))(minter),
        ((i) => (!i) ? xdr.ScVal.scvVoid() : addressToScVal(i))(to),
        ((i) => xdr.ScVal.scvVec(i.map((i) => [
            ((i) => addressToScVal(i))(i[0]),
            ((i) => xdr.ScVal.scvVec(i.map((i) => [
                ((i) => xdr.ScVal.scvU32(i))(i[0]),
                ((i) => xdr.ScVal.scvVec(i.map((i) => xdr.ScVal.scvU32(i))))(i[1])
            ])))(i[1])
        ])))(colors),
        ((i) => xdr.ScVal.scvU32(i))(width)],
    };

    // @ts-ignore Type does exist
    const response = await invoke(invokeArgs);
    return scValStrToJs(response.xdr) as Buffer;
}
```
Screenshot 2023-06-30 at 2 39 59 PM
```rust
pub fn glyph_mint(
    env: &Env,
    minter: Address,
    to: Option<Address>,
    colors: Vec<(Address, Vec<(u32, Vec<u32>)>)>,
    width: u32,
) -> BytesN<32>
```

## Error 3
```ts
function GlyphToXdr(glyph?: Glyph): xdr.ScVal {
    if (!glyph) {
        return xdr.ScVal.scvVoid();
    }
    let arr = [
        new xdr.ScMapEntry({
            key: ((i) => xdr.ScVal.scvSymbol(i))("colors"), val: ((i) => xdr.ScVal.scvVec(i.map((i) => [
                ((i) => addressToScVal(i))(i[0]),
                ((i) => xdr.ScVal.scvVec(i.map((i) => [
                    ((i) => xdr.ScVal.scvU32(i))(i[0]),
                    ((i) => xdr.ScVal.scvVec(i.map((i) => xdr.ScVal.scvU32(i))))(i[1])
                ])))(i[1])
            ])))(glyph["colors"])
        }),
        new xdr.ScMapEntry({ key: ((i) => xdr.ScVal.scvSymbol(i))("length"), val: ((i) => xdr.ScVal.scvU32(i))(glyph["length"]) }),
        new xdr.ScMapEntry({ key: ((i) => xdr.ScVal.scvSymbol(i))("width"), val: ((i) => xdr.ScVal.scvU32(i))(glyph["width"]) })
    ];
    return xdr.ScVal.scvMap(arr);
}
```
Screenshot 2023-06-30 at 2 40 56 PM

tyvdh:
Looks like we're just missing wrapping those bare arrays with a `xdr.ScVal.scvVec`.

willemneal:
@tyvdh Tuple structs should work now.  It will take some more work and thought to update to use @Shaptic, but I think my future efforts should go in the stellar-base not here. However, in the interest of getting this done, I just kept it here.

@paulbellamy @tsachiherman Currently all tests pass, except the fetch test which is the only one that uses futurenet and it fails with:

```
Rpc(JsonRpc(Call(ErrorObject { code: InvalidParams, message: "cannot unmarshal key value AAAABgAAAAG8B08PA5NNAYllO8Fa+agxcEEeEDtMSKY4iDBs+6QayAAAABQAAAABAAAAAA== at index 0", data: None })))
```

Installing a contract to futurenet also has the issue of `error: transaction simulation failed: Could not unmarshal transaction`.

My guess is that the RPC hasn't been updated yet.  But this makes me think we should improve the error messages if the version of the CLI differs from futurenet.

tsachiherman:
@willemneal , I think that for the purposes of our testing here, we shouldn't use any public network. These network conditions aren't deterministic and could cause non-useful results. Instead, we should run the futurenet tests after we've upgraded the network, as part of the smoke testing.

PR #741, opened 2023-07-03
updated rs-env and mono repo versions

sreuland:
### What

set cargo rs-sdk to 0.9.1 release version and pull in latest soroban xdr go lib from go:soroban-xdr-next

### Why

work with latest pre release versions


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/sirupsen/logrus v1.4.1
 	github.com/spf13/cobra v0.0.0-20160830174925-9c28e4bbd74e
 	github.com/spf13/pflag v1.0.5
-	github.com/stellar/go v0.0.0-20230623175925-531797f01226
+	github.com/stellar/go v0.0.0-20230613070852-e20e8bc1619c
```
tsachiherman, line 18: @sreuland this is an older branch.. I think that for the time being we should leave it with tip of tree of soroban-xdr-next-next on the go monorepo.

Cargo.toml
```diff
@@ -43,9 +43,9 @@ version = "0.8.7"
 path = "./cmd/crates/soroban-spec-tools"
 
 [workspace.dependencies.soroban-sdk]
-version = "0.8.4"
+version = "0.9.1"
 git = "https://github.com/stellar/rs-soroban-sdk"
-rev = "e90db9f8009280516375df6b3ee43a65ff524577"
+rev = "v0.9.1"
```
tsachiherman, line 48: could you use the hash here, just to be consistent with the rest of the dependencies ?

tsachiherman (changes requested):

go.mod
sreuland: oh, great catch, my mistake, I updated that to soroban-xdr-next by mistake, I updated this ref to use soroban-xdr-next-next to your point and that resolved the other xdr compile errors also.

Cargo.toml
sreuland: yes, changed that.

tsachiherman (approved):

PR #743, opened 2023-07-03
feat: long listing for identities

chadoh:
### What

`soroban config identity ls -l`, inspired by `soroban config network ls -l`.

Also add a `soroban config identity list` alias.

### Why

I originally expected it at `list`, hence the alias.

The long-listing just shows full path to identity file. I was trying to figure
out where some defined identities were coming from while troubleshooting.

### Known limitations

This is a bit simpler in implementation than the networks long-list. The
important information for identities is, in my opinion, just the path where the
identity file lives. So this just lists those. Happy to discuss the
implementation specifics, though!

Example:

    $ soroban config identity ls -l
    /Users/chadoh/code/soroban/dapp/.soroban/identity/token-admin.toml
    /Users/chadoh/.config/soroban/identity/default.toml

### DIFF
--- a/cmd/soroban-cli/src/commands/config/identity/ls.rs
+++ b/cmd/soroban-cli/src/commands/config/identity/ls.rs
@@ -12,11 +12,23 @@ pub enum Error {
 pub struct Cmd {
     #[command(flatten)]
     pub config_locator: locator::Args,
+    /// Get more info about the identities
+    #[arg(long, short = 'l')]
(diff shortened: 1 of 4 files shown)


paulbellamy (approved):

chadoh:
@paulbellamy looks like this change was lost; is that because I didn't target the preview-10 release branch? Should I create a new PR against `main`?

PR #749, opened 2023-07-05
soroban-cli: Add Auth-next signing support

### LINKED ISSUE #683: soroban-cli: update to support new auth and single host function
### What ?

soroban-cli: update all generated transactions structures

### Epic

https://github.com/stellar/soroban-tools/issues/657

paulbellamy:
Fixes #683 

Adds support for SorobanAuthorizationEntry (i.e. auth-next) multi-party signing in the soroban-cli.

### Usage

```bash
# Here, the local user account `alice` is used to sign the txn,
# but `bob` is the one who's value will be incremented.
#
# The contract method signature for this is:
#     pub fn increment(env: Env, user: Address, value: u32) -> u32 { ... }
$ soroban contract invoke \
    --source alice \
    --id $auth_example_contract \
    -- \
    increment \
    --user bob \
    --value 5
```
### Notes

To get this to work, I've had to change the rpc `simulateTransaction` endpoint. Now, if the txn includes any `SorobanAuthorizationEntries`, the simulation runs in enforcing auth mode. This means that the nonce the user sent will be used, generating the correct footprint and fees.

### Todo

- [x] Try it out locally and check it actually works
- [ ] Write a system-test test for this
- [ ] update the RPC web docs for `simulateTransaction` update
- [ ] create a follow-up issue to support passing keys directly on the CLI in the `S...` format


paulbellamy:
@dmkozh When I'm testing this, I keep getting TxInsufficientFee for any txn which involves auth-next signing...
For example: https://gist.github.com/paulbellamy/5b30db0bb3046c145c921a39f4145e8b
Is there something else we need to do?

dmkozh:
> We probably need to increase some fee for the size of the signatures right?

Yes, as we don't support enforcing preflight yet, there are some gaps in terms of what we can meter in recording mode. Signature verification requires some additional instructions (though IIUC this didn't get to the stage where this is enforced). Also tx size will be increased a bit due to signatures. This should only need a small fee increase though and IIRC preflight already adds about 15% on top of the computed fee. So not sure what else could be missing.

paulbellamy:
It could be that the transaction is quite small, so 15% is not enough for the extra signature in my case.

What is the formula for increasing the size due to the signatures. Is it just adding those bytes onto the `extended_meta_data_size_bytes`? or the `refundable_fee`? or the `fee`? or somewhere else?

dmkozh:
> What is the formula for increasing the size due to the signatures?

The increased tx size needs to be passed to the fee computation library. It will be reflected in the final non-refundable resource fee. To be more specific, it will be included in tx size fee and historical fee. But really if you want to compute fees outside of the rust library, you should rather port the whole library (https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/fees.rs)

cmd/soroban-cli/src/rpc/transaction.rs
```diff
@@ -84,6 +92,189 @@ pub fn assemble(
...
+) -> Result<(Transaction, Vec<SorobanAuthorizationEntry>), Error> {
+    let mut tx = raw.clone();
+
+    if tx.operations.len() != 1 {
+        // This must not be an invokeHostFunction operation, so nothing to do
+        return Ok((tx, Vec::new()));
+    }
+
+    let mut op = tx.operations[0].clone();
+    let OperationBody::InvokeHostFunction(ref mut body) = &mut op.body else {
```
dmkozh, line 154: FYI `CreateContract` calls now can also be signed via soroban auth, so this check is no longer correct..

cmd/soroban-cli/src/rpc/transaction.rs
```diff
@@ -84,6 +92,189 @@ pub fn assemble(
...
+                return Ok(auth);
+            };
+            let SorobanAddressCredentials { ref address, .. } = credentials;
+
+            // See if we have a signer for this authorizationEntry
+            // If not, then we Error
+            let needle = match address {
+                ScAddress::Account(AccountId(PublicKey::PublicKeyTypeEd25519(Uint256(ref a)))) => a,
+                ScAddress::Contract(Hash(c)) => {
+                    // This address is for a contract. There's no way to sign it because it is malformed.
```
dmkozh, line 184: FWIW contract address is not 'malformed', it just means that a custom account is supposed to be used. It's fine that we don't support this initially, but we probably should say that this is not supported yet, not that something is wrong

cmd/soroban-cli/src/rpc/transaction.rs
```diff
@@ -84,6 +92,189 @@ pub fn assemble(
...
+    let SorobanAuthorizationEntry {
+        credentials: SorobanCredentials::Address(ref mut credentials),
+        ..
+    } = auth else {
+        // Doesn't need special signing
+        return Ok(auth);
+    };
+    let SorobanAddressCredentials { nonce, .. } = credentials;
+
+    let preimage = HashIdPreimageSorobanAuthorization {
```
dmkozh, line 236: This should be wrapped into `HashIdPreimage::SorobanAuthorization` (thisi is to ensure we have a seprate 'namespace' for all the signed preimages core might deal with.

cmd/soroban-cli/src/rpc/transaction.rs
paulbellamy: Is this still true, because CreateContract is a type of InvokeHostFunction now?

cmd/soroban-cli/src/rpc/transaction.rs
```diff
@@ -78,12 +79,160 @@ pub fn assemble(
...
+        invocation: auth.root_invocation.clone(),
+        nonce: *nonce,
+        signature_expiration_ledger,
+    })
+    .to_xdr()?;
+
+    let payload = Sha256::digest(preimage);
+    let signature = signer.sign(&payload);
+
+    let map = ScMap::sorted_from(vec![
```
Shaptic, line 204: why sorted?

dmkozh: Only sorted `ScMap`s are considered valid by the Soroban host.

Shaptic: oh what?? good to know :memo: 

but the keys are sorted here anyway so it'd be a waste

cmd/soroban-cli/src/rpc/transaction.rs
dmkozh: It has always been a type of invoke host fn, but now it uses the auth payload instead of bespoke authorization mechanism.

cmd/soroban-cli/src/rpc/transaction.rs
dmkozh: In general, `ScMap`s should be created with `sorted_from`. I'm not even sure if we provide a different constructor. We shouldn't try to optimize something as marginal as sorting a couple of keys, while risking to break compatibility with the host.

Also, I would be much happier if we didn't build `ScVal`s manually at all and used SDK here. That would come with some additional deps and somewhat higher costs, but the benefit would be guaranteed compatibility with host and much more natural structure building. Maybe we should think how we can expose SDK+host for the Rust clients without bringing all the deps...

cmd/soroban-rpc/lib/preflight/src/lib.rs
```diff
@@ -122,14 +123,31 @@ fn preflight_invoke_hf_op_or_maybe_panic(
     })?;
     let host = Host::with_storage_and_budget(storage, budget);
 
-    host.switch_to_recording_auth()?;
+    let needs_auth_recording = invoke_hf_op.auth.is_empty();
+    if needs_auth_recording {
+        host.switch_to_recording_auth()?;
+    } else {
+        host.set_authorization_entries(invoke_hf_op.auth.to_vec())?;
+    }
```
2opremio, line 131: I think a comment here is in order, explaining why we need enforcing mode etc ...

sisuresh (approved):

PR #750, opened 2023-07-05
compile: disable the fetch test

tsachiherman:
### What

Disable a failing test - fetch

### Why

The test is running against futurenet, which wasn't upgraded yet.
This test should be moved to post-deployment testing rather then in here.

### Known limitations

n/a

### DIFF
--- a/cmd/crates/soroban-spec-typescript/src/boilerplate.rs
+++ b/cmd/crates/soroban-spec-typescript/src/boilerplate.rs
@@ -125,6 +125,10 @@ mod test {
         Ok(p)
     }
 
+    // TODO : fix the test below :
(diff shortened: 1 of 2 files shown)


cmd/crates/soroban-spec-typescript/src/boilerplate.rs
```diff
@@ -125,6 +125,10 @@ mod test {
         Ok(p)
     }
 
+    // TODO : fix the test below :
```
sreuland, line 128: should this mention addressing the dependency on futurenet also, the pr's `Why` refers to this as the main reason test is not repeatable?

sreuland (approved):

PR #751, opened 2023-07-06
Fix bug in cli lab xdr decode --type parsing

paulbellamy:
### What

Fix `--type` flag parsing in `soroban xdr lab decode` subcommand.

### Why

I was hitting it when trying to debug other failing transactions. Turns out we need a `try_map` to tell clap how to convert the string into a `TypeVariant`. No idea how it ever worked tbh. Maybe some subtle xdr upgrade change broke it.

Previously you would get:
```
thread 'main' panicked at 'Mismatch between definition and access of `type`. Could not downcast to stellar_xdr::next::generated::TypeVariant, need to downcast to alloc::string::String
', cmd/soroban-cli/src/commands/lab/xdr/decode.rs:9:13
```

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-cli/src/commands/lab/xdr/decode.rs
+++ b/cmd/soroban-cli/src/commands/lab/xdr/decode.rs
@@ -1,11 +1,21 @@
-use clap::{arg, Parser, ValueEnum};
+use clap::{
+    arg,
+    builder::{PossibleValuesParser, TypedValueParser},
+    Parser, ValueEnum,
+};
+use core::str::FromStr;
 use soroban_env_host::xdr;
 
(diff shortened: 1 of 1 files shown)


paulbellamy:
@leighmcculloch not sure if there is a more elegant way to do this. (Or how it ever worked before...)

tsachiherman (approved):

leighmcculloch:
> @leighmcculloch not sure if there is a more elegant way to do this. (Or how it ever worked before...)

Did we update the version of clap we use? The way enums work definitely changed to be more verbose in a more recent version of clap. No more elegant way as far as I know.

Issue #734, opened 2023-06-30
`signAndSend: false` should still return the `simulateTransaction` results


tyvdh:
If `signAndSend` is `false` function call promises return undefined. They should return the simulateTransaction RPC response. Otherwise you can't really make read only calls. Unless I'm missing something. Either way requiring signing in order to see the simulation results greatly slows down the dev experimentation process.

chadoh:
`signAndSend: false` does currently return the simulated _results_, though not the full RPC response. Maybe you tried it with a smart contract function that returns `undefined`?

Let's vidchat about the best ux around returning results directly vs full rpc response!

tyvdh:
That is possible and the two functions I have that do return something are hitting this https://github.com/stellar/soroban-tools/issues/739 so not really able to test this on the contract I've been working with so I'll take your word for it.

janewang:
Closing as sign and simulate are decoupled

PR #753, opened 2023-07-06
feat(typescript): `invoke` improvements

chadoh:
### What

- remove `signAndSend`: if the simulation requires `auth`, then send automatically by default
- hoist `secondsToWait` as an option to all exported functions so that it can actually be used. In addition, if `secondsToWait` is `0`, then do not go into the `while` loop that calls `getTransaction` at all, and instead just return the `sendTransaction` results.
- If `sendTransaction` has any status other than `PENDING`, don't go into the `while` loop; just return the `sendTransaction` results.
- add a `responseType` option: by default this is undefined, which keeps the previous behavior: attempt to parse the returned XDR as the correct type for the exported function (so `balance` parses it as a `bigint`). But now you can set `responseType` to `'simulated'` to have the function only simulate the transaction, even if it requires auth/signing. And you can set it to `full` to get the full RPC results, whether a Simulation (for view/read methods) or a SendResponse (something went wrong with `sendTransaction`, or you set `secondsToWait: 0`) or a GetResponse.
- add `parseResultXdr` function argument: this cleans up the logic, since the result XDR lives in different places depending on whether this is a simulation, an awaited full transaction that uses `getTransaction`, or a non-awaited full transaction that returns faster with just the `sendTransaction` results.

Extensive care has been given to ensure that all the returned types are what you'd expect. So if you use what will probably be the most common function signature, omitting the `responseType` argument altogether such as when you say `balance({ address: 'GAAAA' })`, the type of the value is `bigint`, not a nasty intersection you need to untangle yourself like `bigint | Simulation | SendTx | GetTx`. This required some strange-looking generics (`<R extends ResponseTypes = undefined>`), but seems worth it.

### Why

This is what @tyvdh and I came up with as a result of our conversation about #734

### Known limitations

- Is the logic that replaces `signAndSend` robust enough? It feels like `signAndSend` was a half-measure in which logic was pushed to dapp developers, when this logic can be handled by the generated library instead. The current logic simply checks for the existence and length of an `auth` array. Is that sufficient?

   Even if not, this change makes it easier to troubleshoot functions that don't work as expected.

- We may want to add tests for the inferred types. Maybe this could go in the generated libraries in a `test.ts` file? Something like this would work:

       const stringResult: string = await invoke({ method: 'test' })
       const numberResult: number = await invoke({ method: 'test', parseResultXdr: () => 1 })
       const simulationResult: Simulation = await invoke({ method: 'test', responseType: 'simulated' })
       const rpcResult: SomeRpcResponse = await invoke({ method: 'test', responseType: 'full' })
       console.debug('invoke type-checks; hooray', { stringResult, numberResult, simulationResult, rpcResult })

   Then we build `test.ts` and make sure TypeScript doesn't complain.

   I'm not sure this is worth adding yet.


willemneal:
Other thing to note. Change calls (ei a write to the state) don't have to require auth.  so we should also look at the footprint to decide if there was a write.

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
```diff
@@ -2,8 +2,8 @@ import * as SorobanClient from 'soroban-client';
 import { xdr } from 'soroban-client';
 import { Buffer } from "buffer";
 import { scValStrToJs, scValToJs, addressToScVal, u128ToScVal, i128ToScVal, strToScVal } from './convert.js';
-import { invoke, InvokeArgs } from './invoke.js';
-
+import { invoke } from './invoke.js';
```
Shaptic, line 5: nit: you can drop the `.js` here to match the subsequent line

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
```diff
@@ -231,392 +231,807 @@ function ComplexEnumFromXdr(base64Xdr: string): ComplexEnum {
...
-    } catch (e) {
-        //@ts-ignore
-        let err = getError(e.message);
-        if (err) {
-            return err;
-        } else {
-            throw e;
-        }
-    }
+export async function hello<R extends ResponseTypes = undefined>({hello}: {hello: string}, options: {
```
Shaptic, line 234: would be super cool to run e.g. `prettier` on this code so that it's easier for consumers to grok and read docs inline (plus would help clean up diffs etc.), but I'm not sure how annoying that would be to add to the pipeline

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
```diff
@@ -231,392 +231,807 @@ function ComplexEnumFromXdr(base64Xdr: string): ComplexEnum {
...
+   *   - `'full'` return the full RPC response, meaning either 1. the preflight info, if it's a view/read method that doesn't require auth/signing, or 2. the `sendTransaction` response, if there's a problem with sending the transaction or if you set `secondsToWait` to 0, or 3. the `getTransaction` response, if it's a change method with no `sendTransaction` errors and a positive `secondsToWait`.
+   */
+  responseType?: R
+  /**
+   * If the simulation shows that this invocation requires auth/signing, `invoke` will wait `secondsToWait` seconds for the transaction to complete before giving up and returning the incomplete {@link SorobanClient.SorobanRpc.GetTransactionResponse} results (or attempting to parse their probably-missing XDR with `parseResultXdr`, depending on `responseType`). Set this to `0` to skip waiting altogether, which will return you {@link SorobanClient.SorobanRpc.SendTransactionResponse} more quickly, before the transaction has time to be included in the ledger. Default: 10.
+   */
+  secondsToWait?: number
+} = {}) {
```
Shaptic, line 251: can this `options` member be its own type? so that we can separate out the structure/typedefs/docs from the actual method signature

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/invoke.ts
```diff
@@ -41,22 +31,40 @@ async function getAccount(): Promise<Account | null> {
...
  * Invoke a method on the test_custom_types contract.
  *
  * Uses Freighter to determine the current user and if necessary sign the transaction.
  *
- * @param {string} obj.method - The method to invoke.
- * @param {any[]} obj.args - The arguments to pass to the method.
- * @param {boolean} obj.signAndSend - Whether to sign and send the transaction, or just simulate it. Unless the method requires authentication.
- * @param {number} obj.fee - The fee to pay for the transaction.
- * @returns The transaction response, or the simulation result if signing isn't required.
+ * @returns T, by default, the parsed XDR from either the simulation or the full transaction. If `simulateOnly` or `fullRpcResponse` are true, returns either the full simulation or the result of sending/getting the transaction to/from the ledger.
```
Shaptic, line 53: nit: should be `@returns {T}  by default, <etc...>`, right?

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/invoke.ts
```diff
@@ -70,28 +78,31 @@ export async function invoke({ method, args = [], fee = 100, signAndSend = false
...
-  const auths = simulated.results?.[0]?.auth
-  let auth_len =  auths?.length ?? 0;
-
-  if (auth_len > 1) {
+  // ^^^^ else, is CHANGE method ˅˅˅˅
+  if (authsCount > 1) {
     throw new NotImplementedError("Multiple auths not yet supported")
-  } else if (auth_len == 1) {
+  }
+  if (authsCount === 1) {
```
Shaptic, line 105: seems relevant: https://github.com/stellar/js-stellar-base/pull/632

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/invoke.ts
```diff
@@ -101,17 +112,30 @@ export async function invoke({ method, args = [], fee = 100, signAndSend = false
...
+
+  // if `sendTx` awaited the inclusion of the tx in the ledger, it used
+  // `getTransaction`, which has a `resultXdr` field
+  if ('resultXdr' in raw) return parse(raw.resultXdr)
+
+  // otherwise, it returned the result of `sendTransaction`
+  if ('errorResultXdr' in raw) return parse(raw.errorResultXdr)
+
+  // if neither of these are present, something went wrong
+  console.log("Don't know how to parse result! Returning fullRpcResponse")
```
Shaptic, line 137: suggestion: `console.error`

Shaptic (commented):
I know this is a draft but I just wanted to leave a handful of comments. I can't really comment on the usability of this revamp but I'm assuming this is in line with community feedback!

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
willemneal: The idea here was that doing that doesn't allow the options to appear with the IDE type hover info.  This way devs don't need to look up a type.  And we figured it's pretty much the same since we are generated it anyway.

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/invoke.ts
willemneal: Fair point. Never knew you could reference a generic in a type doc.

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
willemneal: Hmm I remember this being necessary for some targets. So now I'm wondering why the next line worked.

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
willemneal: Yeah we definitely considered it but wanted to keep it simple for the first iteration. Perhaps with https://github.com/rome/tools

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
Shaptic: Hmm true, but it does make the generated code annoying to read (though I guess most people wouldn't do that?). Maybe it's better as a jsdoc on the function itself? i.e. something like

```ts
/**
 * @param {OptionsType} [options]
 * @param {number} [options.fee]  - the fee to pay for the transaction (default: 100). 
 * @param {string} [options.responseType] - the type of response to return:  
 *     - `undefined`, the default, parses the returned XDR as `SimpleEnum`. 
 *       Runs preflight, checks to see if auth/signing is required, and sends the 
 *       transaction if so. If there's no error and `secondsToWait` is positive, 
 *       awaits the finalized transaction.
 *   - `'simulated'` will only simulate/preflight the transaction, even if it's a 
 *      change/set method that requires auth/signing. Returns full preflight info.
 *   - `'full'` return the full RPC response, meaning either 1. the preflight info, 
 *     if it's a view/read method that doesn't require auth/signing, or 
 *     2. the `sendTransaction` response, if there's a problem with sending 
 *     the transaction or if you set `secondsToWait` to 0, or 3. the 
 *     `getTransaction` response, if it's a change method with no `sendTransaction` 
 *     errors and a positive `secondsToWait`
 * @param <and so on...>
 **/
export async function simple<R extends ResponseTypes = undefined>({simple}: {simple: SimpleEnum}, options: { ...
```

or is that **also** incompatible with type-hovers?

Anyway, the code/docs are the important part, less-so where they live. I'm just offering suggestions on readability, but the dev UX is far more crucial, so if this accomplishes that, so be it!

Shaptic: In a separate vein: there should still probably be an `enum ResponseType` for that parameter to not rely on the "magic strings" `'full'` etc.

cmd/crates/soroban-spec-typescript/src/lib.rs
```diff
@@ -107,85 +145,86 @@ pub fn entry_to_ts(entry: &Entry) -> String {
...
-                r#"{ts_doc}export async function {name}({input} {{signAndSend, fee}}: {{signAndSend?: boolean, fee?: number}} = {{signAndSend: false, fee: 100}}){return_type} {{
-    let invokeArgs: InvokeArgs = {{
-        signAndSend,
-        fee,
-        method: '{name}', 
-        {args}
-    }};
-    {output}
+                r#"{ts_doc}export async function {js_name}<R extends ResponseTypes = undefined>({input}options: {options} = {{}}) {{
+    return await invoke({{
```
Shaptic, line 223: can/should this return a `Promise`, instead? unless there's a usability reason to enforce this to be synchronous

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
chadoh: The next line probably doesn't work with NextJS; thanks for noticing!

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
chadoh: I think part of the problem here is that we need to add everything in this `fixtures` folder to our [.gitattributes](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) file so that it doesn't show up in diffs 😅 

The idea is as Willem said: maximize typeahead usability. Unfortunately, TypeScript doesn't give us a way to "expand" definitions of a type like `enum ResponseType` or `Options`, so people would just see those when they type functions in their editor:

image

Then they'd have to go to the definition (if they know how) or look it up the docs online.

cmd/crates/soroban-spec-typescript/src/lib.rs
chadoh: It does return a Promise, because it is `async`. 

If you don't await the `Promise` returned by `invoke`, you can end up with a double-wrapped Promise. At least, I've seen unexpected things happen, when I don't await inner calls like this.

cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
Shaptic: Bah okay fair points, thanks for humoring me. The whole point is giving them usable types to work with, so let's continue to maximize that :+1:

cmd/crates/soroban-spec-typescript/src/lib.rs
Shaptic: > be me
> js sdk maintainer for like 2 years
> still don't fully understand async/await semantics

thanks for elucidating :laughing:

cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts
```diff
@@ -100,18 +115,30 @@ export async function invoke({ method, args = [], fee = 100, signAndSend = false
...
+
+    tx = await signTx(
+      SorobanClient.assembleTransaction(tx, NETWORK_PASSPHRASE, simulated) as Tx
+    );
   }
 
-  tx = await signTx(
-    SorobanClient.assembleTransaction(tx, NETWORK_PASSPHRASE, simulated) as Tx
-  );
+  const raw = await sendTx(tx, secondsToWait);
```
chadoh, line 128: @paulbellamy in an attempt to separate the `sign` and `send` logic, lines 119-125 are now inside the `if (authsCount === 1)` block. Signing & sending, as you'd expect, works just fine. But functions that require `send` without `sign` currently fail with:

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32603,
        "message": "received exception from stellar-core"
    }
}
```

I get this when using the version of the CLI on this branch if I add a silly `become_the_recipient` function to the crowdfund contract. Here's the XDR that gets generated for the `sendTransaction` call: `AAAAAgAAAADSu8MJFOWfcDY1dw+DLEWhCcSrtuevg0onMIUgLsCmswAAAGQAAMFmAAAABQAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAGAAAAAAAAAADAAAAEgAAAAE5Md3nU7e/TCOZn0TipZhXxP3sBVYQT0rLRF9Atiz2qQAAAA8AAAAUYmVjb21lX3RoZV9yZWNpcGllbnQAAAASAAAAAAAAAADSu8MJFOWfcDY1dw+DLEWhCcSrtuevg0onMIUgLsCmswAAAAAAAAAAAAAAAA==`

I can give you more info if you need it, but I'm wondering if it's already obvious to you, based on the code above.

chadoh: I confirmed that if I add an unnecessary signature, the transaction succeeds. So it has something to do with how I'm creating this transaction envelope without a signature.

chadoh: OK, transaction envelope always needs to be signed. That's how the signer pays for the transaction.

cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts
```diff
@@ -41,22 +31,40 @@ async function getAccount(): Promise<Account | null> {
 
 export class NotImplementedError extends Error { }
 
+type Simulation = SorobanClient.SorobanRpc.SimulateTransactionResponse
```
Shaptic, line 34: wait, how does this work if `SorobanClient` isn't default-imported?

Shaptic: ideally we would avoid the default-import to keep bundles slimmer (i.e. only import what we need)

Shaptic (approved):
sick :clap: just some minor comments below :arrow_down:

cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts
chadoh: Great point! Let's do that in a separate PR, though, since this one only maintained the existing import structure

willemneal: +1 always hated default imports.

chadoh: @Shaptic great point! Let's do this in a follow-up

PR #756, opened 2023-07-07
soroban-cli: Implement soroban contract read with real rpc support

paulbellamy:
### What

Add support for talking to soroban-rpc to the `soroban contract read ...` subcommand.

### Why

Previously it only supported against local sandbox ledger.json. But that's not super useful, and we have the new `getLedgerEntries` method we can use.

I wanted this when I was trying to test/debug the `soroban contract bump` and `restore` commands in https://github.com/stellar/soroban-tools/pull/748 (which this is based on)

### Known limitations

[TODO or N/A]


willemneal:
Yay! With some modifications we can also have the sandbox lazily read from RPC.

paulbellamy:
Yeah, the design here needs some work generally. Currently it can only easily read symbol-keyed storage, and only temporary or permanent, not instance storage. And it doesn't show you the remaining expiry on the entry. But it's definitely an improvement as-is.

tsachiherman (approved):

PR #762, opened 2023-07-11
update e2e tests

tsachiherman:
### What

### Why

### Known limitations

[TODO or N/A]


cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/package.json
```diff
@@ -3,9 +3,9 @@
     "name": "test_custom_types",
     "dependencies": {
         "@stellar/freighter-api": "1.5.0",
+        "bigint-conversion": "2.4.1",
         "buffer": "6.0.3",
-        "soroban-client": "git+https://github.com/stellar/js-soroban-client.git#8554b7aff30c91beca7aa89393a81cac9e29894e",
-        "bigint-conversion": "2.4.1"
+        "soroban-client": "github:stellar/js-soroban-client#v0.9.0"
```
Shaptic, line 8: you should be able to use the npm tag here, i.e. just `"soroban-client": "0.9.0"`

tsachiherman: I think that I'm going to rollback my package changes and merge my PR as is. It won't fix all the issues, but it would move it toward the right direction.

PR #767, opened 2023-07-12
soroban-cli: Add better feedback to bump+restore ops

paulbellamy:
### What

Have bump+restore subcommands either output the new expiration ledger seq, or that the entry could not be bumped.

### Why

Previously there was no way to tell if it was successful or not since the core txn is always successful.

### Known limitations

Doesn't really differentiate between entry not found and the entry not being bumped. We could check that by querying the rpc after to see if it exists.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/bump.rs
+++ b/cmd/soroban-cli/src/commands/contract/bump.rs
@@ -6,7 +6,7 @@ use soroban_env_host::xdr::{
     LedgerEntry, LedgerEntryData, LedgerFootprint, LedgerKey, LedgerKeyContractData, Memo,
     MuxedAccount, Operation, OperationBody, Preconditions, ReadXdr, ScAddress, ScSpecTypeDef,
     ScVal, SequenceNumber, SorobanResources, SorobanTransactionData, Transaction, TransactionExt,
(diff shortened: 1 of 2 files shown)


cmd/soroban-cli/src/commands/contract/bump.rs
```diff
@@ -144,10 +151,30 @@ impl Cmd {
...
+        // entry via this command (which we should fix separately, but).
+        if operations.len() == 0 {
+            return Err(Error::LedgerEntryNotFound);
+        }
+
+        if operations[0].changes.len() != 2 {
+            return Err(Error::LedgerEntryNotFound);
+        }
+
+        let (LedgerEntryChange::State(_state), LedgerEntryChange::Updated(LedgerEntry{ data: LedgerEntryData::ContractData(ContractDataEntry{expiration_ledger_seq, ..}), ..})) = (&operations[0].changes[0], &operations[0].changes[1]) else {
```
Shaptic, line 170: what in the abomination is this un `cargo fmt`d line

paulbellamy: It actually *is* cargo fmt'ed

PR #764, opened 2023-07-11
Upgrade generated TypeScript with Preview 10 changes

Shaptic:
### What
This is a first pass attempt at improving the generated TypeScript code to leverage the breaking changes and new features in `soroban-client@0.9.0` (see the [Release Notes](https://github.com/stellar/js-soroban-client/releases/tag/v0.9.0)). I tried to keep it minimal:

 * remove obsolete `scvalToBigInt` (in favor of `SorobanClient.scValToBigInt`)
 * remove bitwise manipulations in `i128ToScVal` and `u128ToScVal` (in favor of `SorobanClient.ScInt`)
 * this drops the need for the `bigint-conversion` library
 * ran the fixture update and updated other deps

I would even argue that `scValToJs` could be replaced by `SorobanClient.scValToNative`, but that would require extensive testing so I'm hesitant to do it here.

### Why
The generated code should be as simple and idiomatic as possible.

### Known limitations
n/a

**Note:** The main "functional" diff should mainly be `cmd/crates/soroban-spec-typescript/src/project_template/src/convert.ts` ([link](https://github.com/stellar/soroban-tools/pull/764/files#diff-157cc5d7c8ff4d63f8c66f141d63258c092d3055ffd090d8489637ab8c0446d6)).

PR #770, opened 2023-07-12
 Upgrade generated TypeScript with Preview 10 changes

Shaptic:
Replaces #764 to target `main`.

The only "real" change is to src/convert.ts [here](https://github.com/stellar/soroban-tools/pull/770/files#diff-157cc5d7c8ff4d63f8c66f141d63258c092d3055ffd090d8489637ab8c0446d6); the rest is just generated code.

cc @willemneal @chadoh

### DIFF
--- a/.github/workflows/soroban-rpc.yml
+++ b/.github/workflows/soroban-rpc.yml
@@ -61,7 +61,9 @@ jobs:
         with:
           go-version: 1.20.1
 
-      - run: rustup target add ${{ matrix.rust_target }}
+      - run: |
+          rustup target add ${{ matrix.rust_target }}
+          rustup update
 
       # On windows, make sure we have the same compiler (linker) used by rust.
       # This is important since the symbols names won't match otherwise.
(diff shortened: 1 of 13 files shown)


paulbellamy (approved):

.github/workflows/soroban-rpc.yml
```diff
@@ -61,7 +61,9 @@ jobs:
         with:
           go-version: 1.20.1
 
-      - run: rustup target add ${{ matrix.rust_target }}
+      - run: |
+          rustup target add ${{ matrix.rust_target }}
+          rustup update
```
sreuland, line 66: do you know where/when rustup is installed initially, only ask b/c depending on how it's installed, the self `update` may not be allowed - https://rust-lang.github.io/rustup/basics.html#keeping-rustup-up-to-date, but doubt that will ever be case here, this seems fine. the other test container setups I've seen so far, they just download rustup with `--default-toolchain` with equates to the latest `stable` which is same as `update`

Shaptic: I had this same question - I initially assumed I should be able to tweak the installed version in some `install-rust`esque action, but no such luck?? I guess it's part of `ubuntu-latest`, which is surprising. I assume this will self-resolve if the image bumps its installed version, too? But since this is a "good enough" solution I'm just gonna merge.

PR #779, opened 2023-07-18
fix CI bug in soroban-rpc testing

tsachiherman:
### What

fix CI bug in soroban-rpc testing

### Why

trivial fix.

### Known limitations

n/a


Shaptic (approved):
why not make it the same line? don't need the `|`. but if it works it works :+1:

PR #798, opened 2023-07-24
soroban-rpc: bump XDR to latest changes (including Go/Rust dependencies and Core version)

2opremio:
Note that:

* This is targetting ` soroban-xdr-next`  and not `main`  in case we want to make more corrections for the P10 release
* This includes a hack in the preflight computation (suggested by @dmkozh ) to move us forward before rent fees are included. The hack consists in artificially bumping the rent fee and minimum suggested fee by 10000.


tsachiherman (approved):
one comment, looks good otherwise.

PR #799, opened 2023-07-24
soroban-rpc: limit number of concurrent requests

tsachiherman:
### What

limit number of concurrent requests

### Why

We need to ensure that the soroban-rpc doesn't build up unreasonable queue of incoming requests.

### Known limitations

n/a


2opremio:
I think that having a setting per request type is overkill. Isn't it enough to set a limit for all request types?

Note that queuing should only be temporary otherwise the service won't be able to keep up anyways.

tsachiherman:
> I think that having a setting per request type is overkill. Isn't it enough to set a limit for all request types?
> 
> Note that queuing should only be temporary otherwise the service won't be able to keep up anyways.

The goal of this PR is to allow us to build the same degree of resiliency we currently have with horizon.
As you recall from the last high-load event, we have been throttling the request rate for a specific method
so that the overall functionality won't get negatively affected.

With the soroban-rpc, we won't be able to do so. The rate control methods we've used for horizon would only
work on the http level. We need to develop our own rpc-level controls. That's exactly what this PR attempts to
achieve. In addition, note that this PR doesn't create any buffers or queues of it's own. It only limits the existing
concurrent execution of rpc methods.

As for the urgency of developing this - I agree that it might seems to be an overkill. However, unlike horizon, the soroban-rpc is a critical piece of infrastructure. Without it, there will be no soroban transactions. Therefore, we have no true choice here - if we need to develop all the control to ensure that we can adapt to any load scenario that we might run into.

sreuland:
> The goal of this PR is to allow us to build the same degree of resiliency we currently have with horizon.

is it feasible to consider dynamic thresholds for resiliency from the rpc based on checking host compute resource thresholds like cpu or ram percentages rather than explicit http max request settings. It could be cumbersome for admins to keep these settings accurate based on the target host deployments and possibly duplicative as deployments may have an external front layer of proxy or content switches applying similar rate limiting rules. The end result ideally provides the same type of response from rpc HTTP endpoints, maybe 503 is better error code in that case for 'server overload' and distinguishes it from 429's that other layers may be responsible for.

tsachiherman:
> > The goal of this PR is to allow us to build the same degree of resiliency we currently have with horizon.
> 
> is it feasible to consider dynamic thresholds for resiliency from the rpc based on checking host compute resource thresholds like cpu or ram percentages rather than explicit http max request settings. It could be cumbersome for admins to keep these settings accurate based on the target host deployments and possibly duplicative as deployments may have an external front layer of proxy or content switches applying similar rate limiting rules. The end result ideally provides the same type of response from rpc HTTP endpoints, maybe 503 is better error code in that case for 'server overload' and distinguishes it from 429's that other layers may be responsible for.

My hope is that admins would not need to change any of these. These settings are created as emergency handlers to allow us to tweak the performance characteristics of the already-deployed soroban-rpc without making any code changes.
If you'll look carefully, you'll see that this implementation protects ( to a degree ) against draining the soroban-rpc resource utilization by quickly suspending incoming traffic that it wouldn't be able to handle otherwise.

Also, I agree with you that 503 seems to be more adequate for this, since it indicate the it's the *server* that is out of resources rather then the *user* that have used all their resources.

2opremio:
@tsachiherman can you rebase?

sreuland:
> My hope is that admins would not need to change any of these. These settings are created as emergency handlers to allow us to tweak the performance characteristics of the already-deployed soroban-rpc without making any code changes.

ok, with fixed max request limits as emergency valves, can that lead un-necessarily to capping of the throughput of rpc server on some deployments where the host's cpu/ram is large enough to allow for more simultaneous requests?

tsachiherman:
> > My hope is that admins would not need to change any of these. These settings are created as emergency handlers to allow us to tweak the performance characteristics of the already-deployed soroban-rpc without making any code changes.
> 
> ok, with fixed max request limits as emergency valves, can that lead un-necessarily to capping of the throughput of rpc server on some deployments where the host's cpu/ram is large enough to allow for more simultaneous requests?

Yes, it can. However, the point where it would be limiting the throughput would be by two order of magnitude greater then the core network capacity..
Example : we're limiting concurrent tx simulations to 50 at a time.
If each such simulation takes 10ms, we're blocking the rpc to support 5000 tx/sec. The network capacity is nowhere that large.
Also, note that core still need to process these transaction sequentially whereas the rpc is using multiple cores.

If we'll get to a point where this is an issue - a simple solution is to adjust the limit upward.

cmd/soroban-rpc/internal/config/options.go
```diff
@@ -271,6 +271,69 @@ func (cfg *Config) options() ConfigOptions {
 			DefaultValue: uint(runtime.NumCPU()),
 			Validate:     positive,
 		},
+		{
+			Name:         "request-backlog-global-queue-limit",
+			Usage:        "Maximum number of outstanding requests",
+			ConfigKey:    &cfg.RequestBacklogGlobalQueueLimit,
+			DefaultValue: uint(1000),
```
sreuland, line 278: is there any room for larger default as it's an aggregate of all requests, like 5000 etc?

tsachiherman: I'm definitely open to modify these baked-in constants. would it be ok if I'll modify that number in the follow-up PR ?

tsachiherman: All these defaults, from my perspective, are a starting point values that we could adjust over time.

cmd/soroban-rpc/internal/network/backlogQ_test.go
```diff
@@ -0,0 +1,222 @@
...
+		for i := queueSize / 2; i < queueSize; i++ {
+			go func() {
+				limiter.ServeHTTP(nil, nil)
+			}()
+		}
+		secondBlockingGroupWg.Wait()
+		// now, try to place additional entry - which should be blocked.
+		var res TestingResponseWriter
+		limiter.ServeHTTP(&res, nil)
+		require.Equal(t, http.StatusServiceUnavailable, res.statusCode)
```
sreuland, line 143: great coverage!

cmd/soroban-rpc/internal/network/backlogQ.go
```diff
@@ -0,0 +1,107 @@
...
+		backlogQLimiter: backlogQLimiter{
+			limit:  limit,
+			gauge:  gauge,
+			logger: logger,
+		},
+	}
+}
+
+func (q *backlogHTTPQLimiter) ServeHTTP(res http.ResponseWriter, req *http.Request) {
+	if newPending := atomic.AddUint64(&q.pending, 1); newPending > q.limit {
```
sreuland, line 55: just for consideration, allowing `0` for limits which would disable limit checking? as way for deployments to disable this in some cases..but not sure exactly when that would come up..

tsachiherman: yes, exactly. Explicitly setting the value to 0 would allow us to disable the endpoint.
( the default is non-zero, so the user would need to explicitly have it set to zero in order to disable the endpoint )

tsachiherman: In the next pr, I'll be adding a pass-through, which would allow the end user to completely skip the limiters.

cmd/soroban-rpc/internal/config/options.go
sreuland: yes, sounds good, I was going to ask if any prior benchmarking was referenced for these initial defaults, and understand this is just first iteration, thanks!

sreuland (approved):
great work!

PR #800, opened 2023-07-24
fix: Enums with tuple of any length now allowed and example is updated

willemneal:
Added a new test to cover this case.

### What

### Why

### Known limitations

[TODO or N/A]


cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -27,6 +27,8 @@ pub enum Error {
     EnumCase(String, String),
     #[error("Enum {0} missing value for type {1}")]
     EnumMissingSecondValue(String, String),
+    #[error("Enum {0} is informed ")]
```
tsachiherman, line 30: typo in `informed` -> `illformed`. Also, drop the space at the end.

tsachiherman (changes requested):

tsachiherman:
also, please review the test failures on the CI - we need to get there resolved before merging this PR.

tsachiherman (approved):

PR #804, opened 2023-07-26
soroban-rpc: limit the execution duration of the jrpc requests

tsachiherman:
### What

Limit the execution duration of the jrpc requests.

### Why

Limiting the execution duration of the jrpc requests allow us to ensure reasonable performance characteristics and provide means for the hosting providers to plan for a desired SLA goals.

### Known limitations

The implementation would still require the underlying handlers to respect the passed-in context, and abort the processing as soon as possible. This is not covered in this PR, and will be handled in a subsequent PR.

### Additional notes

- The unit tests for the backlog request limiter were extensively enhanced 
- The limiters configuration parameters were restricted to the TOML configuration file to avoid cluttering the environment variable and the command line interface.


cmd/soroban-rpc/internal/config/options.go
```diff
@@ -275,7 +275,7 @@ func (cfg *Config) options() ConfigOptions {
 			Name:         "request-backlog-global-queue-limit",
 			Usage:        "Maximum number of outstanding requests",
 			ConfigKey:    &cfg.RequestBacklogGlobalQueueLimit,
-			DefaultValue: uint(1000),
+			DefaultValue: uint(5000),
```
tsachiherman, line 278: per previous code review, I'm upping this to 5000.

cmd/soroban-rpc/internal/config/options.go
```diff
@@ -272,68 +273,128 @@ func (cfg *Config) options() ConfigOptions {
...
-			Name:         "request-backlog-simulate-transaction-queue-limit",
+			TomlKey:      strutils.KebabToConstantCase("request-backlog-simulate-transaction-queue-limit"),
 			Usage:        "Maximum number of outstanding SimulateTransaction requests",
 			ConfigKey:    &cfg.RequestBacklogSimulateTransactionQueueLimit,
 			DefaultValue: uint(100),
 			Validate:     positive,
 		},
+		{
+			TomlKey:      strutils.KebabToConstantCase("request-execution-warning-threshold"),
+			Usage:        "The request execution warning threshold is the predetermined maximum duration of time that a request can take to be processed before a warning would be generated",
```
sreuland, line 340: can this state where the warning is emitted, i.,e. is it a log or metric?

cmd/soroban-rpc/internal/config/options.go
```diff
@@ -272,68 +273,128 @@ func (cfg *Config) options() ConfigOptions {
...
+			ConfigKey:    &cfg.RequestExecutionWarningThreshold,
+			DefaultValue: 5 * time.Second,
+		},
+		{
+			TomlKey:      strutils.KebabToConstantCase("max-request-execution-duration"),
```
sreuland, line 345: absorbing the last review and seeing similar number of new settings repeated for max duration, having a thought, wdyt about a de-duped config model down to 3 settings based on type/profile of request:

'default` is for all endpoints that have no special constraints, ie. they may read the local db and that's it 
`simulation` for simulate transaction
`submit tx` for submitting a transaction

each of the config types has a suffix for `-request-queue-limit` and `-execution-duration`, so the total permutations is 3 types x 2 suffixes = 6 configs

in this way, if new endpoints are added to rpc, don't need to change any configs if the new endpoint aligns with and can be governed by existing type thresholds, or worst case add a new config type for the new endpoint if needed.

this would remove the need to have the catch-all `max-request-execution-duration` as each type provides the notion of most effective threshold for it's grouping.

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -128,89 +128,117 @@ func NewJSONRPCHandler(cfg *config.Config, params HandlerParams) Handler {
...
+			Name: durationWarnCounterName,
+			Help: durationWarnCounterHelp,
+		})
+		requestDurationLimitCounter := prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: params.Daemon.MetricsNamespace(), Subsystem: "network",
+			Name: durationLimitCounterName,
+			Help: durationLimitCounterHelp,
+		})
+		// set the warning threshold to be one third of the limit.
+		requestDurationWarn := handler.requestDurationLimit / 3
```
sreuland, line 233: I was kind of expecting `requestDurationWarn = request-execution-warning-threshold` to be used as that config setting seemed more aligned for warning?

cmd/soroban-rpc/internal/config/options.go
tsachiherman: I think that you've also identified the need to have different vastly different default for different limiters on different methods. The challenge that I see with grouping is that it would create a mapping that might represent a constraint when trying to adjust these values on a deployed platform.

Now, if we could argue that all/most of these endpoints have the exact same characteristics and timing, I would aside with your opinion. ( for instance, if we had 5 methods that are in-memory, immediate value returning ) 
At this time, we have quite a range of method implementations, and I honestly don't know what would be the usage pattern.

I would hope to get to the same level of flexibility, where we could "toggle" a single endpoint/method when a high-load scenario takes place.

cmd/soroban-rpc/internal/jsonrpc.go
tsachiherman: in order to reduce the total number of configuration options, I have included only a configuration for the *duration limit* of all the jrpc handlers. Then, I set a third of that to be the warning time. ( i.e. if it takes more then 30% of the max time, it would considered to be a warning ).

sreuland: oh, further reading below, `request-execution-warning-threshold` is meant for only the global rpc duration limit and then each endpoint dynamically creates it's specific warning by taking a 33% of it's duration max limit? Warnings could be factored into the [config types suggestion](https://github.com/stellar/soroban-tools/pull/804/files#r1278024068) also with a new suffix on the type of:

`-execution-duration-warning`

cmd/soroban-rpc/internal/config/options.go
tsachiherman: sure. it emits it to both. ( for the metrics, it's just a counter ).

cmd/soroban-rpc/internal/jsonrpc.go
tsachiherman: The warning and limit are kind-of related. If we see an uptick in warning, it means that we need to do something before the limit would get hit. ( keep in mind that the two are exclusive : a handler that was `limit`-ed would not be considered `warn`-ed ).
The goal is that in case the server is gradually getting higher load, we would start getting warnings about that before the user would see any service disruption.
I don't mind removing the `RequestExecutionWarningThreshold` and making it a percentage of the `MaxRequestExecutionDuration` ( let's say 33% as well ).

tsachiherman:
@sreuland note that from configuration perspective, it's really important for us to have good and independent control of each of the endpoints. When monitoring our partner's usage, we would be able to tune individual endpoint without any binary change.

tsachiherman:
I'm generally less concerned about the increased overhead here in maintenance : I don't expect to have very frequent change in the API, and following a "pattern" isn't overwhelmingly risky. I am concerned that we would run into a situation where we won't have the right "configuration" tools in place - in particular around server and security configurations.

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -223,13 +251,32 @@ func NewJSONRPCHandler(cfg *config.Config, params HandlerParams) Handler {
...
-		bridge: bridge,
-		logger: params.Logger,
-		Handler: network.MakeHTTPBacklogQueueLimiter(
-			bridge,
-			globalQueueRequestBacklogLimiter,
-			uint64(cfg.RequestBacklogGlobalQueueLimit),
-			params.Logger),
+		bridge:  bridge,
+		logger:  params.Logger,
+		Handler: durationLimitedBridge,
```
sreuland, line 280: is there enough value in having the extra `httpRequestDurationLimiter` to warrant it's additional implementation and complexity, it's mainly effective as a blanket override or stop-gap on top of the per-jrpc endpoint limits/thresholds/warns implemented by `rpcRequestDurationLimiter`? from naive outside observation, it just appears to add a significant amount of code including some duplication for marginal gain, but I might be missing the nuance of it.

cmd/soroban-rpc/internal/jsonrpc.go
sreuland: makes sense, I think we may need to consider adding a MONITORING.md to assist with expectations, the insights you're mentioning are valuable and sound like best practices/recommendations. i.e. use the defaults provided to start with and then the strategy for changing over time?

sreuland: makes sense, I think we may need to consider adding a MONITORING.md to assist with expectations, the insights you're mentioning are valuable and sound like best practices/recommendations. i.e. use the defaults provided to start with and then the strategy for changing over time?

cmd/soroban-rpc/internal/jsonrpc.go
tsachiherman: Great question ! I was thinking about that topic as well.

The major reason that this is beneficial is due to a jrpc feature called "batching", which allow the caller to "pack" multiple calls within a single http call.

Limiting the total execution time per session gives us better opportunity to evenly distribute the available server resources across the users.

Note that http calls *can* be easily throttled and load balanced. This means that by "pushing" users to avoid using big packed jrpc batches, we're allowing ourselves to provide better scaling by taking advantage of http-aware load balancer.

cmd/soroban-rpc/internal/jsonrpc.go
tsachiherman: Yes, I think that adding that would be a great idea. I'll try to come up with one.

sreuland (approved):
thanks for responding on my comments, looks like a great new layer to have in rpc.

cmd/soroban-rpc/internal/jsonrpc.go
tsachiherman: @sreuland I added a MONITORING.md file. I'm sure that you'll have comments ( which are welcome! ) - could we defer the comments and proposed changes to a separate PR ?

PR #819, opened 2023-07-28
feat(bindings-ts): make Wallet injectable

chadoh:
### What

By default, if you do not provide a `wallet` option to one of the
exported functions or `invoke`, then it will continue to import and use
Freighter.

But now you can also pass your own Wallet interface to any of these
functions.

### Why

This facilitates testing the generated libraries with Node, as well as
using them with other Wallet options.

### Known limitations

Does not yet offer similar injectability for network settings & contract
ID.

### DIFF
--- a/cmd/crates/soroban-spec-typescript/src/project_template/src/method-options.ts
+++ b/cmd/crates/soroban-spec-typescript/src/project_template/src/method-options.ts
@@ -2,16 +2,29 @@
 let responseTypes: 'simulated' | 'full' | undefined
 export type ResponseTypes = typeof responseTypes
 
+export type XDR_BASE64 = string
+
+export interface Wallet {
+  isConnected: () => Promise<boolean>,
+  isAllowed: () => Promise<boolean>,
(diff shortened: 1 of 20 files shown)


cmd/crates/soroban-spec-typescript/src/project_template/src/method-options.ts
```diff
@@ -2,21 +2,44 @@
 let responseTypes: 'simulated' | 'full' | undefined
 export type ResponseTypes = typeof responseTypes
 
+export type XDR = string
```
willemneal, line 5: How about XDR_BASE64?

willemneal (approved):
Just one nit.

cmd/crates/soroban-spec-typescript/src/project_template/src/method-options.ts
chadoh: Sure, I like that more. Rebuilding snapshot, then I'll push it up.

tsachiherman (approved):

PR #824, opened 2023-07-31
Feat(CLI) restore/bump --wash_hash

willemneal:
### What

Allow passing a wasm hash when using bump/restore.

Allow restore to take multiple changes per operation.

Also fix issue with trying to store a contract instance (the key flags shouldn't be required).

### Why

### Known limitations

Still currently only support one operation.


2opremio:
Fixes #793 

@tsachiherman we probably want to make a release of the cli after this

2opremio (approved):

tsachiherman:
@willemneal could you please rebase ?

PR #828, opened 2023-08-01
feat(CLI): `config identity fund` to fund accounts on networks

willemneal:
### What

And new subcommand for funding accounts. And allow funding accounts when using `soroban config identity generate`

### Why

Remove the dependency on using tools like curl.

### Known limitations

[TODO or N/A]


2opremio:
The build is failing on some platforms. For aarch64 it seems openssl doesn't build (maybe there is a missing dependency?)

tsachiherman:
> I'm going to disable windows for this command to allow the CI to be happy. Here is an open issue to solve this in the future:
> #843

I *suspect* that you'll need to add this platform-limit also in the cargo.toml

tsachiherman (approved):

PR #829, opened 2023-08-01
feat(CLI): Add two new output types for `contract inspect`

willemneal:
### What

Add `--output` with options `docs` (default), `xdr-base64`, and `xdr-base64-array`

### Why

It can be useful to get the raw values and a json array makes it easier to parse in JS.

### Known limitations

[TODO or N/A]


2opremio (approved):

paulbellamy (approved):
Would be great for consistency (and usefulness) to add `--output xdr-base64` to other commands with a `--output` flag, like `contract read`, `events`, (and eventually `invoke`?)

PR #833, opened 2023-08-02
feat(CLI): Each generated contract method adds <contract-arg>-file-path

willemneal:
### What

This allows passing file paths in place of arguments.  Files are read as strings for all types except Bytes and BytesN which are read as raw bytes.

The info about the new args is in the help doc header for each subcommand, otherwise the new args are hidden to not clog up the help documentation.

### Why

This is particularly useful for passing bytes to contracts and avoids needed to convert them to hex as the current CLI requires.

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/invoke.rs
+++ b/cmd/soroban-cli/src/commands/contract/invoke.rs
@@ -6,7 +6,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::{fmt::Debug, fs, io, rc::Rc};
 
-use clap::{arg, command, Parser};
+use clap::{arg, command, value_parser, Parser};
 use heck::ToKebabCase;
 use soroban_env_host::{
     budget::Budget,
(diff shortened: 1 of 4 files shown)


cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -521,17 +543,26 @@ fn build_custom_cmd(name: &str, spec: &Spec) -> Result<clap::Command, Error> {
...
             .num_args(1)
             .value_parser(clap::builder::NonEmptyStringValueParser::new())
             .long_help(spec.doc(name, type_).unwrap());
 
+        file_arg = file_arg
+            .long(&file_arg_name)
+            .alias(file_arg_name.to_kebab_case())
+            .num_args(1)
+            .hide(true)
+            .value_parser(value_parser!(PathBuf));
```
paulbellamy, line 564: Can we add a "conflicts-with" for the normal arg? That would clarify the handling if you provide both.

willemneal: Good point! Will do!

paulbellamy (approved):

PR #845, opened 2023-08-07
soroban-rpc: Cache all ledger entries queried from DB in read transaction

2opremio:
### What

As a followup to #837 , also cache all entries queried from the DB in a read transaction.

To have finer control over caching, a new contructor (`NewCachedTx`) has been added. This allows disabling caching in situations in which the cache size could grow out of hand.

### Why

The preflight library (which in turn depends on rs-env-host) sometimes queries the same entry more than once.

### Result

The performance improvement is not huge, but it allows us to be more flexible on the preflight library, not needing to be careful about querying the same entry multiple times.
 
Also note that the benchmark is pretty naive and doesn't query a lot of ledger entries from the DB.

Before:

```
goos: darwin
goarch: arm64
pkg: github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/preflight
BenchmarkGetPreflight
BenchmarkGetPreflight/In-memory_storage
BenchmarkGetPreflight/In-memory_storage-12         	    5709	    211948 ns/op
BenchmarkGetPreflight/DB_storage
BenchmarkGetPreflight/DB_storage-12                	    3009	    383341 ns/op
PASS
```

After:

```
goos: darwin
goarch: arm64
pkg: github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/preflight
BenchmarkGetPreflight
BenchmarkGetPreflight/In-memory_storage
BenchmarkGetPreflight/In-memory_storage-12         	    5658	    207791 ns/op
BenchmarkGetPreflight/DB_storage
BenchmarkGetPreflight/DB_storage-12                	    3591	    315048 ns/op
PASS
```

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-rpc/internal/db/ledgerentry.go
+++ b/cmd/soroban-rpc/internal/db/ledgerentry.go
@@ -21,6 +21,7 @@ const (
 type LedgerEntryReader interface {
 	GetLatestLedgerSequence(ctx context.Context) (uint32, error)
 	NewTx(ctx context.Context) (LedgerEntryReadTx, error)
+	NewCachedTx(ctx context.Context) (LedgerEntryReadTx, error)
 }
 
 type LedgerEntryReadTx interface {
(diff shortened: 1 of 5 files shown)


2opremio:
> this will not cache lookups for entries which don't exist.

Uhm, good point, maybe I should fix this.

2opremio:
I fixed it and now the cache performs even better:

```
goos: darwin
goarch: arm64
pkg: github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/preflight
BenchmarkGetPreflight
BenchmarkGetPreflight/In-memory_storage
BenchmarkGetPreflight/In-memory_storage-12         	    6151	    194964 ns/op
BenchmarkGetPreflight/DB_storage
BenchmarkGetPreflight/DB_storage-12                	    4008	    274884 ns/op
PASS
```

we are _only_ 0.7 ms (40% more) away from the in-memory storage.

Issue #844, opened 2023-08-07
Update ecosystem to support updated fees calculation


tsachiherman:
### What problem does your feature solve?

Update signatureArgs, Host Function and arguments amalgamation, fees fields renames, error code representation

Checklist : 

- [x] Go monorepo XDR changes. https://github.com/stellar/go/pull/5009
- [x] Go ingestion changes
- [x] Soroban RPC changes
- [x] Soroban Cli changes
- [x] stellar/js-stellar-base#655

PR #846, opened 2023-08-07
Update CLI and soroban-rpc to the latest XDR changes

2opremio:
### What

Update CLI and soroban-rpc to the latest XDR changes

### Why

Fixes part of https://github.com/stellar/soroban-tools/issues/844

### Known limitations

N/A


cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,193 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+    current_ledger_seq: u32,
 ) -> Result<(SorobanTransactionData, i64), Box<dyn error::Error>> {
+    let ConfigSettingEntry::StateExpiration(state_expiration) =
+        ledger_storage.get_configuration_setting(ConfigSettingId::StateExpiration)?
+        else {
+            return Err(
+                "get_fee_configuration(): unexpected config setting entry for StateExpiration key".into(),
+            );
+        };
+    // FIXME: we are querying the ledger entries from the DB twice (once here and again in calculate_unmodified_ledger_entry_bytes)
```
2opremio, line 411: This will be addressed by https://github.com/stellar/soroban-tools/pull/846

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,193 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
     )?;
-    let soroban_resources = hack_soroban_resources(SorobanResources {
+    let soroban_resources = SorobanResources {
         footprint,
         instructions: 0,
-        // FIXME(fons): this seems to be a workaround a bug in code (the fix is to also count bytes read but not written in readBytes).
+        // FIXME(fons): this seems to be a workaround for a bug in Core (the fix is to also count bytes read but not written in readBytes).
         //        we should review it in preview 11.
         read_bytes: write_bytes,
         write_bytes,
```
2opremio, line 451: @sisuresh does this still hold?

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,193 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+            get_bumpable_entry_expiration_and_durability(&unmodified_entry)?;
+        if durability != Persistent {
+            let err = format!("Non-persistent key ({:?}) in footprint", key).into();
+            return Err(err);
+        }
+        if current_ledger_seq < expiration_ledger {
+            // TODO: is this accurate?
+            //       shouldn't we just check that cd.expiration_ledger_seq < current_ledger_seq + min_persistent_expiration - 1 ?
+            // noop (the entry hadn't expired)
+            continue;
```
2opremio, line 425: @sisuresh @dmkozh thoughts?

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -113,45 +122,54 @@ fn estimate_max_transaction_size_for_operation(
...
+        if !post_storage.map.contains_key::<LedgerKey>(k, budget)? {
+            unmodified_ledger_keys.push(k.clone());
+        }
+    }
+    let unmodified_entries_read_bytes =
+        calculate_unmodified_ledger_entry_bytes(&unmodified_ledger_keys, pre_storage, false)?;
+    let modified_entries_read_bytes: u32 = ledger_changes
+        .iter()
+        .map(|c| c.encoded_key.len() as u32 + c.old_entry_size_bytes)
+        .sum();
```
2opremio, line 144: @dmkozh @2opremio thoughts? The `ledger_changes` don't seem to be enough to compute the full `read_bytes` since they don't seem to contain unmodified (but read) ledger entries.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,193 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+
 pub(crate) fn compute_bump_footprint_exp_transaction_data_and_min_fee(
     footprint: LedgerFootprint,
     ledgers_to_expire: u32,
-    snapshot_source: &ledger_storage::LedgerStorage,
+    ledger_storage: &ledger_storage::LedgerStorage,
     bucket_list_size: u64,
+    current_ledger_seq: u32,
 ) -> Result<(SorobanTransactionData, i64), Box<dyn error::Error>> {
+    // FIXME: we are querying the ledger entries from the DB twice (once here and again in calculate_unmodified_ledger_entry_bytes)
```
2opremio, line 320: This will be addressed by https://github.com/stellar/soroban-tools/pull/846

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,193 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+
+fn get_bumpable_entry_expiration_and_durability(
+    ledger_entry: &LedgerEntry,
+) -> Result<(u32, ContractDataDurability), Box<dyn error::Error>> {
+    // TODO: Are there any other acceptable entry types?
+    //       is the ContractCode entry type always considered persistent?
+    let res = match ledger_entry.data {
+        LedgerEntryData::ContractData(ref cd) => (cd.expiration_ledger_seq, cd.durability),
+        LedgerEntryData::ContractCode(ref cc) => (cc.expiration_ledger_seq, Persistent),
+        _ => {
```
2opremio, line 386: @dmkozh @sisuresh can you check whether this is correct? the durability is used to populate the rent changes.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,198 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+        if durability != Persistent {
+            let err = format!("Non-persistent key ({:?}) in footprint", key).into();
+            return Err(err);
+        }
+        // FIXME: according to https://soroban.stellar.org/docs/fundamentals-and-concepts/state-expiration :
+        //  The entry is considered expired when current_ledger > expirationLedger.
+        //  So, the code below should be `current_ledger_seq <= expiration_ledger`
+        //  However, we observe the expiration happening exactly at expiration_ledger.
+        //  Is this a bug in the ledger updates of soroban-rpc?
+        if current_ledger_seq < expiration_ledger {
```
2opremio, line 426: @sisuresh @dmkozh thoughts?

2opremio: CC @tamirms

willemneal:
Should this PR target main? Shouldn't main stay in line with futurenet's XDR?

cmd/soroban-rpc/lib/preflight/src/fees.rs
dmkozh: That's a bug in recording mode - this hack shouldn't be needed when it's fixed.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,198 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+    current_ledger_seq: u32,
 ) -> Result<(SorobanTransactionData, i64), Box<dyn error::Error>> {
+    // FIXME: we are querying the ledger entries from the DB twice (once here and again in calculate_unmodified_ledger_entry_bytes)
+    let mut rent_changes: Vec<LedgerEntryRentChange> = Vec::new();
+    for key in (&footprint).read_only.as_vec() {
+        let unmodified_entry = ledger_storage.get(key, false)?;
+        let (expiration_ledger, durability) =
+            get_bumpable_entry_expiration_and_durability(&unmodified_entry)?;
+        let new_expiration_ledger = current_ledger_seq + ledgers_to_expire;
+        if new_expiration_ledger <= expiration_ledger {
```
dmkozh, line 327: You also need to check that entry is still alive (current ledger >= expiration ledger). It's not possible to bump an expired entry.

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -292,118 +285,198 @@ fn storage_footprint_to_ledger_footprint(foot: &Footprint) -> Result<LedgerFootp
...
+        &rent_changes,
+        current_ledger_seq,
+        bucket_list_size,
+    )
+}
+
+fn get_bumpable_entry_expiration_and_durability(
+    ledger_entry: &LedgerEntry,
+) -> Result<(u32, ContractDataDurability), Box<dyn error::Error>> {
+    // TODO: Are there any other acceptable entry types?
```
dmkozh, line 381: No, there aren't other expirable entry types.

cmd/soroban-rpc/lib/preflight/src/fees.rs
dmkozh: This is not a bug - the entries to restore are stored in ledger and thus need to be accounted in read bytes as well. The amounts are the same currently.

cmd/soroban-rpc/lib/preflight/src/fees.rs
2opremio: Good to know, would you please answer here once it's fixed?

cmd/soroban-rpc/lib/preflight/src/fees.rs
2opremio: _we observe_ means that soroban-rpc sees the entry expire at `expiration_ledger` and not `expiration_ledger + 1`, as a result if we preflight the restoration right after the expiration and check `current_ledger_seq <= expiration_ledger` soroban-rpc considers the operation a noop and the resulting restoration transaction is not accepted.

2opremio:
> Should this PR target main? Shouldn't main stay in line with futurenet's XDR?

Main already contains XDR changes ahead of futurenet. The futurenet code was tagged at https://github.com/stellar/soroban-tools/tree/v0.9.5

cmd/soroban-rpc/lib/preflight/src/fees.rs
dmkozh: I still don't quite understand; the correct condition is that entry is still alive on `expiration_ledger` and expired (== inaccessible) on `expiraiton_ledger + 1`. Anything else is a bug, so IIUC soroban-rpc should be fixed to respect this logic.

On a side note, I hope that preflight sets the ledger sequence to `last_closed_ledger_seq + 1`, as that would be as close to reality as we can achieve with preflight.

cmd/soroban-rpc/lib/preflight/src/fees.rs
2opremio: That's ensured by the `includeExpired: false` parameter of `ledger_storage.get(key, false)` a few lines above

dmkozh: Ok, sounds good.

cmd/soroban-rpc/lib/preflight/src/fees.rs
2opremio: > We don't emit anything when entry expires

Aha, then we may not be expiring the entries at the right time.

2opremio: Fixed at https://github.com/stellar/soroban-tools/pull/846/commits/db65519b1b747c36701380961fbf54239cd9f7de

dmkozh: Yes, the logic is correct now. But do note, that it is correct only for checking the entry state at LCL. As I've mentioned above, preflight should ideally happen at LCL + 1, as there is clearly no way to include a new transaction in LCL.

willemneal:
Ah cool good to know! So we can't release the CLI again until futurenet is aligned?

cmd/soroban-rpc/internal/ingest/service.go
```diff
@@ -251,6 +251,7 @@ func (s *Service) ingest(ctx context.Context, sequence uint32) error {
 	if err := tx.Commit(sequence); err != nil {
 		return err
 	}
+	s.logger.Debugf("Ingested ledger %d", sequence)
```
Shaptic, line 254: maybe a nitpick and not the time/place for this comment, but still noting here:

would it be better to have something like:

```go
var err error
defer func() {
  s.logger
    .WithError(err)
    .WithField("took", time.Since(startTime))
    .Debugf("Done ingesting ledger %d", sequence)
}()

ledgerCloseMeta, err = s.ledgerBackend.GetLedger(ctx, sequence)
// etc.
```
where each subsequent line assigns that `err` value?

cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -188,26 +194,26 @@ func getInvokeHostFunctionPreflight(params PreflightParameters) (Preflight, erro
...
-	minTempEntryExpiration := uint32(0)
-	minPersistentEntryExpiration := uint32(0)
-	maxEntryExpiration := uint32(0)
-	if hasConfig {
-		setting := stateExpirationConfig.Data.MustConfigSetting().MustStateExpirationSettings()
-		minTempEntryExpiration = uint32(setting.MinTempEntryExpiration)
-		minPersistentEntryExpiration = uint32(setting.MinPersistentEntryExpiration)
-		maxEntryExpiration = uint32(setting.MaxEntryExpiration)
+	if !hasConfig {
+		return Preflight{}, errors.New("state expiration config setting missing in ledger storage")
```
Shaptic, line 198: maybe we use a default here, instead?

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -500,7 +499,7 @@ func TestSimulateInvokeContractTransactionSucceeds(t *testing.T) {
 	require.Contains(t, metrics, "soroban_rpc_json_rpc_request_duration_seconds_count{endpoint=\"simulateTransaction\",status=\"ok\"} 3")
 	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_get_duration_seconds_count{status=\"ok\",type=\"db\"} 3")
 	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_get_duration_seconds_count{status=\"ok\",type=\"all\"} 3")
-	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_entries_fetched_sum 42")
+	require.Contains(t, metrics, "soroban_rpc_preflight_pool_request_ledger_entries_fetched_sum 48")
```
Shaptic, line 502: why did this go up? :detective:

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -771,13 +770,19 @@ func TestSimulateTransactionBumpAndRestoreFootprint(t *testing.T) {
...
+	expired := false
 	for i := 0; i < 50; i++ {
 		err = client.CallResult(context.Background(), "getLedgerEntry", getLedgerEntryrequest, &result)
 		if err != nil {
+			expired = true
+			fmt.Println("ledger entry expired")
 			break
 		}
+		assert.NoError(t, xdr.SafeUnmarshalBase64(result.XDR, &entry))
+		fmt.Println("waiting for ledger entry to expire at ledger", entry.MustContractData().ExpirationLedgerSeq+1)
```
Shaptic, line 782: I think [`t.Logf`](https://pkg.go.dev/testing#T.Log) here and L778 would be better for test outputs

Shaptic (commented):
not qualified to approve this lol but adding some thoughts from an outside perspective

Shaptic:
@willemneal we could branch off of v0.9.5 and cherry-pick, but I don't know if our CI/CD supports publishing crates with that workflow?

cmd/soroban-rpc/internal/preflight/preflight.go
2opremio: The preflight library needs a gazillion other parameters without which cannot function properly. The network should be creating this config parameters on every reset.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: Moar accesses :) 

The preflight algorithm changed. The same entries are queried more than once, but they are heavily cached now.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: Good idea.

cmd/soroban-rpc/internal/preflight/preflight.go
Shaptic: Fair enough!

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
Shaptic: Cool, just double-checking :) I didn't see any big changes to the test itself so I wanted to ensure this was explainable

cmd/soroban-rpc/internal/ingest/service.go
2opremio: The error is surfaced back as a return value, which will be printed later on as a long line so I don't think we should be printing it

The debug printout is to debug possible concurrency problems, which I think does the job

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -253,7 +253,7 @@ func (l *ledgerEntryReadTx) GetLedgerEntry(key xdr.LedgerKey, includeExpired boo
 			if err != nil {
 				return false, xdr.LedgerEntry{}, err
 			}
-			if expirationLedgerSeq <= xdr.Uint32(latestClosedLedger) {
+			if expirationLedgerSeq < xdr.Uint32(latestClosedLedger) {
```
paulbellamy, line 256: Wouldn't this give an extra ledger of visibility? The latestClosedLedger is the one *before* the current ledger we're building now.

2opremio: That’s just how state expiration is defined at https://soroban.stellar.org/docs/fundamentals-and-concepts/state-expiration

2opremio: More specifically 

The entry is considered expired when current_ledger > expirationLedger.

dmkozh: FWIW if the entry expires on `latestClosedLedger + 1` (i.e. it's `expirationLedgerSeq == latestClosedLedger`), then the transaction you preflight should see the entry as expired. The same goes about retrieving contract data  - while the entry has been alive for LCL, it might already be expired at the moment we fetch it.

PR #847, opened 2023-08-07
soroban-rpc: Fix upstert  bug in DB cache

2opremio:
Leftover from #837 

BTW, the fact that the system tests are failing all the time gets the eye used to seeing red-crosses in CI. This lets problems like this leak through the cracks.

We should probably make the unit tests mandatory for merging although I would prefer for the system tests not to be broken so often (CC @tsachiherman @sreuland )

### DIFF
--- a/cmd/soroban-rpc/internal/db/ledgerentry_test.go
+++ b/cmd/soroban-rpc/internal/db/ledgerentry_test.go
@@ -767,6 +767,6 @@ func NewTestDB(tb testing.TB) *DB {
 	})
 	return &DB{
 		SessionInterface: db,
-		ledgerEntryCache: map[string]string{},
+		ledgerEntryCache: newTransactionalCache(),
 	}
 }
--- a/cmd/soroban-rpc/internal/db/transactionalcache.go
+++ b/cmd/soroban-rpc/internal/db/transactionalcache.go
@@ -48,7 +48,8 @@ func (w transactionalCacheWriteTx) commit() {
 	for key, newValue := range w.pendingUpdates {
 		if newValue == nil {
 			delete(w.parent.entries, key)
+		} else {
+			w.parent.entries[key] = *newValue
 		}
-		w.parent.entries[key] = *newValue
 	}
 }


sreuland:
ok, can take a look at e2e tomorrow, in the ci output, the rpc ingestion is logging xdr parse errors, maybe we need to bump the go pkg ref, cor core version, etc.
```
time="2023-08-07T23:07:02.178Z" level=error msg="could not run ingestion. Retrying" error="Error while reading from buckets: Error on XDR record -1 of hash '74585b1a9372de0e01a48318f5734290975fb9381a0990c4fa1128ab132a6d78': decoding LedgerEntry: decoding LedgerEntryExt: decoding LedgerEntryExtensionV1: decoding SponsorshipDescriptor: decoding PublicKey: decoding Uint256: decoding Uint256: xdr:DecodeFixedOpaqueInplace: EOF while decoding 32 bytes - read: '[]'" pid=271
```

tsachiherman:
> ok, can take a look at e2e tomorrow, in the ci output, the rpc ingestion is logging xdr parse errors, maybe we need to bump the go pkg ref, cor core version, etc.
> 
> ```
> time="2023-08-07T23:07:02.178Z" level=error msg="could not run ingestion. Retrying" error="Error while reading from buckets: Error on XDR record -1 of hash '74585b1a9372de0e01a48318f5734290975fb9381a0990c4fa1128ab132a6d78': decoding LedgerEntry: decoding LedgerEntryExt: decoding LedgerEntryExtensionV1: decoding SponsorshipDescriptor: decoding PublicKey: decoding Uint256: decoding Uint256: xdr:DecodeFixedOpaqueInplace: EOF while decoding 32 bytes - read: '[]'" pid=271
> ```

I'd suspect that the js soroban client version doesn't point to a hash that matches the latest xdr.

tsachiherman (approved):

2opremio:
> ok, can take a look at e2e tomorrow, in the ci output, the rpc ingestion is logging xdr parse errors, maybe we need to bump the go pkg ref, cor core version, etc.
> 


Yeah, see https://github.com/stellar/soroban-tools/pull/846

PR #848, opened 2023-08-09
fix(CLI): make aliases visible in help doc

willemneal:
### What

Also fixed generated help doc for contracts. Needed to move the note about <arg>-file-path to use long about instead to avoid it appearing in each subcommand's summary at the top level.

### Why

### Known limitations

[TODO or N/A]


ueco-jb (approved):
Good one!

tsachiherman (approved):

PR #849, opened 2023-08-09
feat(bindings-ts)!: rm --contract-name, add --overwrite

### LINKED ISSUE #752: soroban contract bindings typescript nesting when running > once
When running `soroban contract bindings typescript` more than 1 time the deployed package nests into the existing project.

Yay
Screenshot 2023-07-06 at 3 51 19 PM

Ew
Screenshot 2023-07-06 at 3 51 51 PM

chadoh:
Fixes #752

Removes the `--contract-name` argument from `bindings typescript`, since
this was often confused with `--output-dir` and, really, just using the
file path from the end of `--output-dir` is sufficient.

Adds `--overwrite`. If you specify an `--output-dir` that already
exists, by default it will not be overwritten. Instead, a descriptive error
will be thrown, letting you know that you can add `--overwrite` to
overwrite the directory.

Previously, if you specified an `--output-dir` that already existed, it
would create a new directory with the same name inside this directory,
which can lead to confusing bugs.

This approach was arrived at with @tyvdh

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
+++ b/cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Debug, path::PathBuf};
+use std::{ffi::OsString, fmt::Debug, path::PathBuf};
 
 use clap::{command, Parser};
 use soroban_spec_typescript::{self as typescript, boilerplate::Project};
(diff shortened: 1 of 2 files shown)


tyvdh (approved):
lgtm

willemneal:
My own nit is that paths can be non-UTF8 and you used lossy, but perhaps we could require that the string be utf8 and throw an error instead.

chadoh:
Good point, @willemneal. What possible problems might this cause? All I can come up with is that contract_name might end up looking goofy or truncated, but I'm not sure that matters.

PR #852, opened 2023-08-09
feat(bindings-ts)!: require configuring network settings

chadoh:
### What

Generated libraries now export a `Contract` class that needs to be instantiated by the user with RPC URL, Network Passphrase, and contract ID.

They also export a `networks` object to help initialize `networkPassphrase` and `contractId` with known-valid settings.

**Surprise bonus:** adds a `ts-tests` directory with a test for all `test_custom_types.wasm` exported methods! This exposed a couple easy-to-fix problems that were addressed, while marking other tests as known failures, which we can circle back to later. While adding these tests I also discovered some weird behavior. This is documented in the test cases; we can discuss how to best fix this behavior later.

### Why

Keeps libraries focused on behavior, not specific endpoints, which will make them more flexible. Still gives contract authors a way to tell users about known supported networks.

### Known limitations

To do:

Right now the new tests are not incorporated into CI at all. Do we want to incorporate them into CI? How? It might require running a localnet node within CI and deploying the contract to it. This seems like a lot of work! Is there a project with a similar setup we can copy?


cmd/crates/soroban-spec-typescript/src/boilerplate.rs
```diff
@@ -74,29 +77,48 @@ impl Project {
...
+            "futurenet"
+        } else if network_passphrase == NETWORK_PASSPHRASE_LOCALNET {
+            "localnet"
+        } else {
+            "unknown"
+        };
+        format!(
+            r#"export const networks = {{
+    {network}: {{
+        networkPassphrase: "{network_passphrase}",
```
Shaptic, line 117: you can use e.g. `SorobanClient.Networks.FUTURENET` if you don't want the magic strings repeated

cmd/crates/soroban-spec-typescript/src/project_template/README.md
```diff
@@ -8,11 +8,11 @@ This library was automatically generated by Soroban CLI using a command similar
 soroban contract bindings ts \
   --rpc-url INSERT_RPC_URL_HERE \
   --network-passphrase "INSERT_NETWORK_PASSPHRASE_HERE" \
-  --id INSERT_CONTRACT_ID_HERE \
+  --contract-id INSERT_CONTRACT_ID_HERE \
   --name INSERT_CONTRACT_NAME_HERE
```
Shaptic, line 12: two ux notes:

 - can we make name optional or derived somehow?
 - can we make the id a required positional arg? writing `--contract-id` is mildly annoying

cmd/crates/soroban-spec-typescript/src/project_template/src/method-options.ts
```diff
@@ -15,7 +15,23 @@ export interface Wallet {
...
 
-export type Options<R extends ResponseTypes> = {
+export type ClassOptions = {
+  contractId: string
+  networkPassphrase: string
+  rpcUrl: string
+  /**
+   * A Wallet interface, such as Freighter, that has the methods `isConnected`, `isAllowed`, `getUserInfo`, and `signTransaction`. If not provided, will attempt to import and use Freighter. Example:
+   *
+   * ```ts
```
Shaptic, line 25:
```suggestion
   * @example
```
I _think_ this is the right way to render it? If not, you need the closing ``` anyway

Shaptic (commented):
quick pass, not a formal review as I'm less familiar with this side

tyvdh:
What would a cli command now look like? I read the README but it's still not clear how to set the `networks` object with custom networks

cmd/crates/soroban-spec-typescript/src/boilerplate.rs
chadoh: Good to know! I think the magic string makes more sense in this context, to reduce the reliance on dependencies in the generated client.

cmd/crates/soroban-spec-typescript/src/project_template/README.md
chadoh: Ah, we just removed `--[contract]-name` in #849. Thanks for pointing this out; I'll remove this line.

Interesting question about making contract ID positional. That would make the full command look like:

```bash
soroban contract bindings typescript CB5T6MLZNWJBUBKEQAUVIG5JJWKYSYVVE2OVN25GMX3VX7CZ7OBAPAU4 \
  --network futurenet
  --output-dir path/to/contract-client
```

This introduces ambiguity, though, if you generate from a local `--wasm` rather than a deployed contract. I think maybe it's better to keep them as named arguments, for now.

cmd/crates/soroban-spec-typescript/src/project_template/src/method-options.ts
chadoh: Thanks, [looks like](https://typedoc.org/tags/example/) the best way is to use both `@example` and the backticks together

chadoh:
@tyvdh for now, there is no way to automatically generate a library with multiple configured networks. Library authors will need to manually update the `networks` constant in `index.ts`. Do you think that's acceptable, as a starting point?

Expanding the CLI command to accept multiple `--network` and `--contract-id` arguments would require significant work, and might not be useful enough to warrant that work.

cmd/crates/soroban-spec-typescript/src/project_template/README.md
willemneal: `--id` should be an alias, but for the most part we've opted for named parameters except in places like the config where it's clear what the main argument is.  I have been considering changing the invoke command to take --contract-id as a positional and then we could not need the slop `--` instead it would be seen as a subcommand.

cmd/crates/soroban-spec-typescript/src/boilerplate.rs
paulbellamy: We'll also want to support pubnet and testnet soon, so might be worth doing them as well.

I had expected the  network passphrase & url would come from the `soroban config network` setup the user would have already done.  So if I did:
```bash
$ soroban config network add \
    --name chocolate_factory \
    --passphrase "Oompa Loompa Network ; October 2022" \
    --rpc-url "https://charlie-rpc.com"
$ soroban contract binding typescript --network chocolate_factory --id C1234
```
then it would use the network config settings from that network to generate a network setup like:
```ts
export const networks = {
  readonly "chocolate_factory": {
    // whatever format the client expects so you could pass it directly
    readonly networkPassphrase: "Oompa Loompa Network ; October 2022",
    readonly rpcUrl: "https://charlie-rpc.com",
    readonly contractId: "C1234",
  },
};
```

cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts
```diff
@@ -52,26 +50,31 @@ export async function invoke<R extends ResponseTypes, T = string>({
...
+  contractId,
   wallet,
 }: InvokeArgs<R, T>): Promise<T | string | SomeRpcResponse> {
   wallet = wallet ?? await import('@stellar/freighter-api')
-  const walletAccount = await getAccount(wallet)
+  const server = new SorobanClient.Server(rpcUrl, { allowHttp: rpcUrl.startsWith('http://') });
+  const walletAccount = await getAccount(wallet, server)
 
   // use a placeholder null account if not yet connected to Freighter so that view calls can still work
   const account = walletAccount ?? new SorobanClient.Account('GAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAWHF', '0')
```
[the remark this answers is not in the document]
chadoh: It does view/read calls with the default, and throws an error if it needs to send an actual transaction. See below the `if (isViewCall)` block below; line 117

cmd/crates/soroban-spec-typescript/src/project_template/src/method-options.ts
```diff
@@ -15,7 +15,25 @@ export interface Wallet {
...
+  rpcUrl: string
+  /**
+   * A Wallet interface, such as Freighter, that has the methods `isConnected`, `isAllowed`, `getUserInfo`, and `signTransaction`. If not provided, will attempt to import and use Freighter. Example:
+   *
+   * @example
+   * ```ts
+   * import freighter from "@stellar/freighter-api";
+   *
+   * // later, when calling this function:
+   *   wallet: freighter,
```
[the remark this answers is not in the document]
chadoh: It gets exported as just `Contract`, but you're right, I need to update this example a little bit more for how this `Contract` export is actually used.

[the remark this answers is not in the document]
chadoh: Updated; see latest commit

cmd/crates/soroban-spec-typescript/src/boilerplate.rs
Shaptic: > to reduce the reliance on dependencies in the generated client

but there's no world in which you can do this without SorobanClient, I think?

cmd/crates/soroban-spec-typescript/src/project_template/src/index.ts
```diff
@@ -73,21 +72,20 @@ if (typeof window !== 'undefined') {
...
 
-function getError(err: string): Err<Error_> | undefined {
-    const match = err.match(regex);
+function parseError(message: string): Err<Error_> | undefined {
+    const match = message.match(regex);
     if (!match) {
         return undefined;
     }
-    if (Errors == undefined) {
+    if (Errors === undefined) {
```
Shaptic, line 82: this confused me initially and does still: where is `Errors` supposed to be defined?

cmd/crates/soroban-spec-typescript/ts-tests/src/test.ts
```diff
@@ -0,0 +1,188 @@
+import test from 'ava'
```
Shaptic, line 1: I know this would be obnoxious to do, but do you think there's a benefit to staying consistent with the mocha+chai approach we have in the rest of the SDK? The minimum benefit would be developer familiarity, but maybe that's not worth porting this until it becomes a problem.

cmd/crates/soroban-spec-typescript/ts-tests/src/test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+test('i32', async t => {
+  t.is(await contract.i32({ i32_: 1 }), 1)
+})
+
+test('i64', async t => {
+  t.is(await contract.i64({ i64_: 1n }), 1n)
+})
+
+test("strukt_hel", async (t) => {
```
Shaptic, line 47: `struct` isn't reserved in JS, so why are we doing this weird mangling?

cmd/crates/soroban-spec-typescript/ts-tests/src/test.ts
```diff
@@ -0,0 +1,188 @@
...
+  const tuple = ['hello', 1] as const
+  t.deepEqual(await contract.tuple({ tuple }), tuple)
+})
+
+test.failing('option', async t => {
+  // this makes sense
+  t.deepEqual(await contract.option({ option: 1 }), 1)
+
+  // this passes but shouldn't
+  t.deepEqual(await contract.option({ option: undefined }), 0)
```
Shaptic, line 164: can/should we make this an intentional test failure?

cmd/crates/soroban-spec-typescript/src/project_template/src/index.ts
willemneal: It is injected by the rust. Basically if the contract has errors defined they are used otherwise it's initialized as undefined.

cmd/crates/soroban-spec-typescript/ts-tests/src/test.ts
willemneal: It's the name of the functions in the contract.

cmd/crates/soroban-spec-typescript/ts-tests/src/test.ts
willemneal: This is changed in my updated tests that use the new ContractSpec class to handle the parsing.

willemneal: It returns `undefined`.

cmd/crates/soroban-spec-typescript/ts-tests/src/test.ts
willemneal: We had more recent experience with ava and typescript tests. But now that I was able to get it working for the soroban-client repo I think we could easily port it here too.

chadoh: I like AVA, for [all the reasons](https://github.com/avajs/ava#why-ava). I don't personally see much benefit to keeping consistent tooling across repos.

cmd/crates/soroban-spec-typescript/ts-tests/src/test.ts
chadoh: And this already is an intentional failure; see the `t.failing` above. But it doesn't fail until the line below, when we get to the `assert returns undefined` line, and I wanted to clearly comment on the unexpected behavior of options.

Shaptic (approved):
LGTM from my perspective—yeet it

paulbellamy (approved):

PR #864, opened 2023-08-16
fix: switch xdr dep for spec tools to see if it fixes CI issue

willemneal:


leighmcculloch:
This change is intriguing. Switching to soroban-env-host to access XDR is equivalent to changing the stellar-xdr import to match the stellar-xdr import that soroban-env-common imports. So it's unclear to me how that will fix the issue. If soroban-spec-tools is importing the wrong stellar-xdr lib, we should be able to correct that by changing the version of stellar-xdr in the Cargo.toml, without changing what crates are imported.

willemneal:
I think this was a stab in the dark, and I'm not sure it was ever needed.

2opremio:
Can we close this?

Issue #803, opened 2023-07-25
soroban-rpc: Transaction simulation should assist in determining state expiration


Shaptic:
### What problem does your feature solve?
It can be hard (and slow) to determine when and what to restore when a transaction simulation fails due to expired ledger entries. More specifically, if you run `simulateTransaction()` and encounter a failure, you have to do a bunch of extra work to determine whether or not it's related to state expiration AND figure out which ledgers need to be restored.

### What would you like to see?
An amendment to the `simulateTransaction` response schema:

```diff
+  expiredEntries: string[];
+  restorationFootprint: string;
```

which would be:
 * a list of base64-encoded `xdr.LedgerKey` instances corresponding to entries that the system knows have expired at the time of the simulation
 * a base64-encoded `xdr.SorobanTransactionData` instance that would correspond to a `restoreFootprintOp` transaction's `sorobanData` field (fee info, etc.), so that the client can immediately submit a restoration transaction (without running a sim)

### What alternatives are there?
I don't think there is one. **IF** (and this is a big "if") you have the relevant `xdr.LedgerKey[]`s already, you can

 1. call `getLedgerEntries` to acquire the `xdr.LedgerEntry[]`s that have expired
 3. you then parse+compare that response to determine which `restoreFootprintOp`s are needed

However, this is problematic for several reasons:
 * there is extra latency due to the additional request+processing
 * that extra latency can introduce edge cases in ledger expiration timing
 * **you still need the keys**, which is not feasible without, I dunno, storing all of your tx envelopes client-side

2opremio:
I am not sure we can implement this without changes to the [soroban-env-host](https://github.com/stellar/rs-soroban-env/tree/main/soroban-env-host) library.

Let me explain. Host function preflighting works by trying to execute a host function (through  the `soroban-env-host` library) and recording what ledger entries have been accessed.

When a host function executes, it needs to access ledger entries. To do so, soroban-rpc plugs its ledger storage into `soroban-env-host`.

Although soroban-rpc stores all ledger entries (expired or not) it hides expired ledger entries during host function execution to simulate what really would happen when the preflighted transaction is submitted.

Thus, in the eyes of `soroban-env-host` when the host function tries to access an expired entry, it will simply think it's missing. Leading to missing entry errors.

Now, what we could _maybe_ do is stop hiding expired entries to `soroban-env-host`  and somehow figure out whether the computed ledger-footprint contains expired entries after executing the host function.

[...]

PR #865, opened 2023-08-16
soroban-rpc: simulateTransaction: automatically detect ledger entries which require restoring

2opremio:
This change incorporates two new fields to simulateTransaction's Response:

```
PreRestoreTransactionData string                       `json:"preRestoreTransactionData,omitempty"` // SorobanTransactionData XDR in base64
PreRestoreMinResourceFee  int64                        `json:"preRestoreMinResourceFee,omitempty"`
```

When preflighting an InvokeHostFunction operation, soroban-rpc will now detect ledger entries which require restoring prior to the invocation for it to succeed.

If any such entries are detected, the simulateTransaction response will provide the Soroban Transaction Data and minimum resource fee needed to restore those entries.

The user is then now expected to:
1. Submit a RestoreFootprint operation using the new fields above, if present.
2. Submit the InvokeHostFunction operation normally (i.e. using the pre-existing simulateTransaction response fields)

Closes https://github.com/stellar/soroban-tools/issues/803


2opremio:
@dmkozh does something like [state_expiration.rs](https://github.com/stellar/soroban-tools/pull/865/files#diff-95f4607d5d84262eecf98773e4860dc4bcefab4efc84cb210efc9be1cf9c4d19) already exist? I figure you need something like that upstream.

Otherwise, it may make sense to place it in the env library if you find it appropriate.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -810,4 +783,38 @@ func TestSimulateTransactionBumpAndRestoreFootprint(t *testing.T) {
...
+		},
+		methods.SimulateTransactionResponse{
+			TransactionData: simulationResult.PreRestoreTransactionData,
+			MinResourceFee:  simulationResult.PreRestoreMinResourceFee,
+		},
+	)
+	tx, err = txnbuild.NewTransaction(params)
+	assert.NoError(t, err)
+	sendSuccessfulTransaction(t, client, sourceAccount, tx)
+
```
2opremio, line 813: @Shaptic here's an example of how to use it.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -810,4 +783,38 @@ func TestSimulateTransactionBumpAndRestoreFootprint(t *testing.T) {
...
+	// Finally, we should be able to send the inc host function invocation now that we
+	// have pre-restored the entries
+	params = preflightTransactionParamsLocally(t, invokeIncPresistentEntryParams, simulationResult)
+	tx, err = txnbuild.NewTransaction(params)
+	assert.NoError(t, err)
+	sendSuccessfulTransaction(t, client, sourceAccount, tx)
 }
```
2opremio, line 820: @Shaptic here's an example of how to use it.

2opremio:
@willemneal we should probably use this in the cli once merged

2opremio:
@Shaptic you have an example of how to use this at 

> To me, `PreRestore` sounds misleading, since it's not something you would do before the restore operation. it's the information _for_ the restore operation. I would use the `preamble`/`prelude` prefix for these two fields.

No strong opinions here.

tsachiherman:
thinking about that further, I think that what I'd like to see if I were in @Shaptic 's seat is the following:
```
RestorePreamble  restorePreamble   `json:"restorePeramble,omitempty"` // optional restore preamble
```
where the restorePreamble is defined as
```
type restorePreamble struct {
  TransactionData string           `json:"transactionData"` // non-optional since the struct itself is optional; base64 encoded.
  MinResourceFee  int64            `json:"minResourceFee"` // non-optional since the struct itself is optional.
}
```

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
Shaptic: Thank you! This means the flow is something like:

```
result = simulateTransaction(invoke)
if result.PreRestoreTransactionData != undefined: 
    submitTransaction(restore, transactionData=result.PreRestoreTransactionData)
    result = simulateTransaction(invoke)
submitTransaction(invoke, transactionData=result.transactionData)
```

Correct?

Shaptic:
@tsachiherman from a client perspective, having one omittable structure is better than two, so yes, I agree - I like encapsulating it into a structure like this. Of course we could have said "zero or both of these are present," but "zero or one" is much better!

Shaptic (approved):
you rock :guitar: 

this LGTM as-is but I’d prefer the schema Tsachi outlined

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: Yes, assuming you somehow also use the minfee too.

2opremio:
> this LGTM as-is but I’d prefer the schema Tsachi outlined

sure

tsachiherman (approved):

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -114,21 +120,32 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 			}
 		}
 
+		var results []SimulateHostFunctionResult
+		if result.Result != "" {
```
tsachiherman, line 124: oh - you're omitting the `Results` in case it's empty, as per the response schema.
That's perfectly fine. @Shaptic just wanted to know you're aware of that.

Shaptic: Yep, that's totally fine; we already assume this field is optional: https://github.com/stellar/js-soroban-client/blob/main/src/soroban_rpc.ts#L132-L133.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -28,16 +28,20 @@ type SimulateHostFunctionResult struct {
...
-	LatestLedger              int64                        `json:"latestLedger,string"`
-	PreRestoreTransactionData string                       `json:"preRestoreTransactionData,omitempty"` // SorobanTransactionData XDR in base64
-	PreRestoreMinResourceFee  int64                        `json:"preRestoreMinResourceFee,omitempty"`
+	Error           string                       `json:"error,omitempty"`
+	TransactionData string                       `json:"transactionData"` // SorobanTransactionData XDR in base64
+	MinResourceFee  int64                        `json:"minResourceFee,string"`
+	Events          []string                     `json:"events,omitempty"`          // DiagnosticEvent XDR in base64
+	Results         []SimulateHostFunctionResult `json:"results,omitempty"`         // an array of the individual host function call results
+	Cost            SimulateTransactionCost      `json:"cost"`                      // the effective cpu and memory cost of the invoked transaction execution.
+	RestorePreamble RestorePreamble              `json:"restorePeramble,omitempty"` // If present, it indicates that a prior RestoreFootprint is required
```
Shaptic, line 43: nit: field typo

```suggestion
	RestorePreamble RestorePreamble              `json:"restorePreamble,omitempty"` // If present, it indicates that a prior RestoreFootprint is required
```

Shaptic, line 43: nit: field typo in json

```suggestion
	RestorePreamble RestorePreamble              `json:"restorePreamble,omitempty"` // If present, it indicates that a prior RestoreFootprint is required
```

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: Great

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: Actually, you don't need the second simulation

```
result = simulateTransaction(invoke)
if result.PreRestoreTransactionData != undefined: 
    submitTransaction(restore, transactionData=result.PreRestoreTransactionData)
submitTransaction(invoke, transactionData=result.transactionData)
```

2opremio: @Shaptic did you see my udpate above? You don't need to re-run the transaction simulation

PR #867, opened 2023-08-17
fix(CLI): move plugin logic to Root and pass clap error if plugin's name is close to a built-in subcommand

### LINKED ISSUE #866: CLI: `config` returns incorrect error
soroban 0.9.4 (76cacc7cf885bd7d37756735fab332351d893406)
soroban-env 0.0.17 (400d806387140553e4e685d232deb3a807ec0e36)

### What did you do?

If you don't provide command `address` to `soroban config identity`, then incorrect error appears.

"Command not recognized; available commands..."

### What did you see instead?
```bash
$ soroban config identity address alice
GCH...
$ soroban config identity  alice        # notice missing `address`
error: no such command: `config`

        View all installed plugins with `soroban --list`
```
There must be a bug in a way the command chain is parsed.

willemneal:
fixes #866

### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/crates/soroban-test/tests/it/plugin.rs
+++ b/cmd/crates/soroban-test/tests/it/plugin.rs
@@ -31,6 +31,7 @@ fn list() {
 }
 
 #[test]
+#[cfg(not(unix))]
 fn has_no_path() {
     // Call soroban with the PATH variable set to include just target/bin directory
     assert_cmd::Command::cargo_bin("soroban")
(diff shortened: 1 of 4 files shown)


paulbellamy (approved):
LGTM, but needs an update to be mergeable

paulbellamy:
Seems like the no_path test is failing now

willemneal:
@tsachiherman Testing no path was difficult because unix requires it. However, I was able to test it on windows that if the path is just the plugin bin directory, it still works. But for unixes I added a failure test for if the path doesn't include the plugin.

PR #853, opened 2023-08-10
feat: non-breaking updates to CLI

willemneal:
### What

The updates to xdr and RPC are out of sync with futurenet. So this is a collection of CLI related changes that are non-breaking.

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]

PR #874, opened 2023-08-19
build: CI test `bindings typescript`-generated libs

chadoh:
### What

Add CI test to run the tests in cmd/crates/soroban-spec-typescript/ts-tests

### Why

This will ensure that the generated typescript libraries continue to work as they should.

### Known limitations

You can see that right now this workflow needs to install soroban-cli 0.9.4 rather than the one from the repo itself. This is because the repo itself is currently broken with the quickstart Docker image. Once we start keeping `main` in a working state with quickstart, we can fix this.


chadoh:
retargeting at https://github.com/stellar/soroban-tools/pull/853 in https://github.com/AhaLabs/soroban-tools/pull/11; will re-open in this repo once the target branch is merged

PR #875, opened 2023-08-20
soroban-rpc: refactor libpreflight

2opremio:
* Split out FFI conversions from actual preflight computations. I added a new file `preflight.rs` which contains the actual preflight computations, separate from the `Go<->Rust` ffi interfacing (which is the only thing left in `lib.rs`).
* Use the [`anyhow`](https://docs.rs/anyhow/latest/anyhow/) crate to:
   * Make error handling more readable (`Result<R, Box<dyn Error>>` vs `Result<R>`)
   * Add context to errors ( Closes https://github.com/stellar/soroban-tools/issues/858).
* Some other minor changes, making the code more idiomatic/readable.
* Fix a couple of memory leaks in `free_preflight_result()`

PR #878, opened 2023-08-21
soroban-rpc: simulateTransaction: try to return diagnostic events on failure

2opremio:
Even if `host.invoke_host_function()` fails, leading to an error, we can collect and return diagnostic events.

Note that this implies that the events field of the simulate transaction response can be set when an error is returned.

Also, sneak in some review feedback from #875 which I incorrectly pushed to this branch (sorry @tsachiherman )

(Found out about it in https://github.com/stellar/rs-soroban-env/issues/1008 ).

@willemneal we should probably do the same in the CLI's snapshot mode (in case we don't do it already).

### DIFF
--- a/cmd/soroban-rpc/internal/methods/simulate_transaction.go
+++ b/cmd/soroban-rpc/internal/methods/simulate_transaction.go
@@ -35,11 +35,11 @@ type RestorePreamble struct {
 
 type SimulateTransactionResponse struct {
 	Error           string                       `json:"error,omitempty"`
-	TransactionData string                       `json:"transactionData"` // SorobanTransactionData XDR in base64
-	MinResourceFee  int64                        `json:"minResourceFee,string"`
+	TransactionData string                       `json:"transactionData,omitempty"` // SorobanTransactionData XDR in base64
+	MinResourceFee  int64                        `json:"minResourceFee,string,omitempty"`
 	Events          []string                     `json:"events,omitempty"`          // DiagnosticEvent XDR in base64
 	Results         []SimulateHostFunctionResult `json:"results,omitempty"`         // an array of the individual host function call results
-	Cost            SimulateTransactionCost      `json:"cost"`                      // the effective cpu and memory cost of the invoked transaction execution.
+	Cost            SimulateTransactionCost      `json:"cost,omitempty"`            // the effective cpu and memory cost of the invoked transaction execution.
 	RestorePreamble RestorePreamble              `json:"restorePreamble,omitempty"` // If present, it indicates that a prior RestoreFootprint is required
 	LatestLedger    int64                        `json:"latestLedger,string"`
 }
@@ -136,6 +136,7 @@ func NewSimulateTransactionHandler(logger *log.Entry, ledgerEntryReader db.Ledge
 		}
 
 		return SimulateTransactionResponse{
+			Error:           result.Error,
 			Results:         results,
 			Events:          result.Events,
 			TransactionData: result.TransactionData,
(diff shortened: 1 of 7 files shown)


willemneal:
I have another PR that is targeting the current futurenet branch: https://github.com/AhaLabs/soroban-tools/pull/8

But it should log all simulation events.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -35,11 +35,11 @@ type RestorePreamble struct {
...
 type SimulateTransactionResponse struct {
 	Error           string                       `json:"error,omitempty"`
-	TransactionData string                       `json:"transactionData"` // SorobanTransactionData XDR in base64
-	MinResourceFee  int64                        `json:"minResourceFee,string"`
+	TransactionData string                       `json:"transactionData,omitempty"` // SorobanTransactionData XDR in base64
+	MinResourceFee  int64                        `json:"minResourceFee,string,omitempty"`
 	Events          []string                     `json:"events,omitempty"`          // DiagnosticEvent XDR in base64
 	Results         []SimulateHostFunctionResult `json:"results,omitempty"`         // an array of the individual host function call results
-	Cost            SimulateTransactionCost      `json:"cost"`                      // the effective cpu and memory cost of the invoked transaction execution.
+	Cost            SimulateTransactionCost      `json:"cost,omitempty"`            // the effective cpu and memory cost of the invoked transaction execution.
```
Shaptic, line 42: might be missing something here: won't this always be present since it's a struct? (and same for `RestorePreamble`, below, actually)

Shaptic (approved):
all for this :clap: the more error feedback to clients, the better

cmd/soroban-rpc/internal/methods/simulate_transaction.go
tsachiherman: unless the operation execution was successful, there wouldn't be any `CPUInstructions` or `MemoryBytes`, wouldn't it ?

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -526,6 +526,10 @@ func TestSimulateTransactionError(t *testing.T) {
 	result := simulateTransactionFromTxParams(t, client, params)
 	assert.Greater(t, result.LatestLedger, int64(0))
 	assert.Contains(t, result.Error, "MissingValue")
+	require.Len(t, result.Events, 1)
+	var event xdr.DiagnosticEvent
+	require.NoError(t, xdr.SafeUnmarshalBase64(result.Events[0], &event))
+	fmt.Println(event)
```
tsachiherman, line 532: could you change the `fmt.Println` into `t.Log` ? 
i.e. so that in the success case, it won't write entries to the output.

tsachiherman (changes requested):
one tiny request, looks great otherwise.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: I will just remove the print altogether

cmd/soroban-rpc/internal/methods/simulate_transaction.go
2opremio: If there is an error there is no point in sending cost.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
2opremio: done, ptal

Issue #858, opened 2023-08-14
soroban-rpc: Improve error reporting in preflight library


2opremio:
Namely due to the generalized use of `?`  in Rust, we forward errors without providing any further context on what's wrong.

PR #879, opened 2023-08-21
soroban-rpc: libpreflight: unwrap() errors due to bugs in the Go<->Rust interface

2opremio:
I understand this can be controversial, but there are good reasons for it. I intentionally created a separate followup PR from #875 in case it meets a strong opposition.

* panicking will show the line number where the problem happens whereas otherwise we will get an error with no context at all (as an alternative we could add context to every use of `?` but that would make the code much more unreadable).

* We anyhow capture the panics before returning to Go and thus, the process won't stop.

Further context on why it's OK to use unwrap in these cases https://blog.burntsushi.net/unwrap/

Ideally we would simply have a way to selectively attach a backtrace/line-number to these errors instead, but I don't think that will be possible until https://github.com/rust-lang/rust/issues/53487 lands. Also, even then I am not sure we will be able to pick when we want to show backtraces/line numbers (it seems to be all or nothing).

### DIFF
--- a/cmd/soroban-rpc/lib/preflight/src/lib.rs
+++ b/cmd/soroban-rpc/lib/preflight/src/lib.rs
@@ -16,7 +16,6 @@ use soroban_env_host::xdr::{
     AccountId, InvokeHostFunctionOp, LedgerFootprint, OperationBody, ReadXdr, WriteXdr,
 };
 use soroban_env_host::LedgerInfo;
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
👏🏻

2opremio:
This also addressed part of https://github.com/stellar/soroban-tools/issues/858

Issue #872, opened 2023-08-18
XDR Update : SC_SPEC_TYPE_SET, INSUFFICIENT_REFUNDABLE_FEE, contractEventSizeBytes


tsachiherman:
### Update support for changes in XDR

* Remove SC_SPEC_TYPE_SET ( https://github.com/stellar/soroban-tools/pull/884 )
* Add new error code INVOKE_HOST_FUNCTION_INSUFFICIENT_REFUNDABLE_FEE
* remove contractEventSizeBytes

PR #857, opened 2023-08-11
Update to remove ScSpecTypeSet

graydon:
This is a sketch of the change necessary to adapt to the removal of ScSpecTypeSet upstream. It's incomplete and needs to be further updated with proper git refs as upstream revs land.

Reflects https://github.com/stellar/rs-soroban-sdk/pull/1054 which reflects https://github.com/stellar/rs-stellar-xdr/pull/291 which reflects https://github.com/stellar/stellar-xdr/pull/136 which addresses https://github.com/stellar/stellar-xdr/issues/135 which resulted from https://github.com/stellar/rs-soroban-sdk/pull/979

PR #884, opened 2023-08-22
Update to remove ScSpecTypeSet

2opremio:
Second attempt at #857 , which was reverted due to breaking CI

Parent issue : https://github.com/stellar/soroban-tools/issues/872


paulbellamy (approved):

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -54,7 +50,8 @@ pub(crate) fn compute_host_function_transaction_data_and_min_fee(
             &soroban_resources.footprint,
         )
         .context("cannot estimate maximum transaction size")?,
-        contract_events_size_bytes: soroban_resources.contract_events_size_bytes,
+        contract_events_size_bytes: calculate_event_size_bytes(events)
```
dmkozh, line 53: Events now also include the invocation return value size (naming is a bit misleading, but that's change from before we changed any XDR and I'm not sure it's worth going through the refactoring at this point).

dmkozh, line 53: Not related to the current failures, but you should calculate the size of the contract events only, not all the diagnostic events (i.e. you need to filter based on the event type in that function).

2opremio: You mean the size of the `SCVal` returned by the function? That's really unintuitive

Also, it would had been nice to learn about these sort of changes in advance.

2opremio: Ah, good to know.

2opremio:
@sreuland note that this PR caused an update in the XDR, making the javascript client temporarily incompatible and impacting your system test fixes.

sreuland:
> @sreuland note that this PR caused an update in the XDR, making the javascript client temporarily incompatible and impacting your system test fixes.

thanks, will merge main into https://github.com/stellar/soroban-tools/pull/855 and continue to see how far e2e gets on latest, have been working through cli/rpc/preflight interop. 

do you think this is the last xdr updates, this looks great, as it has updated ver refs of everything in here, go, rs-sdk, rs-env, and core build.

2opremio:
Unfortunately there will be more XDR changes (the state expiration ones)

PR #887, opened 2023-08-22
fix: address feedback; name comparison; no Path test

willemneal:


tsachiherman (approved):

tsachiherman:
@willemneal if you could resolve the conflicts and get this merged in, that would be great.

PR #888, opened 2023-08-22
soroban-rpc: improve HTTPRequestDurationLimiter by adding a recover handling

tsachiherman:
### What

HTTPRequestDurationLimiter is using an internal go routine in order to call the downstream ServeHTTP.
If that ServeHTTP method panics, the server would crash.

### Why

This PR makes sure that we have panic handler on the ServeHTTP method call. If this method fail, a debug output is written to the log.

### Known limitations

n/a

### DIFF
--- a/cmd/soroban-rpc/internal/network/requestDurationLimiter.go
+++ b/cmd/soroban-rpc/internal/network/requestDurationLimiter.go
@@ -2,7 +2,10 @@ package network
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"runtime/debug"
+	"strings"
 	"time"
 
 	"github.com/creachadair/jrpc2"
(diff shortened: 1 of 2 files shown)


cmd/soroban-rpc/internal/network/requestDurationLimiter.go
```diff
@@ -132,14 +135,36 @@ func (q *httpRequestDurationLimiter) ServeHTTP(res http.ResponseWriter, req *htt
...
 		limitCh = time.NewTimer(q.limitThreshold).C
 	}
-	requestCompleted := make(chan interface{}, 1)
+	requestCompleted := make(chan []string, 1)
 	requestCtx, requestCtxCancel := context.WithTimeout(req.Context(), q.limitThreshold)
 	defer requestCtxCancel()
 	timeLimitedRequest := req.WithContext(requestCtx)
 	responseBuffer := makeBufferedResponseWriter(res)
 	go func() {
+		defer func() {
```
sreuland, line 144: this seems like a general panic recover utility method, is it worth pulling this inline func up to a package private method instead like `handlePanic(finally chan []string)`, might help readability here?

cmd/soroban-rpc/internal/network/requestDurationLimiter_test.go
```diff
@@ -305,3 +305,35 @@ func TestJRPCRequestDurationLimiter_NoLimiting_Warn(t *testing.T) {
 	require.Equal(t, [7]int{0, 0, 0, 0, 1, 0, 0}, logCounter.writtenLogEntries)
 	shutdown()
 }
+
+func TestHTTPRequestDurationLimiter_Panicing(t *testing.T) {
```
sreuland, line 309: great test coverage on these corners!

sreuland (approved):

2opremio:
Could I ask to normalize the filenames on the `net` directory while you are at it?

tsachiherman:
> Could I ask to normalize the filenames on the `net` directory while you are at it?

sure.

cmd/soroban-rpc/internal/network/requestDurationLimiter.go
tsachiherman: I think that it makes sense. Let me merge https://github.com/stellar/soroban-tools/pull/856 first, and I'll rebase this & refactor to implement that change.

PR #895, opened 2023-08-23
soroban-rpc: GetPreflight benchmark tweaks

2opremio:
* ~Add walllock benchmark metric~
* Exclude db startup from the benchmark

### DIFF
--- a/cmd/soroban-rpc/internal/preflight/preflight_test.go
+++ b/cmd/soroban-rpc/internal/preflight/preflight_test.go
@@ -362,6 +362,8 @@ func benchmark(b *testing.B, config benchmarkConfig) {
 			disableCache: config.useDB.disableCache,
 		}
 	}
+
+	b.ResetTimer()
 	b.StopTimer()
 	for i := 0; i < b.N; i++ {
 		params := getPreflightParameters(b, dbConfig)


cmd/soroban-rpc/internal/preflight/preflight_test.go
```diff
@@ -362,18 +363,23 @@ func benchmark(b *testing.B, config benchmarkConfig) {
...
+	var fullDuration time.Duration
+	b.ResetTimer()
 	b.StopTimer()
 	for i := 0; i < b.N; i++ {
 		params := getPreflightParameters(b, dbConfig)
+		start := time.Now()
 		b.StartTimer()
 		_, err := GetPreflight(context.Background(), params)
 		b.StopTimer()
+		fullDuration += time.Since(start)
```
leighmcculloch, line 375: The time metric reported by the Go benchmark package is based on the same calculation, a `time.Since` between the `time.Now` in `b.StartTimer` and `b.StopTimer`. For example: https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/testing/benchmark.go;l=127-137. Is the metric being added in this PR different in some other way?

2opremio: Thanks that's why I was noting very similar results. I was assuming it was CPU time.

I removed that part of the PR.

cmd/soroban-rpc/internal/preflight/preflight_test.go
```diff
@@ -362,6 +362,8 @@ func benchmark(b *testing.B, config benchmarkConfig) {
 			disableCache: config.useDB.disableCache,
 		}
 	}
+
+	b.ResetTimer()
 	b.StopTimer()
```
leighmcculloch, line 367: To be really accurate the stop should happen before the reset, otherwise the timer will continue to run between the reset and the stop and so won't actually start at in the first iteration of the for loop from zero.

In practice the tiny amount of time it runs for won't be meaningful for two reasons:
- The amount of time should be truly tiny on its own.
- The extra time is added outside the `b.N` loop, and so it'll be spread over hundreds of thousands of iterations essentially making it zero.

leighmcculloch (approved):
👏🏻

cmd/soroban-rpc/internal/preflight/preflight_test.go
2opremio: Uhm, that won't really as is, since there is some preamble work on the loop

2opremio: but, as you said, the difference is tiny

PR #898, opened 2023-08-24
soroban-rpc: Limit request size to 10MB

2opremio:


tsachiherman (approved):
yes, I think that this one is great. It would be even better to make the magic number ( 10M ) configurable.

PR #903, opened 2023-08-25
soroban-rpc: getLedgerEntries: set maximum number of keys to query for

2opremio:


tsachiherman (approved):

tsachiherman (approved):

tsachiherman (approved):
I think that the value, 200, is a great starting value.
Note that it would be better to have this as a configuration value.
The reasoning is that it would allow us to lower that value on production system more easily if we were to see that this is causing issue.

Issue #889, opened 2023-08-23
Fix --cost parameter on invoke command


leighmcculloch:
### What version are you using?

```
❯ soroban version
soroban 0.9.4 (76cacc7cf885bd7d37756735fab332351d893406)
soroban-env 0.0.17 (400d806387140553e4e685d232deb3a807ec0e36)
soroban-env interface version 85899345971
stellar-xdr 0.0.17 (0f16673441898162c9996da6117be2280ef8fd84)
xdr next (e372df9f677961aac04c5a4cc80a3667f310b29f)
```

### What did you do?

```
soroban contract invoke --id 1 --wasm file.wasm --cost -- hello --to me
```

### What did you expect to see?

I expected to see the cost table outputted, that looks like this:
```
=====================================================================================================================================================================
Cpu limit: 40000000; used: 914627
Mem limit: 52428800; used: 141947
=====================================================================================================================================================================
CostType                 iterations     input          cpu_insns      mem_bytes      const_term_cpu      lin_term_cpu        const_term_mem      lin_term_mem        
WasmInsnExec             344            None           2408           0              7                   0                   0                   0                   
WasmMemAlloc             0              None           0              0              0                   0                   1                   0                   
HostMemAlloc             10             Some(1006)     23500          1086           2350                0                   8                   1                   
HostMemCpy               19             Some(1142)     437            0              23                  0                   0                   0                   
HostMemCmp               7              Some(118)      419            0              43                  1                   0                   0                   
InvokeHostFunction       1              None           928            0              928                 0                   0                   0                   
VisitObject              1              None           19             0              19                  0                   0                   0                   
ValXdrConv               5              None           670            0              134                 0                   0                   0                   
ValSer                   0              Some(0)        0              0              587                 1                   9                   3                   
ValDeser                 0              Some(0)        0              0              870                 0                   4                   1                   
ComputeSha256Hash        0              Some(0)        0              0              1725                33                  40                  0                   
ComputeEd25519PubKey     0              None           0              0              25551               0                   0                   0                   
MapEntry                 29             None           1537           0              53                  0                   0                   0                   
VecEntry                 0              None           0              0              5                   0                   0                   0                   
GuardFrame               2              None           8100           944            4050                0                   472                 0                   
VerifyEd25519Sig         0              Some(0)        0              0              369634              21                  0                   0                   
VmMemRead                1              Some(16)       0              0              0                   0                   0                   0                   
[...]

PR #911, opened 2023-08-31
soroban-cli: Fix `--cost` flag

paulbellamy:
### What

Fix the `--cost` flag. In sandbox mode it logs a `Budget` object (same as before). With a real network it logs a `SorobanResources` (since that is what we have). Previously for the real network it wouldn't log anything.

It also logs the cost output if any flag is set of: `--verbose`, `--very-verbose`, or `--cost`.

### Why

Fixes https://github.com/stellar/soroban-tools/issues/889

### Known limitations

[TODO or N/A]


cmd/soroban-cli/src/log/cost.rs
```diff
@@ -0,0 +1,27 @@
+use soroban_env_host::xdr::SorobanResources;
+use std::fmt::{Debug, Display};
+
+struct Cost<'a>(&'a SorobanResources);
+
+impl Debug for Cost<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // TODO: Should we output the footprint here?
+        writeln!(f, "==================== Cost ====================")?;
+        writeln!(f, "Cpu used: {}", self.0.instructions,)?;
```
tsachiherman, line 10: `Cpu` -> `CPU`

tsachiherman (changes requested):
just one tiny request.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -262,21 +263,24 @@ impl Cmd {
...
-            self.run_in_sandbox()
+            self.run_in_sandbox(global_args)
         } else {
-            self.run_against_rpc_server().await
+            self.run_against_rpc_server(global_args).await
         }
     }
 
-    pub async fn run_against_rpc_server(&self) -> Result<String, Error> {
+    pub async fn run_against_rpc_server(
```
leighmcculloch, line 280: The RPC server when doing simulation should have access to the same budget that the sandbox execution has access to.

Why do we limit ourselves to outputting the `SorobanResource`, and not outputting the `Budget` for the simulated portion?

paulbellamy: It should have that yes. It will. But the RPC doesn't respond with that. So we could do that if we change the RPC API to expose that info, but that's probably a follow-up PR, if we want to do that.

paulbellamy: Which, frankly, if we're trying to provide that level of detail to the user, I'm not convinced the "Budget" object is the best way to do that. A profiling flame graph of heap allocation profile graph would be much more useful tools for improving contract performance, IMO.

tsachiherman (approved):

PR #912, opened 2023-08-31
build: add rust dep checker

tsachiherman:
### What

add rust dep checker

### Why

Verify we don't add references to commit hashes that would "disappear".

### Known limitations

n/a

### DIFF
--- a/.github/workflows/dependency-check.yml
+++ b/.github/workflows/dependency-check.yml
@@ -21,4 +21,5 @@ jobs:
       - run: scripts/check-dependencies.bash
   validate-rust-git-rev-deps:
     steps:
+      - uses: actions/checkout@v3
       - uses: stellar/actions/rust-check-git-rev-deps@main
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-check.yml
```diff
@@ -13,9 +13,12 @@ jobs:
...
-    - run: scripts/check-dependencies.bash
+      - uses: actions/checkout@v3
+      - run: rustup update
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.20.1
+      - run: scripts/check-dependencies.bash
+  validate-rust-git-rev-deps:
+    steps:
+      - uses: stellar/actions/rust-check-git-rev-deps@main
```
leighmcculloch, line 24: The project needs to be checked out.
```suggestion
      - uses: actions/checkout@v3
      - uses: stellar/actions/rust-check-git-rev-deps@main
```

tsachiherman: sure.

leighmcculloch (approved):

PR #916, opened 2023-09-01
soroban-rpc: get expiration ledger sequence at source

tsachiherman:
### What

The current implementation is making multiple requests between the rust and go code when fetching the ledger entries and their corresponding ledger expiration entries.

This PR merge them both, as well as return the expired ledger sequence to the getLedgerEntries endpoint.

Note that this PR is based on https://github.com/stellar/soroban-tools/pull/913/files


2opremio:
I don't know why the comments disappeared but ... in a nutshell: I don't think it makes sense to implement this without hiding the expiration ledger entries from the end user. Otherwise it's redundant.

In addition as it is today, the user can request expiration entries. The hashing can be provided by a library.

Finally, as a UX improvement and library optimization, this isn't in the critical line for the testnet release. Let's please park it and focus on testing until then.

2opremio (changes requested):
As per my comments above, I don't think we should allow accessing raw Expiration ledger entries and providing the expiration in a separate way. I would go one way or the other.

tsachiherman:
per @2opremio 's request, I've restricted the getledgerentries endpoint to block any request for expiration ledger entries.

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -23,8 +24,9 @@ type LedgerEntryReader interface {
 }
 
 type LedgerKeyAndEntry struct {
-	Key   xdr.LedgerKey
-	Entry xdr.LedgerEntry
+	Key                 xdr.LedgerKey
+	Entry               xdr.LedgerEntry
+	ExpirationLedgerSeq *uint32 // optional expiration ledger seq, whne applicable.
```
willemneal, line 29:
```suggestion
	ExpirationLedgerSeq *uint32 // optional expiration ledger seq, when applicable.
```

cmd/soroban-rpc/internal/methods/get_ledger_entries.go
```diff
@@ -64,6 +66,14 @@ func NewGetLedgerEntriesHandler(logger *log.Entry, ledgerEntryReader db.LedgerEn
 					Message: fmt.Sprintf("cannot unmarshal key value %s at index %d", requestKey, i),
 				}
 			}
+			if ledgerKey.Type == xdr.LedgerEntryTypeExpiration {
```
sreuland, line 69: is there a test case to assert this error outcome for request of expiration entry key, it looks like `simulate_transaction_test.go` hits getLedgerEntry/Entries and verifies ExpirationLedger values, but didn't see attempts at this path.

tsachiherman: Ahh.. you're asking if there is a unit test that verifies that expiration ledger keys cannot be requested. Hmm.. no.
We had few such tests that were "fixed" to avoid the error, but I wasn't considering that we should keep these as false positive tests.
I'll add one of these back, just for ensuring complete coverage.

cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -46,7 +46,8 @@ func SnapshotSourceGet(handle C.uintptr_t, cLedgerKey C.xdr_t) C.xdr_t {
 	if err := xdr.SafeUnmarshal(ledgerKeyXDR, &ledgerKey); err != nil {
 		panic(err)
 	}
-	present, entry, err := db.GetLedgerEntry(h.readTx, ledgerKey)
+	// TODO : the expiration sequence here is being ignored for now; it should be passed downstream.
```
sreuland, line 49: if this is out of scope for https://github.com/stellar/soroban-tools/issues/1010, should a new ticket be created to capture this gap of downstream use case of LE expiration seq for tracking and also to reference here, i.e. it can help remove the TODO later?

tsachiherman: yes, I think that it makes sense to create another ticket. As is, it's unblocking the js/cli, so it should good to go.
This comments only suggest that the simulateTransaction performance could improve further here.

tsachiherman: created issue : https://github.com/stellar/soroban-tools/issues/1025

cmd/soroban-rpc/internal/test/cli_test.go
```diff
@@ -211,13 +213,40 @@ func TestCLIRestore(t *testing.T) {
...
+	ledgerExpirationEntryB64, err := xdr.MarshalBase64(ledgerExpirationEntry)
+	require.NoError(t, err)
+	var getLedgerEntryResult methods.GetLedgerEntryResponse
+	err = client.CallResult(context.Background(), "getLedgerEntry", methods.GetLedgerEntryRequest{
+		Key: ledgerExpirationEntryB64,
+	}, &getLedgerEntryResult)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), methods.ErrLedgerExpirationEntriesCannotBeQueriedDirectly)
+
+	// repeat with getLedgerEntries
```
sreuland, line 231: is this verification of rpc `getLedgerEntry/Entries` to check `ErrLedgerExpirationEntriesCannotBeQueriedDirectly`  related to the cli restore test, it's fine here, just seems like it is a separate test appended to the end, noticed there is also a get_ledger_entries_test.go, maybe this could be a test in there.

I was looking to see if the nil case of LedgerEntryResult.ExpirationLedger is validated, i.e. if you request an Account's LedgerEntry, the go response should have LedgerEntryResult.ExpirationLedger=nil and the json response should not have `expirationLedgerSeq ` present, correct?

sreuland (approved):
looks good, I'm light on rust side, on rpc side, left one other comment for consideration related to test coverage of expirationLedgerSeq=nil, such as when getLedgerEntries on a non-expirable LE like Account.

willemneal (approved):

cmd/soroban-rpc/internal/test/cli_test.go
tsachiherman: That's correct for today. It might change in the future, and completely depends on the txmeta.

I'm not sure how to (easily) cover programmatically all the ledger entry types.

tsachiherman:
I'm going to merge this for now, but open to additional unit testing ideas.

2opremio:
@tsachiherman did you modify the docs to include the API changes?

tsachiherman:
> @tsachiherman did you modify the docs to include the API changes?

not yet. I will follow up on that. I wanted to prioritize merging this one in so it won't block the js clients.

PR #922, opened 2023-09-04
soroban-rpc: preflight: remove key sizes from read_bytes computation

2opremio:
Core stopped including including them

Followup from https://github.com/stellar/soroban-tools/pull/913#discussion_r1313608612

### DIFF
--- a/cmd/soroban-rpc/lib/preflight/src/fees.rs
+++ b/cmd/soroban-rpc/lib/preflight/src/fees.rs
@@ -23,9 +23,9 @@ use state_expiration::{get_restored_ledger_sequence, ExpirableLedgerEntry};
 use std::cmp::max;
 use std::convert::{TryFrom, TryInto};
 
-/// Estimate for any `ExpirationEntry` ledger entry, consisting of a 32-byte
-/// hash of the corresponding entry and 4 bytes for expiration ledger.
-const EXPIRATION_ENTRY_SIZE: u32 = 32 + 4;
+// TODO: this should be imported from soroban_env_host::fees instead
+/// Serialize XDR size for any `ExpirationEntry` ledger entry.
+const EXPIRATION_ENTRY_SIZE: u32 = 48;
 
 pub(crate) fn compute_host_function_transaction_data_and_min_fee(
     op: &InvokeHostFunctionOp,
--- a/cmd/soroban-rpc/lib/preflight/src/fees.rs
+++ b/cmd/soroban-rpc/lib/preflight/src/fees.rs
@@ -37,10 +37,8 @@ pub(crate) fn compute_host_function_transaction_data_and_min_fee(
     bucket_list_size: u64,
     current_ledger_seq: u32,
 ) -> Result<(SorobanTransactionData, i64)> {
-    // TODO: is this OK?
-    let init_expiration_entries = ExpirationEntryMap::new();
     let ledger_changes =
-        get_ledger_changes(budget, post_storage, pre_storage, init_expiration_entries)?;
+        get_ledger_changes(budget, post_storage, pre_storage, ExpirationEntryMap::new())?;
     let soroban_resources =
         calculate_host_function_soroban_resources(&ledger_changes, &post_storage.footprint, budget)
             .context("cannot compute host function resources")?;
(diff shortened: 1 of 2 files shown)


tsachiherman (approved):

2opremio:
This has unveiled yet another `ResourceLimitExceeded` error when running the `BumpFootprint` operation. The test fails at https://github.com/2opremio/soroban-cli/blob/remove-keys-from-fee-estimation/cmd/soroban-rpc/internal/test/simulate_transaction_test.go#L748

envelope: `AAAAAgAAAABzdv3ojkzWHMD7KUoXhrPx0GH18vHKV0ZfqpMiEblG1gAAZqAAAAAAAAAABAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAGQAAAAAAAAAUAAAAAQAAAAAAAAABAAAABgAAAAFnrAIfzK644FciM7c4L1q7oxCN5UdDC/Y0lL2onFdNaAAAAA8AAAAHQ09VTlRFUgAAAAABAAAAAAAAAAAAAAB0AAAAAAAAAAAAAE5FAAAAARG5RtYAAABAGAKEmnKibE8ikaCWwDf34EbFh3OeWSWc4gnsF0RY8B8+xLuOsB/EgHjf1Ns1/r0WssqDrf3P8zO2H3V9k/8pAA==`

The soroban transaction data is the following. My guess is that we are not accounting for the written expiration entries. I will try to fix that. 
Screenshot 2023-09-04 at 19 01 00

2opremio:
Unfortunately, adding the write bytes didn't make the test pass. Here is the new soroban transaction data:

Screenshot 2023-09-04 at 19 15 48

2opremio:
Now, thinking about it, the write_bytes couldn't be the problem since they were already 0 before this PR.

2opremio:
If you set `read_bytes` to `128` the test passes but it fails if set to `127` so we are missing `128 - 116 = 12` bytes which I am not sure how to account for.

2opremio:
The serialized length of an expiration ledger entry is 48, and not 36. This explains the missing 12 bytes :S 

I noticed that `fees.rs` in `rs-soroban-env` also gets it wrong (@sisuresh I guess you copy-pased it from there) so it should be corrected.

2opremio:
To be more specific. This prints 48:

```
	entry2 := xdr.LedgerEntry{
		LastModifiedLedgerSeq: 4,
		Data: xdr.LedgerEntryData{
			Type: xdr.LedgerEntryTypeExpiration,
			Expiration: &xdr.ExpirationEntry{
				KeyHash:             xdr.Hash{},
				ExpirationLedgerSeq: 0,
			},
		},
		Ext: xdr.LedgerEntryExt{},
	}
	bin, err := entry.MarshalBinary()
	assert.NoError(t, err)
	fmt.Println(len(bin))
```

cmd/soroban-rpc/lib/preflight/src/fees.rs
```diff
@@ -23,9 +23,9 @@ use state_expiration::{get_restored_ledger_sequence, ExpirableLedgerEntry};
 use std::cmp::max;
 use std::convert::{TryFrom, TryInto};
 
-/// Estimate for any `ExpirationEntry` ledger entry, consisting of a 32-byte
-/// hash of the corresponding entry and 4 bytes for expiration ledger.
-const EXPIRATION_ENTRY_SIZE: u32 = 32 + 4;
+// TODO: this should be imported from soroban_env_host::fees instead
+/// Serialize XDR size for any `ExpirationEntry` ledger entry.
+const EXPIRATION_ENTRY_SIZE: u32 = 48;
```
tsachiherman, line 28: It would be nice to have a unit test that validate that magic number - but not a blocker.

tsachiherman (approved):

cmd/soroban-rpc/lib/preflight/src/fees.rs
2opremio: That should be in the upstream library, no here (check the TODO)

2opremio: I added a test at https://github.com/stellar/rs-soroban-env/pull/1034

2opremio:
This is ready to merge, but let's not do it until @sisuresh / @dmkozh confirm  that the `EXPIRATION_ENTRY_SIZE` fix is correct.

Issue #920, opened 2023-09-01
soroban-cli: should return non-zero exit code when can't find data


leighmcculloch:
### What version are you using?

- stellar/soroban-tools@b1519a4d90502ce3155960541b2a98c88f3dd043 (rpc and cli)
- stellar/stellar-core@22b9bb38416e5ce058eb9c1fadfdd414a4661833
- stellar/rs-soroban-env@eb5a9ba053a7b64a8eff605db625525378f7bea0

### What did you do?

```console
❯ soroban contract read --network local --source me --id CAW7SH5JG45U6YLCFB6YD2KVSOZQHMNEJQPN6KDZOCGSXWR72ZIFBBMP --key abc
```

### What did you expect to see?

An error because no contract data or ledger entry is stored with that key.

### What did you see instead?

Nothing. No output.

### Discussion

It's difficult to compose tools that don't exit with a non-zero code in failure cases.

From a UX point-of-view the user can be left feeling uneasy when a tool returns or outputs nothing.
```[tasklist]
### Tasks
- [ ] https://github.com/stellar/soroban-tools/pull/923
- [ ] https://github.com/stellar/soroban-tools/pull/924
```

willemneal:
The exit code issue is also noted here: https://github.com/stellar/soroban-tools/issues/774 

Currently in the main.rs when doing a `run` it catches the error and prints it, but doesn't exit with a non-zero code. So this seems like two issues. I agree that there should be a proper error, but I think we should also exit with a non-zero so that scripts and other programs invoking the CLI will know there has been an error.

tsachiherman:
I agree with @willemneal here. The fact that the empty group isn't being reported as an error in this particular case is an issue - but not the *same* issue not setting the exit code to 1. Let's tackle these separately ?

PR #924, opened 2023-09-04
soroban-cli: return an error once `contract read` is unable to read any entry.

tsachiherman:
### What

return an error once `contract read` is unable to read any entry.

### Why

https://github.com/stellar/soroban-tools/issues/920

### Known limitations

n/a


cmd/soroban-cli/src/commands/contract/read.rs
```diff
@@ -94,6 +94,8 @@ pub enum Error {
     // TODO: the Display impl of host errors is pretty user-unfriendly
     //       (it just calls Debug). I think we can do better than that
     Host(#[from] HostError),
+    #[error("no ledger key was found for given contract")]
```
leighmcculloch, line 97: "no ledger key" is a little confusing because a developer will have provided a contract data key, and the error sounds like the user has failed to provide one, when in fact no ledger entry can be found that matches the key.

An error that focuses on the fact that no data can be found matching the key would be clearer.

tsachiherman: Hmm. Yes, that's true. I'll update that tomorrow morning.

leighmcculloch (approved):

PR #926, opened 2023-09-05
soroban-rpc: Ingest temp ledger entry evictions

tamirms:
### What

In https://github.com/stellar/go/pull/5037 temporary ledger entries which are evicted are excluded from the change reader. So, we need to handle the deletions explicitly by the EvictedTemporaryLedgerKeys() method on the LedgerCloseMeta struct.

### Known limitations

We aren't able to test eviction of temporary ledger keys through an integration test because stellar-core needs to implement a testing flag to allow evictions to occur more quickly. So instead I have implemented a unit test where I have hard coded a ledger which contains temporary ledger entry evictions.

### DIFF
--- a/cmd/soroban-rpc/internal/ingest/service.go
+++ b/cmd/soroban-rpc/internal/ingest/service.go
@@ -43,10 +43,12 @@ type Config struct {
 }
 
 func NewService(cfg Config) *Service {
-	return newService(cfg, true)
+	service := newService(cfg)
+	startService(service, cfg)
+	return service
 }
 
-func newService(cfg Config, start bool) *Service {
+func newService(cfg Config) *Service {
 	// ingestionDurationMetric is a metric for measuring the latency of ingestion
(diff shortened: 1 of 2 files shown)


2opremio:
Also, I am missing a test.

cmd/soroban-rpc/internal/ingest/service.go
```diff
@@ -81,29 +85,32 @@ func NewService(cfg Config) *Service {
...
-				}
-				return err
-			},
-			contextBackoff,
-			cfg.OnIngestionRetry)
-		if err != nil && !errors.Is(err, context.Canceled) {
-			service.logger.WithError(err).Fatal("could not run ingestion")
-		}
-	})
+	if start {
```
tsachiherman, line 88: I'd have broken that into `newService` and `startService`, but it's ok like that too.

tsachiherman (approved):
overall looks good. I did had quite many comments, but all of these were non-blocking.

PR #937, opened 2023-09-07
soroban-cli: fix contract deploy

tsachiherman:
this is probably going to be scrapped out.


2opremio:
Can you add a test showcasing what was wrong? I added deploy tests at https://github.com/stellar/soroban-tools/pull/940 which are working

tsachiherman:
> Can you add a test showcasing what was wrong? I added deploy tests at #940 which are working

I believe @tamirms fixed the culprit in https://github.com/stellar/soroban-tools/pull/939, but that might mean that in `build_install_contract_code_tx` we should remove the `source_account` from the operation.

willemneal:
I think that https://github.com/stellar/soroban-tools/pull/942 fixes it.

PR #939, opened 2023-09-08
soroban-rpc: Fix set_authorization_entries bug in preflight

tamirms:
### What

Configure the host source account before calling `host.set_authorization_entries()`.

### Why

`host.set_authorization_entries()` will fail if the source account is not set in the source. This is the specific error which will trigger:

https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-host/src/auth.rs#L1366

### Known limitations

[N/A]

### DIFF
--- a/cmd/soroban-rpc/internal/test/simulate_transaction_test.go
+++ b/cmd/soroban-rpc/internal/test/simulate_transaction_test.go
@@ -36,7 +36,7 @@ var (
 func getHelloWorldContract(t *testing.T) []byte {
 	_, filename, _, _ := runtime.Caller(0)
 	testDirName := path.Dir(filename)
-	contractFile := path.Join(testDirName, "../../../../target/wasm32-unknown-unknown/test-wasms/test_hello_world.wasm")
+	contractFile := path.Join(testDirName, helloWorldContractPath)
 	ret, err := os.ReadFile(contractFile)
 	if err != nil {
 		t.Fatalf("unable to read test_hello_world.wasm (%v) please run `make build-test-wasms` at the project root directory", err)
--- a/cmd/soroban-rpc/internal/test/cli_test.go
+++ b/cmd/soroban-rpc/internal/test/cli_test.go
@@ -13,20 +13,45 @@ import (
 	"github.com/stellar/go/keypair"
 	"github.com/stellar/go/txnbuild"
 	"github.com/stellar/go/xdr"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestInstallContractWithCLI(t *testing.T) {
+func TestCLIContractInstall(t *testing.T) {
 	NewCLITest(t)
-	output, err := runCLICommand(t, "contract install --wasm ../../../../target/wasm32-unknown-unknown/test-wasms/test_hello_world.wasm")
-	require.NoError(t, err)
+	output, err := runCLICommand("contract install --wasm " + helloWorldContractPath)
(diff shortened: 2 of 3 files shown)


cmd/soroban-rpc/lib/preflight/src/preflight.rs
```diff
@@ -63,13 +69,6 @@ pub(crate) fn preflight_invoke_hf_op(
             .context("cannot set authorization entries")?;
```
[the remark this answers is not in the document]
2opremio: Ahhh, this is also why we weren't getting debug info!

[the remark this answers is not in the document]
2opremio: Good catch!

2opremio (approved):

cmd/soroban-rpc/lib/preflight/src/preflight.rs
```diff
@@ -46,6 +46,12 @@ pub(crate) fn preflight_invoke_hf_op(
         .context("cannot create budget")?;
     let storage = Storage::with_recording_footprint(ledger_storage_rc.clone());
     let host = Host::with_storage_and_budget(storage, budget);
+    host.set_source_account(source_account.clone())
```
tsachiherman, line 49: The source account passed in here is either the operation source account or the transaction source account, right ? and that logic aligns with the core's implementation, right ? @sisuresh could you confirm that ?

tsachiherman (approved):

cmd/soroban-rpc/lib/preflight/src/preflight.rs
sisuresh: Yeah that aligns with what core passes to the host - https://github.com/stellar/stellar-core/blob/23f3cfb821d79f3958b615e0404d6f6ff446cdaa/src/transactions/OperationFrame.cpp#L220.

cmd/soroban-rpc/internal/test/simulate_transaction_test.go
```diff
@@ -306,6 +306,64 @@ func TestSimulateTransactionSucceeds(t *testing.T) {
...
+	assert.NoError(t, xdr.SafeUnmarshalBase64(response.Results[0].Auth[0], &auth))
+	require.Equal(t, auth.Credentials.Type, xdr.SorobanCredentialsTypeSorobanCredentialsSourceAccount)
+	deployContractOp.Auth = append(deployContractOp.Auth, auth)
+	deployContractParams.Operations = []txnbuild.Operation{deployContractOp}
+
+	// preflight deployContractOp with auth
+	deployContractParams = preflightTransactionParams(t, client, deployContractParams)
+	tx, err = txnbuild.NewTransaction(deployContractParams)
+	assert.NoError(t, err)
+	sendSuccessfulTransaction(t, client, sourceAccount, tx)
```
sreuland, line 364: was kind of expecting this to re-run simulation with the tx now having the auth present?

```
response := simulateTransactionFromTxParams(t, client, deployContractParams)
```

tsachiherman: why ? rerunning the simulateTransaction is only needed in the case of non-source account transaction ( i.e. where you need other address to sign it )

sreuland: ah ok, cool, yeah, this test is essentially replicating the same rpc call flow that cli does during `contract deploy` for coverage.

PR #948, opened 2023-09-11
soroban-rpc: lower max http request size 

tsachiherman:
### What

Per security audit feedback, I'm adjusting the limits. 

### Why

Transaction sizes are limited anyway, there is no reason to set the number so high.

### Known limitations

n/a


leighmcculloch (approved):

cmd/soroban-rpc/internal/jsonrpc.go
```diff
@@ -25,6 +25,11 @@ import (
 	"github.com/stellar/soroban-tools/cmd/soroban-rpc/internal/transactions"
 )
 
+// maxHTTPRequestSize defines the largest request that the http
+// would be willing to accept. The implementation uses the default
+// MaxBytesHandler to limit the request size.
```
2opremio, line 30: I think that the comment is a bit redundant. The identifier is self-descriptive.

Regardless, what do you mean by _the http_ ? You mean _the http *server*_?

tsachiherman: I think that the const name is quite self-explanatory.. the size of the request that's being passed over the http protocol. Clearly, that includes some http-specific artifacts, like http headers, and such.

tsachiherman: oh - I see. you meant the "http" in the variable description. yes, I'll update that.

PR #950, opened 2023-09-11
cli: Remove explicit decoding depths

2opremio:


leighmcculloch (approved):
Good call.

leighmcculloch (commented):
Actually, maybe we should make this an option? i.e. The default being limits in place for untrusted data, but dev can switch it off in other cases?

2opremio:
> Actually, maybe we should make this an option? i.e. The default being limits in place for untrusted data, but dev can switch it off in other cases?

we can do that later if need be

willemneal (approved):

PR #955, opened 2023-09-13
soroban-rpc: Fix integration tests

tamirms:
### What

The first commit fixes failing integration tests caused by a change in the [host](https://github.com/stellar/rs-soroban-env/pull/1053/files). Now, the contract upload operation will validate that the wasm is actually valid. This change in behavior broke our integration tests because in some of our tests we upload contracts with a junk binary payload.

The second commit fixes TestSimulateTransactionSucceeds by updating the fee related fields in `expectedTransactionData`.  I'm not sure if this is the appropriate fix or if this is a sign that we need to update the preflight lib to be consistent with core.

### Known limitations

[N/A]


tsachiherman (approved):
The name of the test, `TestSimulateTransactionSucceeds` is abit misleading, because it's not the invocation of the contract but rather the installation.
The numeric figures you've updated aligns with the actual contract - so it's not a surprize these got bumped up.

PR #956, opened 2023-09-13
Fix dependency checker GitHub workflow

2opremio:
https://github.com/stellar/soroban-tools/pull/912 broke it two weeks ago and it wasn't noticed :S

### DIFF
--- a/.github/workflows/dependency-check.yml
+++ b/.github/workflows/dependency-check.yml
@@ -20,6 +20,7 @@ jobs:
           go-version: 1.20.1
       - run: scripts/check-dependencies.bash
   validate-rust-git-rev-deps:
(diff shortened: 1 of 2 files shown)


2opremio:
In fact, the dependencies are broken :(

```
Go and Rust dependencies are using different revisions of https://github.com/stellar/stellar-xdr

Rust dependencies are using commit 440dc9512b6e72cf84965641c5eb495d6043ed73
Go dependencies are using commit 9ac02641139e6717924fdad716f6e958d0168491
```

tamirms (approved):

2opremio:
It turns out we just had to check the `curr` xdr commit of rs-stellar-xdr instead of next.

Now the check works

Issue #959, opened 2023-09-13
soroban-rpc: data race found by TestCLIContractDeploy


tamirms:
In https://github.com/stellar/soroban-tools/actions/runs/6174861577/job/16760426412 (which was triggered for commit https://github.com/stellar/soroban-tools/tree/06bf51fada16aa8e3ef19b76c6687a48efaaf87c in the main branch), TestCLIContractDeploy failed due to a data race:
 
```
2023-09-13T16:13:20.3583912Z ==================
2023-09-13T16:13:20.3584190Z WARNING: DATA RACE
2023-09-13T16:13:20.3584515Z Write at 0x00c0016fe180 by goroutine 3029:
2023-09-13T16:13:20.3584854Z   bufio.(*Writer).Reset()
2023-09-13T16:13:20.3585415Z       /opt/hostedtoolcache/go/1.20.1/x64/src/bufio/bufio.go:615 +0x110
2023-09-13T16:13:20.3585820Z   net/http.newBufioWriterSize()
2023-09-13T16:13:20.3586385Z       /opt/hostedtoolcache/go/1.20.1/x64/src/net/http/server.go:840 +0x94
2023-09-13T16:13:20.3587133Z   net/http.(*conn).readRequest()
2023-09-13T16:13:20.3587902Z       /opt/hostedtoolcache/go/1.20.1/x64/src/net/http/server.go:1057 +0x1a24
2023-09-13T16:13:20.3588302Z   net/http.(*conn).serve()
2023-09-13T16:13:20.3588844Z       /opt/hostedtoolcache/go/1.20.1/x64/src/net/http/server.go:1920 +0x5d3
2023-09-13T16:13:20.3589247Z   net/http.(*Server).Serve.func3()
2023-09-13T16:13:20.3589799Z       /opt/hostedtoolcache/go/1.20.1/x64/src/net/http/server.go:3089 +0x58
2023-09-13T16:13:20.3590044Z 
2023-09-13T16:13:20.3590233Z Previous write at 0x00c0016fe180 by goroutine 3028:
2023-09-13T16:13:20.3590587Z   bufio.(*Writer).Write()
2023-09-13T16:13:20.3591282Z       /opt/hostedtoolcache/go/1.20.1/x64/src/bufio/bufio.go:668 +0x19d
2023-09-13T16:13:20.3591649Z   net/http.(*response).write()
[...]

PR #961, opened 2023-09-13
soroban-rpc: fix datarace in bufferedResponseWriter.WriteOut

tsachiherman:
### What

Fix data race #959 .

### Why

When implementing `bufferedResponseWriter.WriteOut`, I attempted to implement the writing to the output stream asynchronously, so that in case of context expiration the writing to the channel would be suspended at once.

Unfortuntally, it led to conflict with go http server library : When the `bufferedResponseWriter.WriteOut` could not complete writing to the stream in time, the method would return. However, the writing operation was still going on it's own pace.
then, the http server was attempting to write to the stream, creating a data race.

The solution is to write to the stream in smaller chunks, and continuously check the ctx.Err in order to examine when it expires.

### Known limitations

n/a

### DIFF
--- a/cmd/soroban-rpc/internal/network/requestdurationlimiter.go
+++ b/cmd/soroban-rpc/internal/network/requestdurationlimiter.go
@@ -100,24 +100,33 @@ func (w *bufferedResponseWriter) WriteOut(ctx context.Context, rw http.ResponseW
 	for k, v := range w.header {
 		headers[k] = v
 	}
-	complete := make(chan interface{})
-	go func() {
-		if len(w.buffer) == 0 {
-			if w.statusCode != 0 {
-				rw.WriteHeader(w.statusCode)
-			}
-			return
-		}
+
+	if len(w.buffer) == 0 {
 		if w.statusCode != 0 {
 			rw.WriteHeader(w.statusCode)
 		}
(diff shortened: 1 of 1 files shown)


tsachiherman:
> I would advocate to simply writing all at once. I think responses are small, so I think we can afford to surpass the deadline by the writing time.

I'd like to keep it for protection against malicious client leaving the connection open and draining it at a low rate.

2opremio:
chunks are not going to help you with that either, they can block when reading a single byte, so your solution doesn't really solve the problem

cmd/soroban-rpc/internal/network/requestdurationlimiter.go
```diff
@@ -100,24 +100,20 @@ func (w *bufferedResponseWriter) WriteOut(ctx context.Context, rw http.ResponseW
...
-	complete := make(chan interface{})
-	go func() {
-		if len(w.buffer) == 0 {
-			if w.statusCode != 0 {
-				rw.WriteHeader(w.statusCode)
-			}
-			return
-		}
+
+	if len(w.buffer) == 0 {
```
sreuland, line 104: nice, less complex and same result!

sreuland (approved):

mollykarcher (approved):

PR #965, opened 2023-09-14
soroban-rpc: add jrpc2 batch compliance test

tsachiherman:
### What

add jrpc2 batch compliance test

### Why

Add a test to verify the output of the jrpc library.

### Known limitations

n/a

### DIFF
--- a/cmd/soroban-rpc/internal/jsonrpc_test.go
+++ b/cmd/soroban-rpc/internal/jsonrpc_test.go
@@ -1,0 +1,85 @@
+package internal
+
+import (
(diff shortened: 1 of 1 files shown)


2opremio:
Doesn't this belong to the library itself?

tsachiherman:
> Doesn't this belong to the library itself?

Yes.. but I wanted to confirm that on our end first.

I agree that this test need to move to the library.

tsachiherman:
I spent some time looking into this issue. According to the jrpc2 specification:

PR #975, opened 2023-09-19
Set base prng to zero in preflight and invoke

leighmcculloch:
### What
Set base prng to zero in preflight and invoke.

### Why
The base prng seed need to be configured in the host for the prng host functions to work. Without it being set invocatinos of contract functions that use the prng will fail, and simulations will also.

```[tasklist]
### Todo
- [x] Decide what value should be set: ~zero, or~ **random**, ~or fixed.~
- [x] Find a way to test it. (Thanks to @tsachiherman)
- [x] Get tests passing.
```

### DIFF
--- a/cmd/soroban-rpc/lib/preflight/src/preflight.rs
+++ b/cmd/soroban-rpc/lib/preflight/src/preflight.rs
@@ -52,6 +52,8 @@ pub(crate) fn preflight_invoke_hf_op(
         .context("cannot set debug diagnostic level")?;
     host.set_ledger_info(ledger_info.clone())
         .context("cannot set ledger info")?;
+    host.set_base_prng_seed([0; 32])
+        .context("cannot set base prng seed")?;
 
     // We make an assumption here:
     // - if a transaction doesn't include any soroban authorization entries the client either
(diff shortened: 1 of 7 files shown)


cmd/soroban-rpc/lib/preflight/src/preflight.rs
```diff
@@ -52,6 +52,8 @@ pub(crate) fn preflight_invoke_hf_op(
         .context("cannot set debug diagnostic level")?;
     host.set_ledger_info(ledger_info.clone())
         .context("cannot set ledger info")?;
+    host.set_base_prng_seed([0; 32])
+        .context("cannot set base prng seed")?;
```
leighmcculloch, line 56: @tsachiherman I think we should probably use a random value here so as to better represent the changing value that will occur on chain. Wdyt?

tsachiherman: I think that it would be good idea to use a random number for pubnet/testnet/futurenet and a zero for standalone.
But regardless of that decision, I think that this fix, as is - is a must.

2opremio (approved):

PR #979, opened 2023-09-20
Treat expired temporary entries as not found in the RPC

leighmcculloch:
### What
Treat expired temporary entries as not found in the RPC when looking up data for preflight.

### Why
According to the discussion [here] the RPC is treating temporary entries as found and so once they are expired and until their found the RPC is not identifying that the temporary entry will be a new entry created during the call and the refundable fee is not being calculated to include for that cost.

@tsachiherman asked me to run a test with this small change to see if it would fix the problem. This isn't a complete fix. There's code elsewhere in the file being edited that really should be looked over and tidied up. Also there needs to be tests for this. This change just holds a quick hack to confirm this is the problem.

### DIFF
--- a/cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
+++ b/cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
@@ -186,7 +186,17 @@ impl LedgerStorage {
         }
 
         let entry = LedgerEntry::from_xdr(xdr)?;
-        Ok((entry, expiration_seq))
+        let entry_and_expiration = (entry, expiration_seq);
+
+        if let Ok(expirable_entry) =
+            TryInto::<Box<dyn ExpirableLedgerEntry>>::try_into(&entry_and_expiration)
+        {
+            if expirable_entry.durability() != Persistent {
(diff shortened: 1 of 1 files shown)


cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
```diff
@@ -186,7 +186,17 @@ impl LedgerStorage {
...
 
         let entry = LedgerEntry::from_xdr(xdr)?;
-        Ok((entry, expiration_seq))
+        let entry_and_expiration = (entry, expiration_seq);
+
+        if let Ok(expirable_entry) =
+            TryInto::<Box<dyn ExpirableLedgerEntry>>::try_into(&entry_and_expiration)
+        {
+            if expirable_entry.durability() != Persistent {
+                return Err(Error::NotFound);
```
dmkozh, line 195: I'm not sure I understand this, why don't we do a `has_expired` check here? Seems like every temp entry will be `NotFound` now...

leighmcculloch:
Superseded by:
- https://github.com/stellar/soroban-tools/pull/980.

PR #980, opened 2023-09-20
soroban-rpc: preflight: Exclude temporary expired entries from SnapshotSource

2opremio:


leighmcculloch (approved):

leighmcculloch (approved):
Is this something we can include a test for?

2opremio:
Hold on, this is still broken

2opremio:
Alright, this should be it. 

> Is this something we can include a test for?

Probably, but it's 1 am here :) , feel free to add one

cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
```diff
@@ -227,28 +245,27 @@ impl LedgerStorage {
...
-                Ok(le) => le,
-            };
-        if let Some(ref tracker) = self.restore_tracker {
-            _ = tracker.track_and_restore(self.current_ledger_sequence, key, &entry_and_expiration);
+        let result = <dyn SnapshotSource>::get(self, key);
+        if let Err(ref host_error) = result {
+            if host_error.error == HostError::from(Error::NotFound).error {
+                return Ok(false);
+            }
         }
```
2opremio, line 268: I am not sure this is the best way (or even a correct way) to compare Host errors

@dmkozh / @leighmcculloch  can you take a look?

dmkozh: Typically we verify host errors wit hhost_error.error.is_code/is_type, but for that you'll need to know the mapping from Error::NotFound.

cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
```diff
@@ -227,28 +245,27 @@ impl LedgerStorage {
...
+            let mut entry_and_expiration = self
+                .get_including_persistent_expired(key)
+                .map_err(HostError::from)?;
             // If the entry expired, we modify the expiration to make it seem like it was restored
             entry_and_expiration.1 =
                 tracker.track_and_restore(self.current_ledger_sequence, key, &entry_and_expiration);
+            return Ok((entry_and_expiration.0.into(), entry_and_expiration.1));
         }
-        Ok((entry_and_expiration.0.into(), entry_and_expiration.1))
+        let entry_and_expiration =
```
dmkozh, line 257: Just wanted to double-check: what is the error that occurs when 'auto-restoration' is not enabled and the persistent entry has expired? We need to make sure it's not `NotFound` as to distinguish from a simply non-existent entry.

cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
2opremio: So, is the code correct?

cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
2opremio: The error is also`NotFound`. (`Missing` for the host).

Yeah, we should probably use a different error code but:
1. This is for a separate PR.
2. I don't think the HostError has a suitable error to express it

cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
2opremio: I found a way to improve it. Thanks

cmd/soroban-rpc/lib/preflight/src/ledger_storage.rs
2opremio: Actually, I bit the bullet and included a separate error code for expired entries

dmkozh: Thanks for addressing this as well.

dmkozh (approved):

leighmcculloch (approved):
I'm not an expert in this area of code, but the changes look like they make sense.

PR #997, opened 2023-10-03
Feat!(CLI): remove sandbox

### LINKED ISSUE #983: CLI: Remove sandbox mode
### What

Remove sandbox mode and all functionality supporting it.

### Why

Most folks are using the stellar/quickstart docker image or a deployed soroban-rpc on futurenet/testnet instead of sandbox.

Over time we have not maintained the sandbox functionality consistently with the soroban-rpc and network.

A discussion was started on Discord to field feedback from the ecosystem:
https://discord.com/channels/897514728459468821/1150948802107277392

@namankumar also did user research and discovered that very few folks are using sandbox mode.

We've got limited resources and can reduce the cost of maintaining the CLI by removing this functionality. It should also make some areas of the CLI easier to evolve since there won't be two different code paths for each operation.

cc @namankumar @tomerweller 
```[tasklist]
### Tasks
- [ ] Update docs to use docker image (not testnet) instead of sandbox
- [ ] Remove code from soroban-cli
```

willemneal:
### What

fixes #983 

This also moves all tests that rely on a network to in a single module `integration`, which is ignored by default

### Why

No one was using it and it became a maintenance burden. Perhaps in the future the preflight could be altered to allow lazy state retrieval from a node to allow for offline simulations.

### Known limitations

Tests are now very slow.


2opremio (approved):

2opremio:
can you resolve the conflicts?

PR #1013, opened 2023-10-06
fix: clippy --all

willemneal:
### What

Ensure that clippy used across repo. Also use an action and split CLI integration tests from normal ones. 

### Why

Consistency and make git hooks.

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/crates/soroban-test/tests/fixtures/test-wasms/udt/src/lib.rs
+++ b/cmd/crates/soroban-test/tests/fixtures/test-wasms/udt/src/lib.rs
@@ -34,6 +34,7 @@ pub struct Contract;
 
 #[contractimpl]
 impl Contract {
+    #[allow(clippy::unnecessary_fold)]
     pub fn add(a: UdtEnum, b: UdtEnum) -> i64 {
         let a = match a {
             UdtEnum::UdtA => 0,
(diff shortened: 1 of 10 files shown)


cmd/crates/soroban-test/tests/fixtures/test-wasms/udt/src/lib.rs
```diff
@@ -34,6 +34,7 @@ pub struct Contract;
 
 #[contractimpl]
 impl Contract {
+    #[allow(clippy::unnecessary_fold)]
```
leighmcculloch, line 37: Still needed if we switch to sum?

leighmcculloch (approved):

cmd/crates/soroban-test/tests/fixtures/test-wasms/udt/src/lib.rs
willemneal: I thought it was a nice example for readers that they could do it both ways.

.github/workflows/rust.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     - run: rustup target add wasm32-unknown-unknown
     - if: matrix.target == 'aarch64-unknown-linux-gnu'
       run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
-    - run: cargo clippy --all-targets --target ${{ matrix.target }}
+    - run: cargo clippy --all --all-targets --target ${{ matrix.target }}
```
willemneal, line 65:
```suggestion
    - run: cargo clippy --all-targets --target ${{ matrix.target }}
```

willemneal:
@2opremio, I had to update the budget to pass the tests, but the fee dropped around 50%; this seems extreme, is it correct?

PR #1015, opened 2023-10-09
Flatten the `getEvents` response structure

### LINKED ISSUE #998: Flatten the `getEvents` response schema
The current schema for the event value does not make sense: according to the [getEvents documentation](https://soroban.stellar.org/api/methods/getEvents), we have 

```typescript
{
  // ...other fields...
  value: {
    xdr: string
  }
}
```

where `xdr` is a serialized `ScVal`. This nested structure offers no add'l value.

A flatter structure:

```typescript
{
  value: string;
}
```

It would be highly beneficial to get this in prior to a stable build due to the fact that it's a breaking change.

This can be abstracted client-side, but it's cleaner if it comes from the server.

Shaptic:
### What
This makes `value: { xdr: "a base64-encoded string of an ScVal" }` just `value: "the string"`, because the nested object just adds verbosity and structure complexity.

### Why
Closes #998

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-cli/src/commands/config/events_file.rs
+++ b/cmd/soroban-cli/src/commands/config/events_file.rs
@@ -14,6 +14,7 @@ use std::{
     fs,
     path::{Path, PathBuf},
 };
+use stellar_strkey::{Contract, Strkey};
 
 #[derive(Debug, clap::Args, Clone, Default)]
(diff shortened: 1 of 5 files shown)


tsachiherman (approved):
Lgtm

PR #1016, opened 2023-10-10
fix(CLI): ensuring that read ledger entries match up with ledger keys

willemneal:
### What

Previously it was assumed that `<key1, key1_expiration>` returned with `<LedgerEntry_1, LedgerEntry_expiration>`, but this isn't always the case and it was causing flaky tests.

### Why

This PR ensures that the ledger keys are matched with their expiration. This allows any ordered returned from the RPC.

### Known limitations

[TODO or N/A]


tsachiherman:
@willemneal I'm not sure if this would be needed in light of https://github.com/stellar/soroban-tools/pull/916. What do you think ?

tsachiherman:
@leighmcculloch I think that https://github.com/stellar/soroban-tools/pull/916 conflicts with this one.

willemneal:
close in favor of #1021

PR #1034, opened 2023-10-19
feat(bindings-ts)!: API change: return `AssembledTransaction`s, require signAndSend, and support multi-auth workflows

chadoh:
# Overview

Soroban, and Stellar more generally, supports having multiple parties sign the same transaction.

Prior to now, the libraries generated with `soroban contract bindings typescript` did not.

This adds multi-auth support to the generated libraries, streamlining it as much as possible while adding minimal overhead to the "normal stuff" that most people will do with these libraries 98% of the time.

## Read calls (simulation only)

The biggest change for view calls, which require only a simulation and no signing, is that you now need to get the `result` of a transaction:

```diff
-const balance = await tokenContract.balance({ account: 'G…' })
+const balance = (await tokenContract.balance({ account: 'G…' })).result
```

You can streamline this further using destructuring:

```diff
-const balance = await tokenContract.balance({ account: 'G…' })
+const { result: balance } = await tokenContract.balance({ account: 'G…' })
```

Or just call the variable `result` and it looks very clean:

```diff
-const balance = await tokenContract.balance({ account: 'G…' })
+const { result } = await tokenContract.balance({ account: 'G…' })
```

## Write calls with one signer

For these, you now need to explicitly sign and send the transaction. This helps you build up the mental model that the transaction is first _simulated_, that this is good enough for view calls, and that you also need a `signAndSend` step if you want a write call to actually go through.

```diff
-const result = await tokenContract.send({ to: 'G…', amount: 10n })
+const tx = await tokenContract.send({ to: 'G…', amount: 10n })
+const { result } = tx.signAndSend()
```

This also allows these libraries to expose information and options in more expectable places, rather than cluttering global interfaces. Examples:

- `tx.isReadCall` – make explicit logic around read vs write calls
- `tx.signAndSend({ secondsToWait: 30 })` – specify how many seconds to wait for the transaction to complete (default: 10)
- `tx.raw` – access the raw SorobanClient transaction that this object wraps
- `tx.simulationData` – get the result of the simulation

Once you `signAndSend()`, the object you get back has different fields that are more relevant to the signing and sending behavior. See [SentTransaction](https://github.com/AhaLabs/soroban-tools/blob/feat/bindings-ts/multi-auth/cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts#L500-L514) for more information.

## Write calls with >1 signer

Let's consider an Atomic Swap contract. Alice wants to give 10 of her Token A tokens to Bob for 5 of his Token B tokens. 

```ts
const ALICE = 'G123...'
const BOB = 'G456...'
const TOKEN_A = 'C123…'
const TOKEN_B = 'C456…'
const AMOUNT_A = 10n
const AMOUNT_b = 5n
```

Let's say Alice is also going to be the one signing the final transaction envelope, meaning she is the invoker. So your app, from Alice's browser, simulates the `swap` call:

```ts
const tx = await swapContract.swap({
  a: ALICE,
  b: BOB,
  token_a: TOKEN_A,
  token_b: TOKEN_B,
  amount_a: AMOUNT_A,
  amount_b: AMOUNT_B,
})
```

But your app can't `signAndSend` this right away, because Bob needs to sign it first. You can check this:

```ts
const whoElseNeedsToSign = tx.needsNonInvokerSigningBy()
```

You can verify that `whoElseNeedsToSign` is an array of length `1`, containing only Bob's public key.

Then, still on Alice's machine, you can serialize the transaction-under-assembly:

```ts
const json = tx.toJSON()
```

And now you need to send it to Bob's browser. How you do this depends on your app. Maybe you send it to a server first, maybe you use [WebSockets](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket), or maybe you have Alice text the JSON blob to Bob and have him paste it into your app in his browser (note: this option might be error-prone 😄).

Once you get the JSON blob into your app on Bob's machine, you can deserialize it:

```ts
const tx = swapContract.fromJSON.swap(json)
```

And have Bob sign it. What Bob will actually need to sign is some _auth entries_ within the transaction, not the transaction itself or the transaction envelope. Your app can verify that Bob has the correct wallet selected in Freighter, then:

```ts
await tx.signAuthEntries()
```

Under the hood, this uses `signAuthEntry` from the injected `wallet` (`@stellar/freighter-api` by default).

Now Bob can again serialize the transaction and send back to Alice, where she can finally `signAndSend()`.

To see an even more complicated example, where Alice swaps with Bob but the transaction is invoked by yet another party, check out the new [test-swap.ts](https://github.com/AhaLabs/soroban-tools/blob/feat/bindings-ts/multi-auth/cmd/crates/soroban-spec-typescript/ts-tests/src/test-swap.ts#L46)

# All the nitty gritty details

Here's a full(er) accounting of everything that changed here:

- switch from `ts-tests/package.json` to an `initialize.sh` script that
  uses a `.env` if available or defaults to environment variables. As
  before, this will build, deploy, and generate bindings for the new
  contracts, plus creating an `alice` identity and minting separate
  amounts of two separate tokens to the `root` user and `alice`
- add eslint to catch missing `await`s
- add transaction-rebuilding to increment transaction sequence numbers,
  so that all tests can pass when run in parallel
- add `atomic_swap` and `token` contracts from
  https://github.com/stellar/soroban-examples to `test-wasms`
- add tests for atomic swap functionality inspired by
  https://github.com/stellar/soroban-react-atomic-swap
- let this logic guide needed updates to `bindings typescript`-generated
  libraries:
  - don't return flat values
  - instead, return a `new AssembledTransaction`, a class that has a
    `result` getter that can be used to get the result of the simulation
    right away, or can be used to `await tx.signAndSend()`, which then
    returns a `SentTransaction`
  - `SentTransaction` contains all possibly-relevant fields
    from the logic it performs, such as `sendTransactionAll`
  - `AssembledTransaction` has a `needsNonInvokerSigningBy()` getter
    and a `signAuthEntriesFor(publicKey, signingFunction)` to facilitate
    multi-auth workflows.
  - Since assembling transactions may now take place across multiple
    users' browsers, you can also call `json = tx.toJSON()` on an
    `AssembledTransaction` and then `contract.fromJSON[method](json)` on
    the next machine to continue signing, as shown in `test-swap.ts`


cmd/crates/soroban-spec-typescript/ts-tests/initialize.sh
```diff
@@ -0,0 +1,96 @@
...
+function fund_alice() {
+  exe eval "./soroban config identity generate alice"
+  exe eval "./soroban config identity fund alice"
+}
+function fund() {
+  fund_root
+  fund_alice
+}
+function bindings_custom_types() {
+  exe eval "./soroban contract bindings typescript --contract-id $(cat contract-id-custom-types.txt) --output-dir ./node_modules/test-custom-types --overwrite"
```
willemneal, line 39: why `exe eval`? Also seems like you could break bindings into it's own function and then below call with relevant args instead of a bunch of functions.

willemneal: Yeah but why do you need `exe` at all? why not `./soroban ...`

chadoh: To output the command that's going to be run. It ends up looking like this:

```
npm i

> postinstall
> ./initialize.sh

Network
  RPC:        http://localhost:8000/soroban/rpc
  Passphrase: "Standalone Network ; February 2017"
./soroban config identity fund
./soroban config identity generate alice
./soroban config identity fund alice
2023-10-23T15:10:42.831597Z  WARN soroban_cli::commands::config::network: Account already exists
(./soroban contract deploy --wasm ../../../../target/wasm32-unknown-unknown/test-wasms/test_custom_types.wasm) > contract-id-custom-types.txt
(./soroban contract deploy --wasm ../../../../target/wasm32-unknown-unknown/test-wasms/test_hello_world.wasm) > contract-id-hello-world.txt
(./soroban contract deploy --wasm ../../../../target/wasm32-unknown-unknown/test-wasms/test_swap.wasm) > contract-id-swap.txt
(./soroban contract install --wasm ../../../../target/wasm32-unknown-unknown/test-wasms/test_token.wasm) > contract-token-hash.txt
(./soroban contract deploy --wasm-hash 07ec9b8333159ac477239ff1a54c6fc45c5817e03cf0f45b6c9e51727c0e3dc7) > contract-id-token-a.txt
(./soroban contract deploy --wasm-hash 07ec9b8333159ac477239ff1a54c6fc45c5817e03cf0f45b6c9e51727c0e3dc7) > contract-id-token-b.txt
./soroban contract invoke --id CCE4XXFTF2JJRO3GK4NDEPIZGKZQYXEZUKXIVGDEULUEQZHDD4XOOTGK -- initialize --admin GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4 --decimal 0 --name 'Token A' --symbol 'A'

./soroban contract invoke --id CA63EAYAR75XGN62FLHTYRDKDQSCN2Q7XPNIZ635L3W5RV6GO3BKB4ET -- initialize --admin GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4 --decimal 0 --name 'Token B' --symbol 'B'

./soroban contract bindings typescript --contract-id CDSASQFKGWGL3HUWL6MRWSQMXC4PLCSYQRJYCDGNHKWQWRCQ7V7PLRW2 --output-dir ./node_modules/test-custom-types --overwrite

added 26 packages, and audited 27 packages in 4s

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> test-custom-types@0.0.0 build
> node ./scripts/build.mjs

./soroban contract bindings typescript --contract-id CDBUYGC6MLYU4LNB63THZQJOVVD65GUKWVGKAOOVVQ5E5MHOHEVXAPTB --output-dir ./node_modules/test-hello-world --overwrite

added 26 packages, and audited 27 packages in 1s

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> test-hello-world@0.0.0 build
> node ./scripts/build.mjs

./soroban contract bindings typescript --contract-id CCLE6DN54RHRC4P5TKPIMRUMC3LNCPYJBPL7D5RDEKAOB64JLQ25MWIO --output-dir ./node_modules/test-swap --overwrite

added 26 packages, and audited 27 packages in 1s

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> test-swap@0.0.0 build
> node ./scripts/build.mjs

./soroban contract bindings typescript --contract-id CCE4XXFTF2JJRO3GK4NDEPIZGKZQYXEZUKXIVGDEULUEQZHDD4XOOTGK --output-dir ./node_modules/token --overwrite

added 26 packages, and audited 27 packages in 900ms

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> token@0.0.0 build
> node ./scripts/build.mjs


removed 108 packages, and audited 191 packages in 48s

58 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```

willemneal: Oh I see, you can also do `set -v` at the top of the file. Then using a functions it can be a bit more tidy:
```bash
function bind() {
  ./soroban contract bindings typescript --contract-id "$(cat $1)" --output-dir "./node_modules/$2" --overwrite
}


function bindings() {
  bind contract-id-custom-types.txt test-custom-types
  bind contract-id-hello-world.txt test-hello-world
  bind contract-id-swap.txt test-swap
  bind contract-id-token-a.txt token
}
```

willemneal: Oh nvm that also prints all the functions.

chadoh: I cleaned it up with some functions and added comments + links above the `exe` function definition. I think it looks much tidier; thanks! lmk if you have other ideas for cleaning it up.

cmd/crates/soroban-spec-typescript/ts-tests/initialize.sh
```diff
@@ -0,0 +1,60 @@
...
+echo "  RPC:        $SOROBAN_RPC_URL"
+echo "  Passphrase: \"$SOROBAN_NETWORK_PASSPHRASE\""
+
+# Print command before executing, from https://stackoverflow.com/a/23342259/249801
+# Discussion: https://github.com/stellar/soroban-tools/pull/1034#pullrequestreview-1690667116
+exe() { echo"${@/eval/}" ; "$@" ; }
+
+function fund_all() {
+  exe eval "./soroban config identity fund"
+  exe eval "./soroban config identity generate alice"
```
willemneal, line 22: If the network is set via ENV vars than this should pre-fund the account. Do you see any warnings in the output of the next line?

chadoh: Yes, there's a warning that the account already exists.

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -0,0 +1,584 @@
...
+      this.tx,
+      this.options.networkPassphrase,
+      simulation
+    ).build()
+
+    return this
+  }
+
+  getWallet = async (): Promise<Wallet> => {
+    return this.options.wallet ?? (await import("@stellar/freighter-api"))
```
elizabethengelman, line 142: small thing here, but I think we'll need to add `.default` to the end of this like we did [here](https://github.com/stellar/soroban-tools/blob/733ce062532c45ff96e690f135f6c1193a6d14d3/cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts#L86)

willemneal (approved):

cmd/crates/soroban-spec-typescript/src/lib.rs
```diff
@@ -91,64 +116,100 @@ pub fn generate(spec: &[ScSpecEntry]) -> String {
...
+            .unwrap_or_default();
+        if is_result {
+            output = format!("new Ok({output})");
+        }
+        let parse_result_xdr = if return_type == "void" {
+            r#"() => {}"#.to_owned()
+        } else if is_result {
+            format!(r#"(result: XDR_BASE64 | Err): {return_type} => {{
+            if (result instanceof Err) return result
+            return new Ok(this.spec.funcResToNative("u32_fail_on_even", result))
```
Shaptic, line 190: what purpose is `u32_fail_on_even` here for? :confused:

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -0,0 +1,706 @@
...
+  SorobanRpc,
+  StrKey,
+  TimeoutInfinite,
+  TransactionBuilder,
+  assembleTransaction,
+  hash,
+  nativeToScVal,
+  xdr,
+} from "soroban-client";
+import type { Memo, MemoType, Transaction } from "soroban-client";
```
Shaptic, line 17: we should definitely have another PR either after this or as part of it to port to [`stellar-sdk`](https://github.com/stellar/js-stellar-sdk/releases/tag/v11.0.0-beta.6), to futureproof the bindings (the migration guide is [here](https://gist.github.com/Shaptic/5ce4f16d9cce7118f391fbde398c2f30)).

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -0,0 +1,706 @@
...
+      allowHttp: this.options.rpcUrl.startsWith("http://"),
+    });
+  }
+
+  static async fromSimulation<T>(options: AssembledTransactionOptions<T>): Promise<AssembledTransaction<T>> {
+    const tx = new AssembledTransaction(options)
+    const contract = new Contract(options.contractId);
+
+    tx.raw = new TransactionBuilder(await tx.getAccount(), {
+      fee: options.fee?.toString(10) ?? '100',
```
Shaptic, line 160: using 

```suggestion
      fee: options.fee?.toString(10) ?? BASE_FEE,
```

from the SDK would be better than the magic number

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -0,0 +1,706 @@
...
+  }
+
+  getStorageExpiration = async (storageType: 'temporary' | 'persistent') => {
+    const key = xdr.LedgerKey.contractData(
+      new xdr.LedgerKeyContractData({
+        contract: new Address(this.options.contractId).toScAddress(),
+        key: xdr.ScVal.scvLedgerKeyContractInstance(),
+        durability: xdr.ContractDataDurability[storageType](),
+      }),
+    )
```
Shaptic, line 303: you can use [`getFootprint`](https://stellar.github.io/js-soroban-client/Contract.html#getFootprint) here (see the [code](https://stellar.github.io/js-soroban-client/js-stellar-base_src_contract.js.html#line78)):

```suggestion
    const key = new Contract(this.options.contractId).getFootprint()[1];
```

The instance is always persistent, afaik.

(note this will need a [small amendment](https://github.com/stellar/js-stellar-base/pull/709) after the stable release)

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -0,0 +1,706 @@
...
+
+    const expirationKey = xdr.LedgerKey.expiration(
+      new xdr.LedgerKeyExpiration({ keyHash: hash(key.toXDR()) }),
+    )
+
+    const entryRes = await this.server.getLedgerEntries(expirationKey)
+    if (!(entryRes.entries && entryRes.entries.length)) throw new Error('failed to get ledger entry')
```
Shaptic, line 310: just to note two things:

 * `Server.getLedgerEntries(key)` will automatically fetch the corresponding `expirationKey` if applicable (see the release notes for [`v1.0.0-beta.3`](https://github.com/stellar/js-soroban-client/releases/tag/v1.0.0-beta.3))
 * the RPC server will do this automatically soon as well

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -0,0 +1,706 @@
...
+      )
+
+      // get a reference to the auth entry, then mutate it
+      const addrAuth = entry.credentials().address()
+      addrAuth.signature(xdr.ScVal.scvVec([sigScVal]))
+
+      addrAuth.signatureExpirationLedger(
+        preimage.sorobanAuthorization().signatureExpirationLedger()
+      )
+    }
```
Shaptic, line 485: This code is nearly identical to the code of [`authorizeEntry`](https://stellar.github.io/js-soroban-client/global.html#authorizeEntry) (see the [source code](https://stellar.github.io/js-soroban-client/js-stellar-base_src_auth.js.html#line97)); can we leverage that here to simplify things?

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -0,0 +1,706 @@
...
+
+          this.assembled.raw = new TransactionBuilder(await this.assembled.getAccount(), {
+            fee: this.assembled.raw.fee,
+            networkPassphrase: this.options.networkPassphrase,
+          })
+            .setTimeout(TimeoutInfinite)
+            .addOperation(
+              Operation.invokeHostFunction({ ...op, auth: op.auth ?? [] }),
+            )
+            .build()
```
Shaptic, line 557: I know you mentioned this briefly on Discord, but can you elaborate why

```typescript
this.assembled.raw = TransactionBuilder.cloneFrom(this.assembled.raw);
```

does not work for you? It should be functionally identical.

cmd/crates/soroban-spec-typescript/ts-tests/package.json
```diff
@@ -1,34 +1,31 @@
...
+  },
+  "devDependencies": {
+    "@ava/typescript": "^4.1.0",
+    "@types/node": "^20.4.9",
+    "@typescript-eslint/eslint-plugin": "^6.10.0",
+    "@typescript-eslint/parser": "^6.10.0",
+    "ava": "^5.3.1",
+    "dotenv": "^16.3.1",
+    "eslint": "^8.53.0",
+    "soroban-client": "1.0.0-beta.2",
```
Shaptic, line 17: another PR later to update this to [beta.4](https://github.com/stellar/js-soroban-client/releases/tag/v1.0.0-beta.4)? and/or to stellar-sdk

Shaptic (commented):
first pass!

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
chadoh: It does not increment the transaction sequence number

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
chadoh: Probably! Thanks. I started with an older example and wasn't sure what the new interfaces were; just wanted to get something working.

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
chadoh: For sure! I would like to do it after.

tyvdh:
1. Is there a reason we're using JSON vs XDR for message passing? At the very least XDR should be supported even if we suggest JSON. The rest of the ecosystem uses XDR so we shouldn't special case that here imo.

2. This example block of code seems awkward
```js
import { signAuthEntry } from "@stellar/freighter-api";

await tx.signAuthEntriesFor(
  BOB,
  async payload => await signAuthEntry(payload, {
    accountToSign: BOB,
  })
)
```
Feels like a double sign, If you're signing with `signAuthEntry` what is `signAuthEntriesFor` doing other than applying that signature to the transaction? If that's the case why do we need to include the BOB account as an arg?

cmd/crates/soroban-spec-typescript/src/lib.rs
chadoh: copypasta 🤦🏻

chadoh:
> 1. Is there a reason we're using JSON vs XDR for message passing? At the very least XDR should be supported even if we suggest JSON. The rest of the ecosystem uses XDR so we shouldn't special case that here imo.

This is a special case, though! You are serializing an `AssembledTransaction` object, which will only work with this library, in JavaScript. It wraps a SorobanClient Transaction, which is serialized with XDR, but it also wraps some other stuff that we need to properly get someone else up and editing the transaction in a different browser. See [toJSON](https://github.com/AhaLabs/soroban-tools/blob/feat/bindings-ts/multi-auth/cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts#L115-L125) and [fromJSON](https://github.com/AhaLabs/soroban-tools/blob/feat/bindings-ts/multi-auth/cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts#L127-L147) for more info.

If there's a way to reconstruct everything we need with just a SorobanClient transaction, then we can simplify this and use XDR. But if it's this special-case JavaScript object, then forcing it into XDR seems complicated and unnecessary, imo.

> 2. This example block of code seems awkward
> 
>    ```ts
>    import { signAuthEntry } from "@stellar/freighter-api";
>    
>    await tx.signAuthEntriesFor(
>       BOB,
>       async payload => await signAuthEntry(payload, {
>         accountToSign: BOB,
>       })
>     )
> Feels like a double sign, If you're signing with signAuthEntry what is signAuthEntriesFor doing other than applying that signature to the transaction? If that's the case why do we need to include the BOB account as an arg?

`tx.signAuthEntriesFor` does [a LOT](https://github.com/AhaLabs/soroban-tools/blob/feat/bindings-ts/multi-auth/cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts#L399-L486). The short version is that it loops through _all_ the auth entries in the transaction and signs all the ones for the given public key, using the given signing function. In a Node context like the tests, it looks like this:

```ts
await txAlice.signAuthEntriesFor(
  alice.keypair.publicKey(),
  entry => alice.keypair.sign(entry)
)
```

`signAuthEntry` from Freighter is much more specific: it signs a _particular_ auth entry _payload_, which needs to be constructed in a sorta finicky way (see the `signAuthEntriesFor` implementation, linked above, for details).

I agree, though. I didn't figure out what the Freighter/browser version would look like until I wrote this PR description, and it feels much more repetitive than the Node version.

Maybe we rename it to `tx.signAuthEntries` and remove the first variable, since we can get the current user from the wallet. Then it becomes:

```ts
import { signAuthEntry } from "@stellar/freighter-api";

await tx.signAuthEntries(
  async payload => await signAuthEntry(payload, {
    accountToSign: BOB,
  })
)
```

cmd/crates/soroban-spec-typescript/ts-tests/package.json
chadoh: I think it makes sense to update now. Is stellar-sdk ready and preferred to beta.4?

tyvdh:
> This is a special case, though!

Got it. Yeah I guess that makes sense. Unfortunate it's not more general but I doubt it'll be a major issue. Might be something worth looking into if it can't be reduced to just UX around current `soroban-client` functionality for multi-auth so you could use XDR vs special case JSON.

> Maybe we rename it to `tx.signAuthEntries`

This makes sense to me and I see no downside to this

aristidesstaffieri:
> 1. Is there a reason we're using JSON vs XDR for message passing? At the very least XDR should be supported even if we suggest JSON. The rest of the ecosystem uses XDR so we shouldn't special case that here imo.
> 2. This example block of code seems awkward
> 
> ```js
> import { signAuthEntry } from "@stellar/freighter-api";
> 
> await tx.signAuthEntriesFor(
>   BOB,
>   async payload => await signAuthEntry(payload, {
>     accountToSign: BOB,
>   })
> )
> ```
> 
> Feels like a double sign, If you're signing with `signAuthEntry` what is `signAuthEntriesFor` doing other than applying that signature to the transaction? If that's the case why do we need to include the BOB account as an arg?

want to +1 this comment. I do think it would be better if clients did not have to work with structures that were specific to this lib and not compatible with other tooling(`AssembledTransaction`). IMO signers only need the auth entry xdr and they should be able to use any xdr tooling to parse it and sign it.

That said, I do realize that you can get to the underlying raw tx and this would be a big departure from the patterns here so I don't suggest any changes.

cmd/crates/soroban-spec-typescript/ts-tests/package.json
Shaptic: yes and yes :+1:

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
Shaptic: hm, then you could call 

```
let account = await this.assembled.getAccount(); 
account.incrementSequenceNumber(); 
this.assembled.raw = TransactionBuilder.cloneFrom(...);
```

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
chadoh: Is there anything to change here, at this point?

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
chadoh: I just double-checked this. It does not work. When I run all the tests in parallel, I get errors on most runs if I use the `cloneFrom` approach. Rebuilding it from the operations and re-simulating passes every time.

chadoh:
> IMO signers only need the auth entry xdr and they should be able to use any xdr tooling to parse it and sign it.

@aristidesstaffieri I agree with this assessment, and I think it's worth exploring how to simplify the auth-entry-signing workflow in a follow-up PR. For now, this one works! It's a big change and I'm eager to get it in so we can start building on top of it. Sound good?

@tyvdh @Shaptic @aristidesstaffieri I've addressed all feedback and updated the PR description. lmk what you think!

tyvdh (approved):
Again, haven't gone over the code but the interface lgtm

Shaptic (approved):
yeet it

PR #1039, opened 2023-10-20
rpc: add commandline flag to parameterize (opt-in) preflight debug info

### LINKED ISSUE #963: RPC: Make simulateTransaction debug mode configurable
RPC currently sets the diagnostic level for simulateTransaction (the VM in libpreflight) to debug. We'd like to provide RPC operators the option to enable/disable debug mode, independent of enabling/disabling all diagnostic events.

2opremio:
Fixes #963

### DIFF
--- a/cmd/soroban-rpc/internal/preflight/preflight.go
+++ b/cmd/soroban-rpc/internal/preflight/preflight.go
@@ -79,6 +79,7 @@ type PreflightParameters struct {
 	NetworkPassphrase string
 	LedgerEntryReadTx db.LedgerEntryReadTx
 	BucketListSize    uint64
+	EnableDebug       bool
 }
 
 type Preflight struct {
@@ -224,6 +225,7 @@ func getInvokeHostFunctionPreflight(params PreflightParameters) (Preflight, erro
 		invokeHostFunctionCXDR,
 		sourceAccountCXDR,
 		li,
+		C.bool(params.EnableDebug),
 	)
 	FreeGoXDR(invokeHostFunctionCXDR)
 	FreeGoXDR(sourceAccountCXDR)
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/stellar/soroban-tools
 
-go 1.20
+go 1.21
+
+toolchain go1.21.1
 
 require (
 	github.com/Masterminds/squirrel v1.5.4
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/stellar/soroban-tools
 
-go 1.21
-
-toolchain go1.21.1
+go 1.20
 
 require (
 	github.com/Masterminds/squirrel v1.5.4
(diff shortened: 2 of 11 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/stellar/soroban-tools
 
-go 1.20
+go 1.21
+
+toolchain go1.21.1
```
2opremio, line 5: Why the heck did the IDE change this? :S

tsachiherman: no idea, but I noticed that too with go1.21.

cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -224,6 +225,7 @@ func getInvokeHostFunctionPreflight(params PreflightParameters) (Preflight, erro
 		invokeHostFunctionCXDR,
 		sourceAccountCXDR,
 		li,
+		C.bool(params.EnableDebug),
```
tsachiherman, line 228: huh.. cgo supports c++ bool datatype ?!

I'm pleasantly surprised to see that.

tsachiherman (approved):
I think that we should have a unit test that validate that we're getting *some* functional difference when this flag is on vs off.

leighmcculloch (approved):

go.mod
2opremio: I think it (maybe wrongly) enforces initializing the toolchain field?

https://tip.golang.org/doc/toolchain

2opremio: More info here https://go.dev/blog/toolchain

2opremio: I don't think it's a good idea to enforce the field

2opremio: OK, I think it may have been caused by https://github.com/stellar/soroban-tools/pull/1030 bumping a dependency using 1.21 which cause Go to update the go.mod with the mentioned version.

See https://github.com/dependabot/dependabot-core/issues/7868 for an occurrence of this.

@tsachiherman can you investigate?

2opremio: OK, I found the culprit:

`github.com/creachadair/mds 1.21`

2opremio:
```
go mod why -m github.com/creachadair/mds 
# github.com/creachadair/mds
github.com/stellar/soroban-tools/cmd/soroban-rpc/internal
github.com/creachadair/jrpc2
github.com/creachadair/mds/mlink
```

2opremio: Let's revert https://github.com/stellar/soroban-tools/pull/1030 until this is fixed

leighmcculloch: We have never distinguished this, but that requirement is only meaningful for the Go monorepo because it hosts libraries, the Go SDK, and we need to give folks importing the Go SDK into their own apps time to upgrade. For products, it's reasonable to support only a single version of Go. The soroban-rpc already has a dependence on a single version of Rust so supporting older version of Go seems like a maintenance cost/burden we don't need.

PR #1049, opened 2023-10-24
feat(CLI): Assembled Transaction that handle preparing transaction post simulation

willemneal:
This renames the `transaction` module to `txn`. Adds `txn::Assembled`. `Assembled` handles the different steps in preparing a transaction to be submitted. This is the first step in allowing intermediate transactions be returned and externally signed.
Also in improving error messages from failed transactions.


chadoh (approved):
Love it

2opremio (approved):

PR #1051, opened 2023-10-25
Upgrade E2E tests to work with latest `stellar-sdk`

Shaptic:
### What
This reconfigures the end-to-end tests to use the latest JavaScript SDK, which is `stellar-sdk` now over `soroban-client`.

### Why
The existing way is deprecated moving forward.

### Known limitations
Requires https://github.com/stellar/system-test/pull/81.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest-4-cores
     env:
       # the gh tag of system-test repo version to run
-      SYSTEM_TEST_GIT_REF: 352cc1d4ea96af8cfb2c2b7335c3b51da000f889
+      SYSTEM_TEST_GIT_REF: 30a86b035d588e282895fa14970c346d53cb16ea
 
       # the soroban tools source code to compile and run from system test
       # refers to checked out source of current git hub ref context
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -40,8 +40,8 @@ jobs:
       # option #2, set the version of stellar-sdk used as a ref to a gh repo if
       # a value is set on SYSTEM_TEST_JS_STELLAR_SDK_GH_REPO, it takes
       # precedence over any SYSTEM_TEST_JS_STELLAR_SDK_NPM_VERSION
-      SYSTEM_TEST_JS_STELLAR_SDK_GH_REPO:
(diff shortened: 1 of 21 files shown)


.github/workflows/e2e.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest-4-cores
     env:
       # the gh tag of system-test repo version to run
-      SYSTEM_TEST_GIT_REF: 352cc1d4ea96af8cfb2c2b7335c3b51da000f889
+      SYSTEM_TEST_GIT_REF: master
```
sreuland, line 17: yeah, seems fine to just ref the latest, if it gets too far ahead, then can come back and pin it here.

.github/workflows/e2e.yml
```diff
@@ -32,16 +32,16 @@ jobs:
...
-      # if a value is set on SYSTEM_TEST_JS_SOROBAN_CLIENT_GH_REPO, it takes precedence 
-      # over any SYSTEM_TEST_JS_SOROBAN_CLIENT_NPM_VERSION
-      SYSTEM_TEST_JS_SOROBAN_CLIENT_GH_REPO:
-      SYSTEM_TEST_JS_SOROBAN_CLIENT_GH_REF:
+      # option #1, set the version of stellar-sdk based on a npm release version
+      SYSTEM_TEST_JS_STELLAR_SDK_NPM_VERSION:
+      # option #2, set the version of stellar-sdk used as a ref to a gh repo if
+      # a value is set on SYSTEM_TEST_JS_STELLAR_SDK_GH_REPO, it takes
+      # precedence over any SYSTEM_TEST_JS_STELLAR_SDK_NPM_VERSION
+      SYSTEM_TEST_JS_STELLAR_SDK_GH_REPO:
```
sreuland, line 43: REPO needs to be populated, in this case it would be `https://github.com/stellar/js-stellar-sdk`, the conditional steps below are triggered by it's presence.

Shaptic: oh okay got it, no defaults if tag is specified - fixed!

btw https://github.com/stellar/js-stellar-sdk/pull/877 will let us use an npm tag instead

sreuland: yes, if you want to wait for that to publish, then can just do `SYSTEM_TEST_JS_STELLAR_SDK_NPM_VERSION: 11.0.0-beta.6`, since sdk is changing often, will be changing this to a newer ref soon enough anyways, but there may be a little more value in verifying coverage against the npm repo. as it validates that published code is good.

sreuland (approved):

PR #1052, opened 2023-10-26
soroban-rpc: remove noisy rollback message

tsachiherman:
### What

When stopping soroban-rpc, it would generate a benign warning message:
```
WARN[2023-10-26T11:49:39.561-04:00] could not rollback fillEntriesFromCheckpoint write transactions  error="transaction has already been committed or rolled back" pid=20799
```

### Why

The underlying transaction was already rolled back by the database session. There is no need to roll it back again.
Given that this is the exception rather then the rule, I've silenced this particular case.

### Known limitations

n/a

### DIFF
--- a/cmd/soroban-rpc/internal/ingest/service.go
+++ b/cmd/soroban-rpc/internal/ingest/service.go
@@ -12,6 +12,7 @@ import (
 	"github.com/stellar/go/historyarchive"
 	"github.com/stellar/go/ingest"
 	backends "github.com/stellar/go/ingest/ledgerbackend"
+	supportdb "github.com/stellar/go/support/db"
 	"github.com/stellar/go/support/log"
 	"github.com/stellar/go/xdr"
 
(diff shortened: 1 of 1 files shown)


cmd/soroban-rpc/internal/ingest/service.go
```diff
@@ -208,9 +209,14 @@ func (s *Service) fillEntriesFromCheckpoint(ctx context.Context, archive history
 	if err != nil {
 		return err
 	}
+	transactionCommitted := false
 	defer func() {
-		if err := tx.Rollback(); err != nil {
-			s.logger.WithError(err).Warn("could not rollback fillEntriesFromCheckpoint write transactions")
+		if transactionCommitted == false {
```
stellarsaur, line 214: golangci - prefer `if !transactionCommitted {`

cmd/soroban-rpc/internal/ingest/service.go
```diff
@@ -208,9 +209,14 @@ func (s *Service) fillEntriesFromCheckpoint(ctx context.Context, archive history
...
+			// Internally, we might already have rolled back the transaction. We should
+			// not generate benign error/warning here in case the transaction was already rolled back.
+			if err := tx.Rollback(); err != nil && err != supportdb.ErrAlreadyRolledback {
```
stellarsaur, line 217: golangci - Consider renaming this so it doesn't shadow the declaration of `err` above

stellarsaur (approved):
LGTM (minor golangci issues)

cmd/soroban-rpc/internal/ingest/service.go
tsachiherman: admittedly, it's not my preference, but I'll align with golangci.

cmd/soroban-rpc/internal/ingest/service.go
tsachiherman: yup.

PR #1061, opened 2023-10-31
soroban-cli: Warn or Error When Deploying Contracts Compiled with RC Version of Soroban SDK

### LINKED ISSUE #941: soroban-cli: warn user when deploying a contract that was compiled using release candidate SDK

The feature informs the end user that he is deploying a smart contract that was compiled against a release candidate version of the SDK. 

when user calls
```
soroban contract deploy
```

the following output might be generated:
```
warning: the deployed smart contract <smart contract hash> was complied with a release candidate version of the soroban SDK.
```

We might also want to add a similar feature to `soroban contract inspect`.

stellarsaur:
### What

This PR adds a check to the `soroban contract deploy` command if the Rust SDK version used to compile it is a release candidate SDK version.

If the new flag `--ignore-checks` is present, a warning is displayed. Otherwise, the command will return an error.

### Why

Release candidate versions of the Soroban Rust SDK are not suitable for use with the Stellar Public Network. We should at the very least warn users if they are deploying contracts compiled with release candidate SDK versions.

Closes #941

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/deploy.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy.rs
@@ -10,8 +10,8 @@ use soroban_env_host::{
         AccountId, ContractExecutable, ContractIdPreimage, ContractIdPreimageFromAddress,
         CreateContractArgs, Error as XdrError, Hash, HashIdPreimage, HashIdPreimageContractId,
         HostFunction, InvokeHostFunctionOp, Memo, MuxedAccount, Operation, OperationBody,
-        Preconditions, PublicKey, ScAddress, SequenceNumber, Transaction, TransactionExt, Uint256,
-        VecM, WriteXdr, ScMetaV0, ScMetaEntry,
+        Preconditions, PublicKey, ScAddress, ScMetaEntry, ScMetaV0, SequenceNumber, Transaction,
+        TransactionExt, Uint256, VecM, WriteXdr,
     },
     HostError,
 };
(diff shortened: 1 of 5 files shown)


willemneal:
I've been wanting to remove installing from the deploy command. However, regardless you need to add this check to the install command since the goal is that no incompatible contract should be installed.

So you'll need to add the new `-i` arg to the install command and move the logic there so that it will happen when the following function is called:

https://github.com/stellar/soroban-tools/pull/1061/files#diff-c8ba4158908b240ce43c9bcc7cbce0c5765ca1e1ab73ff09bce6cd26ea9a7b0fR135

cmd/soroban-cli/src/commands/contract/deploy.rs
```diff
@@ -22,6 +22,8 @@ use crate::{
     utils, wasm,
 };
 
+const CONTRACT_META_SDK_KEY: &str = "rssdkver";
```
willemneal, line 25: Is this from the sdk? If so could we import it?

stellarsaur:
> I've been wanting to remove installing from the deploy command. However, regardless you need to add this check to the install command since the goal is that no incompatible contract should be installed.
> 
> So you'll need to add the new `-i` arg to the install command and move the logic there so that it will happen when the following function is called:
> 
> https://github.com/stellar/soroban-tools/pull/1061/files#diff-c8ba4158908b240ce43c9bcc7cbce0c5765ca1e1ab73ff09bce6cd26ea9a7b0fR135

Thanks for taking a look! I moved the logic to `install` and propagated `-i` from the `deploy` command to `install`

cmd/soroban-cli/src/commands/contract/deploy.rs
stellarsaur: I found this [contractmetabuiltin macro](https://github.com/stellar/rs-soroban-sdk/blob/main/soroban-sdk-macros/src/lib.rs#L267) but it doesn't look like we can import the contract meta key for SDK version in a straightforward way.

tsachiherman (approved):
LGTM

willemneal (approved):

PR #1062, opened 2023-11-02
Upgrade Ubuntu to 22.04 from 20.04

satyamz:
### What:
Upgrade Ubuntu to 22.04 from 20.04

### Why:

We want to upgrade Ubuntu base image to 22.04.


mollykarcher:
We'll also want to update this in our [CI matrix]([.github/workflows/soroban-rpc.yml](https://github.com/stellar/soroban-tools/blob/c791a683af8d950a5b4b3105fd516502a0d8677e/.github/workflows/soroban-rpc.yml#L17))

satyamz:
@mollykarcher

 > We'll also want to update this in our [CI matrix](https://github.com/stellar/soroban-tools/pull/%5B.github/workflows/soroban-rpc.yml%5D(https://github.com/stellar/soroban-tools/blob/c791a683af8d950a5b4b3105fd516502a0d8677e/.github/workflows/soroban-rpc.yml#L17))

I am not sure if it would break anything there if we change Github actions workflow to use 22.04. Do you think it will break any workflow? Usually this change is for Docker base image which doesn't care about the host OS.

jacekn:
closing in favour of https://github.com/stellar/soroban-tools/pull/1158

PR #1063, opened 2023-11-02
optimize in-memory transaction store

tsachiherman:
### What

Increasing the transaction retention window would increase the memory utilization.
This optimization helps to reduce that.

### Why

Storing the expanded go-XDR in memory could be a memory hog compared to encoded XDR.

### Known limitations

The additional encoding during ingestion would slow down the ingestion ( a bit ).
This is a tradeoff in CPU & ingestion time vs memory utilization.

## Memory Consumption

| Branch | Transaction Count | Memory |
| --------------: | :--------- | ----------: |
| main | 10,000,000 | 22.7 GB |
| current | 10,000,000 | 8.8 GB |

The transaction count of 10M was selected since it aligns with the expected daily transaction volume of 500 transactions per ledger /w roughly 20,000 transactions per day ( 4.32 seconds / ledger )

### DIFF
--- a/cmd/soroban-rpc/internal/transactions/transactions.go
+++ b/cmd/soroban-rpc/internal/transactions/transactions.go
@@ -14,10 +14,11 @@ import (
 
 type transaction struct {
 	bucket           *ledgerbucketwindow.LedgerBucket[[]xdr.Hash]
-	result           xdr.TransactionResult
-	meta             xdr.TransactionMeta
-	envelope         xdr.TransactionEnvelope
+	result           []byte // encoded XDR of xdr.TransactionResult
+	meta             []byte // encoded XDR of xdr.TransactionMeta
+	envelope         []byte // encoded XDR of xdr.TransactionEnvelope
 	feeBump          bool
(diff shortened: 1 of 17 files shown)


cmd/soroban-rpc/internal/transactions/transactions.go
```diff
@@ -135,11 +143,12 @@ type LedgerInfo struct {
...
 type Transaction struct {
-	Result           xdr.TransactionResult
-	Meta             xdr.TransactionMeta
-	Envelope         xdr.TransactionEnvelope
+	Result           []byte // XDR encoded xdr.TransactionResult
+	Meta             []byte // XDR encoded xdr.TransactionMeta
+	Envelope         []byte // XDR encoded xdr.TransactionEnvelope
 	FeeBump          bool
 	ApplicationOrder int32
+	Successfull      bool
```
2opremio, line 151: Mispelling

2opremio:
Two minor things, otherwise LGTM

2opremio (approved):

PR #1065, opened 2023-11-02
feat: create soroban-rpc and break up send_transaction to fix not always submitting a transaction.

willemneal:
This creates a new crate for the RPC which is been a long time coming. Perhaps this crate and the go rpc could move to a separate repo. But either way it makes it clearer when the rpc is updated and allowing for updates to the rpc without updating the CLI's dep.

There is some response parsing done by subcommands that could move to the rpc, but this change is minimal to the current RPC API.


cmd/crates/soroban-rpc/Cargo.toml
```diff
@@ -0,0 +1,59 @@
+[package]
+name = "soroban-rpc"
```
leighmcculloch, line 2: I have pushed an empty crate to crates.io for this crate so that we know we can use it and its ready to go.

cmd/crates/soroban-rpc/README.md
```diff
@@ -0,0 +1,3 @@
+# soroban-rpc
+
+Tools and utilities for soroban rpc.
```
stellarsaur, line 3: Would be ideal to have a more comprehensive README but this is fine to get the crate created/published

Cargo.toml
```diff
@@ -61,26 +61,52 @@ rev = "822ce6cc3e461ccc925275b472d77b6ca35b2cd9"
 version = "20.0.2"
 path = "cmd/soroban-cli"
 
+[workspace.dependencies.soroban-rpc]
+version = "20.0.0-rc4"
```
stellarsaur, line 65: Should this be updated now that `20.1.1` is released?

stellarsaur (approved):

stellarsaur:
Thanks for taking this on! Please resolve merge conflicts and make sure all tests are passing and we can get this merged.

willemneal:
in lieu of #1174

PR #1066, opened 2023-11-02
Fix/typescript bindings

elizabethengelman:
### What
* update the `@stellar/freighter-api` import to use the `default` object if there is no wallet connected yet
* update the soroban-client version to `1.0.0-beta.3` for the cli typescript bindings generation

### Why

* update the `@stellar/freighter-api` import to use the `default`: 
    * Before this change, I was seeing an error when trying to use the generated bindings in the [Getting Started demo app](https://soroban.stellar.org/docs/getting-started/create-an-app#call-the-contract-from-the-frontend) when a wallet isn't connected yet:
`error   wallet.isConnected is not a function`

    * `@stellar/freighter-api` returns a null prototype with a default that has the expected methods, `isConnected`, etc. This should be sufficient for view calls, so this change just uses the `default` until later in the invoke method, when an account is needed (for change calls).

* update the soroban-client version to `1.0.0-beta.3` for the cli typescript bindings generation: the one update that seemed necessary to make this upgrade work was to change the `errorResultXdr` from `sendTransaction` result to `errorResult`.

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts
+++ b/cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts
@@ -82,7 +82,7 @@ export async function invoke<R extends ResponseTypes, T = string>({
   contractId,
   wallet,
 }: InvokeArgs<R, T>): Promise<T | string | SomeRpcResponse> {
-  wallet = wallet ?? (await import("@stellar/freighter-api"));
+  wallet = wallet ?? (await import("@stellar/freighter-api")).default;
   let parse = parseResultXdr;
   const server = new SorobanClient.Server(rpcUrl, {
     allowHttp: rpcUrl.startsWith("http://"),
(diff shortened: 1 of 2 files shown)


cmd/crates/soroban-spec-typescript/src/project_template/src/invoke.ts
```diff
@@ -17,6 +17,7 @@ import type {
 
 export type Tx = Transaction<Memo<MemoType>, Operation[]>;
 
+export class SendFailedError extends Error { }
```
elizabethengelman, line 20: This error /the pattern of throwing the error is being added in https://github.com/stellar/soroban-tools/pull/1034 as well

stellarsaur (approved):
LGTM

willemneal (approved):
Thanks! Glad to have human readable errors!

PR #1073, opened 2023-11-08
chore: don't force rebuild during build-test-wasms

chadoh:
### What

Remove the `Cargo.lock` prereq from the `build-test-wasms` task and the `check` from the `all:` directive

### Why

I don't see a good reason to include these, and they cause most tasks to take significantly longer to run. If there's no strong reason to include them, then it would be nice to save the build time.

### Known limitations

These were originally added here https://github.com/stellar/soroban-tools/pull/251 and here https://github.com/stellar/soroban-tools/commit/0c51782cb8ee9cf4e7b019e4edb807b3c52d963d. I don't see a description of why these particular changes were added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-all: check build test
+all: build test
 
 export RUSTFLAGS=-Dwarnings -Dclippy::all -Dclippy::pedantic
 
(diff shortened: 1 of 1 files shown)


leighmcculloch (commented):
I agree with the change. I'm somewhat confused why we're even manually calling `cargo update` since most cargo commands should do this as part of the typical work. So I'd be curious if we could pull it out entirely, but would like to hear from @2opremio @mollykarcher and other platform folks what the problem being solved is.

chadoh:
@leighmcculloch I think the rebuild might be caused by the `all: check` at the top, added by you in 0c51782cb. I have confirmed that the current change is insufficient to prevent `make build-snapshot` from recompiling everything every time.

leighmcculloch:
@chadoh I'm going to look into this today.

chadoh:
Resolved in a [Discord discussion](https://discord.com/channels/897514728459468821/1174787866530947072/1174787866530947072)

PR #1085, opened 2023-11-15
soroban-rpc: Ensure that the key corresponds to the value in getLedgerEntries

### LINKED ISSUE #1084: soroban-rpc: `getLedgerEntries` may interleave incorrect keys and values.
Testnet, which I believe is [`v20.0.0-rc.4.1`](https://github.com/stellar/soroban-tools/releases/tag/v20.0.0-rc.4.1).

### What did you do?
I requested two ledger entries: a `ContractData` key alongside its corresponding `ExpirationData` key. Here is a bash script:

```bash
#! /bin/bash

set -e
curl -s -H 'Content-Type: application/json' --data '{
  "jsonrpc": "2.0",
  "id": 12345,
  "method": "getLedgerEntries",
  "params": [[
    "AAAABgAAAAHm/PXiV0Tr/3LrABRnZtg8JYtnHiDjyuf2WN1Wx6xuBAAAABAAAAABAAAAAgAAAA8AAAAHUG9vbEVQUwAAAAASAAAAARlJVwVvl0UjM1CLV/d+MOmzOCgtTG85KqcIOlvOYObNAAAAAQ==",
    "AAAACa/lybv+NkUUDNmdxpiLnu6JZif6hKYdbMERMpzUNQLL"
  ]]
}' 'https://soroban-testnet.stellar.org' | jq '.result.entries'
```

I expected to see two returned `entries`: one with `key` as the same key and `xdr` as the corresponding `LedgerDataEntry`.

### What did you see instead?
**On occasion** (approx. one in a dozen runs), the values would interleave incorrectly with the keys. In other words, the `LedgerKeyContractData` in `key` would have the `ExpirationData` in `xdr` and vice-versa. Observe:

```jsonc
// $ ./repro.sh
[
  {
    "key": "AAAABgAAAAHm/PXiV0Tr/3LrABRnZtg8JYtnHiDjyuf2WN1Wx6xuBAAAABAAAAABAAAAAgAAAA8AAAAHUG9vbEVQUwAAAAASAAAAARlJVwVvl0UjM1CLV/d+MOmzOCgtTG85KqcIOlvOYObNAAAAAQ==",
    "xdr": "AAAABgAAAAAAAAAB5vz14ldE6/9y6wAUZ2bYPCWLZx4g48rn9ljdVsesbgQAAAAQAAAAAQAAAAIAAAAPAAAAB1Bvb2xFUFMAAAAAEgAAAAEZSVcFb5dFIzNQi1f3fjDpszgoLUxvOSqnCDpbzmDmzQAAAAEAAAAKAAAAAAAAAAAAAAAAABt3QA==",
    "lastModifiedLedgerSeq": "2535782"
  },
  {
    "key": "AAAACa/lybv+NkUUDNmdxpiLnu6JZif6hKYdbMERMpzUNQLL",
    "xdr": "AAAACa/lybv+NkUUDNmdxpiLnu6JZif6hKYdbMERMpzUNQLLACpiZg==",
    "lastModifiedLedgerSeq": "2535782"
  }
]
// $ ./repro.sh
[
  {
    "key": "AAAABgAAAAHm/PXiV0Tr/3LrABRnZtg8JYtnHiDjyuf2WN1Wx6xuBAAAABAAAAABAAAAAgAAAA8AAAAHUG9vbEVQUwAAAAASAAAAARlJVwVvl0UjM1CLV/d+MOmzOCgtTG85KqcIOlvOYObNAAAAAQ==",
    "xdr": "AAAACa/lybv+NkUUDNmdxpiLnu6JZif6hKYdbMERMpzUNQLLACpiZg==",
    "lastModifiedLedgerSeq": "2535782"
  },
  {
    "key": "AAAACa/lybv+NkUUDNmdxpiLnu6JZif6hKYdbMERMpzUNQLL",
    "xdr": "AAAABgAAAAAAAAAB5vz14ldE6/9y6wAUZ2bYPCWLZx4g48rn9ljdVsesbgQAAAAQAAAAAQAAAAIAAAAPAAAAB1Bvb2xFUFMAAAAAEgAAAAEZSVcFb5dFIzNQi1f3fjDpszgoLUxvOSqnCDpbzmDmzQAAAAEAAAAKAAAAAAAAAAAAAAAAABt3QA==",
    "lastModifiedLedgerSeq": "2535782"
  }
]
// $ ./repro.sh
[
  {
    "key": "AAAABgAAAAHm/PXiV0Tr/3LrABRnZtg8JYtnHiDjyuf2WN1Wx6xuBAAAABAAAAABAAAAAgAAAA8AAAAHUG9vbEVQUwAAAAASAAAAARlJVwVvl0UjM1CLV/d+MOmzOCgtTG85KqcIOlvOYObNAAAAAQ==",
    "xdr": "AAAABgAAAAAAAAAB5vz14ldE6/9y6wAUZ2bYPCWLZx4g48rn9ljdVsesbgQAAAAQAAAAAQAAAAIAAAAPAAAAB1Bvb2xFUFMAAAAAEgAAAAEZSVcFb5dFIzNQi1f3fjDpszgoLUxvOSqnCDpbzmDmzQAAAAEAAAAKAAAAAAAAAAAAAAAAABt3QA==",
    "lastModifiedLedgerSeq": "2535782"
  },
  {
    "key": "AAAACa/lybv+NkUUDNmdxpiLnu6JZif6hKYdbMERMpzUNQLL",
    "xdr": "AAAACa/lybv+NkUUDNmdxpiLnu6JZif6hKYdbMERMpzUNQLLACpiZg==",
    "lastModifiedLedgerSeq": "2535782"
  }
]
```

The first and third execution are correct. However, in the second execution, the `xdr` fields do not match their `key` fields. Again, this happens sporadically.

## Additional Investigation
I believe the culprit lies in these lines of code:

https://github.com/stellar/soroban-tools/blob/3fcab5b4d46c441548c78b003c2c0c2513aa070f/cmd/soroban-rpc/internal/methods/get_ledger_entries.go#L111-L115

There is an inherent assumption that the earlier call to `tx.GetLedgerEntries(ledgerKeys...)` ([line 90](https://github.com/stellar/soroban-tools/blob/3fcab5b4d46c441548c78b003c2c0c2513aa070f/cmd/soroban-rpc/internal/methods/get_ledger_entries.go#L90)) will return the entries in the same order as they were requested. However, this empirically does not appear to be the case. Instead, we should encode the key the same way we encode the XDR when building the entry rather than relying on the `i` index.

I believe this should also be backported to the current testnet release (i.e. some `rc.4.2`), which should not be difficult because that file has not changed since then, so there shouldn't be conflicts.

Shaptic:
### What
Encodes the key alongside the value because `request.Keys[i]` may not always match it.

### Why
It should fix #1084.

### Known limitations
It's not tested because it's a bit of a [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug) (more accurately, it's non-deterministic).

### DIFF
--- a/cmd/soroban-rpc/internal/methods/get_ledger_entries.go
+++ b/cmd/soroban-rpc/internal/methods/get_ledger_entries.go
@@ -89,28 +89,40 @@ func NewGetLedgerEntriesHandler(logger *log.Entry, ledgerEntryReader db.LedgerEn
 		ledgerEntryResults := make([]LedgerEntryResult, 0, len(ledgerKeys))
 		ledgerKeysAndEntries, err := tx.GetLedgerEntries(ledgerKeys...)
 		if err != nil {
-			logger.WithError(err).WithField("request", request).
-				Info("could not obtain ledger entryies from storage")
+			message := "could not obtain ledger entries from storage"
+			logger.WithError(err).WithField("request", request).Info(message)
 			return GetLedgerEntriesResponse{}, &jrpc2.Error{
 				Code:    jrpc2.InternalError,
-				Message: "could not obtain ledger entryies from storage",
+				Message: message,
 			}
 		}
 
-		for i, ledgerKeyAndEntry := range ledgerKeysAndEntries {
-			ledgerXDR, err := xdr.MarshalBase64(ledgerKeyAndEntry.Entry.Data)
+		for _, ledgerKeyAndEntry := range ledgerKeysAndEntries {
+			entryXdr, err := xdr.MarshalBase64(ledgerKeyAndEntry.Entry.Data)
 			if err != nil {
-				logger.WithError(err).WithField("request", request).
-					Infof("could not serialize ledger entry data for ledger entry %v", ledgerKeyAndEntry.Entry)
+				message := fmt.Sprintf(
+					"could not serialize ledger entry data for ledger entry %v",
+					ledgerKeyAndEntry.Entry)
+				logger.WithError(err).WithField("request", request).Info(message)
 				return GetLedgerEntriesResponse{}, &jrpc2.Error{
 					Code:    jrpc2.InternalError,
-					Message: fmt.Sprintf("could not serialize ledger entry data for ledger entry %v", ledgerKeyAndEntry.Entry),
+					Message: message,
+				}
+			}
+
+			keyXdr, err := xdr.MarshalBase64(ledgerKeyAndEntry.Key)
+			if err != nil {
+				message := fmt.Sprintf("could not serialize ledger key %v", ledgerKeyAndEntry.Key)
+				logger.WithError(err).WithField("request", request).Infof(message)
+				return GetLedgerEntriesResponse{}, &jrpc2.Error{
+					Code:    jrpc2.InternalError,
+					Message: message,
 				}
 			}
 
 			ledgerEntryResults = append(ledgerEntryResults, LedgerEntryResult{
-				Key:                request.Keys[i],
-				XDR:                ledgerXDR,
+				Key:                keyXdr,
+				XDR:                entryXdr,
 				LastModifiedLedger: int64(ledgerKeyAndEntry.Entry.LastModifiedLedgerSeq),
 			})
 		}


2opremio:
I will look into it

2opremio:
@Shaptic I created https://github.com/stellar/soroban-tools/pull/1093 which should override this PR

Issue #1018, opened 2023-10-11
soroban-cli: replace lab xdr subcommand with embedding the stellar-xdr cli


leighmcculloch:
The soroban-cli has a `lab xdr` sub command that supports decoding XDR.

However, the stellar-xdr cli tool has more features and is likely to be developed further than the soroban-cli's xdr command.

We should embed the stellar-xdr cli inside the soroban-cli, so that the soroban-cli benefits from the stellar-xdr.

```[tasklist]
### Tasks
- [x] Export the CLI from the stellar-xdr lib. https://github.com/stellar/rs-stellar-xdr/pull/326
- [x] Export the CLI root command run fns from the stellar-xdr lib. https://github.com/stellar/rs-stellar-xdr/pull/337
- [x] Release stellar-xdr. https://github.com/stellar/rs-stellar-xdr/pull/338
- [x] Update patch of soroban-env. https://github.com/stellar/rs-soroban-env/pull/1305
- [x] Release patch of soroban-env. https://github.com/stellar/rs-soroban-env/pull/1304
- [x] Update patch of soroban-sdk. https://github.com/stellar/rs-soroban-sdk/pull/1190
- [x] Release patch of soroban-sdk. https://github.com/stellar/rs-soroban-sdk/pull/1191
- [ ] Import that exported CLI into the soroban-cli. https://github.com/stellar/soroban-tools/pull/1135
```

stellarsaur:
Just to clarify, the scope of this issue is to essentially merge [stellar/rs-stellar-xdr](https://github.com/stellar/rs-stellar-xdr) into soroban-cli, right? Should we deprecate/remove the original repo in that case?

leighmcculloch:
No, the scope of the issue to embed the tool.

PR #1086, opened 2023-11-16
feat!: move config commands to top level

### LINKED ISSUE #934: soroban-cli: refers to 'stellar asset contract' as 'token'

9615897821492671f951169f8f442d69cafb3d81

```console
❯ soroban version
soroban 0.9.4 (v0.9.5-70-g9615897821492671f951169f8f442d69cafb3d81)
soroban-env 0.0.17 (eb5a9ba053a7b64a8eff605db625525378f7bea0)
soroban-env interface version 85899345977
stellar-xdr 0.0.17 (39904e09941046dab61e6e35fc89e31bf2dea1cd)
xdr next (65afa63b7f52c898143ebbe9541ef91fcf290ade)
```

### What did you do?

```console
❯ soroban lab -h
Experiment with early features and expert tools

Usage: soroban lab <COMMAND>

Commands:
  token  Wrap, create, and manage token contracts
  xdr    Decode xdr

❯ soroban lab token -h
Wrap, create, and manage token contracts

Usage: soroban lab token <COMMAND>

Commands:
  wrap  Deploy a token contract to wrap an existing Stellar classic asset for smart contract usage
  id    Compute the expected contract id for the given asset

```

Instructions about deploying and interacting with a stellar asset contract for a stellar asset.

### What did you see instead?
```
  token  Wrap, create, and manage token contracts
```

### Discussion

Quite a few releases ago we stopped referring to the stellar asset contract as the token contract.

There are two distinct but related concepts in Soroban:
1. token interface;
2. and, the stellar asset contract.

The stellar asset contract implements the token interface for all stellar assets.

Contract developers can create their own contracts that implement the token interface.

The CLI provides a confusing interface because it mixes this terminology, and uses terms that don't exactly parallel the XDR or SDK. We should make these things more consistent.

A few issues:
- The CLI refers to wrapping and creating as if there are two distinction actions. However there is only one action.
- The CLI refers to the stellar asset contract at a high level as 'token'.
- The CLI uses the term 'wrap' to describe an action where no new token is being created, a contract for an existing token (the stellar asset) is being deployed.

### LINKED ISSUE #1053: soroban-cli: rename `config identity` to `config key`
I think we should consider renaming `config identity` to `config key`.

Keys are buried in the soroban-cli under the `config identity` command. It's easy to miss it and think the soroban command doesn't do key generation or key management.

While identity is a nice name for a number of reasons of why we picked it, it reduces the discoverability.

### LINKED ISSUE #1054: soroban-cli: move `config *` commands to top level
I think we should consider moving the commands that live under `soroban config` to the top level.

i.e.:

`config identity` => `key` (assuming #1053)
`config network` => `network`

The `network` and `identity` commands are buried under config which makes them hard to find. They're important concepts that devs need to access regularly.

We should bury things under config if they're a set once and forget type of thing.

In reality devs need to interact with keys frequently. And likely need to switch and use multiple networks frequently too.

### LINKED ISSUE #1079: soroban-cli: uses the same identity/account/key for everybody by default

```console
$ soroban version
soroban 20.0.0-rc.4.2 (v20.0.0-rc.4.1-4-g4e9a48756b96e4e1a070105bad202f39901139ad-dirty)
soroban-env 20.0.0-rc2 (8c63bff68a15d79aca3a705ee6916a68db57b7e6)
soroban-env interface version 85899345977
stellar-xdr 20.0.0-rc1 (d5ce0c9e7aa83461773a6e81662067f35d39e4c1)
xdr curr (9ac02641139e6717924fdad716f6e958d0168491)
```

### What did you do?

Used the default identity:

```console
$ soroban config identity address
```

A non-shared address. Maybe randomly generated on first use.

### What did you see instead?

An address and key that everybody who is using the soroban-cli will use by default.

```
GDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCR4W4
```

### Why is this a security bug?

It's surprising that any key is shared.

Tooling that provides any shared key runs the risk of causing someone to think they have a unique key when they don't. Somebody may use the key on mainnet thinking it is a key unique to themselves.

### Why is this a general bug?

Developers who use the default identity will be contending for access to the network. The network will only allow one tx per account in each ledger.

willemneal:
### What

- [x] fixes #1079
- [x] fixes #1054 
- [x] fixes #1053 
- [x] fixes #934
    - `contract asset` was added with `deploy` and `id` subcommands.
    - Initially `contract deploy` had two subcommands `wasm` and `asset`, but this was a breaking change and it was hard to have clap allow `deploy` without a subcommand to default to `wasm`. So perhaps we will add this when introducing breaking changes.
- [x] Removed `xdr` subcommand, but will address later. #1018

### Why

### Known limitations

[TODO or N/A]


willemneal:
@leighmcculloch The last step is to split identity into keys and address book. Should I put them under the network too?

That is store them in a folder for each network?

leighmcculloch:
I thought the identity command today was just keys that have a secret, not address only identities. I don't think we need to add the address book now if so.

leighmcculloch:
> That is store them in a folder for each network?

I'm concerned that separating keys by network in the CLI UI will create for a confusing or unintuitive interface. I think we should do some UI design in a doc before embarking on it in code.

willemneal:
Yeah the addresses become tricky because contract id's are derived from the network passsphrase and are obviously only deployed on one network. So we could keep keys as is and add the address book per network. But either way I'll just update this PR to just rename `identity` to `keys`.

2opremio:
> @2opremio Looks like the prefight debug test is failing:
> 
> ```
> === RUN   TestGetPreflightDebug
>     preflight_test.go:388: 
>                 Error Trace:    /Users/willem/c/s/soroban-cli/cmd/soroban-rpc/internal/preflight/preflight_test.go:388
>                 Error:          "host invocation failed\n\nCaused by:\n    HostError: Error(WasmVm, MissingValue)\n    DebugInfo not available\n    " does not contain "Backtrace"
> ```

Uhm, can you try to debug it? I am off this week. Thanks!

willemneal:
> Food for thought, and not this PR. Would be nice to be able view accounts in a table-like view with its respective network futurenet|testnet|mainnet
> 
> ```
> Network Name Public key 
> ─────── ──── ─────── 
> testnet alice G-
> testnet bob G-
> ```

I agree, but we currently don't store keys under their network so it's not possible. So I'll make a future PR that addresses this.

willemneal:
I purposely depended on the stellar-xdr` repo at a different version in the soroban-cli crate to get access to the`cil` feature. However, when I initially used the version at the top level it broke some things.  I can either update the whole workspace to the newest xdr and then merge the xdr CLI stuff or we can ignore this for now? https://github.com/stellar/soroban-tools/actions/runs/7104234281/job/19338741451?pr=1086

I vote for shelving the xdr cli until the whole workspace is updated to the newest stellar-xdr, as it doesn't make sense that the rest of the workspace has a different version.

willemneal:
@leighmcculloch @mollykarcher 

The CI dep checker failed with just the following:
> Error: Process completed with exit code 6.

leighmcculloch:
Would be great to get this included in the next release, but a couple things:

1. Looks like there's one system test failing that might be important.

2. There are some minor conflicts before it can merge.

3. This change is a breaking change to the CLI's UI, and we've released v20 stable already, so can we add hidden subcommands, or subcommands marked as deprecated, into the old locations for the commands being moved/renamed so that anyone already using them will continue to be able to use them?

janewang:
Let's make this change not breaking please https://github.com/stellar/soroban-tools/issues/1146

PR #1091, opened 2023-11-20
soroban-rpc: Force On-Disk Mode by Removing --captive-core-use-db

stellarsaur:
### What

This PR removes the `--captive-core-use-db` flag and forces on-disk mode.

### Why

Eventually, the core working-state size will grow, and eventually get too big for RAM. This means in-memory captive-core will not work long-term. Therefore, we should force users to use on-disk mode.

### DIFF
--- a/cmd/soroban-rpc/internal/config/config.go
+++ b/cmd/soroban-rpc/internal/config/config.go
@@ -15,7 +15,6 @@ type Config struct {
 	Strict bool
 
 	StellarCoreURL         string
-	CaptiveCoreUseDB       bool
 	CaptiveCoreStoragePath string
 	StellarCoreBinaryPath  string
 	CaptiveCoreConfigPath  string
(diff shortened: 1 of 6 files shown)


Shaptic (approved):

2opremio:
Maybe we shouldn't be worrying about this since we haven't reached pubnet but ... shouldn't we deprecate the parameter first? (and/or maybe change it's default value if necessary)

stellarsaur:
> Maybe we shouldn't be worrying about this since we haven't reached pubnet but ... shouldn't we deprecate the parameter first? (and/or maybe change it's default value if necessary)

That's a valid point. If we were in pubnet I would want to deprecate the parameter first and set the default to `true`, but in this case I think simply removing it is fine. Don't feel too strongly about it though

mollykarcher:
While I agree in general the deprecation pattern is what we should follow, since we're technically "pre-release" here, I'd prefer to just kill this off. We've run into a lot of problems with partner support scenarios because of Horizon supporting (and until recently, defaulting to) in-memory mode, despite that not being our/core team's recommendation. There's almost no advantage to running it, and if we support it out of the gate, we likely also need to profile hardware and define scaling guidelines for both configurations.

PR #1093, opened 2023-11-20
rpc: fix multi-entry responses for getLedgerEntries method

2opremio:
Closes https://github.com/stellar/soroban-tools/issues/1084 

The key->entry correspondence was broken since the code assumed a 1-to-1 match (in size and ordering) between the keys in the request and the keys in the response.

This didn't hold because:

1. Some keys may be missing in the response (if not found in the db).
2. The db access method (`LedgerEntryReadTx.GetLedgerEntries()`) returned a result whose ordering wasn't consistent with its arguments (due to traversing a map to build the result).

This change:

1. Fixes the key->entry mapping
2. Makes `getLedgerEntries` response ordering stable and consistent with the keys provided.
3. Make a few small performance improvements (mostly allocation related) along the way.
4. Increases test coverage.

### DIFF
--- a/cmd/soroban-rpc/internal/db/ledgerentry.go
+++ b/cmd/soroban-rpc/internal/db/ledgerentry.go
@@ -260,16 +260,22 @@ func entryKeyToTTLEntryKey(key xdr.LedgerKey) (xdr.LedgerKey, error) {
 }
 
 func (l *ledgerEntryReadTx) GetLedgerEntries(keys ...xdr.LedgerKey) ([]LedgerKeyAndEntry, error) {
-	encodedKeys := make([]string, len(keys), 2*len(keys))
-	encodedKeyToKey := make(map[string]xdr.LedgerKey, len(keys))
-	encodedKeyToEncodedTTLLedgerKey := make(map[string]string, len(keys))
-	for _, k := range keys {
+	encodedKeys := make([]string, 0, 2*len(keys))
+	type keyToEncoded struct {
+		key           xdr.LedgerKey
+		encodedKey    string
+		encodedTTLKey *string
+	}
+	keysToEncoded := make([]keyToEncoded, len(keys))
+	for i, k := range keys {
+		k2 := k
+		keysToEncoded[i].key = k2
 		encodedKey, err := encodeLedgerKey(l.buffer, k)
 		if err != nil {
 			return nil, err
 		}
+		keysToEncoded[i].encodedKey = encodedKey
 		encodedKeys = append(encodedKeys, encodedKey)
-		encodedKeyToKey[encodedKey] = k
 		if !hasTTLKey(k) {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


2opremio:
Overrides https://github.com/stellar/soroban-tools/issues/1084

cmd/soroban-rpc/internal/db/ledgerentry.go
```diff
@@ -260,16 +260,22 @@ func entryKeyToTTLEntryKey(key xdr.LedgerKey) (xdr.LedgerKey, error) {
...
-	encodedKeys := make([]string, len(keys), 2*len(keys))
-	encodedKeyToKey := make(map[string]xdr.LedgerKey, len(keys))
-	encodedKeyToEncodedTTLLedgerKey := make(map[string]string, len(keys))
-	for _, k := range keys {
+	encodedKeys := make([]string, 0, 2*len(keys))
+	type keyToEncoded struct {
+		key           xdr.LedgerKey
+		encodedKey    string
+		encodedTTLKey *string
+	}
```
Shaptic, line 268: Can you explain why one is a pointer while the other isn't? Is it because not every ledger key has a corresponding TTL key (e.g. an account entry)? I thought this was the explanation but I can't find the part where we skip over non-expiring entries in the code.

Shaptic (approved):
Awesome! Thanks fons :clap: 

Two logistical questions:
 * Do we typically add changelog entries in PRs or are they summarized as part of the release?
 * Should this be targeting `rc.4.2` so we can release it to testnet?

cmd/soroban-rpc/internal/db/ledgerentry.go
2opremio: >  Is it because not every ledger key has a corresponding TTL key (e.g. an account entry)? 

Exactly

They are skipped at:

```go
		if !hasTTLKey(k) {
			continue
		}
```

2opremio:
> Awesome! Thanks fons 👏
> 
> Two logistical questions:
> 
>     * Do we typically add changelog entries in PRs or are they summarized as part of the release?

We don't even have a CHANGELOG file. It seems tsachi has been using github releases as a changelog. 

But ... we could change that if you think it's better.

>     * Should this be targeting `rc.4.2` so we can release it to testnet?

Uhm, we can do it if necessary.

But ... the XDR and symbol names have considerably diverged from what's on testnet since https://github.com/stellar/soroban-tools/pull/1029 , so I am not sure it's worth it.

The bug is pretty horrible though, but I think it has gone unnoticed until because most people do single key requests.

Shaptic:
@2opremio agreed with your point about divergence! Let's merge to main for now. And we can keep it in the release notes for now, but we should start a changelog once we have a stable release imo :+1: 

I'll look into cherry picking it afterward: this has bigger effects because of the workaround @sreuland added in https://github.com/stellar/js-soroban-client/pull/161 so that users can get expiration data after it moved into its own entry, so most requests will have 2 keys :sweat_smile: if it's too messy I'll just abandon it

Issue #1081, opened 2023-11-14
Sequence JSON return types are inconsistent across methods


ElliotFriend:
I've noticed that the ledger sequence number isn't always returned as the same type in some of the Soroban-RPC methods. The `getLatestLedger` method returns the `sequence` as a `uint32`. Elsewhere when a ledger's sequence number is returned, it's stringified in the JSON response.

I could've sworn I noticed this same kind of thing with a `protocolVersion` field in two of the methods, but I'm not finding it now. I might've just been conflating some things in my head.

ElliotFriend:
For some additional details and context, here are all the occurrences I can find of a ledger sequence number in the methods, and how it's returned

| method name | return name | struct type | json type |
| --- | --- | --- | --- |
| `getLatestLedger` | `sequence` | `uint32` | n/a |
| `getLedgerEntries` | `lastModifiedLedgerSeq` | `int64` | `string` |
| `getLedgerEntries` | `liveUntilLedgerSeq` | `*uint32` | `string` |
| `getLedgerEntries` | `latestLedger` | `int64` | `string` |
| `getEvents` | `ledger` | `int32` | `string` |
| `getTransaction` | `latestLedger` | `int64` | `string` |
| `getTransaction` | `oldestLedger` | `int64` | `string` |
| ` getTransaction` | `ledger` | `int64` | `string` |
| `sendTransaction` | `latestLedger` | `int64` | `string` |
| `simulateTransaction` | `latestLedger` | `int64` | `string` |

Seems like it's only returned as a number in the `getLatestLedger` method. I am curious if there's a significance or reasoning behind the use of `int64` vs. `uint32` vs. `int32` vs. `*uint32` in the structs, though? Maybe that's just how the data is retrieved from stellar core?

leighmcculloch:
Ledger sequences are uint32 in the xdr, so I don't think we need to be communicating the sequence as a int64 or as a string. It looks like most of the fields need updating.

See:
https://github.com/stellar/stellar-xdr/blob/6a620d160aab22609c982d54578ff6a63bfcdc01/Stellar-ledger.x#L82

ElliotFriend:
[...]

PR #1097, opened 2023-11-23
rpc: unify all ledger sequence types to uint32 and stop stringyfying integers < 53-bits wide

2opremio:
### What

Closes https://github.com/stellar/soroban-tools/issues/1081

### Why

Consistent is good plus it's cleaner to not stringify fields we don't need to (since they can be treated by javascript).

### Known limitations

Downstream tickets are needed.

### DIFF
--- a/cmd/soroban-cli/src/rpc/mod.rs
+++ b/cmd/soroban-cli/src/rpc/mod.rs
@@ -111,10 +111,7 @@ pub struct SendTransactionResponse {
         default
     )]
     pub error_result_xdr: Option<String>,
-    #[serde(
-        rename = "latestLedger",
-        deserialize_with = "deserialize_number_from_string"
-    )]
+    #[serde(rename = "latestLedger")]
     pub latest_ledger: u32,
     #[serde(
         rename = "latestLedgerCloseTime",
@@ -174,10 +171,7 @@ impl TryInto<GetTransactionResponse> for GetTransactionResponseRaw {
 pub struct LedgerEntryResult {
     pub key: String,
     pub xdr: String,
-    #[serde(
-        rename = "lastModifiedLedgerSeq",
-        deserialize_with = "deserialize_number_from_string"
-    )]
+    #[serde(rename = "lastModifiedLedgerSeq")]
     pub last_modified_ledger: u32,
     #[serde(
         rename = "liveUntilLedgerSeq",
@@ -191,10 +185,7 @@ pub struct LedgerEntryResult {
 #[derive(serde::Deserialize, serde::Serialize, Debug)]
 pub struct GetLedgerEntriesResponse {
     pub entries: Option<Vec<LedgerEntryResult>>,
-    #[serde(
-        rename = "latestLedger",
-        deserialize_with = "deserialize_number_from_string"
-    )]
+    #[serde(rename = "latestLedger")]
     pub latest_ledger: i64,
 }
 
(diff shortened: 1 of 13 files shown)


leighmcculloch:
> javascript doesn't handle uints that large

JS handles numbers safely up to [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), that is  `9,007,199,254,740,991`.

Uint32 max is `4,294,967,295`.

So I think JS will handle uint32's fine. Anything over 53-bit is a problem though.

2opremio:
OK, so the problem happened with int64 numbers, which (AFAICT) is overkill for sequence numbers. And (AFAICT) for all practical purposes `uint32` is enough for all practical purposes (in fact I think the XDR uses uint32 at some places?).

However, also if I recall properly, Horizon uses `int64` everywhere. @Shaptic @tamirms , do you know why that is?

leighmcculloch:
> in fact I think the XDR uses `uint32` at some places?

This is correct. The XDR uses `uint32` for ledger seq.

https://github.com/stellar/stellar-xdr/blob/6a620d160aab22609c982d54578ff6a63bfcdc01/Stellar-ledger.x#L82

https://github.com/search?q=repo:stellar/stellar-xdr%20ledgerSeq&type=code

I imagine it would be such a heavy lift to change from uint32 to uint64 in the XDR that for a future change like that updating Horizon and Soroban RPC's APIs in support of that would probably be a small job in comparison to everything else.

But I also think it's reasonable to have the RPC API use strings for all integers irrespective of their bit size. I just wanted to surface that 32bit would fit if it was preferred to be a number.

leighmcculloch:
It appears Horizon uses 32-bit for ledger seq too:
- https://github.com/stellar/go/blob/69266de4154fdf4dd5e8f39ada37f642cb421ca1/protocols/horizon/main.go#L344C10-L348
- https://github.com/stellar/go/blob/69266de4154fdf4dd5e8f39ada37f642cb421ca1/protocols/horizon/main.go#L505

Shaptic:
We may have conflated _account_ sequence with _ledger_ sequence: the former needs to be `int64` because it can be set to any value, while the latter is guaranteed to grow sequentially and is `uint32` in the XDR as you guys have mentioned. 

I don't think there's anything wrong with leaving it stringified, but it is not necessary because all languages can handle them being encoded as integers (while JS has issues with values over 53 bits, like Leigh said).

Shaptic (approved):
I will leave it up to you whether or not we want to keep them JSON-encoded as strings! I'm just glad we are consistent, now.

2opremio:
It think it's cleaner to not stringify them but ... at this point it may cause too much turmoil.

@Shaptic you are dealing with downstream tickets nowadays, wdyt?

Shaptic:
@2opremio imo it's all the same: we have to issue a downstream ticket no matter what for at least a single field, so we might as well fix them all while we're there, and the maintainers are all very responsive.

2opremio:
Alright. Uint32 it is then!

2opremio:
@Shaptic can you please create downstream tickets? I will update the documentation.

Shaptic:
@2opremio draft is [here](https://gist.github.com/Shaptic/d8cac9464f7dce6489511f4f16370957), please double-check it! I will also cross-reference it with the doc updates once you make them.

PR #1100, opened 2023-11-25
Make output consistent for all ways to get version

### LINKED ISSUE #795: version, --version, -V, -v weirdness
```bash
% soroban -V
soroban 0.9.1 (339cc91f6db1752fed4220df2a2fafe499644fbf)
```
```bash
% soroban --version
soroban 0.9.1 (339cc91f6db1752fed4220df2a2fafe499644fbf)
soroban-env 0.0.17 (400d806387140553e4e685d232deb3a807ec0e36)
soroban-env interface version 85899345971
stellar-xdr 0.0.17 (0f16673441898162c9996da6117be2280ef8fd84)
```

`-V` vs `-v` seems like an anti-pattern. If we want to shorten the `--verbose` command it should be `-V` and `-v` should be version and it should always print out all the version not just the `soroban` cli version.

leighmcculloch:
### What
Make output of any way to get at the version consistent.

### Why
We should have one version output for folks to provide consistent version info when reporting bugs.

Close #795

### DIFF
--- a/cmd/soroban-cli/src/commands/mod.rs
+++ b/cmd/soroban-cli/src/commands/mod.rs
@@ -45,11 +45,10 @@ Full CLI reference: https://github.com/stellar/soroban-tools/tree/main/docs/soro
 #[derive(Parser, Debug)]
 #[command(
     name = "soroban",
-    version = version::short(),
-    long_version = version::long(),
     about = ABOUT,
     long_about = ABOUT.to_string() + LONG_ABOUT,
     disable_help_subcommand = true,
+    disable_version_flag = true,
 )]
 pub struct Root {
     #[clap(flatten)]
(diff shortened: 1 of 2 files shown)


cmd/soroban-cli/src/commands/mod.rs
```diff
@@ -45,11 +45,10 @@ Full CLI reference: https://github.com/stellar/soroban-tools/tree/main/docs/soro
 #[derive(Parser, Debug)]
 #[command(
     name = "soroban",
-    version = version::short(),
-    long_version = version::long(),
     about = ABOUT,
     long_about = ABOUT.to_string() + LONG_ABOUT,
     disable_help_subcommand = true,
+    disable_version_flag = true,
```
chadoh, line 51: We used to disable `--version` / `-v` and added it back in https://github.com/stellar/soroban-tools/pull/484

I agree that removing the difference between `version` and `long_version` seems useful to cut down on troubleshooting convo length, but personally find it unexpected when a CLI doesn't support `--version` or `-v`.

leighmcculloch: > I agree that removing the difference between `version` and `long_version` seems useful to cut down on troubleshooting convo length

Good point, this is really the most important thing to change. If there are extra other ways to get at version it doesn't matter so much. I'll reduce this change to that then I think we should move on to other things.

PR #1103, opened 2023-11-28
all: Bump core, Rust dependencies and update code

2opremio:


willemneal (approved):
Not sure what the mint contract #2 error is, but I can say that after updating locally I'm able to finally run tests and they are passing, except the one failing on CI.

PR #1107, opened 2023-12-04
feat: start network docker container with cli

elizabethengelman:
### What
This PR adds the `network start` and `network stop` subcommands to allow users to start and stop the stellar docker container using the CLI. 

#### Start help output
```
$ soroban network start --help
Start network

This command allows for starting a stellar quickstart container. To run it, you can use the following command: `soroban network start <NETWORK> [OPTIONS]`

By default, when starting a testnet container, without any optional arguments, it will run the equivalent of the following docker command: docker run --rm -p 8000:8000 --name stellar stellar/quickstart:testing --testnet --enable-soroban-rpc

Usage: soroban network start [OPTIONS] <NETWORK>

Arguments:
  <NETWORK>
          Network to start

          [possible values: local, testnet, futurenet, pubnet]

Options:
  -d, --docker-socket-path <DOCKER_SOCKET_PATH>
          Optional argument to override the default docker socket path. This is useful when you are using a non-standard docker socket path for your Docker-compatible container runtime, e.g. Docker Desktop defaults to $HOME/.docker/run/docker.sock instead of /var/run/docker.sock

  -l, --limit <LIMIT>
          Optional argument to specify the limits for the local network only

  -p, --ports-mapping <PORTS_MAPPING>...
          Argument to specify the HOST_PORT:CONTAINER_PORT mapping

          [default: 8000:8000]

  -r, --disable-soroban-rpc
          Optional argument to turn off soroban rpc

  -t, --image-tag-override <IMAGE_TAG_OVERRIDE>
          Optional argument to override the default docker image tag for the given network

  -v, --protocol-version <PROTOCOL_VERSION>
          Optional argument to specify the protocol version for the local network only

  -h, --help
```

#### Stop help output
```
$ soroban network stop --help
Stop network

Usage: soroban network stop [OPTIONS] <NETWORK>

Arguments:
  <NETWORK>  Network to stop [possible values: local, testnet, futurenet, pubnet]

Options:
  -d, --docker-socket-path <DOCKER_SOCKET_PATH>
          Optional argument to override the default docker socket path. This is useful when you are using a non-standard docker socket path for your Docker-compatible container runtime, e.g. Docker Desktop defaults to $HOME/.docker/run/docker.sock instead of /var/run/docker.sock
  -h, --help
          Print help
```

#### Output when the command is not able to connect to the docker daemon
```
$ soroban network start testnet
ℹ️  Starting testnet network
⛔️ Failed to connect to the Docker daemon at "/var/run/docker.sock". Is the docker daemon running?
ℹ️  Running a local Stellar network requires a Docker-compatible container runtime.
ℹ️  Please note that if you are using Docker Desktop, you may need to utilize the `--docker-socket-path` flag to pass in the location of the docker socket on your machine.

error: ⛔ ️Failed to start container: error trying to connect: No such file or directory (os error 2)
```

#### Output when the command starts the container successfully
```
ℹ️  Starting testnet network
✅ Container started: stellar-testnet
ℹ️  To stop this container run: soroban network stop testnet --docker-socket-path /Users/elizabethengelman/.docker/run/docker.sock
```

### Why
To make it easier for users to run a stellar node for soroban development.

### Known limitations

I learned that when using Docker Desktop, it starts the docker daemon on a non-standard socket 😵‍💫 . The Docker Desktop UI allows a user to overwrite this, in an advanced setting. But I figured allowing the user to do this all via the CLI would probably be easier, so I added the `--docker-socket-path` flag. I think that it may be helpful to allow the user to set an env var for this, instead of always passing a flag, but I'll plan to allow for this in a followup PR.

resources that helped identify this:
    - https://github.com/fussybeaver/bollard/issues/345 (suggestion of using DOCKER_HOST env var did not work)
    - https://docs.docker.com/desktop/mac/permission-requirements/
    - https://devops.stackexchange.com/questions/17905/ctop-does-not-work-with-latest-version-of-docker-get-http-unix-sock-info-d#:~:text=This%20is%20probably%20because%20Docker,sock%20anymore.&text=If%20you%20want%20to%20have,used%20and%20press%20Apply%20%26%20restart%20.

### Todo
- [x] handle unwraps more gracefully
- [x] revisit errors / error handling
- [x] pull shared stuff out into a shared module (specifically docker connect stuff)
- [x] address the rest of https://github.com/stellar/soroban-tools/pull/1107/files#r1456367830
- [x] document how / why you need to override the socket path for docker desktop
- [x] make sure i have the correct default image tags
- [x] address https://github.com/stellar/soroban-tools/pull/1107#issuecomment-1896637620
- [x] address https://github.com/stellar/soroban-tools/pull/1107#issuecomment-1897505257

### To handle in a followup PR
- allow for mounting volumes
- add a healthcheck, and don’t release the control back to the command line until container is healthy, see https://github.com/stellar/soroban-tools/pull/1107#discussion_r1456367830
- add the ability to have it be non-detached mode. and allowing it to stream the output of the container. Or just tell them how to attach and see the logs?
- put a link to some docs explaining more about connecting to docker, see https://github.com/stellar/soroban-tools/pull/1107#issuecomment-1897505257
- allow for users to pass in their non-standard docker socket path in an env var instead of just `--docker-socket-path`


cmd/soroban-cli/src/commands/network/stop.rs
```diff
@@ -0,0 +1,44 @@
...
+fn build_docker_command(container_name: &str) -> String {
+    format!("docker stop {container_name}")
+}
+
+fn run_docker_command(docker_command: &str) {
+    println!("Running docker command: `{docker_command}`");
+    let output = Command::new("sh")
+        .args(["-c", &docker_command])
+        .output()
+        .expect("Failed to execute command");
```
willemneal, line 35: Could add this case to the Error enum. Then make this function `Result<(), Error>`, then you can drop the `;` and `Ok(())` from the run function.

willemneal:
Looks great! Just one nit about using expect vs using an error. The error lets you capture the failed command.

You can use `.map_error(|_| Error::CommandFailed(docker_command.clone()))?`. This let's you use your own error and the inner anonymous function only executes if there was an error.

cmd/soroban-cli/src/commands/network/stop.rs
elizabethengelman: thanks for the suggestion, ill get a fix pushed up shortly!

cmd/soroban-cli/src/commands/network/stop.rs
```diff
@@ -18,27 +23,29 @@ impl Cmd {
...
-    }
+        .map_err(|err| Error::CommandError { error: err })
+        .and_then(|output| {
+            if output.status.success() {
+                let result = String::from_utf8_lossy(&output.stdout);
+                println!("Docker image stopped: {result}");
+                Ok(())
+            } else {
+                let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+                Err(Error::ContainerNotFoundErr { error: stderr })
```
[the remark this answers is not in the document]
willemneal: Makes sense to me!

cmd/soroban-cli/src/commands/network/stop.rs
```diff
@@ -1,7 +1,12 @@
 use std::process::Command;
 
 #[derive(thiserror::Error, Debug)]
-pub enum Error {}
+pub enum Error {
+    #[error("Failed to execute docker command: {error}")]
```
willemneal, line 5: This is totally fine way to do it, but my usual preference is to use {0} and then the enum case can be a tuple:
```rust
CommandError(std::id::Error)
```
then the map becomes:
```rust
.map_err(Error::CommandError)
```

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,151 @@
...
+
+/// This command allows for starting a stellar quickstart container. To run it, you can use the following command:
+/// `soroban network start <NETWORK> [OPTIONS] -- [DOCKER_RUN_ARGS]`
+///
+/// OPTIONS: refer to the options that are available to the quickstart image:
+/// --enable-soroban-rpc - is enabled by default
+/// --protocol-version (only for local network)
+/// --limits (only for local network)
+
+/// `DOCKER_RUN_ARGS`: These are arguments to be passed to the `docker run` command itself, and should be passed in after the slop `--`. Some common options are:
```
elizabethengelman, line 19: using the slop arg here to allow users to pass in options to `docker run`

i had considered using the slop arg for stellar/quickstart specific arguments instead, but it seemed to make more sense to use it for `docker run` options since the stellar/quickstart image has way fewer options than `docker run`

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,151 @@
...
+        protocol_version = protocol_version,
+        limits = limits,
+    );
+
+    docker_command
+}
+
+fn get_image_name(cmd: &Cmd) -> String {
+    // this can be overriden with the `-t` flag
+    let mut image_tag = match cmd.network.as_str() {
```
elizabethengelman, line 103: @chadoh and I discussed that it is important to make sure that at any given time the `soroban network start` command will be compatible with a version of the quickstart image, for each network.

one idea we had was to hardcode specific image tag digests, but I wonder if this may be error-prone since it would be easy to forget to update the sha for each release. 

also, as I was digging into how the quickstart images are built and pushed to docker hub, I realized that `latest`, `testing` and `soroban-dev` seem to be pushed to docker hub on every push to main, even if the underlying image is the same, iIwonder if this still affects the digest. 

Anyway, I could use some help with this!

elizabethengelman: Another thing I noticed as I was looking through the quickstart repo is that `latest`, `testing` and `soroban-dev` all seem to be using the same versions of xdr, core, horizon, etc. Does this mean that `[latest](https://github.com/stellar/quickstart/blob/master/.github/workflows/build-latest.yml)` is running protocol 20?

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,151 @@
...
+/// -p <`FROM_PORT`>:<`TO_PORT`> - this maps the port from the container to the host machine. By default, the port is 8000.
+/// -d - this runs the container in detached mode, so that it runs in the background
+
+// By default, without any optional arguments, the following docker command will run:
+// docker run --rm -p 8000:8000 --name stellar stellar/quickstart:testing --testnet --enable-soroban-rpc
+
+#[derive(Debug, clap::Parser, Clone)]
+pub struct Cmd {
+    /// Network to start, e.g. local, testnet, futurenet, pubnet
+    pub network: String,
```
chadoh, line 29: As written, this allows passing any string for the `network` argument. I wonder if we should make this an `enum` and hard-code an allowed list. Not sure what I would expect it to do, if I wrote something like `soroban network start lolfactory`. 

I also kinda like the idea of using this as the `--name` argument to `docker run`, so that the commands become parallel. `soroban network start local` / `soroban network stop local`. 

Not sure where to say this next bit, but it's sorta related to this, so I'll put it here for now: I also think we should start in detached mode by default. BUT, with the modification that we only "release" back to the prompt after the container is _healthy_ — once RPC and Friendbot are up and running. 

Right now, when I start a network in detached mode, I see this:

image

Not sure what that `0f50...` output is.

In my ideal version, we'd show a [loading animation](https://lib.rs/crates/spinners) and a message like "waiting for quickstart to fully boot...". It might be worth copying colima's approach here, where you can see the full log of what's happening scroll by in a limited ~7 line space, but greyed-out . Once RPC and Friendbot are ready, then we output a message like:

```
Stellar local quickstart container is up and running!
  Soroban RPC: http://localhost:8000/soroban/rpc
  Friendbot: http://localhost:8000/friendbot
  Horizon: ...
  etc: ...
  etc: ... [do we need more than RPC and Friendbot??]
To stop this container: `soroban network stop local`
```

And drop them back to their prompt.

chadoh (changes requested):
Solid start! Marking as "request changes" for the various doc comment stuff. The "start in detached by default with more complex health checks" thing (see comment below) would be _awesome_, but we could also do it in a follow-up to this one. It's already v useful to have these commands, so it might not be worth holding it up. lmk if you think we should do it at all, and if so, if we should do it now or wait for later.

chadoh:
I also think it would be really cool if we checked if Docker Daemon is running and, if not, rather than allowing Docker's error message to bubble through:

```
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

...we could instruct them on [how to install colima](https://github.com/abiosoft/colima/blob/main/docs/INSTALL.md). OR, we could automatically install it for them? (We could offer, and if they say `y`, then we could go ahead with it. But even then it seems error-prone, doesn't it?) If `colima` is already installed, then maybe we just automatically start a colima network called `stellar`? Not sure if we would also want to stop the colima network when we `soroba network stop` the last network—probably not?

I don't think we need to be all-in on colima, though. If there are other alternatives to Docker Desktop/Daemon like it, then we could check for those, too. And in any case, rather than bubbling the `cannot connect to docker daemon` error, we could throw a message that's more instructive & actionable. Maybe something along the lines of:

```
Running a local Stellar network requires a Docker-compatible container runtime. You can:
- install/open Docker Desktop: https://www.docker.com/products/docker-desktop/
- install/run colima: https://github.com/abiosoft/colima
```

Or maybe we sniff their OS and recommend colima only if they're on a Mac, podman if linux, and containerd if Windows, which are the suggestions [in this blog post](https://konfuzio.com/en/docker-desktop-alternative/). I don't know. But I want to make sure we make this feel doable to total Docker & Stellar newbies who try to start a local dev environment with `soroban network start`.

leighmcculloch:
Rather than hardcoding specific recommendations into the CLI, I think we should have a message similar to what's displayed above, but link to the soroban.stellar.org docs and place a page there with links out to docker/container runtimes folks can try, with recs based on mac, linux, windows. It means we can update the list as needed.

I'd also keep the original error. Never know what type of useful information might be found by passing through errors:

```
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
Running a local Stellar network requires a docker-compatible container runtime. See this url for a list of some common container runtimes to install: https://soroban.stellar.org/container-runtimes
```

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,151 @@
+use std::process::{Command, Stdio};
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {}
+
+const FROM_PORT: i32 = 8000;
+const TO_PORT: i32 = 8000;
+const CONTAINER_NAME: &str = "stellar";
+const DOCKER_IMAGE: &str = "stellar/quickstart";
```
leighmcculloch, line 9: Many alternatives to docker will require the fully qualified image name:
```suggestion
const DOCKER_IMAGE: &str = "docker.io/stellar/quickstart";
```

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,151 @@
...
+
+        run_docker_command(&docker_command);
+        Ok(())
+    }
+}
+
+fn run_docker_command(docker_command: &str) {
+    println!("Running docker command: `{docker_command}`");
+    let mut cmd = Command::new("sh")
+        .args(["-c", &docker_command])
```
leighmcculloch, line 66: Running docker commands will not only require a docker daemon, but a docker compatible CLI. While most container runtimes do provide docker compatible CLIs.

The daemon has a standardized API we should be able to call instead of shelling out.

I have less experience with this in the land of Rust, but in Go CLIs that use docker it's common to just import the docker libs and use them to talk to the daemon directly.

I would think there's a way we could do this in Rust too.

Possibly one of these projects would be similar. I haven't used any though:
- https://github.com/softprops/shiplift
- https://github.com/fussybeaver/bollard

cmd/soroban-cli/src/commands/network/start.rs
leighmcculloch: Latest has the software that supports protocol 20, but the public network (pubnet/mainnet) is running protocol 19. stellar-core knows how to run with each protocol version and when the validators vote on running protocol 20 the functionality for that protocol will become enabled.

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // option argument to specify the protocol version for the local network only
+    #[arg(short = 'p', long)]
+    pub protocol_version: Option<String>,
+
+    // option argument to specify the limits for the local network only
+    #[arg(short = 'l', long)]
+    pub limit: Option<String>,
+
+    // optional arguments to pass to the docker run command
```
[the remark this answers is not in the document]
elizabethengelman: nice catch!

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,151 @@
...
+/// --protocol-version (only for local network)
+/// --limits (only for local network)
+
+/// `DOCKER_RUN_ARGS`: These are arguments to be passed to the `docker run` command itself, and should be passed in after the slop `--`. Some common options are:
+/// -p <`FROM_PORT`>:<`TO_PORT`> - this maps the port from the container to the host machine. By default, the port is 8000.
+/// -d - this runs the container in detached mode, so that it runs in the background
+
+// By default, without any optional arguments, the following docker command will run:
+// docker run --rm -p 8000:8000 --name stellar stellar/quickstart:testing --testnet --enable-soroban-rpc
+
```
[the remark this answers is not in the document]
elizabethengelman: Great catch! I'll move an edited version of these comments to the mod.rs file.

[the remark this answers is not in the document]
elizabethengelman: updated in [57790a4](https://github.com/stellar/soroban-tools/pull/1107/commits/57790a4ef465c418ba5e1ade4745beb73f815a3e)

cmd/soroban-cli/src/commands/network/start.rs
elizabethengelman: Great call! I was able to get this working with bollard in commit [51fbdce](https://github.com/stellar/soroban-tools/pull/1107/commits/51fbdcee45a9d43af11d7bcd8c6fa90884aae9c3) for start, and [198c070](https://github.com/stellar/soroban-tools/pull/1107/commits/198c07040921307723972d78176fa3aac1bc93e7) for stop.

I use docker desktop, and learned that it puts the docker socket file in a different location than the expected default. I was able to find a couple of ways around this that I'll make sure to document somewhere before merging this in. But the TLDR; is that there is now a `--docker-socket-path` flag the user can pass in the socket location on their machine.

cmd/soroban-cli/src/commands/network/start.rs
elizabethengelman: * > As written, this allows passing any string for the network argument. I wonder if we should make this an enum and hard-code an allowed list. Not sure what I would expect it to do, if I wrote something like soroban network start olfactory.

    Good call - this should be an enum! Updated in 8457ecbd1000098a71673816b931f604e2754113

* > I also kinda like the idea of using this as the --name argument to docker run, so that the commands become parallel. soroban network start local / soroban network stop local.

    Ah, interesting idea! I like `soroban network start local` and `soroban network stop local` being the same! 
  
    If we are prioritizing the commands being the same, I'll remove the ability for a user to pass in their own container name for now (`--container-name`), since it complicates having `soroban network start local` and `soroban network stop local` be in sync if the user is making up their own docker container name.

Update in [6687e37](https://github.com/stellar/soroban-tools/pull/1107/commits/6687e378570c9005465ab6b1b1d4192131606280)

cmd/soroban-cli/src/commands/network/start.rs
elizabethengelman: Thanks! That's very helpful 👍

cmd/soroban-cli/src/commands/network/start.rs
elizabethengelman: I didn't get a chance to respond to the other part your comment last week, but here it is!
> I also think we should start in detached mode by default. BUT, with the modification that we only "release" back to the prompt after the container is healthy — once RPC and Friendbot are up and running....

This is a great call! Once switching to using bollard to handle the docker interaction, it seems to default to run in detached mode. So I am thinking that I'll keep with that, and then in a future PR we can add the ability to have it be non-detached mode, and stream the output of the container.. if we decide that's important.

I love the idea of running a healthcheck, and having some more helpful messaging, instead of just spitting out the container id, which it was it used to do. I've made a small improvement, but I’m leaning toward pushing the health check piece to a follow-up PR so we can get this one in.

It now looks like this when the container is started successfully:
```
ℹ️  Starting testnet network
✅ Container started: stellar-testnet
ℹ️  To stop this container run: soroban network stop testnet --docker-socket-path /Users/elizabethengelman/.docker/run/docker.sock
```

elizabethengelman:
@chadoh @leighmcculloch Regarding the error case when the docker daemon can’t be found - I realized that when using bollard (and the underlying Docker API) it doesn’t seem to bubble up the original error ( `docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`) but instead returns a kind of vague Hyper error:  `error trying to connect: No such file or directory (os error 2)`

So instead, I’m priting out a message similar to `Cannot connect…`, and that is where we can also include a link to the docs when we have a page for that message, which I can work on in a followup. 

```bash

ℹ️  Starting testnet network
⛔️ Failed to connect to the Docker daemon at "/var/run/docker.sock". Is the docker daemon running?
Running a local Stellar network requires a Docker-compatible container runtime.
Please note that if you are using Docker Desktop, you may need to utilize the `--docker-socket-path` flag to pass in the location of the docker socket on your machine.

error: ⛔ ️Failed to start container: error trying to connect: No such file or directory (os error 2)
```

chadoh (approved):
Heroic.

A couple small nits/suggestions, but I'm ok with it being merged whether you decide to incorporate these changes or not.

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,196 @@
...
+    #[error("⛔ ️Failed to start container: {0}")]
+    StartContainerError(#[from] bollard::errors::Error),
+}
+
+#[derive(Debug, clap::Parser, Clone)]
+pub struct Cmd {
+    /// Network to start
+    pub network: Network,
+
+    #[arg(short = 'd', long, help = DOCKER_SOCKET_PATH_HELP)]
```
willemneal, line 26: This is the official ENV for docker tooling to check. It lets it be flexible to both sockets and http endpoints.
```suggestion
    #[arg(short = 'd', long, help = DOCKER_SOCKET_PATH_HELP, env = "DOCKER_HOST")]
```

cmd/soroban-cli/src/commands/network/stop.rs
```diff
@@ -0,0 +1,27 @@
...
+    #[error("Failed to stop container: {0}")]
+    StopContainerError(#[from] bollard::errors::Error),
+}
+
+#[derive(Debug, clap::Parser, Clone)]
+pub struct Cmd {
+    /// Network to stop
+    pub network: Network,
+
+    #[arg(short = 'd', long, help = DOCKER_SOCKET_PATH_HELP)]
```
willemneal, line 14:
```suggestion
    #[arg(short = 'd', long, help = DOCKER_SOCKET_PATH_HELP, env = "DOCKER_HOST")]]
```

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,196 @@
...
+    .cloned()
+    .collect()
+}
+
+fn get_image_name(cmd: &Cmd) -> String {
+    // this can be overriden with the `-t` flag
+    let mut image_tag = match cmd.network {
+        Network::Testnet => "testing",
+        Network::Futurenet => "future",
+        _ => "latest", // default to latest for local and pubnet
```
leighmcculloch, line 147: 👍🏻

docs/soroban-cli-full-docs.md
```diff
@@ -65,6 +67,8 @@ This document contains the help content for the `soroban` command-line program.
 * [`soroban network add`↴](#soroban-network-add)
 * [`soroban network rm`↴](#soroban-network-rm)
 * [`soroban network ls`↴](#soroban-network-ls)
+* [`soroban network start`↴](#soroban-network-start)
+* [`soroban network stop`↴](#soroban-network-stop)
```
leighmcculloch, line 71: Something that might be confusing with bundling this under `network` is that folks configure a network, and then we also use network to mean this thing which isn't the same thing. In fact we can configure networks called the same name that aren't the same.

Looking at some other tools in other ecosystems, ganache calls these things "instances" instead of "network". Maybe it would make sense to use a similar term, a term that doesn't imply this thing is a network, but a local instance.

Or maybe there's another option. Or maybe not confusing for others...

leighmcculloch (commented):
This is looking great. Couple thoughts inline.

leighmcculloch (commented):
This is looking great. Couple thoughts inline.

cmd/soroban-cli/src/commands/network/start.rs
elizabethengelman: nice!!

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,196 @@
...
+    StartContainerError(#[from] bollard::errors::Error),
+}
+
+#[derive(Debug, clap::Parser, Clone)]
+pub struct Cmd {
+    /// Network to start
+    pub network: Network,
+
+    #[arg(short = 'd', long, help = DOCKER_SOCKET_PATH_HELP)]
+    pub docker_socket_path: Option<String>,
```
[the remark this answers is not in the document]
elizabethengelman: I just changed this to `docker_host`, but am now noticing a small issue/nuance with how bollard is understanding the `DOCKER_HOST` env var in its current release. It seems to only consider `DOCKER_HOST` for HTTP, https and tcp connections, not socket connections. 

That's why I think I made this `docker_socket_path` originally, so that it was clear this was only to be used for socket connections. I think this may be confusing for users, since they may assume that they can use the `DOCKER_HOST` var like you can for the CLI, which allows it to be for socket, http, tcp, etc connections.

Bollard made some [updates](https://github.com/fussybeaver/bollard/commit/ba8a0ec415a878405fd3c90bad6dd2821190709c) in `master` that haven't been released yet and included a func called `connect_with_defaults`, but this still ignores the case that I was trying to account for... if the user has a non-standard socket path that they want to pass in via `DOCKER_HOST`. 🤔 

I wonder if we can use a similar thing as they did in that update, and switch on the value of `DOCKER_HOST`. And the default can be to try to connect via sockets. I hope this makes sense - it was a bit tricky to unwind, but I'll push a commit up that should hopefully solve this issue.

cmd/soroban-cli/src/commands/network/start.rs
```diff
@@ -0,0 +1,196 @@
...
+    #[arg(short = 'l', long)]
+    pub limit: Option<String>,
+
+    /// Argument to specify the HOST_PORT:CONTAINER_PORT mapping
+    #[arg(short = 'p', long, num_args = 1.., default_value = DEFAULT_PORT_MAPPING)]
+    pub ports_mapping: Vec<String>,
+
+    /// Optional argument to turn off soroban rpc
+    #[arg(short = 'r', long)]
+    pub disable_soroban_rpc: bool,
```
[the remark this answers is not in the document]
elizabethengelman: updated in [7aec389](https://github.com/stellar/soroban-cli/pull/1107/commits/7aec3895d3513366e26f0cccd0cb15f40d13f8c9)

cmd/soroban-cli/src/commands/network/shared.rs
```diff
@@ -0,0 +1,76 @@
...
+    docker_socket_path: &Option<String>,
+) -> Result<Docker, bollard::errors::Error> {
+    let docker = if docker_socket_path.is_some() {
+        let socket = docker_socket_path.as_ref().unwrap();
+        let connection = Docker::connect_with_socket(socket, DEFAULT_TIMEOUT, API_DEFAULT_VERSION)?;
+        check_docker_connection(&connection).await?;
+        connection
+    } else {
+        let connection = Docker::connect_with_socket_defaults()?;
+        check_docker_connection(&connection).await?;
```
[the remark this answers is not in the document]
elizabethengelman: Love this idea! For now I am going to try to connect to `$HOME/.docker/run/docker.sock"` if connecting to DOCKER_HOST fails. But I like the idea of trying in parallel, and will add that to my followup task list!

Also, this is what it looks like. I thought it made sense to alert the user that we failed to connect to DOCKER_HOST and were trying $HOME/.docker/run/docker.sock instead. But lmk if you all think including that is confusing!
```
ℹ️  Starting local network
⛔️ Failed to connect to the Docker daemon at "/var/run/docker.sock". Is the docker daemon running?
ℹ️  Running a local Stellar network requires a Docker-compatible container runtime.
ℹ️  Please note that if you are using Docker Desktop, you may need to utilize the `--docker-host` flag to pass in the location of the docker socket on your machine.

Failed to connect to DOCKER_HOST: unix:///var/run/docker.sock.
Trying to connect to the default Docker Desktop socket at /Users/elizabethengelman/.docker/run/docker.sock.
✅ Container started: stellar-local
ℹ️  To stop this container run: soroban network stop local
```

cmd/soroban-cli/src/commands/network/start.rs
elizabethengelman: Marking this comment thread as resolved, and have added some follow up tasks in https://github.com/orgs/stellar/projects/50?pane=issue&itemId=54158722

docs/soroban-cli-full-docs.md
elizabethengelman: @leighmcculloch Hm, interesting point. Just to make sure I'm understanding correctly - you're talking about the distinction between:
- this PR, which is starting up nodes instances for different networks vs
- the other things that the `network` subcommand is doing is configuring connections to different networks

is that right?

I wonder if it would be worth putting this functionality under `soroban network instance`, so that is is more clear that this is an instance of node that is connecting to a specific network? So it would be `soroban network instance start testnet`. 

What do you think @chadoh ?

cmd/soroban-cli/src/commands/network/shared.rs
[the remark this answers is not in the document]
elizabethengelman: Marking this as resolved, and have the idea of checking the connections in parallel in the follow up task: https://github.com/orgs/stellar/projects/50?pane=issue&itemId=54158722

docs/soroban-cli-full-docs.md
janewang: @leighmcculloch Could you suggest what the subcommand could be? Thanks

leighmcculloch: I was thinking `soroban instance list|start|stop`.

But `soroban network instance ...` could work too, especially if launching an instance auto configures a network for that instance.

elizabethengelman: Based on the conversation in the meeting last week, it seems like folks were good with moving forward with `soroban network start | stop`. I'll plan to go ahead with that, and can always make an update in the future.

elizabethengelman: Added a link to this comment conversation in the follow up issue: https://github.com/orgs/stellar/projects/50?pane=issue&itemId=54158722

PR #1108, opened 2023-12-04
soroban-cli: Only Give RC SDK Contract Compilation Warning/Errors if Using Pubnet

### LINKED ISSUE #1077: soroban-cli: prevents deployment of contracts built with an rc sdk to any network and not just public network

```console
$ soroban version
soroban 20.0.0-rc.4.2 (v20.0.0-rc.4.1-4-g4e9a48756b96e4e1a070105bad202f39901139ad-dirty)
soroban-env 20.0.0-rc2 (8c63bff68a15d79aca3a705ee6916a68db57b7e6)
soroban-env interface version 85899345977
stellar-xdr 20.0.0-rc1 (d5ce0c9e7aa83461773a6e81662067f35d39e4c1)
xdr curr (9ac02641139e6717924fdad716f6e958d0168491)
```

### What did you do?

```console
$ docker run --rm -i -p 8000:8000 stellar/quickstart:testing --testnet --enable-soroban-rpc
$ export SOROBAN_RPC_URL=http://localhost:8000/soroban/rpc
$ export SOROBAN_NETWORK_PASSPHRASE='Test SDF Network ; September 2015'
$ soroban config identity fund
$ soroban contract install --wasm out/contract.optimized.wasm
```

I expected the contract to deploy, because I'm submitting the contract to a test network, not a public network.

### What did you see instead?

```
error: the deployed smart contract out/contract.optimized.wasm was built with Soroban Rust SDK v20.0.0-rc2#0992413f9b05e5bfb1f872bce99e89d9129b2e61, a release candidate version not intended for use with the Stellar Public Network. To deploy anyway, use --ignore-checks
```

### Discussion

The error message is confusing because it explicitly says it won't let me deploy to the Stellar Public Network, which is a specific network. However, it appears to display message for any network I'm connected to, including a local or test network.

Looking at the code it looks like it displays this error message for any network:
https://github.com/stellar/soroban-tools/blob/c68d03cd8f0588263e9a47b93ed8b2dc28d32f2f/cmd/soroban-cli/src/commands/contract/install.rs#L77-L85

Which was introduced in:
- https://github.com/stellar/soroban-tools/pull/1061

Either the error message shouldn't specify the Stellar Public Network, or the error message should only be displayed for when the Network Passphrase is configured to be `Public Global Stellar Network ; September 2015`. The latter would be my preference because blocking folks from deploying release candidates to a local or test network doesn't seem like it would be of significant benefit.

stellarsaur:
### What

Allow users to deploy contracts to Testnet or Futurenet without a warning/error if they compiled their contract with a RC version of the Rust SDK.

### Why

Release Candidate SDK versions are only not intended to be used with Pubnet, so we shouldn't prevent users from deploying contracts compiled with RC SDK versions to Testnet or Futurenet

Closes #1077


willemneal (approved):
Makes more sense! Thanks!

willemneal:
As for the unit test, it seems that the debug option isn't working. I suggest ignoring the test for now with a high priority issue for it's resolution.

Issue #1110, opened 2023-12-06
Fix check-dependencies script


tamirms:
There is a bug in the check dependencies [script](https://github.com/stellar/soroban-tools/blob/main/scripts/check-dependencies.bash#L35) which is causing the "Dependency sanity checker" CI failure. The script relies on the fact that the git and rev properties would be supplied for the stellar-xdr dependency. However, those properties were commented out of the stellar-xdr dependency in the rs-soroban-env repo https://github.com/stellar/rs-soroban-env/blob/main/Cargo.toml#L31-L32 .

mollykarcher:
Link to [slack discussion](https://stellarfoundation.slack.com/archives/C030Z9EHVQE/p1701793185697229)

We should likely be able to scrap the script (or at least part of it) in favor of cargo-deny, which other rust repos currently use. However, that script is also checking go dependencies (not just rust) to make sure xdr versions match up, so we may need some hybrid or an alternative if that's still needed.

PR #1116, opened 2023-12-07
build: make dependency-check script more verbose

chadoh:
### What

Add more printf and echo statements to the dependency-check script so we can figure out where it fails.

### Why

Right now if something goes wrong it's really hard to tell what.

### Known limitations

This branch should probably also fix whatever is causing the dependency-check script to fail on CI.

### DIFF
--- a/scripts/check-dependencies.bash
+++ b/scripts/check-dependencies.bash
@@ -2,13 +2,18 @@
 
 set -e
 
+echo "Checking dependencies."
+
 SED=sed
 if [ -z "$(sed --version 2>&1 | grep GNU)" ]; then
-    SED=gsed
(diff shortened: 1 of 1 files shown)


stellarsaur (approved):

sreuland:
@chadoh , this seems closely aligned to https://github.com/stellar/soroban-tools/issues/1110, do you want to take that issue and add `Closes #1110` here?

chadoh:
@leighmcculloch mentioned in Slack that we need to remove this custom script and use cargo-deny as in other repos; I think he was working on that?

leighmcculloch:
I'm not working on it 😄.

PR #1113, opened 2023-12-06
Update E2E tests to references pending stable JS releases

Shaptic:
### What
This updates system-test references to the latest SDK versions.

### Why
To ensure that end-to-end tests work.
Closes #1048 

### Known limitations
There's probably a bunch of other versions that need to be bumped here.

PR #1117, opened 2023-12-07
Don't Convert Start Ledger to String for Events Command

stellarsaur:
### What

`soroban events` takes a `uint32` for the `start-ledger` param. We should not convert it to a string before calling the RPC endpoint.

### DIFF
--- a/cmd/soroban-cli/src/rpc/mod.rs
+++ b/cmd/soroban-cli/src/rpc/mod.rs
@@ -356,7 +356,7 @@ pub struct Event {
     #[serde(rename = "type")]
     pub event_type: String,
 
-    pub ledger: String,
+    pub ledger: u32,
     #[serde(rename = "ledgerClosedAt")]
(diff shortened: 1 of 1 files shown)


sreuland:
should the target be a v20.0.1 branch off the v20.0.0 tag to enable hotfix, then can port to main afterwards?

stellarsaur:
> should the target be a v20.0.1 branch off the v20.0.0 tag to enable hotfix, then can port to main afterwards?

Yeah that's a better idea; done

sreuland (approved):
may be good to create a ticket for tracking later on local/unit test coverage that could assert some of the rpc schema expectations.

sreuland:
@stellarsaur , is ok to proceed with merging release/v20.0.1 back to main?, then we can [renable e2e on main and pass](https://github.com/stellar/soroban-tools/pull/1113), thanks!

PR #1120, opened 2023-12-11
Backport v20.0.1

sreuland:
### What

back port the v20.0.1 to main

### Why

the gh bot pr to do the same backport has fallen behind main and would have a conflict - https://github.com/stellar/soroban-tools/pull/1119 , doing it here as a rebase on top of latest main instead.


mollykarcher (approved):
I don't see the conflict you're referencing with https://github.com/stellar/soroban-tools/pull/1119, but also don't see an issue with merging this one instead of that one

sreuland:
ok, I'll merge, I saw conflict on cargo.lock when I rebased main locally and assumed that would have popped up when merging main into release/v20.0.1, but didn't verify that

PR #1121, opened 2023-12-11
bumped the preflight fees padding on instrs to 3M

sreuland:
### What

increased the padding on preflight instr fees to 3M

### Why

to enable more advanced contracts that demand more compute to still get through, rather than get a fees exception based on newer v20 cost model.


mollykarcher:
We need to create the release branch via this [workflow](https://github.com/stellar/actions/blob/main/README-rust-release.md#step-1-run-the-bump-version-workflow)

sreuland:
going to redo this.

PR #1122, opened 2023-12-11
fix(bindings-ts): latest RPC & stellar-sdk

chadoh:
### What

- update to `stellar-sdk@11.0.1`
- fetch expiration using new `getFootprint` logic
- improve `initialize.sh` to show network status
  - latest RPC returns better error messages
  - deal with these and parse `getHealth` response

### Why

current `bindings typescript` logic is broken with latest Futurenet &
`quickstart:soroban-dev`

### Known limitations

Please double-check my `entryRes.entries[0].liveUntilLedgerSeq` logic in
the `getStorageExpiration` function.

### DIFF
--- a/cmd/crates/soroban-spec-typescript/src/project_template/package.json
+++ b/cmd/crates/soroban-spec-typescript/src/project_template/package.json
@@ -4,7 +4,7 @@
     "dependencies": {
         "@stellar/freighter-api": "1.7.1",
         "buffer": "6.0.3",
-        "stellar-sdk": "11.0.0-beta.6"
+        "stellar-sdk": "11.0.1"
     },
     "scripts": {
         "build": "node ./scripts/build.mjs"
--- a/cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
+++ b/cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
@@ -297,16 +297,15 @@ export class AssembledTransaction<T> {
   }
 
   getStorageExpiration = async () => {
-    const key = new Contract(this.options.contractId).getFootprint()[1]
-
-    const expirationKey = xdr.LedgerKey.expiration(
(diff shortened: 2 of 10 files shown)


cmd/crates/soroban-spec-typescript/src/project_template/package.json
```diff
@@ -4,7 +4,7 @@
     "dependencies": {
         "@stellar/freighter-api": "1.7.1",
         "buffer": "6.0.3",
-        "stellar-sdk": "11.0.0-beta.6"
+        "stellar-sdk": "11.0.0"
```
Shaptic, line 7: 11.0.1 has a bugfix related to getEvents! should be a trivial upgrade

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
```diff
@@ -297,16 +297,15 @@ export class AssembledTransaction<T> {
   }
 
   getStorageExpiration = async () => {
-    const key = new Contract(this.options.contractId).getFootprint()[1]
-
-    const expirationKey = xdr.LedgerKey.expiration(
-      new xdr.LedgerKeyExpiration({ keyHash: hash(key.toXDR()) }),
+    const entryRes = await this.server.getLedgerEntries(
+      new Contract(this.options.contractId).getFootprint()
```
Shaptic, line 301: keep in mind that this doesn't include the wasm code expiration

Shaptic (approved):
small fixup needed but otherwise :shipit:

cmd/crates/soroban-spec-typescript/src/project_template/src/assembled-tx.ts
chadoh: We previously used the persistent ttl specifically. How would I get that now? How would I get instance ttl?

Shaptic: this will give you the instance, just not the wasm

PR #1131, opened 2023-12-14
soroban-rpc: Allow Budget Instruction Leeway to be Configured by Clients

### LINKED ISSUE #1128: soroban-rpc: make preflight fudge factors configurable by the client

Preflight estimations have been off, requiring us to repeatedly issue releases to up the fudge factor used in these calculations. When users are unable to create contracts due to these issues, there aren't any good workarounds for them.

The preflight request will accept new parameters that translate into the [fudge factor used by preflight](https://github.com/stellar/soroban-tools/blob/397cc90590d1d74f493e5707fcb6b50ddc84fe6d/cmd/soroban-rpc/lib/preflight/src/fees.rs#L141). TBD whether we should support/add both the percentage and flat addition factors, or if we should just support one or the other. End users could then tweak this up or down depending on their specific needs. These parameters should be optional and the default should be the current hardcoded values. Optionally, we could make these defaults configurable RPC/server side as well. 

Alternatively discussed making the fudge factor(s) configurable RPC-side. This would still allow us and other RPC providers to move the fudge factor without having to release/deploy. However, given a single RPC instance will service many different profiles of dapps/users, and it's impossible to provide a one-size-fits-all default that will work for all contracts, it makes more sense for it to be client provided.

stellarsaur:
### What

This PR allows budget instruction leeway used in preflight calculations to be configured by passing in a new `ResourceConfig` struct to a `SimulateTransactionRequest`. Right now we're only allowing the flat addition factor to be configured, but we could add support for both by just adding another field to `ResourceConfig`.

If not provided the leeway defaults to what we were previously using (`3000000` instructions)

NOTE: This PR does not include updating the CLI to support configuring this value through contract command line args.

### Why

Preflight estimations have been off, requiring us to repeatedly issue releases to increase the leeway used in these calculations. When users are unable to create contracts due to these issues, there aren't any good workarounds for them.

Closes #1128


cmd/soroban-rpc/internal/methods/simulate_transaction.go
```diff
@@ -46,7 +47,7 @@ type SimulateTransactionResponse struct {
 }
 
 type PreflightGetter interface {
-	GetPreflight(ctx context.Context, readTx db.LedgerEntryReadTx, bucketListSize uint64, sourceAccount xdr.AccountId, opBody xdr.OperationBody, footprint xdr.LedgerFootprint) (preflight.Preflight, error)
+	GetPreflight(ctx context.Context, readTx db.LedgerEntryReadTx, bucketListSize uint64, sourceAccount xdr.AccountId, opBody xdr.OperationBody, footprint xdr.LedgerFootprint, resourceConfig *preflight.ResourceConfig) (preflight.Preflight, error)
```
2opremio, line 50: It's probably time to pass a structure instead of so many parameters.

2opremio: (the context should still be a parameter, though)

cmd/soroban-rpc/internal/preflight/preflight.go
```diff
@@ -79,6 +87,7 @@ type PreflightParameters struct {
 	NetworkPassphrase string
 	LedgerEntryReadTx db.LedgerEntryReadTx
 	BucketListSize    uint64
+	ResourceConfig    *ResourceConfig
```
2opremio, line 90: I don't think this should be a pointer. If it was ommitted in the request, we can initialize it to the default value.

In fact ... we should probably always use a base default value and add (or substract) the passed value on top of it

Shaptic: I think it should be optional in the request, but internally it can just be set to a default like you said. I don't think it should be additive, though: I'd rather we document the default and let users override it.

cmd/soroban-rpc/internal/methods/simulate_transaction.go
stellarsaur: Agreed; done

cmd/soroban-rpc/internal/preflight/preflight.go
stellarsaur: Initialized this to a default value instead of being a pointer here. Since it's a pointer in `SimulateTransactionRequest` it will be optional (and set to default if `nil`)

stellarsaur:
Pending [updating the SDK to send params as an object](https://github.com/stellar/js-stellar-sdk/issues/895) so tests will pass

Shaptic (approved):

Shaptic (approved):

PR #1152, opened 2024-01-04
rpc: Add diagnostic events to sendTransaction response

2opremio:
### What

Add new `diagnosticEventsXdr` field to `sendTransaction`'s RPC method response.

Closes https://github.com/stellar/soroban-tools/issues/1157

Similar to https://github.com/stellar/go/pull/5148

Also, use the new Core `ENABLE_DIAGNOSTICS_FOR_TX_SUBMISSION` flag in integration tests.

### Why

Improve debug-ability of Soroban transaction submission.

### Known limitations

N/A


cmd/soroban-rpc/internal/test/stellar-core-integration-tests.cfg
```diff
@@ -1,4 +1,5 @@
 ARTIFICIALLY_ACCELERATE_TIME_FOR_TESTING=true
+ENABLE_DIAGNOSTICS_FOR_TX_SUBMISSION=true
```
sreuland, line 2: this core cfg needs to be added to rpc's captive core instance for any deployment correct? I assume is the case, we'll need to update the rpc deployment manifests on kube to include this and i spun up [quickstart/#543](https://github.com/stellar/quickstart/issues/543)

dmkozh: > only if the more general `ENABLE_SOROBAN_DIAGNOSTIC_EVENTS` isn’t already enabled

These flags are independent and control different things. Here is the brief description:

`ENABLE_SOROBAN_DIAGNOSTIC_EVENTS`:
 - Adds the *apply-time* diagnostic events to the tx meta stream
 - Meaningless if tx meta is not enabled and is not allowed at all on validator nodes
 - Should be enabled for the captive core nodes that are being used for tx meta ingestion

`ENABLE_DIAGNOSTICS_FOR_TX_SUBMISSION`
 - Adds the diagnostics directly to the Core's `tx` endpoint JSON response. 
 - The diagnostics are only populated for the transactions that get immediately rejected by Core (i.e. they don't have a chance to be applied)
 - Can be enabled on any node, including validators
 - Should be enabled on the nodes which accept transactions (from Horizon/soroban-rpc). Typically these are not captive core nodes, but in some setups I can imagine both flags being enabled (notably, for standalone network mode both flags are useful).

sreuland: @dmkozh , awesome, thanks for further clarification, yes, in some deployment configs, the captive core instance is the node used for tx sub. btw, is there a separate core cfg variables doc that has this same content for ref?

cmd/soroban-rpc/internal/methods/send_transaction.go
```diff
@@ -21,6 +21,9 @@ type SendTransactionResponse struct {
 	// ErrorResultXDR is a TransactionResult xdr string which contains details on why
 	// the transaction could not be accepted by stellar-core.
 	ErrorResultXDR string `json:"errorResultXdr,omitempty"`
+	// DiagnosticEventsXDR is present only if Status is equal to proto.TXStatusError.
+	// DiagnosticEventsXDR is a base64-encoded slice of xdr.DiagnosticEvent
+	DiagnosticEventsXDR []string `json:"diagnosticEventsXdr,omitempty"`
```
sreuland, line 26: is there a separate process initiated as part of rpc api resource changes, i.e. the associated soroban-docs updates and sdk blast/mention on this new field so they can handle? such as for js-stellar-sdk [SendTransactionResponse](https://github.com/stellar/js-stellar-sdk/blob/master/src/soroban/api.ts#L187-L207)

cmd/soroban-rpc/internal/test/transaction_test.go
```diff
@@ -151,6 +151,58 @@ func TestSendTransactionBadSequence(t *testing.T) {
...
+
+	assert.Equal(t, proto.TXStatusError, result.Status)
+	var errorResult xdr.TransactionResult
+	assert.NoError(t, xdr.SafeUnmarshalBase64(result.ErrorResultXDR, &errorResult))
+	assert.Equal(t, xdr.TransactionResultCodeTxSorobanInvalid, errorResult.Result.Code)
+
+	assert.Greater(t, len(result.DiagnosticEventsXDR), 0)
+	var event xdr.DiagnosticEvent
+	err = xdr.SafeUnmarshalBase64(result.DiagnosticEventsXDR[0], &event)
+	assert.NoError(t, err)
```
sreuland, line 202: is there any detail of the diagnostic event payload that would be worthwhile to assert on related to insufficient fee error?

sreuland (approved):
lgtm, left comment on thoughts for propagating the new api field to upstream(docs, sdks).

cmd/soroban-rpc/internal/methods/send_transaction.go
2opremio: I will create a ticket in this repo and the docs repo once this is merged

cmd/soroban-rpc/internal/test/transaction_test.go
2opremio: I don't want to rely on that in case it changes

cmd/soroban-rpc/internal/test/stellar-core-integration-tests.cfg
dmkozh: It doesn't seem like we have a doc like that, but we have inline comments for most of the configuration flags in the example config (https://github.com/stellar/stellar-core/blob/f9999b8d31aef2b3e96226685b55a8506ba3c721/docs/stellar-core_example.cfg#L610). I'll open an issue to update it with the new soroban-related flags.

Issue #656, opened 2023-05-24
soroban-cli: add 'new' / 'init' command that creates a new project with the boilerplate


leighmcculloch:
Developers would benefit from having a command they can run in soroban-cli that creates a new project with the dependencies and boilerplate already filled out, ready to go.

As evidenced by [this discussion](https://discord.com/channels/897514728459468821/1108412441932611684/1108412441932611684), it is easy to miss the simple things like including `#![no_std]` in the contract.

If developers have a command to run that creates a basic contract to start with, it'll help them get started without getting bogged down in the boilerplate basics.

The command could follow the behaviors of the `cargo new` and `cargo init` commands, although for the sake of reducing the surface area of our CLI, it would be nice if we could go with one or the other.

elizabethengelman:
* First PR: https://github.com/stellar/soroban-tools/pull/1156
* I'll plan to add the following in a follow up PR:
    * `--frontend-template`
    * generate the possible example contracts to be included at runtime (https://github.com/stellar/soroban-tools/pull/1156#discussion_r1443144404https://github.com/stellar/soroban-tools/pull/1156#discussion_r1443144404)

PR #1156, opened 2024-01-04
feat: soroban init command

elizabethengelman:
### What

This PR adds a `soroban init` command to the CLI to allow for initializing a new soroban rust project. 

Addresses https://github.com/stellar/soroban-tools/issues/656

### Why

The intention of this change is to make it easier to get up and running with a soroban project, and to allow for the Getting Started guide to be more concise by being able to remove some of the boilerplate Rust setup. 

### Known limitations

- Currently, this command creates a plain Rust project, there are plans to have a follow-up PR that will allow a user to pass in a specific frontend (`--frontend-template`) to use so that the project can be set up for frontend development as well. The first FE will be Astro, since we already have an example Astro project working.
- I am seeing a pre-push check failing locally for this branch as well as on main. 
```
error[E0046]: not all trait items implemented, missing: `escalate_error_to_panic`
    --> /Users/elizabethengelman/.cargo/git/checkouts/rs-soroban-sdk-0ab58a716c671239/93b09e4/soroban-sdk/src/env.rs:1449:1
     |
1449 | impl internal::EnvBase for Env {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `escalate_error_to_panic` in implementation
     |
     = help: implement the missing item: `fn escalate_error_to_panic(&self, _: <Self as soroban_env_host::EnvBase>::Error) -> ! { todo!() }`
```


cmd/soroban-cli/Cargo.toml
```diff
@@ -93,6 +93,13 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 cargo_metadata = "0.15.4"
 pathdiff = "0.2.1"
 dotenvy = "0.15.7"
+gix = { version = "0.55.2", default-features = false, features = [
+  "blocking-http-transport-reqwest-rust-tls",
+  "worktree-mutation",
+] }
+anyhow = "1.0.75"
```
willemneal, line 100:
```suggestion
```
You ended up using thiserror right?

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+    #[arg(short, long, num_args = 1..=20, default_value = "none")]
+    pub with_contract: Vec<ExampleContract>,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Io error: {0}")]
+    CreateDirError(#[from] io::Error),
+
+    // the gix::clone::Error is too large to include in the error enum as is, so we wrap it in a Box
```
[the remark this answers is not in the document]
willemneal: Box makes sense to me!

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+use std::{fs, io};
+
+use clap::Parser;
+use std::num::NonZeroU32;
+use std::sync::atomic::AtomicBool;
+use toml_edit::{Document, Formatted, InlineTable, TomlError, Value};
+
+#[derive(Clone, Debug, PartialEq, clap::ValueEnum)]
+pub enum ExampleContract {
+    Account,
```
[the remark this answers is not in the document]
willemneal: Yeah I think it's a good start to hard code. It is possible to generate the help at runtime using a builder for args after a `--` like we do for the invoke custom CLI. I think perhaps the best next step would be to generate this at build time.

cmd/soroban-cli/Cargo.toml
elizabethengelman: yep, nice catch!

cmd/soroban-cli/src/commands/init.rs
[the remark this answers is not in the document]
elizabethengelman: Ah, cool! I'll plan to get this into a follow up PR.

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+    FetchError(#[from] Box<gix::clone::fetch::Error>),
+
+    #[error("Failed to checkout the template repository: {0}")]
+    CheckoutError(#[from] gix::clone::checkout::main_worktree::Error),
+
+    #[error("Failed to parse Cargo.toml: {0}")]
+    TomlParseError(#[from] TomlError),
+}
+
+const TEMPLATE_URL: &str = "https://github.com/AhaLabs/soroban-init-template.git";
```
stellarsaur, line 95: This is fine for now, but may be worth having this be a command line flag in the future to be able to use different templates

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    pub project_path: String,
+
+    /// optional flag to specify soroban example contracts to include
+    #[arg(short, long, num_args = 1..=20, default_value = "none")]
```
stellarsaur, line 71: Is it worth including a set of common contracts by default (e.g. `hello-world`) rather than have the default include no example contracts?

stellarsaur (approved):
Nice addition to the CLI!

cmd/soroban-cli/src/commands/init.rs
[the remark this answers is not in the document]
leighmcculloch: Some folks developing will be developing locally. Leaning towards local only solutions where possible makes sense and makes the cli [more accessible to a wider variety of people](https://stellar.org/foundation). If we do have it pull from a remote source would be nice if it gracefully fell back to an internal list.

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+    // clone the template repo into the temp dir
+    clone_repo(template_url, template_dir.path())?;
+
+    // create the project directory and copy the template contents into it
+    std::fs::create_dir_all(project_path)?;
+    copy_contents(template_dir.path(), project_path)?;
+
+    // if there are with-contract flags, include the example contracts
+    if include_example_contracts(with_contracts) {
+        println!("Including example contracts: {with_contracts:?}");
```
leighmcculloch, line 127: Can we print a comma separated list of contract names instead of the array?


```suggestion
        println!("Including example contracts: {}", with_contracts.iter().map(ToString::to_string).join(", "));
```

Or, if you replace the `impl ToString for ExampleContract` with a `impl Display for ExampleContract` you'll get the `ToString` impl for free and you'll be able to write the join as:

```suggestion
        println!("Including example contracts: {}", with_contracts.iter().join(", "));
```

_For both you'll need to add `use itertools::Itertools;` to the file._

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        if contents_to_exclude_from_copy.contains(&entry_name.as_str()) {
+            continue;
+        }
+
+        if path.is_dir() {
+            std::fs::create_dir_all(&new_path)?;
+            copy_contents(&path, &new_path)?;
+        } else {
+            std::fs::copy(&path, &new_path)?;
```
leighmcculloch, line 228: It appears writing contracts and initializing the project is destructive and will overwrite files that already exist. I think we should default to either erroring or skipping files that already exist. The latter would probably be good for us if the intent is to allow rerunning the command with an existing repo to add additional contracts.

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+        .fetch_then_checkout(gix::progress::Discard, &AtomicBool::new(false))
+        .map_err(Box::new)?;
+
+    let (_repo, _outcome) =
+        prepare.main_worktree(gix::progress::Discard, &AtomicBool::new(false))?;
+
+    Ok(())
+}
+
+fn copy_contents(from: &Path, to: &Path) -> Result<(), Error> {
```
leighmcculloch, line 212: Could we log to console each file and directory being written? If we skip files that already exist we can indicate so in the output.

leighmcculloch (commented):
👏🏻 This is looking nice. I tried it out locally and left some comments inline.

cmd/soroban-cli/src/commands/init.rs
leighmcculloch: e.g.:

```
❯ soroban init . --with-contract hello-world --with-contract custom-types
ℹ️ Initializing project at .
ℹ️ Skipped creating .gitignore as it already exists
ℹ️ Skipped creating Cargo.toml as it already exists
ℹ️ Skipped creating Cargo.lock as it already exists
ℹ️ Skipped creating README.md as it already exists
ℹ️ Initializing example contract: hello_world
➕ Writing contracts/hello_world/Cargo.toml
➕ Writing contracts/hello_world/src/lib.rs
➕ Writing contracts/hello_world/src/test.rs
ℹ️ Initializing example contract: custom_types
ℹ️ Skipped creating contracts/custom_types as it already exists
✅ Project initialized
```

cmd/soroban-cli/src/commands/init.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    pub project_path: String,
+
+    /// optional flag to specify soroban example contracts to include
+    #[arg(short, long, num_args = 1..=20, default_value = "none")]
+    pub with_contract: Vec<ExampleContract>,
```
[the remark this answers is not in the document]
elizabethengelman: I like that!

cmd/soroban-cli/src/commands/init.rs
elizabethengelman: Yep, good call! I'll push a fix up shortly.

elizabethengelman: c10d79a32798b945b0e92f7e308b99e6c739ee31

cmd/soroban-cli/src/commands/init.rs
elizabethengelman: Great point! I'll plan to wrap a `--template` flag into a follow up PR.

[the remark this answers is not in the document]
elizabethengelman: Nice catch, I really like the goal of making this accessible to users without an internet connection. Are you thinking that we could include the code from https://github.com/AhaLabs/soroban-init-template directly in this repo? And maybe just include the `hello-world` contract as an example.

cmd/soroban-cli/src/commands/init.rs
leighmcculloch: Instead of hello world, which tbh is a bit of an odd contract (my fault, I wrote it), I think we should include something like the increment / storage example. We could chop it down a bit, but Im thinking something simple but functional that anyone can understand like adding two numbers of incrementing a stored number. And ideally we also include a test in a test file, and a fuzz test in a fuzz setup. Thoughts?

leighmcculloch: Also I realize fuzzing is an advanced thing, but I think we've gotta push hard on fuzzing to make it adopted.

cmd/soroban-cli/src/commands/init.rs
elizabethengelman: Great catch! I agree that skipping existing files makes sense - I'll get a change pushed up shortly.

cmd/soroban-cli/src/commands/init.rs
elizabethengelman: Yeah, great idea!

elizabethengelman: updated in [31fdbf9](https://github.com/stellar/soroban-tools/pull/1156/commits/31fdbf999f092fd0d49d14fc812ca4a4635aabe1)

cmd/soroban-cli/src/commands/mod.rs
```diff
@@ -121,6 +123,8 @@ pub enum Cmd {
     Config(config::Cmd),
     /// Watch the network for contract events
     Events(events::Cmd),
+    /// Initialize a new Soroban project
+    Init(init::Cmd),
```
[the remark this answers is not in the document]
elizabethengelman: updated in [73b0ab7](https://github.com/stellar/soroban-tools/pull/1156/commits/73b0ab7ff0d2c99388f71f7cba102aa1fc209766)

cmd/soroban-cli/src/commands/init.rs
elizabethengelman: updated in [31fdbf9](https://github.com/stellar/soroban-tools/pull/1156/commits/31fdbf999f092fd0d49d14fc812ca4a4635aabe1)

elizabethengelman:
I just pushed up a change that adds the template code in this repo with the increment example contract, which I hope addresses a couple of the outstanding pieces on this PR. [d71493f](https://github.com/stellar/soroban-tools/pull/1156/commits/d71493f99f8739145ca78908baffb8e92071438c) and [4daf40b](https://github.com/stellar/soroban-tools/pull/1156/commits/4daf40b6ad29f49325675e253345fc1e8030987e)

- users should be able to run `soroban contract init` without the need for an internet connection since the template code is in this repo
- the project will include an example contract by default, so that we're not initializing an empty project
- I'm thinking that we can make the inclusion of other examples via `--with-example` from the examples repo be dependent on an internet connection, and in a follow-up PR we can generate the list instead of hardcoding it, and fall back to just including `increment`

Right now I just added the template code in `cmd/soroban-cli/src/utils/`, but in the future, maybe we can think about including it via git submodule or something. But I think that can be an add-on later. Also, happy to rethink if `utils` is the best place for it. 

@leighmcculloch @stellarsaur curious to see what you think!

cmd/soroban-cli/src/commands/init.rs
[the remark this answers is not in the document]
elizabethengelman: Marking this as resolved, as I plan to handle this in a follow-up PR.

cmd/soroban-cli/src/commands/init.rs
elizabethengelman: Making this as resolved - I've added the increment contract to be included by default. I'll plan to look at the fuzz setup piece in a follow-up PR.

cmd/soroban-cli/Cargo.toml
```diff
@@ -93,6 +93,13 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 cargo_metadata = "0.15.4"
 pathdiff = "0.2.1"
 dotenvy = "0.15.7"
+gix = { version = "0.55.2", default-features = false, features = [
```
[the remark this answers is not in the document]
elizabethengelman: planning to handle this in a follow up PR

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -40,6 +41,9 @@ pub enum Cmd {
     #[command(subcommand)]
     Id(id::Cmd),
 
+    /// Initialize a Soroban project with an example contract
+    Init(init::Cmd),
```
chadoh, line 45: I think we want this to be a top-level command, right? If we nest it under `contract`, it makes it seem like it is for initializing a contract. But it's more general than that.

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    pub project_path: String,
+
+    /// optional flag to specify soroban example contracts to include
+    #[arg(short, long, num_args = 1..=20, default_value = "none")]
```
chadoh, line 72: `increment` gets installed no matter what, currently. If I omit all `-w` options, it gets installed. If I only say `-w hello-world`, both `increment` and `hello-world` get added. If I say `-w increment`, then I get this output:

```
cargo run -- contract init ../lol -w increment
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/soroban contract init ../lol -w increment -w hello-world`
ℹ️  Initializing project at ../lol
➕  Writing ../lol/Cargo.toml
➕  Writing ../lol/contracts/increment/Cargo.toml
➕  Writing ../lol/contracts/increment/src/test.rs
➕  Writing ../lol/contracts/increment/src/lib.rs
➕  Writing ../lol/README.md
➕  Writing ../lol/.gitignore
ℹ️  Initializing example contract: increment
ℹ️  Skipped creating ../lol/contracts/increment/Cargo.toml as it already exists
ℹ️  Skipped creating ../lol/contracts/increment/src/test.rs as it already exists
ℹ️  Skipped creating ../lol/contracts/increment/src/lib.rs as it already exists

chadoh (changes requested):
Noticed two small things that I hope are quick to fix!

cmd/soroban-cli/src/commands/contract/mod.rs
elizabethengelman: @leighmcculloch suggested we move it under `contract` as we are starting to think about merging soroban-cli and stellar-cli: https://github.com/stellar/soroban-tools/pull/1156#discussion_r1446852810

I agree that moving this from the top level may make it easier to merge the CLIs, but I also understand that putting it under `contract` may make it feel more specific than the intended command. 🤔

cmd/soroban-cli/src/commands/contract/init.rs
elizabethengelman: Ah, nice catch! In some of the convo above on this PR we had talked about always including the increment contract by default so that the user would have an example to go off. But I can rework this interaction, so it's a bit more intuitive.

leighmcculloch: Could we make the default contract that is written out a new contract template rather than one of the examples? It could contain the same logic as increment, but named 'contract/' so it is more a boilerplate for someone to work in, rather than an example named something else.

elizabethengelman: Just pushed up a change in [4f7eac4](https://github.com/stellar/soroban-tools/pull/1156/commits/4f7eac47456d8724a9f0ce86770c4b34415eac83) and [ffc4d80](https://github.com/stellar/soroban-tools/pull/1156/commits/ffc4d80f2d12198fde747995ee64116ffb11a498) that will include hello-world by default, and removes it from the list of options so no longer have that confusing `Skipped creating...` for it.

cmd/soroban-cli/src/commands/contract/mod.rs
chadoh: @leighmcculloch I'd like to see us keep this as a top-level command. As we increase support for classic stellar functionality in the CLI, I think the right thing to do with `init` would be to change the template so that it also has better support for classic Stellar. Nesting this under `contract` feels a little confusing to me, since it's not for initializing a contract. Are you ok with moving it back to the top level?

leighmcculloch: I guess I'm not seeing what an init command for classic stellar would contain. Classic stellar applications are typically built using SDKs, building XDR txns, and calling the Horizon API. There's no common shape that I think we've ever found applications doing these things, although maybe we could generate a project that ingests data, or a project that is a frontend for using freighter.

Is that what you're thinking?

I think if we go that path we really need a flag that identifies a "type" of project the init function is generating. In the case of this PR, it adds a "contract" category.

chadoh: Ok well in that case I think keeping it where it is for now also makes sense.

chadoh (approved):

cmd/soroban-cli/src/commands/contract/mod.rs
leighmcculloch: Thinking about this more, I think this could be quite useful especially as we encourage patterns for ingesting meta and building indexers. It could be helpful to init a project that is an indexer. So I think you're onto something and `soroban init --type contract` isn't too dissimilar of `soroban contract init`.

@chadoh @elizabethengelman I defer to you folks.

elizabethengelman: Thanks for the good convo on this! Leaving it as `soroban contract init` for now seems good to me.

I like the idea of creating templates for indexers or ingesting metadata, but maybe we shouldn't include the `--type` flag yet, just in case we end up not building those additional templates. Also, I could see a world where `soroban indexer init` or `soroban project init` could be a thing. So maybe we can wait and see how it progresses to add the idea of different initializers.

PR #1159, opened 2024-01-09
Run CI for the msrv and latest rust version

leighmcculloch:
### What
Run CI for the minimum supported rust version (msrv) and latest rust version.

Also make the msrv consistent in this repo.

### Why
To ensure that the package continues to build on the minimum supported rust version that is defined in the Cargo.toml, which is currently Rust v1.70, as well as continues to be tested on the latest version.

For stellar/rs-soroban-env#1311


Cargo.toml
```diff
@@ -12,6 +12,7 @@ exclude = ["cmd/crates/soroban-test/tests/fixtures/hello"]
 
 [workspace.package]
 version = "20.1.1"
+rust-version = "1.74.0"
```
leighmcculloch, line 15: The packages in this repo claimed to minimally supported rust 1.69, 1.70, and 1.72, however they actually require 1.74. This is part of why we're adding these new CI runs that test against the msrv. The new CI runs failed because of this. The crates in this repo require 1.74 because they're dependent on other crates that have a msrv of 1.74.

stellarsaur (approved):

PR #1160, opened 2024-01-09
Bump Go, Rust and Core dependencies

2opremio:


.github/workflows/soroban-rpc.yml
```diff
@@ -112,7 +112,7 @@ jobs:
     env:
       SOROBAN_RPC_INTEGRATION_TESTS_ENABLED: true
       SOROBAN_RPC_INTEGRATION_TESTS_CAPTIVE_CORE_BIN: /usr/bin/stellar-core
-      PROTOCOL_20_CORE_DEBIAN_PKG_VERSION: 20.0.3-1648.7c7dcefd3.focal
+      PROTOCOL_20_CORE_DEBIAN_PKG_VERSION: 20.0.3-1655.114b833e7.focal
```
sreuland, line 115: could also bump the [core git ref in gh e2e.yml](https://github.com/stellar/soroban-tools/blob/main/.github/workflows/e2e.yml#L25) to same `114b833e755400178a57142f45b7fb892ddb034f`, or maybe e2e can wait to be bumped later when 20.1.0 tag refs are available, looks like e2e is passing currently with its older 20.0.2 core ref.

2opremio: Ah, good point. I will do that

PR #1167, opened 2024-01-12
soroban-rpc: Remove panics from internal codebase

### LINKED ISSUE #500: soroban-rpc: Centralize error handling and failure

There are currently a few cases where soroban-rpc will `panic`/`os.Exit` from deep within the code.

For example:
```
cmd/soroban-rpc/internal/methods/simulate_transaction.go:62
// FIXME: SourceAccount() panics, so, the user can doctor an envelope which makes the server crash
```

Audit and fix these where possible. Where not possible, make the fns return errs up and recover. If no recovery is possible, return an error up to `main`. 

@paulbellamy 's Opinion: `main` is the only place which should be doing a `panic` or `os.Exit`. This makes the rpc server easier to integration test.

aditya1702:
### What

Fixes #500 

There are instances in soroban-rpc where the code will panic/exit from within. This PR fixes the code and returns error to main instead of exiting internally.

### Why

Makes it easier to test the code and makes it consistent in handling errors.

### Known limitations

NA


cmd/soroban-rpc/internal/methods/send_transaction.go
```diff
@@ -74,7 +74,11 @@ func NewSendTransactionHandler(daemon interfaces.Daemon, logger *log.Entry, stor
 		}
 		txHash := hex.EncodeToString(hash[:])
 
-		ledgerInfo := store.GetLatestLedger()
+		ledgerInfo, err := store.GetLatestLedger()
+		if err != nil {
+			return SendTransactionResponse{}, err
```
sreuland, line 79: just for thought, would be great if could add a test assertion for how this error is expected to look from json-rpc response, transaction_test.go has some other tests that check error responses, but in this case, not sure how to inject a mock or trigger GetLatestLedger to return an err during the test. 

Do you think it's worthwhile to wrap the err further into a jrpc2 struct, to include more Message denoting this was due to GetLatestLedger attempt similar to [invalid params above](https://github.com/stellar/soroban-tools/pull/1167/files#diff-73f019a9e55037d6a4812f6b0e82b5ed6e56ab3b8e0a0d6151406838b9cd2edeR61-R63)

cmd/soroban-rpc/internal/config/config_option.go
```diff
@@ -101,7 +99,7 @@ func (o *ConfigOption) setValue(i interface{}) (err error) {
 		}
 	}()
 	parser := func(option *ConfigOption, i interface{}) error {
-		panic(fmt.Sprintf("no parser for flag %s", o.Name))
+		return errors.Errorf("no parser for flag %s", o.Name)
```
sreuland, line 102: can you add an assertion in config_option_test.go that demonstrates the expected outcome if this gets triggered?

cmd/soroban-rpc/internal/methods/send_transaction.go
aditya1702: @sreuland Yes I was also thinking the same. It would be better for a user to see where it failed from.

cmd/soroban-rpc/internal/ledgerbucketwindow/ledgerbucketwindow.go
```diff
@@ -66,11 +66,11 @@ func (w *LedgerBucketWindow[T]) Len() uint32 {
 }
 
 // Get obtains a bucket from the window
-func (w *LedgerBucketWindow[T]) Get(i uint32) *LedgerBucket[T] {
+func (w *LedgerBucketWindow[T]) Get(i uint32) (*LedgerBucket[T], error) {
 	length := w.Len()
 	if i >= length {
-		panic(fmt.Errorf("index out of range [%d] with length %d", i, length))
+		return nil, fmt.Errorf("index out of range [%d] with length %d", i, length)
```
2opremio, line 72: I think panics like these are justified (just like `s[i]` panics if `i` is out of bounds).

aditya1702: Yeah makes sense

2opremio (approved):

PR #1168, opened 2024-01-16
rpc: Use external soroban-simulation library for preflight computations

2opremio:
### What

Move preflight computations to `rs-soroban-env` repo (followup to https://github.com/stellar/rs-soroban-env/pull/1300 )

### Why

Moving it closer to the host library will help maintain the code and (longer term) test it against the metering done by core.

### Known limitations

N/A


stellarsaur (approved):
LGTM! I think it's good we're using an external library for this.

PR #1060, opened 2023-10-30
soroban-rpc: unify transaction and events retention window

tsachiherman:
### What

unify transaction and events retention window

### Why

see https://github.com/stellar/soroban-rpc/issues/15

### Known limitations

n/a

Issue #1056, opened 2023-10-26
soroban-rpc: increase transaction retention window to 24 hours


tsachiherman:
### What ?

Requested by product - increase transaction retention window to 24 hours. This would reduce confusion about retention window.
Before increasing the retention window, we should evaluate what the expected memory utilization is going to be and re-align hardware specifications as needed.

tsachiherman:
@mollykarcher I was asked to add this to the next release; would that be possible for P0 ?

tomerweller:
Just to clarify: this is the default value, operators can still configure a different value, right? 

Regardless, I suggest having a single configuration value for retention window for both transactions and events

mollykarcher:
> Just to clarify: this is the default value, operators can still configure a different value, right?

Correct, this is just the default value.

> Regardless, I suggest having a single configuration value for retention window for both transactions and events

So this sounds a little different than the title/description here, but I'd agree that this is what would actually reduce the user confusion with it. We want to align the retention windows, not simply increase the default transaction retention window. So I think that would mean we'd _remove_ both `transaction-retention-window` and `event-retention-window` configs, and replace those with a single `retention-window` that controls the retention for both.

> @mollykarcher I was asked to add this to the next release; would that be possible for P0 ?

I'm in favor. If we go with what I've described above, we'd be removing configuration options which is technically a breaking change, so I'd prefer to get it out before the stable release. It also sounds like minimal work, assuming the memory footprint doesn't take a bit hit. Either way, I think getting it into the stable release is definitely doable.

ire-and-curses:
> and re-align hardware specifications as needed.
To be clear: we need to understand the expected memory utilisation and then _decide if that is acceptable_ for hardware requirements.

I've updated the ticket body to cover this, and to describe the product rationale, because in the future we might want to remember why we did this.

tsachiherman:
With https://github.com/stellar/soroban-tools/pull/1063 merged in, I think it would be safe to have a 16GB machine running a pubnet node with a 24 hour transaction history.

tsachiherman:
I have not worked on the in-memory events store, but expect that a similar gains could be duplicated there.

mollykarcher:
@stellarsaur can you run the same profiling you did for RPC with the updated transaction window (to 24 hours), so we can get a sense of how this will affect the memory consumption?

stellarsaur:
[...]

PR #1169, opened 2024-01-16
soroban-rpc: Add common retention window for events and transactions.

psheth9:
### What

Add a config for common retention window for events and transactions

### Why

More detail here: https://github.com/stellar/soroban-rpc/issues/15

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-rpc/internal/config/config.go
+++ b/cmd/soroban-rpc/internal/config/config.go
@@ -39,6 +39,7 @@ type Config struct {
 	PreflightEnableDebug                        bool
 	SQLiteDBPath                                string
 	TransactionLedgerRetentionWindow            uint32
+	LedgerRetentionWindow                       uint32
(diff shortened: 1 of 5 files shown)


2opremio:
I know it's still a draft, but we shouldn't merge this until we can gauge the storage implications of increasing the tx window to 24 hours.

psheth9:
Yes agreed @2opremio !! I won't merge :) 

I think @stellarsaur already did some initial profiling [here](https://github.com/stellar/soroban-tools/issues/1056#issuecomment-1841875651)  

Do we also want to make sure that retention window should not go beyond 24hrs? or no such upper bound limitation here? cc @mollykarcher (not sure if we are doing that already)

2opremio:
BTW, isn’t this a dup of #1060 ?

leighmcculloch:
Closing since the rpc has moved to another repo, this change needs taking to that other repo if the change is still relevant.

PR #1174, opened 2024-01-19
[Epic] Separating soroban-rpc to prepare for repo change

willemneal:
### What

- [x] feat: move ContractSpec to spec tools crate
- [x] feat: add methods to GetTransaction and break up submitting a transaction
- [x] feat: create soroban-rpc crate
- [x] ~~Use testcontainers to test CLI instead of through go.~~ In follow up PR

### Why

### Known limitations

[TODO or N/A]


willemneal:
Currently the rust is [failing](https://github.com/stellar/soroban-tools/actions/runs/7720380869/job/21045175902?pr=1174) when building windows because the rust version is 1.70.0, but soroban-sdk requires 1.74.0.

e2e tests are still failing because of the changes introduced by this PR. Currently it relies on the fact that previously all invokes where signed and submitted, regardless if they had auth and had no writes. I added a flag with an env var to get around this. I had hoped that `SYSTEM_TEST_VERBOSE_OUTPUT` would be true for the invoke call.

I added to the RPC tests to test the same case that is failing for e2e tests and those are passing. Should I make a branch of the e2e? 

Or should I make signed the default so that this isn't a breaking change?

cmd/crates/soroban-rpc/Cargo.toml
```diff
@@ -0,0 +1,59 @@
+[package]
```
stellarsaur, line 1: IMO this crate should live in the new `soroban-rpc` repo, but it should be easy to move it over once this PR is merged

cmd/crates/soroban-rpc/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+ed25519-dalek = { workspace = true }
+tracing = { workspace = true }
+
+
+# networking
+jsonrpsee-http-client = { workspace = true }
+jsonrpsee-core = { workspace = true }
+http = { workspace = true }
+
+# soroban-ledger-snapshot = { workspace = true }
```
stellarsaur, line 53: Should we just remove these commented out lines?

cmd/crates/soroban-rpc/README.md
```diff
@@ -0,0 +1,3 @@
+# soroban-rpc
+
+Tools and utilities for soroban rpc.
```
stellarsaur, line 3: Same comment above; this should live in `soroban-rpc` eventually but it's fine for now

cmd/crates/soroban-rpc/src/lib.rs
```diff
@@ -21,15 +21,20 @@ use std::{
...
 use termcolor::{Color, ColorChoice, StandardStream, WriteColor};
 use termcolor_output::colored;
 use tokio::time::sleep;
 
-use crate::utils::contract_spec;
-
+pub mod log;
 mod txn;
 
+pub use txn::*;
```
stellarsaur, line 32: Looks like we're only using `txn::Assembled` so would `pub use tnx::Assembled` be better so we're not pulling everything in? Disclaimer: I'm not a Rust expert

cmd/crates/soroban-rpc/src/lib.rs
```diff
@@ -170,6 +178,42 @@ impl TryInto<GetTransactionResponse> for GetTransactionResponseRaw {
...
+    ///
+    /// # Errors
+    pub fn return_value(&self) -> Result<xdr::ScVal, Error> {
+        if let Some(xdr::TransactionMeta::V3(xdr::TransactionMetaV3 {
+            soroban_meta: Some(xdr::SorobanTransactionMeta { return_value, .. }),
+            ..
+        })) = self.result_meta.as_ref()
+        {
+            Ok(return_value.clone())
+        } else {
```
stellarsaur, line 191: I don't think we need the `else` here (also in `events` function)

stellarsaur (approved):
A few very minor comments, but LGTM. Thanks for the PR!

cmd/crates/soroban-rpc/Cargo.toml
willemneal: I agree. Perhaps it's best to move it after the first publish?

cmd/crates/soroban-rpc/src/lib.rs
willemneal: It won't cause the dependent crate to bloat since rust has tree shaking by default. However, you raise a good point: since this crate is now a library, it's probably best to limit the API and the blanket * could cause issues in the future if anything is added to `txn`.

cmd/crates/soroban-rpc/Cargo.toml
willemneal: Updated deps to not use workspace to make this transition easier.

PR #1183, opened 2024-01-30
rpc: Reduce event memory footprint

2opremio:
1. Get rid of the in-memory operation index (since it was always set to zero anyways)
2. Keep events serialized while in memory (it saves quite a bit of space due to the inneficient representation of unions in golang).


Shaptic (approved):
noice! :clap:  are there benchmarks? not that they're necessary to merge by any means but it'd be cool to see what kind of improvement this adds

2opremio:
@psheth9 is working on the benchmarks separately

Issue #1182, opened 2024-01-26
rpc: add `transactionId` to event resource


tomerweller:
There is currently no way to associate an event from the `getEvents` endpoint with the transaction that caused it to be emitted. 

([context](https://discord.com/channels/897514728459468821/1199438502098645103/1200431449829023824) from discord)

PR #1185, opened 2024-01-31
rpc: Store and serve the event transaction ID

2opremio:
### What

Add the transaction hash pertaining to each event in the `getEvents` response

### Why

Fixes https://github.com/stellar/soroban-tools/issues/1182

### Known limitations

It increases the size of the in-memory event storage (but this should be mitigated by #1183 )

It would be good to have specific numbers (@psheth9 is working on it)


psheth9 (approved):
Looks good to me !!

PR #1188, opened 2024-02-02
rpc: Make sure we internally use uint32 for latestledger in getEvents

2opremio:
Followup to https://github.com/stellar/soroban-tools/pull/1097


stellarsaur:
If it's not too much work, please close this PR and reopen in `soroban-rpc`. We could also merge this PR here (after tests pass) and cherry pick over to `soroban-rpc`, but it would be easier to directly merge this change to `soroban-rpc`.

PR #1196, opened 2024-02-08
fix: embed the init template files in the build

elizabethengelman:
### What

Embed the contract_init_template files in the binary using `rust-embed` to make sure that they are accessible to the `init` command to copy to the user's filesystem.

### Why

Fixes the issue reported here: https://discord.com/channels/897514728459468821/1204799895014346842/1204799895014346842

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/init.rs
+++ b/cmd/soroban-cli/src/commands/contract/init.rs
@@ -1,13 +1,23 @@
-use clap::builder::{PossibleValue, PossibleValuesParser, ValueParser};
-use clap::{Parser, ValueEnum};
+use std::{
+    ffi::OsStr,
+    fs::{copy, create_dir_all, metadata, read_dir, read_to_string, write},
+    io,
+    num::NonZeroU32,
+    path::Path,
+    str,
+    sync::atomic::AtomicBool,
+};
+
+use clap::{
+    builder::{PossibleValue, PossibleValuesParser, ValueParser},
+    Parser, ValueEnum,
+};
+use gix::{clone, create, open, progress, remote};
 use rust_embed::RustEmbed;
 use serde::Deserialize;
-use std::fs::read_to_string;
-use std::num::NonZeroU32;
-use std::path::Path;
-use std::sync::atomic::AtomicBool;
-use std::{fs, io};
-use toml_edit::{Document, Formatted, InlineTable, TomlError, Value};
+use serde_json::{from_str, json, Error as JsonError, Value as JsonValue};
+use toml_edit::{Document, Formatted, InlineTable, Item, TomlError, Value as TomlValue};
+use ureq::{get, Error as UreqError};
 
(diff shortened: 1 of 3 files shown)


cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -127,20 +133,21 @@ impl Cmd {
...
-    let template_dir_path = Path::new(cli_cmd_root)
-        .join("src")
-        .join("utils")
-        .join("contract-init-template");
-
     // create a project dir, and copy the contents of the base template (contract-init-template) into it
-    std::fs::create_dir_all(project_path)?;
-    copy_contents(template_dir_path.as_path(), project_path)?;
+    std::fs::create_dir_all(project_path).map_err(|e| {
+        eprintln!("Error creating new project directory: {project_path:?}");
```
elizabethengelman, line 147: Not sure if this is overkill, but I ended up printing out a message in addition to the error message for a bunch of potential errors, since a lot of them end up being pretty generic `IO` errors, and it may be difficult to track down the origin of the error. 

Happy to rethink this if folks have some other ideas!

willemneal: I think this is fine for now.  It would be nice if you were using init as a library to have the more specific error, but most CLI's use anyhow, which acts similar to this.

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -1,12 +1,12 @@
-use std::fs::read_to_string;
-use std::path::Path;
-use std::{env, fs, io};
-
 use clap::builder::{PossibleValue, PossibleValuesParser, ValueParser};
 use clap::{Parser, ValueEnum};
+use rust_embed::RustEmbed;
 use serde::Deserialize;
+use std::fs::read_to_string;
```
willemneal, line 5: I personally like all std imports at the top, then external crates and then workspace/crate. Also the imports can be combined. 
```rust
use std::{fs::..., path::Path, ...}
```

elizabethengelman: I'll get a fix pushed up for this shortly!

Also, in general, do you prefer to have external crates imported at the top of the file with `use`, over qualifying them in-line? I've realized I have a bit of both here, and am thinking that some of the qualified imports could be updated.

cmd/soroban-cli/src/commands/contract/init.rs
elizabethengelman: Interesting! Are we using any other commands as libraries at the moment?

chadoh (approved):

cmd/soroban-cli/src/commands/contract/init.rs
willemneal: It's used  by the soroban-test crate. But I've used it a lot with other CLI projects. Makes it easy to plug an existing CLI into a subcommand and link them all statically.

elizabethengelman: Cool, i'll take a look!

PR #1201, opened 2024-02-14
soroban-cli: Update Release Docs

stellarsaur:
### What

This PR updates release docs for Soroban CLI only.

### Why

Previously our release process was for Soroban CLI & RPC together. Now that we release separately we should have separate docs.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,5 +1,18 @@
 # Releasing
 
-The process for how to release the crates in this repository are documented here:
+To release Soroban CLI, follow this process **in order**:
 
-https://github.com/stellar/actions/blob/main/README-rust-release.md
+## Follow Rust Workflow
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -1,5 +1,18 @@
 # Releasing
 
-The process for how to release the crates in this repository are documented here:
+To release Soroban CLI, follow this process **in order**:
 
-https://github.com/stellar/actions/blob/main/README-rust-release.md
```
leighmcculloch, line 5: The release process this repo should follow is still the linked rust release process, so I'm not sure this change is needed. The rust release process has instructions for how to create a release, which shouldn't be the git tag commands, but instead using the steps in the referenced markdown.

leighmcculloch (changes requested):

PR #1203, opened 2024-02-16
Repoint soroban-spec-tools to the moved one in stellar/soroban-rpc

Shaptic:
### What
Removal of the spec-tools crate (`rm -rf ./cmd/crates/soroban-spec-tools`) and repointing it to the [soroban-rpc repository](https://github.com/stellar/soroban-rpc).

### Why
This is complementary to https://github.com/stellar/soroban-rpc/pull/68 in order to resolve a circular dependency between CLI and RPC.


psheth9 (approved):
Looks good !!

.github/workflows/e2e.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 jobs:
   integration:
+    if: false
```
mollykarcher, line 10: Was this intentional?

Shaptic: yerp

2opremio:
I made more progress after reverting #1196 (which hopefully won't be needed if https://github.com/stellar/soroban-cli/pull/1206 fixes the problem).

However, now:
1. we have run into https://github.com/sfackler/rust-openssl/issues/2179
2. The typescript tests are also failing

Shaptic:
It only took 5 hours to get the correct version of `make` used but EVERYTHING PASSES NOW.

PR #1206, opened 2024-02-21
Rename Cargo.toml to Cargo.text in the init template files

elizabethengelman:
### What

Updates the init template Cargo file from `Cargo.toml` to `Cargo.text`.

### Why
We were seeing a failure with the dry-run:
https://github.com/stellar/soroban-cli/actions/runs/7983365311/job/21798405675

Slack thread with discussion about the problem: https://aha-labs.slack.com/archives/C04B02ABF37/p1708486449196609

This is due to a known cargo issue: https://github.com/rust-lang/cargo/issues/9017

If a subdirectory includes a `Cargo.toml` file, it will be viewed as a sub-package, and therefore the contents of the directory will not be included in the package. Since we are including this `Cargo.toml` to be used as a template for writing a file to the user's fs, and it is not to be used as a sub-package, I've changed the name of the file to `Cargo.text` in this source code. Then when the file is being copied into the project that is being created on the user's machine, the file name is written as `Cargo.toml`, as expected.

From the [cargo book docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields):
```
Regardless of whether exclude or include is specified, the following files are always excluded:

Any sub-packages will be skipped (any subdirectory that contains a Cargo.toml file).
A directory named target in the root of the package will be skipped.
```

I tested this locally by recreating the steps in the dry-run, that @2opremio outlined in the slack convo, and was able to get through `cargo-hack hack --feature-powerset --ignore-private --config "source.crates-io.replace-with = 'vendored-sources'" --config "source.vendored-sources.directory = 'vendor'" package` without error.


2opremio:
I would personally use something more obvious, in the lines of `Cargo.toml.removeextension` or `Cargo.toml.extensionworkaround`

elizabethengelman:
> I would personally use something more obvious, in the lines of Cargo.toml.removeextension or Cargo.toml.extensionworkaround

Great point. I'll push up a change shortly.

Shaptic (approved):

2opremio:
@elizabethengelman I just noticed that the `Cargo.toml` file at https://github.com/stellar/soroban-cli/tree/main/cmd/soroban-cli/src/utils/contract-init-template/contracts should had been renamed too.

PR #1208, opened 2024-02-21
Update references in end-to-end tests to point to the latest releases

Shaptic:
### What
See title.

### Why
Confirm that tests pass against current latest release.

### Known limitations
n/a


.github/workflows/e2e.yml
```diff
@@ -45,45 +45,48 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         name: checkout system-test
         with:
           repository: stellar/system-test
           ref: ${{ env.SYSTEM_TEST_GIT_REF }}
           path: system-test
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
sreuland, line 67: I don't think need to checkout local copy of soroban-tools, cli refs to gh workspace directory, and rpc refs soroban-rpc repo now

.github/workflows/e2e.yml
```diff
@@ -45,45 +45,48 @@ jobs:
...
         name: checkout soroban-tools
         with:
           repository: stellar/soroban-tools
           path: soroban-tools
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
sreuland, line 72: I think can remove this local checkout of soroban-rpc, SYSTEM_TEST_SOROBAN_RPC_REF provides where that source is coming from.

sreuland:
the e2e is failing to [build the system-test image based on cli source location](https://github.com/stellar/soroban-cli/actions/runs/7996724112/job/21839786586#step:8:7739), [SYSTEM_TEST_SOROBAN_CLI_REF](https://github.com/stellar/soroban-cli/pull/1208/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR21) should point at the gha workspace folder that soroban-cli source is checked out, which I think in this case of it being the pr repo, can be obtained from just `${{ github.workspace }}` , no need to append an additional 'soroban-cli' path onto it.

.github/workflows/e2e.yml
Shaptic: I think it's needed for pathing? unsure though

.github/workflows/e2e.yml
```diff
@@ -45,45 +45,48 @@ jobs:
...
       # configured in standalone network mode(core, rpc)
-      SYSTEM_TEST_QUICKSTART_GIT_REF: https://github.com/stellar/quickstart.git#412bb828ddb4a93745227ab5ad97c623d43f3a5f
+      SYSTEM_TEST_QUICKSTART_GIT_REF: https://github.com/stellar/quickstart.git#eeb44ee99c564275654800edea2e426d73e966b8
 
       # triggers system test to log out details from quickstart's logs and test steps
       SYSTEM_TEST_VERBOSE_OUTPUT: "true"
 
       # the soroban test cases will compile various contracts from the examples repo
-      SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_HASH: "v20.0.0"
+      SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_HASH: "v20.2.0"
```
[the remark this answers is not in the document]
Shaptic: nice call, added :+1:

.github/workflows/e2e.yml
sreuland: system-test make/build will not be referring to the soroban-tools path, it resolves rpc and cli source paths from the urls or local paths set on  `SYSTEM_TEST_SOROBAN_CLI_REF` and `SYSTEM_TEST_SOROBAN_RPC_REF`

sreuland: does soroban-cli need to use a pinned ref to soroban-rpc for testing purposes? [current config on SYSTEM_TEST_SOROBAN_RPC_REF](https://github.com/stellar/soroban-cli/pull/1208/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR22) is not spec'ing a git ref on the repo url, using mainline/master ref then.

PR #1209, opened 2024-02-22
Bump version of soroban rpc and spec tools

psheth9:
We need to bump these versions because fee padding fix available in the latest release of rpc [here](https://github.com/stellar/soroban-rpc/releases/tag/v20.3.3)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -263,9 +263,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.15.1"
+version = "3.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c764d619ca78fccbf3069b37bd7af92577f044bb15236036662d79b6559f25b7"
+checksum = "a3b1be7772ee4501dba05acbe66bb1e8760f6a6c474a36035631638e4415f130"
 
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -2619,18 +2619,18 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
...
-checksum = "4bf7e82ffd6d3d6e6524216a0bfd85509f68b5b28354e8e7800057e44cefa9b4"
+checksum = "5cff92b6f71555b61bb9315f7c64da3ca43d87531622120fea0195fc761b4843"
 dependencies = [
  "cc",
 ]
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.90"
+version = "0.9.101"
```
Shaptic, line 2631: This isn’t supposed to change 😅 but hopefully it doesn’t affect dry run. Just keep it in mind if it fails later:

cargo update openssl-sys —precise 0.9.90 

will undo this.

Shaptic (approved):

Cargo.lock
psheth9: reverting changes for openssl and bumpalo !!

PR #1210, opened 2024-02-22
feat: have users preview transaction before signing

willemneal:
### What

This adds a preview of the transaction and prompts the user before signing.
  - [ ] Show output

### Why

Currently the CLI signs transactions and auth entries automatically. This can cause security issues, for example, a nested cross contract call which authorizes a FT transfer.

### Known limitations

[TODO or N/A]


willemneal:
closed in favor of #1283

PR #1211, opened 2024-02-22
Update references in end-to-end tests to point to the latest releases

sreuland:
### What
trying e2e changes

### Why

e2e is not building from source on #1208


sreuland:
@Shaptic , @psheth9 , got the e2e passing, should we land this pr and close #1208 ?

.github/workflows/e2e.yml
```diff
@@ -18,12 +18,12 @@ jobs:
 
       # the soroban CLI & RPC source code to compile and run from system test
       # refers to checked out source of current git hub ref context
-      SYSTEM_TEST_SOROBAN_CLI_REF: ${{ github.workspace }}
-      SYSTEM_TEST_SOROBAN_RPC_REF: https://github.com/stellar/soroban-rpc.git
+      SYSTEM_TEST_SOROBAN_CLI_REF: ${{ github.workspace }}/soroban-cli
+      SYSTEM_TEST_SOROBAN_RPC_REF: https://github.com/stellar/soroban-rpc.git#main
```
sreuland, line 22: I'm not sure on this ref to main, probably will want to pin this to last stable release tag?

Shaptic: yeah I can set it to the right tags and merge the PR! no need to wait for it to pass since main is the same as the tags right now imo

Shaptic: did it!

Shaptic (approved):

PR #1215, opened 2024-02-26
Add soroban-spec-tools

leighmcculloch:
### What
Add soroban-spec-tools.

### Why
Part of moving the crate back to this repo.

Related:
- https://github.com/stellar/soroban-rpc/pull/80

PR #1216, opened 2024-02-26
Output interface version parts

leighmcculloch:
### What
Output interface version parts in the contract inspect command.

### Why
The interface version stored in the contract meta is a 64-bit value where the high 32-bits is the protocol version and the low 32-bits is the actual interface version.

It was pointed out that the large integer is confusing.

We should break it down so that folks understand what it means.

Related:
- https://discord.com/channels/897514728459468821/1211454547524853761

Dependent on:
- #1215


leighmcculloch:
The CI failure is unrelated to this change and is being tracked here:
- https://github.com/stellar/soroban-cli/issues/1217

psheth9 (approved):

Issue #1217, opened 2024-02-26
Fails to build after previously working


leighmcculloch:
The `gix-ref` dependency has started to fail when building.

The failure is seen on CI and locally (macos).

I'm seeing it on Rust v1.74.0, 1.74.1, 1.75.0, and 1.76.0, so I don't think this isn't due to a Rust version release, unless I have a wicked caching issue across these tests.

```
   Compiling gix-ref v0.41.0
   Compiling gix-index v0.29.0
   Compiling rustls v0.22.2
   Compiling gix-refspec v0.22.1
   Compiling gix-odb v0.57.0
   Compiling openssl v0.10.55
   Compiling openssl-sys v0.9.101
   Compiling gix-negotiate v0.12.0
error[E0599]: the method `context` exists for fn item `fn(&mut &[u8]) -> Result<SignatureRef<'_>, ErrMode<_>> {decode::<'_, _>}`, but its trait bounds were not satisfied
   --> /Users/leighmcculloch/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gix-ref-0.41.0/src/store/file/log/line.rs:142:46
    |
142 |                 gix_actor::signature::decode.context(StrContext::Expected("<name> <<email>> <timestamp>".into())),
    |                                              ^^^^^^^ method cannot be called due to unsatisfied trait bounds
    |
    = note: the following trait bounds were not satisfied:
            `<for<'a> fn(&'a mut &[u8]) -> Result<SignatureRef<'_>, winnow::error::ErrMode<_>> {gix_actor::signature::decode::<'_, _>} as FnOnce<(&mut _,)>>::Output = Result<_, ErrMode<_>>`
            which is required by `for<'a> fn(&'a mut &[u8]) -> Result<SignatureRef<'_>, winnow::error::ErrMode<_>> {gix_actor::signature::decode::<'_, _>}: winnow::Parser<_, _, _>`
            `<&for<'a> fn(&'a mut &[u8]) -> Result<SignatureRef<'_>, winnow::error::ErrMode<_>> {gix_actor::signature::decode::<'_, _>} as FnOnce<(&mut _,)>>::Output = Result<_, ErrMode<_>>`
            which is required by `&for<'a> fn(&'a mut &[u8]) -> Result<SignatureRef<'_>, winnow::error::ErrMode<_>> {gix_actor::signature::decode::<'_, _>}: winnow::Parser<_, _, _>`
[...]

PR #1222, opened 2024-02-27
Make builds and tests consistent

leighmcculloch:
### What
Change the install make target to use the lock file, by adding --locked to all cargo install commands. Also encourage --locked usage elsewhere in CLI help output, and fix old make targets that are no longer needed since the split.

### Why
Cargo install won't use the Cargo.lock file at all unless the --locked option is passed. This means that installs in CI are pulling in whatever the latest versions of software that match the Cargo.toml specification. This allows any developer of any dependency to have control over what runs in CI, or on developer machines. It is the main reason to rely on the Cargo.lock file and only update the lock file intentionally.

Cargo build automatically uses the lock file and only modifies it if it needs modifying. Install is one command that unfortunately doesn't use the lock file at all.

Recently the fact that the lock file wasn't in use caused CI to break:
- #1217

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,36 +24,28 @@ ifeq ($(shell uname -s),Darwin)
 	MACOS_MIN_VER = -ldflags='-extldflags -mmacosx-version-min=13.0'
 endif
 
-# update the Cargo.lock every time the Cargo.toml changes.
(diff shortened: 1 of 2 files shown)


psheth9 (approved):
Nice !!

PR #1224, opened 2024-02-27
Fix/update init template

elizabethengelman:
### What

A companion PR to https://github.com/stellar/soroban-docs/pull/730. A few tweaks to the `init` command and template to make it a bit easier to work with.

- add testutils feature to template Cargo.toml
- allow `init` to be called from within an existing directory - previously we were looking at the passed-in directory name to use as the "name" in `package.json`. But if the user calls `soroban contract init ./` we wouldn't have a project name. This PR handles this scenario.
- Include the `.soroban/contract-ids` file in the template with a `.gitkeep`

### Why

While updating the Getting Started Guide doc, I discovered some things that could change to make `soroban contract init` a bit nicer to work with.

### Known limitations
N/A


chadoh (approved):
Cool!

PR #1228, opened 2024-02-29
Feat/init/merge readmes

elizabethengelman:
### What

For `soroban contract init` if the user passes in the `--frontend-template` flag, the init command will now merge the built-in base README with the frontend template's readme.

### Why

This is helpful so that the user has access to both of the READMEs while developing their project.

### Known limitations
N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/init.rs
+++ b/cmd/soroban-cli/src/commands/contract/init.rs
@@ -1,8 +1,8 @@
 use std::{
     env,
     ffi::OsStr,
-    fs::{copy, create_dir_all, metadata, read_dir, read_to_string, write},
-    io,
+    fs::{copy, create_dir_all, metadata, read_dir, read_to_string, write, File, OpenOptions},
+    io::{self, Read, Write},
     num::NonZeroU32,
     path::Path,
     str,
@@ -158,9 +158,6 @@ fn init(
         // clone the template repo into the temp dir
         clone_repo(frontend_template, fe_template_dir.path())?;
 
-        // IF the frontend has a README.md, append it to the project's README.md and remove it
-        // `if stat fe_template_dir.path()/README.md > /dev/null 2>&1; then cat fe_template_dir.path()/README.md >> project_path/README.md && rm fe_template_dir.path()/README.md; fi` 
-
         // copy the frontend template files into the project
         copy_frontend_files(fe_template_dir.path(), project_path)?;
     }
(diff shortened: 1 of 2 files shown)


cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -265,18 +265,11 @@ fn copy_contents(from: &Path, to: &Path) -> Result<(), Error> {
...
-                            "Error copying from {:?} to {:?}",
-                            path.to_string_lossy(),
-                            new_path
-                        );
-
-                        e
-                    })?;
-                    continue;
+                if new_path.to_string_lossy().contains(".gitignore") {
+                    append_contents(&path, &new_path)?;
```
elizabethengelman, line 269: @chadoh I changed this method a bit from when we were working on it together this morning - I moved the check to see if the path is ".gitigore" or "README.md" here instead of in the `append_contents` method itself so that we don't call that method for every file that's being copied in the loop.

elizabethengelman: Another thing that we may need to handle is that if a user calls `soroban contract init` more than once with the `-f` flag, we'll append the frontend template README contetnst more than once, resulting in duplicate info. This is likely an edge case, but something to consider. i don't have a solution in mind yet 🤔

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -440,6 +433,41 @@ fn check_internet_connection() -> bool {
...
+    let mut to_file = OpenOptions::new().read(true).append(true).open(to)?;
+    let mut to_content = String::new();
+    to_file.read_to_string(&mut to_content)?;
+
+    let delimiter = get_merged_file_delimiter(to)?;
+    // if the to file already contains the delimiter, we don't need to append the contents again
+    if to_content.contains(&delimiter) {
+        return Ok(());
+    }
+
```
elizabethengelman, line 451: This piece is so that if a user runs `soroban contract init -f <template URL>` more than once, they won't continue to append the FE template's README to the end of their projects README. Not sure if this is an edge case that is likely to happen often, so I'm happy to rethink the inclusion of this.

cmd/soroban-cli/src/commands/contract/init.rs
chadoh: I think it's enough of an edge case that we just don't worry about solving it for now? Now and then people will end up with goofy READMEs when they have goofy workflows ¯\\\_(ツ)_/¯

chadoh (approved):
Sorry for the delay in reviewing!

cmd/soroban-cli/src/commands/contract/init.rs
elizabethengelman: I forgot to update this comment thread - but I ended up solving for this edge case in `append_contents`. I am thinking I'll leave this code in and get this puppy merged! But I totally hear ya that this is an edge for people with goofy workflows, and that's probably not something to worry about.

chadoh: Oh nice! Even better then.

PR #1244, opened 2024-03-05
fetch possible example contracts at compile time

elizabethengelman:
### What

This came out of https://github.com/stellar/soroban-cli/issues/1229 - the cli was displaying an error when it could not fetch the example contracts from `soroban-examples`, every time a user ran soroban. This was because the way we had setup the clap cli help info, it would try to fetch valid example contracts everytime help was invoked.

https://github.com/stellar/soroban-cli/pull/1232 moved the examples back to being hardcoded, and this PR bakes the list of example contracts into the package - fetching the list at compile time.


cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,45 @@
 fn main() {
     crate_git_revision::init();
+    let w = &mut std::io::stdout();
+    __set_example_contracts(w).unwrap();
+}
+
+fn __set_example_contracts(w: &mut impl std::io::Write) -> std::io::Result<()> {
+    // let example_contracts = get_example_contracts().unwrap().join(",");
+    let example_contracts = get_example_contracts().unwrap().join(",");
```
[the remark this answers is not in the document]
elizabethengelman: Great idea! Are you thinking something like in [24f80d2](https://github.com/stellar/soroban-cli/pull/1244/commits/24f80d2a0c695fac4fe7dcb45540c260094b54ee)?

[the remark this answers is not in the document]
willemneal: Yep!

cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,125 @@
...
+        // Create or open the cache file
+        let mut file = File::create(cache_file_path)?;
+
+        // Write the data to the cache file
+        file.write_all(data.as_bytes())?;
+
+        Ok(())
+    }
+
+    fn file_exists(file_path: &str) -> bool {
```
willemneal, line 118:
```suggestion
    fn file_exists(file_path: &Path) -> bool {
```

cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,125 @@
...
+    pub enum Error {
+        #[error("Failed to complete get request")]
+        UreqError(#[from] Box<ureq::Error>),
+
+        #[error("Io error: {0}")]
+        IoError(#[from] std::io::Error),
+    }
+
+    fn get_example_contracts() -> Result<String, Error> {
+        if file_exists(cached_example_contracts_file_path().to_str().unwrap()) {
```
willemneal, line 44:
```suggestion
        if file_exists(&cached_example_contracts_file_path().unwrap()) {
```

cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,125 @@
...
+        let mut file = File::create(cache_file_path)?;
+
+        // Write the data to the cache file
+        file.write_all(data.as_bytes())?;
+
+        Ok(())
+    }
+
+    fn file_exists(file_path: &str) -> bool {
+        if let Ok(metadata) = metadata(file_path) {
```
willemneal, line 119: One thing to note is you can convert this function to one line (which will end up being multiple with `cargo fmt`).
```rust
    metadata(p).as_ref().map(Metadata::is_file).unwrap_or(false)
```

First `as_ref` for an Option or Result means `Result<Metadata,..>` becomes `Result<&Metadata>`, which we can then map over since `Metadata::is_file` expects a reference.

cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,124 @@
 fn main() {
     crate_git_revision::init();
+    build_helper::set_example_contracts();
+}
+
+mod build_helper {
```
elizabethengelman, line 6: I wonder if this should be pulled out into its own file, or if it's okay here for now.

cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,124 @@
...
+        // Create or open the cache file
+        let mut file = File::create(cache_file_path)?;
+
+        // Write the data to the cache file
+        file.write_all(data.as_bytes())?;
+
+        Ok(())
+    }
+
+    fn file_exists(file_path: &Path) -> bool {
```
elizabethengelman, line 118: This fn as well as `project_root` above are also used in other files, I'm thinking that a general-purpose utils mod may be helpful. I may do this in a separate PR if folks have any other functions that could be moved there.

cmd/soroban-cli/build.rs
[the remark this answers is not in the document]
chadoh: Is this worth doing? If you only run it once, then it won't use the cache. And if you run it multiple times, then a cache may result in stale data, and it'll be hard to figure out why (or even notice). Does the release process fetch it over and over, to a point where it's worth caching for a short period of time?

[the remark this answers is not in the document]
willemneal: While developing, you could have a watch, which would compile and fetch each time a file changed. In this case stale data isn't an issue as it will always happen on a `cargo install` or when built for release. If the data is stale for the local dev, I'm sure they aren't relying on it unless they are developing with it at which case they can delete the cache if needed.

cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,124 @@
...
+        metadata(file_path)
+            .as_ref()
+            .map(Metadata::is_file)
```
willemneal, line 121:
```suggestion
            .map_or(false, Metadata::is_file)
```

cmd/soroban-cli/build.rs
```diff
@@ -1,3 +1,124 @@
...
+            .unwrap_or(false)
```
willemneal, line 122:
```suggestion
```
Forgot we can make it even better ;-)

willemneal (approved):

cmd/soroban-cli/build.rs
elizabethengelman: Good question - just tested this locally and haven't noticed a compile time compared to `main`. So I'll plan to merge it in as-is, and will keep an eye out in case it becomes an issue.

Issue #1251, opened 2024-03-10
Funding accounts errors with invalid uri error


leighmcculloch:
### What version are you using?

Broken in 38d865040dd1db59104675ac7dc38a7d1017943d (main)

### What did you do?

Generated a new key that also automatically gets funded:

```
$ soroban keys generate \
    --rpc-url https://soroban-testnet.stellar.org:443 \
    --network-passphrase 'Test SDF Network ; September 2015' \
    me
```

### What did you expect to see?

Successfully funded account, like it worked in ae5446f63ca8a275e61912019199254d598f3bd5 (v2.3.1).

### What did you see instead?

```
$ soroban keys generate \
    --rpc-url https://soroban-testnet.stellar.org:443 \
    --network-passphrase 'Test SDF Network ; September 2015' \
    me
2024-03-10T00:32:26.229192Z ERROR soroban_cli::commands::network: invalid uri character
2024-03-10T00:32:26.229550Z  WARN soroban_cli::commands::keys::generate: fund_address failed: Invalid URL https://friendbot.stellar.org/
```

willemneal:
Can confirm locally! Looking into it! I changed the way that friendbot url is parsed from RPC because if using the local quickstart the friendbot url is hardcoded to use port 8000.

PR #1257, opened 2024-03-12
fix: #1251

willemneal:
This fixes a issue caused by quotes around the friendbot url caused by using Debug `{uri:?} instead of Display `{uri}`.

### DIFF
--- a/cmd/soroban-cli/src/commands/network/mod.rs
+++ b/cmd/soroban-cli/src/commands/network/mod.rs
@@ -172,7 +172,6 @@ impl Network {
         if self.network_passphrase.as_str() == LOCAL_NETWORK_PASSPHRASE {
             let auth = rpc_uri.authority().unwrap().clone();
             let scheme = rpc_uri.scheme_str().unwrap();
-            // format!("{scheme}://{auth}/friendbot");
             Ok(Uri::builder()
                 .authority(auth)
                 .scheme(scheme)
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -181,7 +180,7 @@ impl Network {
         } else {
             let client = Client::new(&self.rpc_url)?;
             let uri = client.friendbot_url().await?;
-            Uri::from_str(&format!("{uri:?}?addr={addr}")).map_err(|e| {
+            Uri::from_str(&format!("{}?addr={addr}", uri.trim_matches('"'))).map_err(|e| {
```
leighmcculloch, line 183: I think we should fix this in the client. The `friendbot_url` function shouldn't be returning a URL that is surrounded by quotes.

willemneal: Just realized that the issue was with the debug string.

cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -181,7 +180,7 @@ impl Network {
         } else {
             let client = Client::new(&self.rpc_url)?;
             let uri = client.friendbot_url().await?;
-            Uri::from_str(&format!("{uri:?}?addr={addr}")).map_err(|e| {
+            Uri::from_str(&format!("{uri}?addr={addr}")).map_err(|e| {
```
leighmcculloch, line 183: Woah. Nice one 💪🏻 .

leighmcculloch (approved):
✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅

PR #1267, opened 2024-03-20
feat: add no-build option for fee::Args

### LINKED ISSUE #1265: Separate tx build from simulate/sign/send
I'm splitting this component of #1180 into its own issue, because I think it needs to be prioritised separately.

### What
Add support for the ability to not sign and send transactions to the network and instead output their transaction XDR as base64 to stdout.

### Why
The CLI signs every transactions it submits to the network. It also has the option to simulate only.

Users who wish to build a transaction and submit it at a later date or sign it on another machine need a way to perform all the operations that the CLI supports today without doing those final two steps of signing and sending. These users need the CLI to output the transaction XDR rather than performing those steps.

### How
Add a `--build-only` option wherever the `--source` option exists.

When `--build-only` is present, commands do not simulate, sign, or send the transaction to the network. Instead the base64 encoded transaction XDR is outputted to stdout.

Stdout is intentional so that the output can be piped to other programs.

### Example
For example, today the following command will simulate a deploy, then build a transaction that performs the deploy, the sign the transaction with the source, then send it to the network.
```
$ soroban contract asset deploy \
    --asset CDE:GCYLVLL2FC7VGFHNU4MCSQ4GP3LFVK5NPSGH5HVMLHBBWJ56WOAVSIPY \
    --source me
CAFRE77725EPX4ISTZHEDVMGHASMLBM35PURHMQXRDOWWDJLSND76NS6
```

For example, with this change it will be possible to build the tx without signing and submitting:
```
$ soroban contract asset deploy \
    --asset CDE:GCYLVLL2FC7VGFHNU4MCSQ4GP3LFVK5NPSGH5HVMLHBBWJ56WOAVSIPY \
    --source me \
    --build-only
AAAAAgAAAACUdJTjk6FaYdoQ/LNQtVN9VXKdysiWyB+6ZwjPaxxxgQDfXAwACRCAAAAAAwAAAAAAAAAAAAAAAQAAAAAAAAAYAAAAAQAAAAEAAAABQ0RFAAAAAACwuq16KL9TFO2nGClDhn7WWqutfIx+nqxZwhsnvrOBWQAAAAEAAAAAAAAAAAAAAAA=
```

For example, after #1180 is also complete we'd have a composable set of commands:
```
$ soroban contract asset deploy \
    --asset CDE:GCYLVLL2FC7VGFHNU4MCSQ4GP3LFVK5NPSGH5HVMLHBBWJ56WOAVSIPY \
    --source me \
    --build-only \
  | soroban tx simulate \
  | soroban tx sign --signer me \
  | soroban tx send
CAFRE77725EPX4ISTZHEDVMGHASMLBM35PURHMQXRDOWWDJLSND76NS6
```

### Related
This issue is similar and related to #1264.

willemneal:
Fixes #1265.
Requires: #1173 

Commands that build transactions now include the following args to output and exit early:
- [x] `--build-only` prints the base64 xdr before simulation
- [x] `--sim-only` prints the base64 xdr after simulation

For example,

```
soroban contract invoke --build-only .. -- hello --world world
```
prints:
[`
AAAAAMwWyQMtAWO6nW4s+cKMRunx+kiaKfjJwIRNZEEclDvRAAAAZAAIgAEAAAANAAAAAAAAAAAAAAABAAAAAAAAABgAAAAAAAAAAQ3DO777M8sdGtPbMau+cuxAles4UILfUkZXOWTsqKvOAAAABWhlbGxvAAAAAAAAAQAAAA8AAAAFd29ybGQAAAAAAAAAAAAAAA==
`](https://laboratory.stellar.org/#xdr-viewer?input=AAAAAMwWyQMtAWO6nW4s+cKMRunx+kiaKfjJwIRNZEEclDvRAAAAZAAIgAEAAAANAAAAAAAAAAAAAAABAAAAAAAAABgAAAAAAAAAAQ3DO777M8sdGtPbMau+cuxAles4UILfUkZXOWTsqKvOAAAABWhlbGxvAAAAAAAAAQAAAA8AAAAFd29ybGQAAAAAAAAAAAAAAA==&type=Transaction&network=test)

```
soroban contract invoke --sim-only .. -- hello --world world
```

[`AAAAAMwWyQMtAWO6nW4s%2BcKMRunx%2BkiaKfjJwIRNZEEclDvRAAD4WAAIgAEAAAANAAAAAAAAAAAAAAABAAAAAAAAABgAAAAAAAAAAQ3DO777M8sdGtPbMau%2BcuxAles4UILfUkZXOWTsqKvOAAAABWhlbGxvAAAAAAAAAQAAAA8AAAAFd29ybGQAAAAAAAAAAAAAAQAAAAAAAAACAAAABgAAAAENwzu%2B%2BzPLHRrT2zGrvnLsQJXrOFCC31JGVzlk7KirzgAAABQAAAABAAAAB3mMSQSdZWZNBqeYQchZLRMLUhSFaxsEJAOkX%2Bel%2B%2BMzAAAAAABcQYMAABsoAAAAAAAAAAAAANeQ
`](https://laboratory.stellar.org/#xdr-viewer?input=AAAAAMwWyQMtAWO6nW4s%2BcKMRunx%2BkiaKfjJwIRNZEEclDvRAAD4WAAIgAEAAAANAAAAAAAAAAAAAAABAAAAAAAAABgAAAAAAAAAAQ3DO777M8sdGtPbMau%2BcuxAles4UILfUkZXOWTsqKvOAAAABWhlbGxvAAAAAAAAAQAAAA8AAAAFd29ybGQAAAAAAAAAAAAAAQAAAAAAAAACAAAABgAAAAENwzu%2B%2BzPLHRrT2zGrvnLsQJXrOFCC31JGVzlk7KirzgAAABQAAAABAAAAB3mMSQSdZWZNBqeYQchZLRMLUhSFaxsEJAOkX%2Bel%2B%2BMzAAAAAABcQYMAABsoAAAAAAAAAAAAANeQ&type=Transaction&network=test)

# Transaction result

The NetworkRunable trait a result type of `TxnResult`, either a raw xdr string or a T. This allows returning early.

## update `is-view`
Ensure that `is-view` doesn't fetch account info as a source account doesn't need to exist for a simulation.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
@@ -14,7 +14,9 @@ use std::{array::TryFromSliceError, fmt::Debug, num::ParseIntError};
 use crate::{
     commands::{
         config::{self, data},
-        global, network, NetworkRunnable,
+        global, network,
+        txn_result::TxnResult,
+        NetworkRunnable,
     },
     rpc::{Client, Error as SorobanRpcError},
     utils::{contract_id_hash_from_asset, parsing::parse_asset},
@@ -80,7 +82,7 @@ impl NetworkRunnable for Cmd {
         &self,
         args: Option<&global::Args>,
         config: Option<&config::Args>,
-    ) -> Result<String, Error> {
+    ) -> Result<TxnResult<String>, Error> {
         let config = config.unwrap_or(&self.config);
         // Parse asset
         let asset = parse_asset(&self.asset)?;
@@ -108,8 +110,11 @@ impl NetworkRunnable for Cmd {
             network_passphrase,
             &key,
         )?;
+        if self.fee.build_only {
+            return Ok(TxnResult::from_xdr(&tx)?);
+        }
         let txn = client.create_assembled_transaction(&tx).await?;
-        let txn = self.fee.apply_to_assembled_txn(txn);
+        let txn = self.fee.apply_to_assembled_txn(txn)?;
         let get_txn_resp = client
             .send_assembled_transaction(txn, &key, &[], network_passphrase, None, None)
             .await?
@@ -118,7 +123,9 @@ impl NetworkRunnable for Cmd {
             data::write(get_txn_resp, &network.rpc_uri()?)?;
         }
 
-        Ok(stellar_strkey::Contract(contract_id.0).to_string())
+        Ok(TxnResult::Xdr(
+            stellar_strkey::Contract(contract_id.0).to_string(),
+        ))
     }
 }
 
--- a/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
@@ -114,11 +114,10 @@ impl NetworkRunnable for Cmd {
             return Ok(TxnResult::Txn(tx));
         }
         let txn = client.create_assembled_transaction(&tx).await?;
-        let txn = self.fee.apply_to_assembled_txn(txn)?;
-        let txn = match txn {
-            TxnResult::Txn(raw) => return Ok(TxnResult::Txn(raw)),
-            TxnResult::Res(txn) => txn,
-        };
+        let txn = self.fee.apply_to_assembled_txn(txn);
+        if self.fee.sim_only {
+            return Ok(TxnResult::Txn(txn.transaction().clone()));
+        }
         let get_txn_resp = client
             .send_assembled_transaction(txn, &key, &[], network_passphrase, None, None)
             .await?
--- a/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
@@ -76,13 +76,13 @@ impl Cmd {
 #[async_trait::async_trait]
 impl NetworkRunnable for Cmd {
     type Error = Error;
-    type Result = String;
+    type Result = stellar_strkey::Contract;
 
     async fn run_against_rpc_server(
         &self,
         args: Option<&global::Args>,
         config: Option<&config::Args>,
-    ) -> Result<TxnResult<String>, Error> {
+    ) -> Result<TxnResult<Self::Result>, Error> {
         let config = config.unwrap_or(&self.config);
         // Parse asset
         let asset = parse_asset(&self.asset)?;
@@ -115,6 +115,10 @@ impl NetworkRunnable for Cmd {
         }
         let txn = client.create_assembled_transaction(&tx).await?;
         let txn = self.fee.apply_to_assembled_txn(txn)?;
+        let txn = match txn {
+            TxnResult::Xdr(raw) => return Ok(TxnResult::Xdr(raw)),
+            TxnResult::Res(txn) => txn,
+        };
         let get_txn_resp = client
             .send_assembled_transaction(txn, &key, &[], network_passphrase, None, None)
             .await?
@@ -123,9 +127,7 @@ impl NetworkRunnable for Cmd {
             data::write(get_txn_resp, &network.rpc_uri()?)?;
         }
 
-        Ok(TxnResult::Xdr(
-            stellar_strkey::Contract(contract_id.0).to_string(),
-        ))
+        Ok(TxnResult::Res(stellar_strkey::Contract(contract_id.0)))
     }
 }
 
--- a/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
@@ -76,13 +76,13 @@ impl Cmd {
 #[async_trait::async_trait]
 impl NetworkRunnable for Cmd {
     type Error = Error;
-    type Result = stellar_strkey::Contract;
+    type Result = TxnResult<stellar_strkey::Contract>;
 
     async fn run_against_rpc_server(
         &self,
         args: Option<&global::Args>,
         config: Option<&config::Args>,
-    ) -> Result<TxnResult<Self::Result>, Error> {
+    ) -> Result<Self::Result, Error> {
         let config = config.unwrap_or(&self.config);
         // Parse asset
         let asset = parse_asset(&self.asset)?;
@@ -111,12 +111,12 @@ impl NetworkRunnable for Cmd {
             &key,
         )?;
         if self.fee.build_only {
-            return Ok(TxnResult::from_xdr(&tx)?);
+            return Ok(TxnResult::Txn(tx));
         }
         let txn = client.create_assembled_transaction(&tx).await?;
         let txn = self.fee.apply_to_assembled_txn(txn)?;
         let txn = match txn {
-            TxnResult::Xdr(raw) => return Ok(TxnResult::Xdr(raw)),
+            TxnResult::Txn(raw) => return Ok(TxnResult::Txn(raw)),
             TxnResult::Res(txn) => txn,
         };
         let get_txn_resp = client
--- a/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
@@ -184,11 +184,10 @@ impl NetworkRunnable for Cmd {
         }
 
         let txn = client.create_assembled_transaction(&txn).await?;
-        let txn = self.fee.apply_to_assembled_txn(txn)?;
-        let txn = match txn {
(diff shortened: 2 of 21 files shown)


cmd/soroban-cli/src/fee.rs
```diff
@@ -16,15 +17,39 @@ pub struct Args {
     /// Number of instructions to simulate
     #[arg(long, help_heading = HEADING_RPC)]
     pub instructions: Option<u32>,
+    /// Build the transaction only write the base64 xdr to stdout
+    #[arg(long, help_heading = HEADING_RPC)]
+    pub build_only: bool,
+    /// Simulation the transaction only write the base64 to stdout
+    #[arg(long, help_heading = HEADING_RPC, conflicts_with = "build_only")]
+    pub sim_only: bool,
```
leighmcculloch, line 25: Why do the new options get attached to the fee instead of to the `commands::config::Args` type which is where the --source-account is attached today? These options feel more similar to source account than fees. Fees could also arguably live in the same place as source account, so maybe this is an area that would benefit of a refactor.

cmd/soroban-cli/src/fee.rs
```diff
@@ -16,15 +17,39 @@ pub struct Args {
...
 }
 
 impl Args {
-    pub fn apply_to_assembled_txn(&self, txn: Assembled) -> Assembled {
-        if let Some(instructions) = self.instructions {
+    pub fn exit_if_build_only(&self, txn: &xdr::Transaction) -> Result<(), xdr::Error> {
+        if self.build_only {
+            println!("{}", txn.to_xdr_base64(xdr::Limits::none())?);
+            std::process::exit(0);
```
leighmcculloch, line 32: Deep exits can be hard to track in code over time, is there a reason we can't `if build_only { return Ok(()) }` or similar in callers? It'd result in code flow in callers that's more obvious.

leighmcculloch (commented):
That was fast! This is great 👏🏻.

Couple questions inline.

The `--sim-only` might be unnecessary, given that folks could do `soroban contract deploy --build-only | soroban tx simulate` in the future, and that leans on command composability and keeps the options simpler. Thoughts from others on this? cc @fnando @chadoh

cmd/soroban-cli/src/fee.rs
leighmcculloch: I missed the comment here https://github.com/stellar/soroban-cli/issues/1265#issuecomment-2009578081, makes sense, I'll continue that conversation over there.

leighmcculloch:
Will take another look once #1173 is merged, but this looked good when I last reviewed other than the couple minor things which are addressed so I think we'll be able to fast merge this once the other is on its way.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -115,18 +119,20 @@ impl NetworkRunnable for Cmd {
         &self,
         global_args: Option<&global::Args>,
         config: Option<&config::Args>,
-    ) -> Result<String, Error> {
+    ) -> Result<TxnResult<String>, Error> {
         let config = config.unwrap_or(&self.config);
         let wasm_hash = if let Some(wasm) = &self.wasm {
+            let mut fee = self.fee.clone();
+            fee.build_only = false;
```
leighmcculloch, line 126: This results in a tx being sent to the network, to do the install, when --build-only is passed which would be surprising for someone wanting to only produce a tx. When building only we should probably skip install and build just the deployment transaction. Then they can produce the install and deploy txs independently with the two commands. Wdyt?

cmd/soroban-cli/src/commands/contract/deploy/asset.rs
```diff
@@ -118,7 +123,9 @@ impl NetworkRunnable for Cmd {
             data::write(get_txn_resp, &network.rpc_uri()?)?;
         }
 
-        Ok(stellar_strkey::Contract(contract_id.0).to_string())
+        Ok(TxnResult::Xdr(
+            stellar_strkey::Contract(contract_id.0).to_string(),
```
leighmcculloch, line 127: Why is the contract address being wrapped in a TxnResult::Xdr? The fact that strings are convertible to XDR seems a bit odd for this use case. Contract addresses are not normally encoded into XDR.

cmd/soroban-cli/src/fee.rs
```diff
@@ -16,15 +17,31 @@ pub struct Args {
...
             add_padding_to_instructions(txn)
+        };
+        if self.sim_only {
+            println!(
+                "{}",
+                simulated_txn
+                    .transaction()
+                    .to_xdr_base64(xdr::Limits::none())?
+            );
+            std::process::exit(0);
```
leighmcculloch, line 42: It's difficult to follow the application flow because of this exit buried in the code. And the code path of the --build-only option doesn't result in this. Is this something that's too difficult to change?

leighmcculloch (commented):
Few questions inline. Looks like this isn't far off.

cmd/soroban-cli/src/fee.rs
willemneal: Nope! This should use the same `TxnResult` type since it has the notion of item vs base64 xdr. Will update!

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
willemneal: Yeah it would be nice to have these as two operations in the same transaction, but I agree I've disliked the install step because it is an unexpected side effect.

cmd/soroban-cli/src/commands/contract/deploy/asset.rs
willemneal: This should be a `TxnResult::Res` as this should return a strkey and you're right it shouldn't be encoded as a String.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
leighmcculloch: I think we can skip the install when deploy is build only. We don't actually need anything from the install. The hash is all the deploy needs and it can be calculated with a sha256 of the binary.

cmd/soroban-cli/src/commands/txn_result.rs
```diff
@@ -0,0 +1,34 @@
+use std::fmt::{Display, Formatter};
+
+use soroban_env_host::xdr::{Limits, Transaction, WriteXdr};
+
+pub enum TxnResult<R> {
+    Txn(Transaction),
```
leighmcculloch, line 6: I pushed a change changing the `Xdr(String)` variant to a `Txn(Transaction)` variant. This is so that we don't have functions that return values of String or u32, or String or String, which is pretty confusing, and was I think the source of one bug (https://github.com/stellar/soroban-cli/pull/1267#discussion_r1591957598) that got fixed.

We don't need to eagerly convert the Transactions to XDR base64 strings, and that defers the conversion to string to the point where the code that cares most about the presentation layer can do so. It results in a little duplicated to_xdr_base64 code, but that seems really immaterial.

Also on its own the clarity gained as a reader feels like a worthwhile tradeoff.

cmd/soroban-cli/src/commands/contract/deploy/asset.rs
```diff
@@ -108,8 +110,14 @@ impl NetworkRunnable for Cmd {
...
             &key,
         )?;
+        if self.fee.build_only {
+            return Ok(TxnResult::Txn(tx));
+        }
         let txn = client.create_assembled_transaction(&tx).await?;
         let txn = self.fee.apply_to_assembled_txn(txn);
+        if self.fee.sim_only {
+            return Ok(TxnResult::Txn(txn.transaction().clone()));
+        }
```
leighmcculloch, line 120: I pushed a change moving the `if sim_only` check to the same level as the `if build_only` check. If you look above at the few lines captured on this comment, the sameness and likeness of the two (I think) make it really easy to understand the story they form.

The other reason I suggest this change is because the `apply_to_assembled_txn` was ambiguous when it returned a `Transaction`/XDR String or an `Assembled`. It's difficult to know what state that tx is in, and why it's being passed back separately from the Assembled, given that the Assembled actually contains the Transaction anyway. This way we let each command on its own control their flow.

leighmcculloch (approved):
I pushed a couple changes to the PR in the last 3 commits, and I left comments inline to outline why. Please take a look and let me know if you agree with them.

leighmcculloch:
@willemneal If you're good with the changes I pushed, can you resolve the unresolved conversations which are just the notes I left explaining why each change, and once they're all resolved that should trigger GitHub to auto-merge the PR.

If possible don't force push the branch in case we need to make more changes.

leighmcculloch:
I pushed a bug fix in 457a54b for how sim only works with install. I left a comment in the code about it for more information.

PR #1273, opened 2024-03-25
fix: clippy in ft tests

willemneal:
### What

Fix issue of a function over 100 lines.

### Why

Make clippy happy

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/crates/soroban-test/tests/fixtures/test-wasms/token/src/test.rs
+++ b/cmd/crates/soroban-test/tests/fixtures/test-wasms/token/src/test.rs
@@ -14,6 +14,7 @@ fn create_token<'a>(e: &Env, admin: &Address) -> TokenClient<'a> {
     token
 }
 
+#[allow(clippy::too_many_lines)]
 #[test]
 fn test() {
     let e = Env::default();


cmd/crates/soroban-test/tests/fixtures/test-wasms/token/src/test.rs
```diff
@@ -14,183 +13,154 @@ fn create_token<'a>(e: &Env, admin: &Address) -> TokenClient<'a> {
...
             admin1.clone(),
-            AuthorizedInvocation {
-                function: AuthorizedFunction::Contract((
-                    token.address.clone(),
-                    symbol_short!("mint"),
-                    (&user1, 1000_i128).into_val(&e),
-                )),
-                sub_invocations: std::vec![]
-            }
+            auth_invoke(e, token, "mint", (user1, 1000_i128).into_val(e))
```
leighmcculloch, line 45: I think this code would be clearer if you kept the type and just added a `#[allow(clippy::...)]`.

It feels rather arbitrary to move this block out into a function, especially on an arbitrary number of lines.

cmd/crates/soroban-test/tests/fixtures/test-wasms/token/src/test.rs
```diff
@@ -14,6 +14,7 @@ fn create_token<'a>(e: &Env, admin: &Address) -> TokenClient<'a> {
     token
 }
 
+#[allow(clippy::too_many_lines)]
```
leighmcculloch, line 17: 💯

leighmcculloch (approved):
💯

PR #1277, opened 2024-04-01
Update Cargo.toml template.

fnando:
### What

Update Cargo.toml template to use generic values.

### Why

So it's generic and not tied to SDF. Also update the contract name to match what's currently in our docs.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/utils/contract-init-template/contracts/hello_world/Cargo.toml.removeextension
+++ b/cmd/soroban-cli/src/utils/contract-init-template/contracts/hello_world/Cargo.toml.removeextension
@@ -14,5 +14,5 @@ doctest = false
 [dependencies]
 soroban-sdk = { workspace = true }
 
-[dev_dependencies]
+[dev-dependencies]
 soroban-sdk = { workspace = true, features = ["testutils"] }
--- a/cmd/soroban-cli/src/utils/contract-init-template/contracts/hello_world/Cargo.toml.removeextension
+++ b/cmd/soroban-cli/src/utils/contract-init-template/contracts/hello_world/Cargo.toml.removeextension
@@ -1,10 +1,7 @@
 [package]
 name = "hello-world"
 version = "0.0.0"
-authors = ["John Doe <john@example.com>"]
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/utils/contract-init-template/contracts/hello_world/Cargo.toml.removeextension
```diff
@@ -15,4 +15,4 @@ doctest = false
 soroban-sdk = { workspace = true }
 
 [dev_dependencies]
```
willemneal, line 17: Since we are here, this should also be updated. `cargo add --dev ...` will error if there is an `_`.
```suggestion
[dev-dependencies]
```

willemneal (approved):
Just one suggestion to have the crate work well with `cargo add`

cmd/soroban-cli/src/utils/contract-init-template/contracts/hello_world/Cargo.toml.removeextension
```diff
@@ -1,7 +1,7 @@
 [package]
-name = "soroban-hello-world-contract"
+name = "hello-world"
 version = "0.0.0"
-authors = ["Stellar Development Foundation <info@stellar.org>"]
+authors = ["John Doe <john@example.com>"]
```
leighmcculloch, line 4: Suggest changing the package part of the file to only the following which is more consistent with what `cargo init` does:
```
[package]
name = "hello-world"
version = "0.0.0"
edition = "2021"
publish = false
```
This means dropping `authors`, `license`, `rust-version`.

leighmcculloch (commented):
One suggestion mentioned inline, +1 the suggestion @willemneal mentioned too.

cmd/soroban-cli/src/utils/contract-init-template/contracts/hello_world/Cargo.toml.removeextension
fnando: done!

leighmcculloch (approved):

Issue #1264, opened 2024-03-20
Add support for building transactions with a different signer to the source account


leighmcculloch:
### What
The CLI needs to support for adding a different signer to a transaction without the signer being the source account.

### Why
The CLI today assumes that the signer of a transaction is the master key of the source account. This is fine for development and for simpler user key setups, but doesn't support slightly more complex setups or where people have their key stored on a hardware device.

Note that there are even more complex setups that might require signatures by multiple keys, or multiple parties and this issue isn't intended to address those more complex scenarios. Those scenarios will be address by the addition of the `tx sign` command in:
- https://github.com/stellar/stellar-cli/issues/1490

### How
Add a new option `--sign-with-*` that is supported anywhere that `--source` is supported.

When a `--sign-with-*` is not present, the `--source` is the signer.

When a `--sign-with-*` is present, the `--source` is not a signer, the signer is looked up separately and signs the transaction.

Initially this issue captures the need for `--sign-with-key` which can accept any of the key types currently passed to `--source`.

### Examples
For example, today the CLI supports, and it would continue to support:
```
soroban contract deploy --wasm ... --source me
```

And for example, a user should be able to sign the tx with a different key than the source:
```
soroban contract deploy --wasm ... --source me --sign-with-key myotherkey
```

This change makes the first example in the issue above shorthand for:
```
soroban contract deploy --wasm ... --source me --sign-with-key me
```

tomerweller:
What if the user doesn't want to sign with _any_ key? they just want to go through the build->simulate flow and then take the tx XDR and sign elsewhere?

leighmcculloch:
I think we should capture that in a separate issue. I'm hoping once #1180 gets broken up into two issues that the issue that comes out of that for skipping signing on any tx built will involve adding a `--no-sign` or `--sign false`.

leighmcculloch:
@tomerweller A separate issue now captures supporting that scenario: https://github.com/stellar/soroban-cli/issues/1265

leighmcculloch:
After reading through https://github.com/stellar/stellar-cli/issues/1490 I think we can probably keep this the same as `tx sign` and only ever sign by one device/key at a time, with more complex multi-signature flows using the `tx sign` command multiple times.

I'm updating this issue to remove the intent to make for multiple signers, but keeping the issue open because I think from a usability pov it's important we support signing with other devices natively on commands so that users can run commands like:
```
soroban contract deploy --wasm ... --source me --sign-with-key mykey
[...]

PR #1283, opened 2024-04-15
feat: txn command to simulate

willemneal:
Part of https://github.com/stellar/soroban-cli/issues/1265
Adds 

- `txn simulate` - simulate and assemble transaction

Future PR:
- `signer::Stellar` trait and a default `InMemory` implementation needed for `txn sign`
- `txn sign` - sign a given transaction
- `txn inspect`
- `txn send` - sends a signed transaction to the PRC (not useful without sign)

Left
- [x] Tests


cmd/soroban-cli/src/commands/tx/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+pub mod sign;
+pub mod simulate;
+pub mod xdr;
+
+use stellar_xdr::cli as xdr_cli;
+
+#[derive(Debug, Parser)]
+pub enum Cmd {
+    /// Add a new identity (keypair, ledger, macOS keychain)
+    Inspect(xdr_cli::Root),
```
leighmcculloch, line 13: It's a bit odd that `tx inspect` works like this: `tx inspect decode --type TransactionEnvelope`.

I think we should replace it with `tx decode` and `tx encode` where the type parameter is already filled in.

See this issue that details the commands:
- https://github.com/stellar/soroban-cli/issues/1180

cmd/soroban-cli/src/commands/tx/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Given an identity return its address (public key)
+    Sign(sign::Cmd),
```
leighmcculloch, line 15: The doc comments on these two commands look like a copy-paste from somewhere else and don't make much sense in this context.

cmd/soroban-cli/src/commands/tx/send.rs
```diff
@@ -0,0 +1,35 @@
...
+    #[clap(flatten)]
+    pub xdr_args: super::xdr::Args,
+    #[clap(flatten)]
+    pub config: super::super::config::Args,
+}
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
+        let response = self.send().await?;
+        println!("{response:#?}");
```
leighmcculloch, line 25: If we're going to output the response we should probably output it in JSON rather than the rust debug format, so it is at least usable.

I'm also wondering what's the right thing to return from here. The response makes sense I guess, it actually contains a lot of data that could be useful when scripting.

cmd/soroban-cli/src/commands/tx/sign.rs
```diff
@@ -0,0 +1,97 @@
...
+    UserCancelledSigning,
+}
+
+#[derive(Debug, clap::Parser, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// Confirm that a signature can be signed by the given keypair automatically.
+    #[arg(long, short = 'y')]
+    yes: bool,
+    #[clap(flatten)]
```
leighmcculloch, line 36: We should leave out the yes, and require the user to specifier the list of signers who should sign, using the same format as expressed here:
- https://github.com/stellar/soroban-cli/issues/1264

And then address interactivity at a later time holistically.

For starters in this PR we could allow just one signer to begin with, and defer to #1264 adding support for multiple.

cmd/soroban-cli/src/commands/tx/sign.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub xdr_args: super::xdr::Args,
+    #[clap(flatten)]
+    pub config: super::super::config::Args,
+}
+
+impl Cmd {
+    #[allow(clippy::unused_async)]
+    pub async fn run(&self) -> Result<(), Error> {
+        let envelope = self.sign()?;
+        println!("{}", envelope.to_xdr_base64(Limits::none())?.trim());
```
leighmcculloch, line 46: What is the `.trim()` needed for?

cmd/soroban-cli/src/commands/tx/xdr.rs
```diff
@@ -0,0 +1,67 @@
...
+        conflicts_with = "xdr_file"
+    )]
+    pub xdr_base64: Option<String>,
+    //// File containing Binary encoded data
+    #[arg(
+        long = "xdr-file",
+        env = "STELLAR_TXN_XDR_FILE",
+        conflicts_with = "xdr_base64"
+    )]
+    pub xdr_file: Option<PathBuf>,
```
leighmcculloch, line 38: Can we replace these xdr inputs with reading from stdin only, for consistency with the xdr commands? Would rather keep the interface smaller, and then look at improving the interface for passing XDR around separately and holistically across all the XDR commands to make sure what we do works everywhere and is consistent. The API for parameter passing is likely to be different to this to be consistent with the how the XDR subcommand does format selection.

cmd/soroban-cli/src/signer.rs
```diff
@@ -0,0 +1,277 @@
...
+            network_id: self.network_hash(),
+            tagged_transaction: TransactionSignaturePayloadTaggedTransaction::Tx(txn.clone()),
+        };
+        let hash = Sha256::digest(signature_payload.to_xdr(Limits::none())?).into();
+        let decorated_signature = self.sign_txn_hash(hash, source_account)?;
+        Ok(TransactionEnvelope::Tx(TransactionV1Envelope {
+            tx: txn,
+            signatures: vec![decorated_signature].try_into()?,
+        }))
+    }
```
leighmcculloch, line 84: The sign function assumes the tx hasn't been wrapped in a fee bump transaction, but ideally it would support all Stellar txs.

leighmcculloch (commented):
I pushed some commits getting main merged in.

This PR looks like it's still a little ways away from being ready. Left some inline comments. 

Would be a good idea to add tests for these commands too, from the outside-in.

leighmcculloch:
Given the sensitivity of the sign command, I'd like to shift that work into a separate PR, so that we can get the other commands merged sooner, then follow up and take our time with the sign command. cc @janewang

cmd/soroban-cli/src/commands/tx/xdr.rs
```diff
@@ -0,0 +1,67 @@
...
+                let _ = stdin()
+                    .read_to_string(&mut buf)
+                    .map_err(|_| Error::StdinDecode)?;
+                T::from_xdr_base64(buf.trim(), Limits::none()).map_err(|_| Error::StdinDecode)
+            }
+        }
+    }
+
+    pub fn txn(&self) -> Result<Transaction, Error> {
+        self.xdr::<Transaction>()
```
leighmcculloch, line 61: It should always be TransactionEnvelope's that we pass around, not Transactions on their own.

cmd/soroban-cli/src/commands/tx/sign.rs
willemneal: Had issues with piping to a send command. But perhaps it should be on the input.

cmd/soroban-cli/src/commands/tx/mod.rs
willemneal: `decode::Cmd` is not [public](https://github.com/stellar/rs-stellar-xdr/blob/efe941d4f0dc55d555273ff7da2b46626f7f985b/src/cli/mod.rs#L1) so I went with root, which is as you point out less than optimal.

cmd/soroban-cli/src/commands/tx/xdr.rs
willemneal: That's okay with me. `type` and `cat` can work for now. And then when we can properly build a txn envolope it will be the best way to start the piping.

cmd/soroban-cli/src/commands/tx/mod.rs
leighmcculloch: Making it public here:
- https://github.com/stellar/rs-stellar-xdr/pull/369

We could remove the inspect/encode/decode from this PR and revisit it once that is merged.

willemneal:
Friend bot seems to be failing.

cmd/crates/soroban-test/tests/it/integration/tx.rs
```diff
@@ -0,0 +1,40 @@
...
+use soroban_test::{AssertExt, TestEnv};
+
+use crate::integration::util::{deploy_contract, DeployKind, HELLO_WORLD};
+
+#[tokio::test]
+async fn txn_simulate() {
+    let sandbox = &TestEnv::new();
+    let xdr_base64_build_only = deploy_contract(sandbox, HELLO_WORLD, DeployKind::BuildOnly).await;
+    let xdr_base64_sim_only = deploy_contract(sandbox, HELLO_WORLD, DeployKind::SimOnly).await;
+    println!("{xdr_base64_build_only}");
```
chadoh, line 14: Should we remove this?

cmd/crates/soroban-test/tests/it/integration/tx.rs
```diff
@@ -0,0 +1,40 @@
...
+    };
+    let assembled = cmd.simulate(tx, &sandbox.client()).await.unwrap();
+    let assembled_str = sandbox
+        .new_assert_cmd("tx")
+        .arg("simulate")
+        .write_stdin(xdr_base64_build_only.as_bytes())
+        .assert()
+        .success()
+        .stdout_as_str();
+    println!("{assembled_str}");
```
chadoh, line 29: And this?

chadoh (commented):
Nice clean code! I really like the tests. I have some questions, before I approve.

cmd/crates/soroban-test/tests/it/integration/tx.rs
```diff
@@ -0,0 +1,40 @@
...
+    assert_eq!(xdr_base64_sim_only, assembled_str);
+    let txn_env = TransactionEnvelope::Tx(TransactionV1Envelope {
+        tx: assembled.transaction().clone(),
+        signatures: VecM::default(),
+    });
+
+    assert_eq!(
+        txn_env.to_xdr_base64(Limits::none()).unwrap(),
+        assembled_str
```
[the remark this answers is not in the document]
chadoh: Having looked at the code itself, I think this is more a question about where the logic in `run` happens. 

```rs
TransactionEnvelope::Tx(TransactionV1Envelope {
    tx,
    signatures: VecM::default()
})
.to_xdr_base64(xdr::Limits::none())?
```

Why do this in `run`, but not in `run_against_rpc_server`? Why not in `simulate`? 

You might have a great reason for this, but coming in fresh, all of these functions seem like they should do the same thing.

cmd/crates/soroban-test/tests/it/integration/tx.rs
willemneal: It's only printed if it fails, but it was just for debugging.

cmd/soroban-cli/src/commands/tx/mod.rs
```diff
@@ -0,0 +1,28 @@
+use clap::Parser;
+
+use super::global;
+
+pub mod simulate;
+pub mod xdr;
```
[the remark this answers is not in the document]
willemneal: It was pending a release of rs-stellar-xdr. For the sake of closing this I say we do it in a follow up PR. I'm pretty sure it's released now.

cmd/soroban-cli/src/commands/tx/simulate.rs
```diff
@@ -0,0 +1,70 @@
...
+                tx,
+                signatures: VecM::default()
+            })
+            .to_xdr_base64(xdr::Limits::none())?
+        );
+        Ok(())
+    }
+
+    pub async fn simulate(
+        &self,
```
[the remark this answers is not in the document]
willemneal: Very true!

cmd/soroban-cli/src/commands/tx/mod.rs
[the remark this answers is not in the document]
chadoh: oh, you mean the file itself? I just saw that we have a `tx/inspect.rs` that's not included in the module. But yeah, up to you!

cmd/crates/soroban-test/tests/it/integration/tx.rs
```diff
@@ -0,0 +1,40 @@
...
+async fn txn_simulate() {
+    let sandbox = &TestEnv::new();
+    let xdr_base64_build_only = deploy_contract(sandbox, HELLO_WORLD, DeployKind::BuildOnly).await;
+    let xdr_base64_sim_only = deploy_contract(sandbox, HELLO_WORLD, DeployKind::SimOnly).await;
+    println!("{xdr_base64_build_only}");
+    let cmd = tx::simulate::Cmd::default();
+    let tx_env =
+        TransactionEnvelope::from_xdr_base64(&xdr_base64_build_only, Limits::none()).unwrap();
+    let TransactionEnvelope::Tx(TransactionV1Envelope { tx, .. }) = &tx_env else {
+        panic!("Only transaction v1 is supported")
```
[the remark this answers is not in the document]
willemneal: I should! But it currently expects stdin. I'll refactor so that you can pass it instead of assuming stdin.

cmd/crates/soroban-test/tests/it/integration/tx.rs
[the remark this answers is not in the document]
willemneal: My thought was that as a library after a simulation you probably wanted an assembled transaction. However, as the output from the CLI you want a transaction envelope to keep the expected type consistent across all `tx` commands.

cmd/soroban-cli/src/commands/tx/mod.rs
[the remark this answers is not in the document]
willemneal: Yeah that should be removed.

cmd/soroban-cli/src/commands/tx/simulate.rs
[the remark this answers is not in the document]
willemneal: removed it! Was a bit silly since it was just one line.

chadoh (approved):
Looks great, thanks!

PR #1284, opened 2024-04-16
fix: depend on a real, deployed RPC version

chadoh:
### What

Roll back RPC dependency

### Why

`main` and the latest released version of the CLI are totally broken
because the 3.3.5 release of the RPC was yanked. While we are waiting
for that yank to be rolled back (it was related to a bug in simulation,
which does not introduce security vulnerabilities and which did not need
to be yanked), rolling back the RPC version _may_ work well enough. I
say "may" because I don't really know if there are other dependencies on
the latest RPC here. Let's see if CI passes, i guess?

### Known limitations

Unknown!


leighmcculloch:
The RPC dep has been unyanked so we shouldn't need to do this now. Please reopen if I'm misunderstanding.

PR #1292, opened 2024-04-25
feat!: update to protocol 21

willemneal:


willemneal:
@leighmcculloch I remember we discussed something similar, but perhaps we should have different branches that would act as a beta/nightly for testnet and futurenet?

leighmcculloch:
> Are we going to need to create some sort of logic for installing futurenet, testnet and mainnet enabled cli versions or branches?

If you need a version of the CLI that only works on protocol 20, you can install v20 of the CLI. The XDR will always be binary backwards compatible from v21 to v20 so all v20 XDRs should be decodable by v21.

```
cargo install --locked --version '^20' soroban-cli
```

I think in the future we can add a version modifier and support multiple versions installed or old behavior, such as:

```
soroban +20 contract invoke
soroban +21 contract invoke
```

But that functionality doesn't exist today, and because of the complexity to add something like that we need more information about its usefulness.

Cargo.toml
```diff
@@ -38,29 +38,29 @@ version = "20.3.4"
...
 
 [workspace.dependencies.stellar-xdr]
-version = "=20.1.0"
+version = "=21.0.1"
 default-features = true
 
 [workspace.dependencies]
 stellar-strkey = "0.0.8"
-sep5 = "0.0.2"
+sep5 = "0.0.3"
```
leighmcculloch, line 63: Reviewed this version bump and seems like a no-op as the only difference between sep5 0.0.2 and 0.0.3 is the change in dependence on stellar-strkey 0.0.7 to 0.0.8.

leighmcculloch (approved):

Cargo.toml
willemneal: Yeah it was causing issues because the version differences led to different types, but I had fixed this by going back to the raw type instead of returning the strkey type directly.

PR #1296, opened 2024-04-26
fix(init): don't add two `dev_dependencies` sections

chadoh:
When initializing a new project with `soroban contract init`, we need to
clean up the `Cargo.toml` of example projects a little. We need to make
sure `soroban-sdk` is always inherited from the workspace, and we need
to make sure there's no `profile` section.

We had a couple issues with the old logic.

1. It was adding a `dev_dependencies` section, with an underscore, which
   is deprecated. If the Cargo.toml already had a `dev-dependencies`
   section, it would add a second `dev_dependencies`, which cargo would
   complain about.
2. The new setting would only set `{ workspace = true }`, even if the
   old one included other details like `features = ['testutils']`.

This adds a new crate, `cargo_toml`, which allows us to parse and
manipulate Cargo.toml files in a more sophisticated way, which allows us
to make sure we get the `dev-dependencies` section even if it's called
`dev_dependencies`. It also makes it straightforward to add the correct
inherited dependency.

### Known limitations

The format of the new `Cargo.toml` is a little unconventional:

```toml
[package]
name = "soroban-account-contract"
edition = "2021"
version = "0.0.0"
publish = false
[dependencies.soroban-sdk]
workspace = true
[dev-dependencies.ed25519-dalek]
version = "1.0.1"

[dev-dependencies.rand]
version = "0.7.3"

[dev-dependencies.soroban-sdk]
features = ["testutils"]
workspace = true
```

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/init.rs
+++ b/cmd/soroban-cli/src/commands/contract/init.rs
@@ -1,3 +1,11 @@
+use cargo_toml;
+use clap::{
+    builder::{PossibleValue, PossibleValuesParser, ValueParser},
+    Parser, ValueEnum,
+};
+use gix::{clone, create, open, progress, remote};
+use rust_embed::RustEmbed;
+use serde_json::{from_str, json, to_string_pretty, Error as JsonError, Value as JsonValue};
 use std::{
     env,
     ffi::OsStr,
@@ -11,15 +19,6 @@ use std::{
     str,
     sync::atomic::AtomicBool,
 };
-
-use clap::{
-    builder::{PossibleValue, PossibleValuesParser, ValueParser},
-    Parser, ValueEnum,
-};
-use gix::{clone, create, open, progress, remote};
-use rust_embed::RustEmbed;
-use serde_json::{from_str, json, to_string_pretty, Error as JsonError, Value as JsonValue};
-use toml_edit::{Document, Formatted, InlineTable, Item, TomlError, Value as TomlValue};
 use ureq::get;
 
 const SOROBAN_EXAMPLES_URL: &str = "https://github.com/stellar/soroban-examples.git";
@@ -73,7 +72,7 @@ pub enum Error {
     CheckoutError(#[from] clone::checkout::main_worktree::Error),
 
     #[error("Failed to parse toml file: {0}")]
-    TomlParseError(#[from] TomlError),
+    CargoTomlParseError(#[from] cargo_toml::Error),
 
     #[error("Failed to parse package.json file: {0}")]
     JsonParseError(#[from] JsonError),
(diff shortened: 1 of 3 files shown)


willemneal:
Looks good to me. Love a crate that makes everything smoother.

willemneal (approved):
We could use some tests, but this seems straight forward enough not to warrant one before the merge. Perhaps we can add more tests for the init in a future PR.

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -637,8 +658,32 @@ mod tests {
...
+            .detail()
+            .unwrap();
+        assert!(
+            dev_dep.inherited,
+            "expected [dev_dependencies.soroban-sdk] to be a workspace dependency"
+        );
+        assert!(
+            dev_dep.features.len() > 0,
+            "expected [dev_dependencies.soroban-sdk] to have some features set"
+        );
```
chadoh, line 686: Tests here, @willemneal

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -637,8 +658,32 @@ mod tests {
...
         let contract_dir = project_dir.join("contracts").join(contract_name);
         let cargo_toml_path = contract_dir.as_path().join("Cargo.toml");
-        let cargo_toml_str = read_to_string(cargo_toml_path).unwrap();
-        assert!(cargo_toml_str.contains("soroban-sdk = { workspace = true }"));
+        println!(
+            "Cargo.toml:\n{}",
+            read_to_string(cargo_toml_path.clone()).unwrap()
+        );
+        let toml = cargo_toml::Manifest::from_path(cargo_toml_path).unwrap();
+        if let Some(dep) = toml.dependencies.get("soroban-sdk").unwrap().detail() {
```
chadoh, line 666: One weird thing: I had to add this “if let“ here, because the hello-world Cargo.toml is evaluated as having no “dependencies.soroban-sdk”, which seems obviously wrong. I ran out of time to investigate yesterday, but it might be worth looking into more before we merge.

cmd/soroban-cli/src/commands/contract/init.rs
willemneal: 🤦

cmd/soroban-cli/src/commands/contract/init.rs
chadoh: Did the testing with `toml` crate instead of `cargo_toml`. Still using `cargo_toml` to construct the file, which works well, but for some reason when parsing the (correctly-formatted) files, it was parsing `dependencies.soroban-sdk.workspace = true` into its internal version of`dependencies.soroban-sdk.version = "20.3.2"`. Checking it with just `toml` demonstrates that it is well-formatted.

PR #1301, opened 2024-04-30
Embed examples contract list into source and remove from build script

leighmcculloch:
### What
Embed examples contract list into source and remove from build script.

### Why
When running a lot of builds repeatedly, such as in CI, we're hitting GitHub rate limits on almost every build. It's convenient that the list is auto updated at every build, but we don't really need it to be updated that frequently. We'll get 90% of the same benefit with a reminder that it's got out of sync to update it from time to time with a command.

The added check will run once a commit, but won't block builds if we do rate limit or if it's out of date. Folks can run the generate make target as needed to get the check back to green.

This also means that cargo installing the crate will no longer phone out to GitHub and potentially get blocked too.


elizabethengelman (approved):
This looks good to me! 👏 🙏

PR #1307, opened 2024-05-01
Add xdr command to root

leighmcculloch:
### What
Add xdr command to root of cli command structure.

### Why
It is currently under the lab, but at some point in the past we had commented about moving it to the root because really the majority of the CLI is "lab-like" commands and so having some commands under a lab subcommand makes little sense. XDR is also so prevalent that functionality to use it should be accessible. As a first step this PR adds it to the root. A follow up PR can remove it from the lab command.


fnando (approved):
noticed that rpc tests are (still?) failing.

leighmcculloch:
> noticed that rpc tests are (still?) failing.

RPC tests are a bit flakey, it needs investigation separately.

PR #1278, opened 2024-04-03
Feat/ledger signing

elizabethengelman:
### What

Add support for signing transactions on a Ledger device. 
https://github.com/stellar/soroban-cli/issues/1178

### Why

Work toward key protection when submitting txns from the CLI. ([https://github.com/stellar/soroban-cli/issues/1255](https://github.com/stellar/soroban-cli/issues/1255))

### Known limitations

This does not implement the full `Stellar` signer trait yet. The [stellar app](https://github.com/LedgerHQ/app-stellar) deployed on the ledger device does not yet support signing soroban auth entries yet, so in the mean time we will only be able to do hash signing for soroban tx.

### Todo
- [x] get tests working on CI
- [x] cleanup 
- [x] add tests for other Ledger devices - the tests currently focus on Nano S (will plan to do this in another PR so as not to hold this one up: https://github.com/stellar/soroban-cli/pull/1321)
- [x] error handling
- [x] make sure that the Stellar trait is in sync with https://github.com/stellar/soroban-cli/pull/1311

PR #1311, opened 2024-05-02
feat: use ledger in `txn sign`

willemneal:
This depends on #1283 and #1278.

There is a lot of cleaning up to do and move some work into the ledger PR, but this currently works!
```
cargo r -- contract install --build-only --wasm ./target/wasm32-unknown-unknown/test-wasms/test_hello_world.wasm | 
cargo r -- txn simulate | 
cargo r -- txn sign --signer ledger | 
cargo r -- txn send
```


willemneal:
close in favor of: https://github.com/stellar/stellar-cli/pull/1353

PR #1320, opened 2024-05-08
Add stellar-cli crate alongside soroban-cli

### LINKED ISSUE #1187: soroban-cli: rename the Soroban CLI to Stellar CLI
- Explore we can have two CLIs (Soroban CLI and Stellar CLI)
- Sunset Soroban CLI once the download trends down

leighmcculloch:
### What
Add stellar-cli crate alongside soroban-cli

### Why
Close #1187

### DIFF
--- a/.github/workflows/full-help-docs.yml
+++ b/.github/workflows/full-help-docs.yml
@@ -18,8 +18,8 @@ jobs:
       # this looks goofy to get GITHUB_OUTPUT to work with multi-line return values;
       # see https://stackoverflow.com/a/74266196/249801
       run: |
-        cargo md-gen
-        raw_diff=$(git diff docs/soroban-cli-full-docs.md)
+        make generate-full-help-doc
+        raw_diff=$(git diff FULL_HELP_DOCS.md)
         if [ "$raw_diff" != "" ]; then echo ""; echo "Unexpected docs change:"; echo "======================="; echo ""; echo "$raw_diff"; echo ""; echo "======================="; echo ""; fi
         echo diff=$raw_diff >> $GITHUB_OUTPUT
       id: doc-gen
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ install_rust: install
 
 install:
 	cargo install --locked --path ./cmd/soroban-cli --debug
+	cargo install --locked --path ./cmd/stellar-cli --debug
 	cargo install --locked --path ./cmd/crates/soroban-test/tests/fixtures/hello --root ./target --debug --quiet
 
 # regenerate the example lib in `cmd/crates/soroban-spec-typsecript/fixtures/ts`
@@ -40,6 +41,9 @@ build-test-wasms:
 	cargo build --package 'test_*' --profile test-wasms --target wasm32-unknown-unknown
 
 build-test: build-test-wasms install
+
+generate-full-help-doc:
+	cargo run --bin doc-gen --features clap-markdown
 
 generate-examples-list:
 	curl -sSL https://api.github.com/repos/stellar/soroban-examples/git/trees/main \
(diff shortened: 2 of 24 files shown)


.github/workflows/full-help-docs.yml
```diff
@@ -18,8 +18,8 @@ jobs:
       # this looks goofy to get GITHUB_OUTPUT to work with multi-line return values;
       # see https://stackoverflow.com/a/74266196/249801
       run: |
-        cargo md-gen
-        raw_diff=$(git diff docs/soroban-cli-full-docs.md)
+        make generate-full-help-doc
+        raw_diff=$(git diff FULL_HELP_DOCS.md)
```
leighmcculloch, line 22: Renamed the doc file to disconnect it from the term "soroban-cli" and refer to it with a consistent single term "full help docs" throughout the repo and elsewhere.

Makefile
```diff
@@ -28,6 +28,7 @@ install_rust: install
 
 install:
 	cargo install --locked --path ./cmd/soroban-cli --debug
+	cargo install --locked --path ./cmd/stellar-cli --debug
```
leighmcculloch, line 31: Both CLIs will exist for a time, or maybe forever.

Makefile
```diff
@@ -41,6 +42,9 @@ build-test-wasms:
 
 build-test: build-test-wasms install
 
+generate-full-help-doc:
+	cargo run --bin doc-gen --features clap-markdown
```
leighmcculloch, line 46: For some reason this one target was built not with a makefile target, but I moved it here for consistency.

cmd/soroban-cli/src/cli.rs
```diff
@@ -0,0 +1,76 @@
...
+        "SECRET_KEY",
+        "CONFIG_HOME",
+    ];
+    for var in vars {
+        let soroban_key = format!("SOROBAN_{var}");
+        let stellar_key = format!("STELLAR_{var}");
+        if let Ok(val) = std::env::var(soroban_key) {
+            std::env::set_var(stellar_key, val);
+        }
+    }
```
leighmcculloch, line 32: This new file is really just a move of the old `main.rs` so it can be shared. The only new code is these lines that allow us to continue to support the SOROBNA_ env vars by mapping them to the STELLAR_ env vars.

cmd/soroban-cli/src/utils.rs
```diff
@@ -71,16 +71,32 @@ pub fn contract_id_from_str(contract_id: &str) -> Result<[u8; 32], stellar_strke
...
-                "soroban directory not found",
-            ));
+            break;
         }
     }
-    Ok(soroban_dir(&pwd))
+    Err(std::io::Error::new(
+        std::io::ErrorKind::Other,
+        "stellar directory not found",
+    ))
```
leighmcculloch, line 99: This is the only new significant logic as part of the change. Both CLIs will look for a .stellar or a .soroban, prefer the .stellar, and warn if both are present. But won't warn about the old .soroban, and it's fine if people continue to keep using it. There's really little advantage to messing with folks otherwise.

cmd/stellar-cli/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+[package.metadata.binstall]
+pkg-url = "{ repo }/releases/download/v{ version }/{ name }-{ version }-{ target }{ archive-suffix }"
+bin-dir = "{ bin }{ binary-ext }"
+
+[features]
+default = []
+opt = ["soroban-cli/opt"]
+
+[dependencies]
+soroban-cli = { workspace = true }
```
leighmcculloch, line 28: For the moment the stellar-cli is just a wrapper around the soroban-cli. It won't be critical to change this anytime soon as it's unlikely this will cause technical debt.

leighmcculloch (commented):
This is still a wip but some explanations inline of the approach I took.

leighmcculloch:
The failing rpc tests are flakey and pass locally sometimes, then mostly don't pass on CI. From my investigation it looks like a race condition between when generating accounts completes and is fully executed vs when a following test runs seeing the account is not gone. So I don't think the test breakage should block this PR.

fnando (approved):

PR #1323, opened 2024-05-10
Output TransactionEnvelope instead of Transaction for --build-only

leighmcculloch:
### What
Output TransactionEnvelope instead of Transaction for --build-only

### Why
Tools should always output, and accept as input, full envelopes because otherwise other tools that the inputs/outputs are used with won’t know what type of transaction we’ve given them.

It’s a bit confusing because tools always use the term ‘transaction’ to talk about what they accept as input, but in reality when tools like the Lab and others say they accept or build a transaction, they actually mean a transaction envelope.


willemneal:
My thought here was that signing should result in a transaction envelope.

leighmcculloch:
Yeah I understand that logic and approach, but the way that other tools work is that you always pass around an envelope, because there are more than one type of transaction and the way to distinguish between them is to pass around the envelope and then within it it can have the different types. Other tools like the Lab pass around the envelope so we should do the same.

fnando (approved):

PR #1328, opened 2024-05-16
fix: switch to stderr directly for when account already exists

willemneal:
### What

Switch from `tracing::warn` to `eprintln`

### Why

It didn't make it into stderr in tests.

### Known limitations

[TODO or N/A]


chadoh (approved):
Great! Such a small change! Can you explain it a little? Why did `tracing::warn` break the test?

chadoh:
I see your what/why, but I'm curious to know more. Why had it been `tracing::warn`?

leighmcculloch (commented):
We should probably fix the tests, rather change the output. I think we want the output to go to the warning logs don't we?

willemneal:
Yeah but this was blocked other PRs and it should go towards logs. Rather not sure why the log wasn't being picked up by stdout.

PR #1332, opened 2024-05-17
Fix/stellar asset bindings

BlaineHeffron:
### What

Bindings command can now generate for a StellarAsset type contract. Fixes https://github.com/stellar/stellar-cli/issues/1312
Logic for checking if its a StellarAsset was extracted from `invoke.rs` to `get_spec.rs` used by both `invoke` and `bindings`.

### Why

Bindings were failing when trying to retrieve the wasm for a StellarAsset type from the contract Id. 

### Known limitations

none

### DIFF
--- a/cmd/crates/soroban-test/tests/it/integration/bindings.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/bindings.rs
@@ -29,4 +29,7 @@ async fn invoke_test_generate_typescript_bindings() {
 
     let files = std::fs::read_dir(output_dir).expect("Failed to read output directory");
     assert!(files.count() > 0, "No files generated in the output directory");
+
+    // Clean up: remove the output directory and its contents
+    std::fs::remove_dir_all(OUTPUT_DIR).expect("Failed to delete output directory");
 }
--- a/cmd/crates/soroban-test/tests/it/integration/bindings.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/bindings.rs
@@ -1,0 +1,32 @@
+use soroban_test::{TestEnv, LOCAL_NETWORK_PASSPHRASE};
+use super::util::deploy_swap;
+
+pub const OUTPUT_DIR: &str = "./bindings-output";
+
+
+#[tokio::test]
+async fn invoke_test_generate_typescript_bindings() {
+    let sandbox = &TestEnv::new();
+    let contract_id = deploy_swap(sandbox).await;
+    let cmd = sandbox.cmd_arr::<soroban_cli::commands::contract::bindings::typescript::Cmd>(&[
+        "--network-passphrase",
+        LOCAL_NETWORK_PASSPHRASE,
+        "--rpc-url",
+        &sandbox.rpc_url,
+        "--output-dir",
+        OUTPUT_DIR,
+        "--overwrite",
+        "--contract-id",
+        &contract_id.to_string(),
+    ]);
+
+    let result = sandbox.run_cmd_with(cmd, "test").await;
+
+    assert!(result.is_ok(), "Failed to generate TypeScript bindings");
+
+    let output_dir = std::path::Path::new(OUTPUT_DIR);
+    assert!(output_dir.exists(), "Output directory does not exist");
+
+    let files = std::fs::read_dir(output_dir).expect("Failed to read output directory");
+    assert!(files.count() > 0, "No files generated in the output directory");
+}
--- a/cmd/crates/soroban-test/tests/it/integration/bindings.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/bindings.rs
@@ -31,5 +31,4 @@ async fn invoke_test_generate_typescript_bindings() {
         files.count() > 0,
         "No files generated in the output directory"
     );
-
 }
--- a/cmd/crates/soroban-test/tests/it/integration/bindings.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/bindings.rs
@@ -1,8 +1,7 @@
-use soroban_test::{TestEnv, LOCAL_NETWORK_PASSPHRASE};
 use super::util::deploy_swap;
+use soroban_test::{TestEnv, LOCAL_NETWORK_PASSPHRASE};
 
 const OUTPUT_DIR: &str = "./bindings-output";
-
 
 #[tokio::test]
 async fn invoke_test_generate_typescript_bindings() {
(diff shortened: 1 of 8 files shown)


cmd/soroban-cli/src/get_spec.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use soroban_spec::read::FromWasmError;
+pub use soroban_spec_tools::contract as contract_spec;
+
+use crate::rpc;
+use crate::commands::{config::data, global};
+
+
+#[derive(thiserror::Error, Debug)]
+pub enum GetSpecError {
```
willemneal, line 15: This should just be `Error`, since we can refer to it externally with the module `get_spec::Error`.

cmd/crates/soroban-test/tests/it/integration/bindings.rs
```diff
@@ -0,0 +1,35 @@
...
+    assert!(result.is_ok(), "Failed to generate TypeScript bindings");
+
+    let output_dir = std::path::Path::new(OUTPUT_DIR);
+    assert!(output_dir.exists(), "Output directory does not exist");
+
+    let files = std::fs::read_dir(output_dir).expect("Failed to read output directory");
+    assert!(files.count() > 0, "No files generated in the output directory");
+
+    // Clean up: remove the output directory and its contents
+    std::fs::remove_dir_all(OUTPUT_DIR).expect("Failed to delete output directory");
```
willemneal, line 34:
```suggestion
```
`TestEnv` uses temp_dir, which when dropped deletes the folder! This can be an issue if it's dropped early. This is why having an interface that lets you pass a function, which has the `&TestEnv` as a argument is preferred since it can't be dropped by the test.

It's currently hard to do with async functions though or at least last time I tried.

cmd/crates/soroban-test/tests/it/integration/bindings.rs
```diff
@@ -0,0 +1,35 @@
+use soroban_test::{TestEnv, LOCAL_NETWORK_PASSPHRASE};
+use super::util::deploy_swap;
+
+pub const OUTPUT_DIR: &str = "./bindings-output";
```
willemneal, line 4:
```suggestion
const OUTPUT_DIR: &str = "./bindings-output";
```
Is this used outside this file?

willemneal (approved):
Just a few nits. Also looks like you have to run `cargo fmt`. Good first real Rust PR! 🎉

cmd/soroban-cli/src/get_spec.rs
BlaineHeffron: fixed

cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
```diff
@@ -61,22 +67,41 @@ pub enum Error {
...
+    async fn run_against_rpc_server(
+        &self,
+        global_args: Option<&global::Args>,
+        _config: Option<&config::Args>,
+    ) -> Result<(), Error> {
         let spec = if let Some(wasm) = &self.wasm {
             let wasm: wasm::Args = wasm.into();
             wasm.parse()?.spec
         } else {
-            let fetch = contract::fetch::Cmd {
```
leighmcculloch, line 72: It would be convenient if we had a `spec` command, that outputted just the spec as XDR / JSON, given a network and contract id, network and wasm hash, or wasm file, and then if we used that here too like we were using the fetch command. Maybe some clean up that we could follow up with.

willemneal: Good point! Could even add the schema!

cmd/crates/soroban-test/tests/it/integration/bindings.rs
BlaineHeffron: fixed

cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
```diff
@@ -61,22 +62,38 @@ pub enum Error {
...
-                locator: self.locator.clone(),
-                network: self.network.clone(),
-            };
-            let bytes = fetch.get_bytes().await?;
-            contract_spec::Spec::new(&bytes)?.spec
+            let contract_id = soroban_spec_tools::utils::contract_id_from_str(&self.contract_id)
+                .map_err(|e| Error::CannotParseContractId(self.contract_id.clone(), e))?;
+            let spec_entries = get_remote_contract_spec(
+                &contract_id,
+                self.network.rpc_url.as_deref().ok_or(Error::MissingRpcUrl)?,
```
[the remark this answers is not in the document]
BlaineHeffron: I implemented your recommendation. It does make the function call have a few more arguments; perhaps there is a way to make things more concise.

willemneal:
@elizabethengelman Any ideas on the failing emulator tests?

cmd/soroban-cli/src/get_spec.rs
```diff
@@ -0,0 +1,80 @@
...
+    let ContractDataEntry {
+        val: ScVal::ContractInstance(ScContractInstance { executable, .. }),
+        ..
+    } = r
+    else {
+        return Err(Error::MissingResult);
+    };
+
+    // Get the contract spec entries based on the executable type
+    let spec_entries = match executable {
```
willemneal, line 61: This could just be `Ok(match ...`

Instead of needing the local var.

cmd/soroban-cli/src/get_spec.rs
```diff
@@ -0,0 +1,80 @@
...
+    #[error(transparent)]
+    Config(#[from] config::Error),
+}
+
+///
+/// # Errors
+pub async fn get_remote_contract_spec(
+    contract_id: &[u8; 32],
+    locator: locator::Args,
+    network: network::Args,
```
willemneal, line 38:
```suggestion
    locator: &locator::Args,
    network: &network::Args,
```
This will prevent the cloning in the caller.

willemneal (approved):
Just two small nits, but otherwise this looks great!

cmd/soroban-cli/src/get_spec.rs
BlaineHeffron: fixed

elizabethengelman:
> Any ideas on the failing emulator tests?

It looks like the emulator tests are passing now, but definitely something we should keep an eye on in case they are flakey.

cmd/soroban-cli/src/get_spec.rs
```diff
@@ -0,0 +1,78 @@
...
+/// # Errors
+pub async fn get_remote_contract_spec(
+    contract_id: &[u8; 32],
+    locator: &locator::Args,
+    network: &network::Args,
+    global_args: Option<&global::Args>,
+    config: Option<&config::Args>,
+) -> Result<Vec<ScSpecEntry>, Error> {
+    let network = config.map_or_else(
+        || network.get(&locator).map_err(Error::from),
```
willemneal, line 43:
```suggestion
        || network.get(locator).map_err(Error::from),
```

PR #1333, opened 2024-05-20
fix(ts-bindings): package.json needs type: module

chadoh:
### What

Since we started using `module: nodenext` in tsconfig.json, we need to
tell libraries that import these ones that the exported code is an ES
Module.

### Why

This fixes github.com/stellar/soroban-astro-template to work with
libraries generated with `soroban contract bindings typescript`

### Known limitations

N/A


BlaineHeffron (approved):
I just validated that it works with the astro template. Looks good.

kalepail (approved):

elizabethengelman (approved):

PR #1339, opened 2024-05-27
Remove the contract-ids dir

leighmcculloch:
### What

Remove the contract-ids directory from the .soroban directory prefilled in initialised contracts.

### Why

The contract-ids directory isn't used in examples, and isn't a pattern being encouraged since there are plans to add aliases.

cc @elizabethengelman 

### Known limitations

N/A


elizabethengelman (approved):
Makes sense!

Issue #1335, opened 2024-05-22
Test


willemneal:
### What version are you using?
Current main

### What did you do?

Try to pass CI

### What did you expect to see?

passes

### What did you see instead?

https://github.com/stellar/stellar-cli/actions/runs/9196175043/job/25293677925#step:9:11

leighmcculloch:
This is the test:
```
thread 'integration::dotenv::cli_args_have_priority' panicked at cmd/crates/soroban-test/tests/it/integration/util.rs:49:10:
called `Result::unwrap()` on an `Err` value: Install(Rpc(TransactionSubmissionFailed("Some(\n    TransactionResult {\n        fee_charged: 182117,\n        result: TxFailed(\n            VecM(\n                [\n                    OpInner(\n                        InvokeHostFunction(\n                            ResourceLimitExceeded,\n                        ),\n                    ),\n                ],\n            ),\n        ),\n        ext: V0,\n    },\n)")))
```

PR #1340, opened 2024-05-28
fix: dotenv test is flaky

willemneal:
fixes: #1335
This is currently blocking CI and failing on main. Given that clap handles ensuring that CLI args have higher priority than ENV vars, this test is more of a sanity check so it could also just be removed.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,6 +2,7 @@
 
 [alias] # command aliases
 f = "fmt"
+md-gen = "run --bin doc-gen --features clap-markdown"
 # b = "build"
 # c = "check"
 # t = "test"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,7 +2,6 @@
 
 [alias] # command aliases
 f = "fmt"
-md-gen = "run --bin doc-gen --features clap-markdown"
 # b = "build"
 # c = "check"
 # t = "test"
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -547,7 +547,7 @@ Initialize a Soroban project with an example contract
 
 * `-w`, `--with-example <WITH_EXAMPLE>`
 
-  Possible values: `account`, `alloc`, `atomic_multiswap`, `atomic_swap`, `auth`, `cross_contract`, `custom_types`, `deep_contract_auth`, `deployer`, `errors`, `eth_abi`, `events`, `fuzzing`, `increment`, `liquidity_pool`, `logging`, `mint-lock`, `simple_account`, `single_offer`, `timelock`, `token`, `ttl`, `upgradeable_contract`, `workspace`
+  Possible values: `account`, `alloc`, `atomic_multiswap`, `atomic_swap`, `auth`, `cross_contract`, `custom_types`, `deep_contract_auth`, `deployer`, `errors`, `eth_abi`, `events`, `fuzzing`, `increment`, `liquidity_pool`, `logging`, `mint-lock`, `simple_account`, `single_offer`, `timelock`, `token`, `upgradeable_contract`, `workspace`
 
 * `-f`, `--frontend-template <FRONTEND_TEMPLATE>`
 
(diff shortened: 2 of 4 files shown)


willemneal:
@leighmcculloch This is definitely a bad code smell and deserves an investigation why sleeps fix the concurrency issue, but do you think it's fine to just let it pass for now since it's a blocker for other PRs?

.cargo/config.toml
```diff
@@ -2,6 +2,7 @@
 
 [alias] # command aliases
 f = "fmt"
+md-gen = "run --bin doc-gen --features clap-markdown"
```
leighmcculloch, line 5:
```suggestion
```

FULL_HELP_DOCS.md
```diff
@@ -547,7 +547,7 @@ Initialize a Soroban project with an example contract
 
 * `-w`, `--with-example <WITH_EXAMPLE>`
 
-  Possible values: `account`, `alloc`, `atomic_multiswap`, `atomic_swap`, `auth`, `cross_contract`, `custom_types`, `deep_contract_auth`, `deployer`, `errors`, `eth_abi`, `events`, `fuzzing`, `increment`, `liquidity_pool`, `logging`, `mint-lock`, `simple_account`, `single_offer`, `timelock`, `token`, `upgradeable_contract`, `workspace`
+  Possible values: `account`, `alloc`, `atomic_multiswap`, `atomic_swap`, `auth`, `cross_contract`, `custom_types`, `deep_contract_auth`, `deployer`, `errors`, `eth_abi`, `events`, `fuzzing`, `increment`, `liquidity_pool`, `logging`, `mint-lock`, `simple_account`, `single_offer`, `timelock`, `token`, `ttl`, `upgradeable_contract`, `workspace`
```
leighmcculloch, line 550:
```suggestion
  Possible values: `account`, `alloc`, `atomic_multiswap`, `atomic_swap`, `auth`, `cross_contract`, `custom_types`, `deep_contract_auth`, `deployer`, `errors`, `eth_abi`, `events`, `fuzzing`, `increment`, `liquidity_pool`, `logging`, `mint-lock`, `simple_account`, `single_offer`, `timelock`, `token`, `upgradeable_contract`, `workspace`
```

cmd/soroban-cli/example_contracts.list
```diff
@@ -19,5 +19,6 @@ simple_account
 single_offer
 timelock
 token
+ttl
```
leighmcculloch, line 22:
```suggestion
```

leighmcculloch (approved):
Looks good, just going to strip out changes that seem unrelated.

leighmcculloch:
> @leighmcculloch This is definitely a bad code smell and deserves an investigation why sleeps fix the concurrency issue, but do you think it's fine to just let it pass for now since it's a blocker for other PRs?

There's not much point removing this test because other tests have sleeps too. There's something up with the cost to do txns and the resources required which are racing and getting it wrong. Likely because we're reusing the same account over and over. In other projects where we use a live network we create new accounts on every test, so the tests can run entirely isolate each other. That's the fix I think.

willemneal:
Each test should be a different account. It's just in this case they are all trying to deploy the same contract... But so do other tests so I'm really at a loss for the cause.

leighmcculloch:
I was too specific, the creation of a resource that has a different resource usage on first creation vs update. Accounts are a common culprit, but the WASM install / contract deploy would be similar.

leighmcculloch:
One way to see if that would fix it and allow us to get rid of the sleeps, would be to make the .wasm files unique for each test. Which can be done by writing some random data into a custom section of the wasm:
```
$ echo $RANDOM | wasm-cs file.wasm write makeunique
```

willemneal:
Great idea! Though I'm wondering if perhaps we can instead change the install step to first check if the ledger entry exists. Though this too could run into a race condition. So I will look into using `wasm-cs` to remove these and in general make all tests more isolated.

leighmcculloch:
There are some conditions when install should be allowed even though it is already installed. V21 added some situations where if you reinstall your contract will be updated with new data calculated at install time that make it more optimal ([CAP-54], [CAP-56]). 

[CAP-54]: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0054.md
[CAP-56]: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0056.md

PR #1343, opened 2024-05-29
Fix/ledger tests org update

elizabethengelman:
### What
Moves stellar-ledger test-only files into a `test_fixtures` directory.

### Why

Hopefully, this will make it clearer what files are used just for testing, and which ones are source code.

### Known limitations

Once [this ticket](https://github.com/stellar/stellar-cli/issues/1341) is complete this test-only code will be pulled out into its own crate.


cmd/crates/stellar-ledger/src/test_fixtures/speculos.rs
```diff
@@ -3,6 +3,8 @@ use testcontainers::{core::WaitFor, Image, ImageArgs};
```
leighmcculloch, line 3: "test fixtures" are usually files that are there to support tests, so things like binaries, or generated code to regenerate to compare against. So I'd keep code supporting the tests in a tests/test directory, while keep artifacts that tests use in the test_fixtures directory.

cmd/crates/stellar-ledger/src/emulator_tests.rs
```diff
@@ -4,9 +4,7 @@ use soroban_env_host::xdr::Transaction;
 use soroban_env_host::xdr::{self, Operation, OperationBody, Uint256};
```
willemneal, line 4: Do we need this file here anymore? Could it too be moved to `tests/test`

leighmcculloch (approved):
Looks good to me.

You could make the modifications @willemneal mentioned too, but either way this looks like a good improvement 👍🏻.

I don't feel strongly that you need to do this now, but we normally keep the test and test_fixtures modules under the src directory, not at the top level, so that we are still only linking one binary during tests. While Rust does support outside in testing via the tests/ folder at the top level a problem with it is every test file becomes a separate link target, which means the crate has to be relinked over and over, which slows down running tests. For crates this size it may not matter much.

willemneal:
@leighmcculloch From my understanding if you have `tests/it/main.rs`, then it acts as a single target. See https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html#Rules-of-Thumb

So really it's a question about the surface area, should it be external or internal?

leighmcculloch:
TIL nice, I didn't know about the module/main.rs approach. 🎉

Also yeah this lib is small so probably not a big deal to worry about test organisation too much. The main goal here I think is reducing confusion about what helper functionality is for tests or not, which this does. 👍🏻

cmd/crates/stellar-ledger/src/emulator_tests.rs
elizabethengelman: Good point! I'll move this to `tests`. I may leave it at the top level of `tests` since it is the only integration test file so far.

cmd/crates/stellar-ledger/src/lib.rs
```diff
@@ -304,12 +301,19 @@ fn test_network_hash() -> Hash {
     use sha2::Digest;
     Hash(sha2::Sha256::digest(TEST_NETWORK_PASSPHRASE).into())
 }
+
 #[cfg(test)]
 mod test {
```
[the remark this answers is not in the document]
elizabethengelman: These tests don't need anything non-public, but I kind of like the idea of leaving these unit tests in `src/lib.rs` - I like the separation of internal:unit vs external:integration like is mentioned in that article you shared. And then we also have a way to run some quick tests via `cargo run --lib` which avoids the longer running integration tests in `tests/`. 

The only thing weird is that they reach out to `tests/test/emulator_http_transport.rs`. That is just so there is something that implements the `Exchange` trait to pass into the `LedgerSigner` initializer. I am going to leave this for now... but this could also be mocked out pretty easily too in the future.

elizabethengelman:
Great discussion, and thanks for sharing that article @willemneal - super helpful! 

I also think that if we have `tests/module/mod.rs` Cargo won't treat the files in that dir as integration tests, and won't be compiled as separate crates, so I think we may also avoid the relinking for each test helper issue... if I'm reading [this](https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests) correctly.

cmd/crates/stellar-ledger/src/emulator_tests.rs
elizabethengelman: Oh actually, i see now how that's being done in `soroban-test` - i like that! It also makes it a bit easier to turn on/off the emulator-tests via a feature.

PR #1353, opened 2024-05-31
feat: add txn sign/send and ledger support

willemneal:
Currently relies on #1283.

Adds:
- tx sign - use local key or ledger to sign a transaction (envelope)
- tx send - submit a signed transaction envelope to the network
- `Stellar` trait which handles the signing of transactions and auth-entries


.github/workflows/binaries.yml
```diff
@@ -36,6 +36,9 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
+    - name: install libudev-dev
+      run: |
+        sudo apt install -y libudev-dev
```
fnando, line 41: You need to match the arch, like we're currently doing a [few lines below](https://github.com/stellar/stellar-cli/pull/1353/files#diff-bdfe89ee8491e51e3f81ada685c51d24188416d8b529b443ee6085a0e0fd4a4aL40).

willemneal:
Closing in favor of #1406 as it is cleaner to review.

PR #1355, opened 2024-06-04
Use safe unwrapping in option unwrap

leighmcculloch:
### What
Use safe unwrapping in option unwrap in code.

### Why
In Rust options can be unwrapped safely where the compiler ensures that the value is not used if set to none. Avoids the use of panic causing calls like .unwrap.

There's no functional change about this code. The code was perfectly fine as is, but over time under change it will be less brittle if we rely on the compiler being able to check that it is safe for us to unwrap a value.


elizabethengelman (approved):
👏 🙏

PR #1356, opened 2024-06-04
Add support for contract id alias name when deploying and invoking contracts.

fnando:
#### `soroban contract deploy`

This pr introduces two different switches:

* `--alias ALIAS`: the alias name, effectively the file name that will be created under `.soroban/contract-ids/:network_name/:alias.txt`. If `--global` is provided, then the alias will be created under `~/config/.soroban/contract-ids/:network_name/:alias.txt`.

#### `soroban contract invoke`

When invoking a contract, if the user provides a valid alias to `--id`, then the respective contract id will be loaded. If the alias doesn't exist, then an error message like `error: cannot parse contract ID foo: the strkey is invalid` will be returned.

### Why

When deploying wasms, it's common to redirect the contract id to a file so it can be reused later on. To simplify the whole process and make this a non-*nix feature, let's introduce the concept of contract id aliasing.

### Known limitations

* This pr optimistically tries to create and read the contract id file.

### Tickets

- https://github.com/stellar/stellar-cli/issues/1248


cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +109,68 @@ pub enum Error {
...
+
+        let _ = create_dir_all(dir);
+
+        let mut to_file = OpenOptions::new()
+            .create(true)
+            .write(true)
+            .open(file_path)
+            .unwrap();
+        let _ = to_file.write_all(contract.as_bytes());
+    }
```
willemneal, line 173: Could you move this into `config::Args`. Eventually it would be part of the config crate.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +109,68 @@ pub enum Error {
...
+        } else {
+            Ok(())
+        }
+    }
+
+    fn alias(&self) -> String {
+        match &self.alias {
+            Some(n) => n.to_string(),
+            None => String::new(),
+        }
```
willemneal, line 151: Just for your info.
```suggestion
        self.alias.as_ref().map(Clone::clone).unwrap_or_default()
 ```
 
 However, I suggest removing method.

File: cmd/soroban-cli/src/commands/contract/deploy/wasm.rs [hunk 3]
```diff
@@ -100,19 +109,68 @@ pub enum Error {
...
+    }
+
+    fn alias_path(&self) -> PathBuf {
+        let config_dir = self.config.config_dir().unwrap();
+        let name = format!("{}.txt", self.alias());
+
+        config_dir.join("contract-ids").join(name)
+    }
```
willemneal, line 159:
```suggestion
    fn alias_path(&self) -> Result<Option<PathBuf>, Error> {
        let config_dir = self.config.config_dir()?;
        Ok(self.alias.as_ref().map(|alias| config_dir.join("contract-ids").join(name)))
    }
```
We already added the errors from config to this files `Error` enum. This means we can use `?` if we return a `Result`. Then we can also return an `Option` in case there was no alias. Also I suggest removing the `.txt`.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
willemneal, line 173:
```suggestion
    fn save_contract_id(&self, contract: &str) -> Result<(), Error> {
        let Some(file_path) = self.alias_path()? else { return Ok(()); };
        let dir = file_path.parent()?;
        create_dir_all(dir)?;
        let mut to_file = OpenOptions::new()
            .create(true)
            .write(true)
            .open(file_path)?;
        to_file.write_all(contract.as_bytes())?;
        Ok(())
    }
```
`unwrap`, while nice when prototyping (still prefer `expect` since it contains a message) does panic, which is not what we want to do.  So it's better to use a Result type or return and option so that the caller can deal with it.

willemneal:
Overall a great first big rust PR. As mentioned above I do think that these functions would work best in the config so that we can separate them from the commands.

And I mentioned yesterday that I think that aliases should be tied to the network where the contract was deployed to. Perhaps for now we can use `local`, `mainnet`, `testnet`, and `futurenet` as subdirectories. I know you had another issue to stream line adding networks using these.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: Would the best place to put these be `cmd/soroban-cli/src/commands/config/`? Or is it unrelated to the config command?

willemneal: I'd do `config/mod.rs` I would add all alias methods to the `impl`.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -305,8 +305,33 @@ impl Cmd {
...
+            .config
+            .config_dir()
+            .unwrap()
+            .join("contract-ids")
+            .join(file_name);
+
+        if path.exists() {
+            match fs::read_to_string(path) {
+                Ok(content) => Some(content),
+                _ => None,
```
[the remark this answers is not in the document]
fnando: thanks for the explanation. This is the hardest part of learning a new language: knowing the idioms that make sense to the language.

> Wait what? A boolean value has methods! Yep!

That doesn't surprise me anymore! I'm from Ruby-land, where all things have methods.

> ```rust
> Ok(path.exists().then(|| fs::read_to_string(path)).transpose()?)
> ```

This is nice!

fnando:
@willemneal can you please do a second review pass? Will extract everything to config::Args in a separate pr, because I noticed the change surface it's way too big.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -54,6 +57,12 @@ pub struct Cmd {
     #[arg(long, short = 'i', default_value = "false")]
     /// Whether to ignore safety checks when deploying contracts
     pub ignore_checks: bool,
+    /// The alias that will be used to save the contract's id.
+    #[arg(long)]
+    pub alias: Option<String>,
+    /// Force saving the contract id file even if it already exists.
+    #[arg(long, short = 'f', default_value = "false")]
+    pub force: bool,
```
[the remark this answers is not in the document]
fnando: i like the idea of always replacing the alias file. As for the command to remove an alias, I'll work on that in a separate pr.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -305,8 +305,33 @@ impl Cmd {
...
+            Some(id) => id,
+            None => self.contract_id.clone(),
+        };
+
+        soroban_spec_tools::utils::contract_id_from_str(&contract_id)
+            .map_err(|e| Error::CannotParseContractId(contract_id.clone(), e))
+    }
+
+    fn load_contract_id(&self) -> Option<String> {
+        let file_name = format!("{}.txt", self.contract_id);
```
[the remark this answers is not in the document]
fnando: better safe than sorry. I'll make sure it only contains `[a-z0-9-_]`.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +109,68 @@ pub enum Error {
...
+        let dir = file_path.parent().unwrap();
+
+        let _ = create_dir_all(dir);
+
+        let mut to_file = OpenOptions::new()
+            .create(true)
+            .write(true)
+            .open(file_path)
+            .unwrap();
+        let _ = to_file.write_all(contract.as_bytes());
```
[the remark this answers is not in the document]
leighmcculloch: I would also avoid it being possible to assign the same alias multiple times with different networks, and instead take the same approach that keys has that the alias "id space" is across all networks, just so that it is less confusing and to remove the ability to have multiple values assigned to a single alias. But something that is different for aliases than keys, is that contract IDs will never be the same across networks, so now or in the future if we store the network passphrase (not network name) in the .json, we should error if someone is trying to use an alias on the wrong network because we're just saving them from doing stuff that won't work.

leighmcculloch (commented):
Just had the thought, that install and deploy are separate, and maybe there's also value in aliases for contracts installed, and not just aliases for deployments of installed contracts. I don't think it changes anything significant about what we're doing here, but just keeping that in mind that we may also end up with aliases pointing to different things. In this case they could still be separate ID spaces so I think what we're doing here, making aliases specifically for contract IDs makes sense.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,91 @@ pub enum Error {
...
+    #[error("cannot access config dir for alias file")]
+    CannotAccessConfigDir,
+    #[error(
+        "alias must be 1-30 chars long, and have only letters, numbers, underscores and dashes"
+    )]
+    InvalidAliasFormat { alias: String },
+    #[error(transparent)]
+    JSONSerialization(#[from] serde_json::Error),
+    #[error(transparent)]
+    IO(#[from] std::io::Error),
```
leighmcculloch, line 120:
```suggestion
    JsonSerialization(#[from] serde_json::Error),
    #[error(transparent)]
    Io(#[from] std::io::Error),
```
Because in Rust caps are reserved for statics and consts, and single letter caps are reserved for generics.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,91 @@ pub enum Error {
...
+
+        if regex.is_match(&alias) {
+            Ok(())
+        } else {
+            Err(Error::InvalidAliasFormat { alias })
+        }
+    }
+
+    fn alias(&self) -> String {
+        self.alias.as_ref().map(Clone::clone).unwrap_or_default()
```
leighmcculloch, line 155: This is the same thing:
```suggestion
        self.alias.clone().unwrap_or_default()
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,91 @@ pub enum Error {
...
+            return Err(Error::CannotAccessConfigDir);
+        };
+
+        match create_dir_all(dir) {
+            Ok(()) => {}
+            _ => return Err(Error::CannotAccessConfigDir),
+        };
+
+        let mut to_file = OpenOptions::new()
+            .create(true)
```
leighmcculloch, line 183: Clippy is telling me this line has undefined behaviour because it isn't accompanied by `.truncate(bool)` to indicate if the file should be truncated or writing should be appended. I think this should be truncated because the goal is to write over:
```suggestion
            .create(true)
            .truncate(true)
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,91 @@ pub enum Error {
...
+            .open(file_path)
+            .map_err(Error::IO)?;
+
+        let payload = AliasData {
+            id: contract.into(),
+        };
+
+        let content = serde_json::to_string(&payload).map_err(Error::JSONSerialization)?;
+
+        to_file.write_all(content.as_bytes()).map_err(Error::IO)
```
leighmcculloch, line 194: Something interesting is happening here that initially wasn't obvious to me.

When no alias is set, the code still runs to write it to a file with an empty name. This happens because the alias is unwrap_or_default resulting in an empty string. Some locations that then use that value are guarded from the case it is empty, but not all, leading to the file writing code running.

I think should remove the unwrap_or_default, and instead use the Option<String> in all the places.

leighmcculloch (commented):
I think there's a bug, that doesn't actually result in negative outcome for the user as it is invisible really, but the behaviour doesn't seem correct. See below.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: interesting… my clippy is running clean!

```console
$ cargo clippy -- -Dwarnings -Dclippy::all -Dclippy::pedantic
warning: /Users/fnando/Projects/stellar/stellar-cli/cmd/stellar-cli/Cargo.toml: file `/Users/fnando/Projects/stellar/stellar-cli/cmd/stellar-cli/src/bin/main.rs` found to be present in multiple build targets:
  * `bin` target `stellar`
  * `bin` target `soroban`
warning: /Users/fnando/Projects/stellar/stellar-cli/cmd/soroban-cli/Cargo.toml: file `/Users/fnando/Projects/stellar/stellar-cli/cmd/soroban-cli/src/bin/main.rs` found to be present in multiple build targets:
  * `bin` target `stellar`
  * `bin` target `soroban`
   Compiling soroban-cli v21.0.0-rc.1 (/Users/fnando/Projects/stellar/stellar-cli/cmd/soroban-cli)
    Checking soroban-spec-typescript v21.0.0-rc.1 (/Users/fnando/Projects/stellar/stellar-cli/cmd/crates/soroban-spec-typescript)
    Checking soroban-spec-tools v21.0.0-rc.1 (/Users/fnando/Projects/stellar/stellar-cli/cmd/crates/soroban-spec-tools)
    Checking soroban-spec-json v21.0.0-rc.1 (/Users/fnando/Projects/stellar/stellar-cli/cmd/crates/soroban-spec-json)
    Checking soroban-test v21.0.0-rc.1 (/Users/fnando/Projects/stellar/stellar-cli/cmd/crates/soroban-test)
    Finished dev [unoptimized + debuginfo] target(s) in 2.78s
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: huh. I couldn't reproduce that. What file name is being used, given that `alias` is not set? I'd expect `.soroban/contract-ids/testnet/.json`, but again, couldn't reproduce it.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: This is more readable imo. Thanks!

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -305,8 +312,39 @@ impl Cmd {
...
+    }
+
+    fn load_contract_id(&self) -> Result<Option<String>, Error> {
+        let file_path = self.alias_path()?;
+
+        if !file_path.exists() {
+            return Ok(None);
+        }
+
+        let content = fs::read_to_string(file_path).map_err(Error::Io)?;
```
willemneal, line 342:
```suggestion
        let content = fs::read_to_string(file_path)?;
```

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -305,8 +312,39 @@ impl Cmd {
...
+
+        let data: AliasData =
+            serde_json::from_str(content.as_str()).map_err(Error::JsonDeserialization)?;
```
willemneal, line 345:
```suggestion
            serde_json::from_str(content.as_str())?;
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,92 @@ pub enum Error {
...
+            .truncate(true)
+            .write(true)
+            .open(file_path)
+            .map_err(Error::Io)?;
+
+        let payload = AliasData {
+            id: contract.into(),
+        };
+
+        let content = serde_json::to_string(&payload).map_err(Error::JsonSerialization)?;
```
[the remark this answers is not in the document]
fnando: removing all `map_err` as suggested prevent us from showing better error messages, right? Any suggestions for that?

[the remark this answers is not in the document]
willemneal: The `#[error(transparent)]` and the `#[from]` automatically generates a `impl From<OtherError> for Error`.  This means that you can do `other_error.into()` to convert to `Error`. And `?` does this for you! If this ends up not being a nice error message, you can create a new one and then use `map_err`.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: I swear I had a check in place. It probably got removed during a refactor. Fixed!

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,96 @@ pub enum Error {
...
+        Ok(config_dir
+            .join("contract-ids")
+            .join(network)
+            .join(file_name))
+    }
+
+    fn save_contract_id(&self, contract: &String) -> Result<(), Error> {
+        if self.alias().is_empty() {
+            return Ok(());
+        }
```
leighmcculloch, line 173: This fixed the bug, but not the root cause, that we're not leaning on the compiler to help us write safe logic with an Option<String>. So I would replace each of these if empty guards with a `if let Some(alias) = &self.alias { /* existing logic */ }`.

leighmcculloch (approved):
Looks good. I would make the changed noted below if you're up for it. It is a more common idiom in Rust than using empty strings to signal an absent value. I hadn't mentioned it before, but mentioning it now because I think that was the root cause of the bug that was fixed.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: Do you think the reliance on `self.alias()`, which ensures a string is always returned, is a anti-pattern in Rust? Should I drop it entirely so I can use `match` over the optional value in all places?

leighmcculloch: I think it would be considered an anti-pattern. It's a more common approach in languages without a type like `Option<T>`. If an empty alias was a meaningful thing I'd say it was fine, in this case an empty alias is just another form of `Option::None`, so we're converting from one form to another form, and we're converting away from a form that the compiler is there to help protect us from those types of mistakes.

leighmcculloch: I'd replace it with `if let Some(alias) = &self.alias {` or `match &self.alias {`.

leighmcculloch:
The RPC tests look like they're failing for reasons that aren't obvious. They aren't failing on other PRs that are open, so it might be something to do with this change.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,96 @@ pub enum Error {
...
+        let file_path = self.alias_path()?;
+
+        let Some(dir) = file_path.parent() else {
+            return Err(Error::CannotAccessConfigDir);
+        };
+
+        match create_dir_all(dir) {
+            Ok(()) => {}
+            _ => return Err(Error::CannotAccessConfigDir),
+        };
```
willemneal, line 184:
```suggestion
        create_dir_all(dir).map_err(|_| Err(Error::CannotAccessConfigDir))?;
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,96 @@ pub enum Error {
...
+        let alias = self.alias();
+        let file_name = format!("{alias}.json");
+
+        Ok(config_dir
+            .join("contract-ids")
+            .join(network)
+            .join(file_name))
+    }
+
+    fn save_contract_id(&self, contract: &String) -> Result<(), Error> {
```
willemneal, line 170:
```suggestion
    fn save_contract_id(&self, contract: &str) -> Result<(), Error> {
```
See other comment about this.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,96 @@ pub enum Error {
...
+        if self.alias().is_empty() {
+            return Ok(());
+        }
+
+        let file_path = self.alias_path()?;
+
+        let Some(dir) = file_path.parent() else {
+            return Err(Error::CannotAccessConfigDir);
+        };
```
willemneal, line 179:
```suggestion
        let dir = file_path.parent().ok_or(Err(Error::CannotAccessConfigDir))?;
```
In this case you can convert the `Option` to `Result`, where Some -> Ok, and then you provide the error for the `None` case. There is also `ok_or_else` which allows you to pass a function that returns an error.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,96 @@ pub enum Error {
...
+            Ok(()) => {}
+            _ => return Err(Error::CannotAccessConfigDir),
+        };
+
+        let mut to_file = OpenOptions::new()
+            .create(true)
+            .truncate(true)
+            .write(true)
+            .open(file_path)
+            .map_err(Error::Io)?;
```
willemneal, line 191:
```suggestion
            .open(file_path)?;
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
willemneal: I would suggest. This acts as an assignment, but has the escape hatch.
```rust
    let Some(alias) = &self.alias else { return Ok(()); }
```

I'd also add alias as an argument to `alias_path` so
```rust
```rust
    let Some(alias) = &self.alias else { return Ok(()); }
    let file_path = self.alias_path(alias); // it's already a reference here
```
This way you don't need to worry about the existence of the alias in the other method.

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -165,3 +167,8 @@ pub enum SpecOutput {
     /// Pretty print of contract spec entries
     Docs,
 }
+
+#[derive(Serialize, Deserialize)]
+pub struct AliasData {
+    id: String,
+}
```
willemneal, line 174: It would require creating a new Error type (perhaps it's worth moving to a separate file under `contract/config`, but then you could move your validation logic to parsing
```
impl FromStr for AliasData {

   type Error = Error; // The local error type which will include `Error::InvalidAliasFormat`
   
    fn from_str(s: &str) -> Result<Self, Self::Error> {
        let regex = Regex::new(r"^[a-zA-Z0-9_-]{1,30}$")?; // Here you should add the regex error (also there should be a way to have this done at compile time to avoid a runtime check on the validity of the regex)
        let alias = s.to_owned(); // My personal preferred way of going `&str` to `String`, as `String` is the owned version of the reference.
        if regex.is_match(&alias) {
            Ok(Self { id: alias } )
        } else {
            Err(Error::InvalidAliasFormat { alias })
        }
    }

}
```
See https://doc.rust-lang.org/std/str/trait.FromStr.html
There is then the `parse` method on `&str` which generic but requires that the `T` implements `FromStr`.

Then you would have above:

```rust
        let payload = AliasData {
            id: contract.into(),
        };
```
Becomes
```rust
        let payload: AliasData = contract.parse()?;
```
The compiler can then infer the type for `parse` so it's the same as 
```rust
        let payload = contract.parse::<AliasData>()?;
```

This article called [Parse, don't Validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) uses haskell to explain it, but the TL;DR is that by parsing a type, it pushing the validation of the type to it's creation avoid footguns and creating a clear error.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,83 @@ pub enum Error {
...
             TxnEnvelopeResult::Res(contract) => {
+                self.save_contract_id(&contract)?;
                 println!("{contract}");
             }
         }
         Ok(())
     }
+
+    fn validate_alias(&self) -> Result<(), Error> {
+        match self.alias.clone() {
```
willemneal, line 139: Below I suggest moving this to parsing the `AliasData` type instead of validating. However, in this case you should
not clone (only in the case of the `Error`). And again in this case `let Some(alias) = &self.alias else { return Ok(()) }`.

willemneal (changes requested):

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -100,19 +108,83 @@ pub enum Error {
...
+                    Err(Error::InvalidAliasFormat { alias })
+                }
+            }
+            None => Ok(()),
+        }
+    }
+
+    fn alias_path_for(&self, alias: &str) -> Result<PathBuf, Error> {
+        let config_dir = self.config.config_dir()?;
+        let network = self.config.network.network.clone().expect("must be set");
```
willemneal, line 155: This is the case of the RPC failures since a lot of tests don't use a network alias, rather the rpc_url and network_passphrase directly. I think the better solution is to use `config.get_network()`, which returns the resolved network (the name is used to look it up). Then you could use the network passphrase to determine which network it is. Perhaps you could even add it to the `AliasData` type. e.g.

```rust
struct AliasData {
    ids: BTreeMap<String, String>;
}
```
This would then map the network passphrase to ids, then you could have the contract-ids folder be flat.

cmd/soroban-cli/src/commands/contract/mod.rs
willemneal: Oops just realized that here `id` is a contract id! In this case you would want to move the contract_id parsing logic here. 
```rust
impl FromStr for AliasData {

   type Error = Error; // The local error type which will include `Error::InvalidAliasFormat`
   
    fn from_str(s: &str) -> Result<Self, Self::Error> {
        let id = soroban_spec_tools::utils::contract_id_from_str(&contract_id).map_err(|e| Error::CannotParseContractId(contract_id.clone(), e))?;
        Ok(Self{ id })
    }

}
```


Then you could create a method 

```rust
impl AliasData {
    fn write_to_file(&self, alias: &str, dir: &Path) -> Result<(), Error> {
        // Check the alias here
    }
}

File: cmd/soroban-cli/src/commands/contract/deploy/wasm.rs (hunk 22 above)
Comment (fnando) [reply cmd/soroban-cli/src/commands/contract/deploy/wasm.rs:155]: Good call. Leigh made a comment somewhere saying that arbitrary networks wouldn't work as well, so the fair assumption we can make (and state in the docs), is that alias work for `testnet`, `pubnet`, `futurenet` and `local`.

Comment (willemneal) [reply cmd/soroban-cli/src/commands/contract/deploy/wasm.rs:155]: We could do `alias.testnet` etc. However, the network passphrases are consistent so would be an okay key to use in the map IMO. @leighmcculloch might think otherwise though (I normally avoid spaces in string keys), but we still could also map the network passphrase to `testnet` and use those as keys.

File: cmd/soroban-cli/src/commands/contract/mod.rs (hunk 20 above)
Comment (willemneal) [reply cmd/soroban-cli/src/commands/contract/mod.rs:174]: Also note that the `FromStr` wouldn't make sense if we changed the struct to have a Map.  So I would instead create a `new` method that returns `Result<Self, Error>`.

Comment (fnando) [reply cmd/soroban-cli/src/commands/contract/mod.rs:174]: I don't think we can follow this pattern, as we need to validate the alias before we make the call to RPC. Otherwise you'd need to fail on a operation that has already succeeded.

File: cmd/soroban-cli/src/commands/contract/deploy/wasm.rs (hunk 22 above)
Comment (leighmcculloch) [reply cmd/soroban-cli/src/commands/contract/deploy/wasm.rs:155]: Arbitrary networks would work fine for this. I see right now the alias is stored in file inside a 'network' directory. I think you can store aliases in a single namespace for all networks. This is similar to how we store aliases for keys across all networks. Then store the network _passphrase_ not the network name, in the .json for the alias. If anywhere we use the alias we use it with a network with a passphrase not matching the recorded network passphrase, we can show an error.

Comment (leighmcculloch) [reply cmd/soroban-cli/src/commands/contract/deploy/wasm.rs:155]: I would avoid a map where a single aliases can be mapped to contracts on different networks. It feels like it steps over a complexity boundary where it starts to become harder for a user to grasp what's happening and how it works. A simple 1-1 where alias maps to contract ID on one network is easier to understand. I understand it is not as powerful, but we can make this more powerful in the future if evidence shows it's a worthy tradeoff.

**Comment by fnando** [comment]
@leighmcculloch @willemneal I've changed the alias file to be scoped by network passphrases. This means we'll have only one file for everything.

```console
$ soroban contract deploy --wasm ./target/wasm32-unknown-unknown/release/hello_world.wasm --network standalone --source alice --alias alice --global
CCOZHG2BHABPGVIYBJP675PHDFCT5LPVVY5GAX5CGOWI274DAN6PPUZ5

$ cat ~/.config/soroban/contract-ids/alice.json | jq
{
  "ids": {
    "Test SDF Network ; September 2015": "CDMXLF7V4XQ4QZ62D6MT2J2SGY4YLBHPOQTT3WPXRHTGQ4CUFUE3JW5E",
    "Standalone Network ; February 2017": "CCOZHG2BHABPGVIYBJP675PHDFCT5LPVVY5GAX5CGOWI274DAN6PPUZ5"
  }
}

$ soroban contract invoke --network testnet --source alice --id alice --global -- hello --to fnando
["Hello","fnando"]

$ soroban contract invoke --network standalone --source alice --id alice --global -- hello --to fnando
["Hello","fnando"]
```

Can you please review the latest changes and suggest only extremely required changes? I'd like to merge this in so I can fix the contract init's generated file before the next CLI release (today-tomorrow). :)

leighmcculloch (approved):

PR #1359, opened 2024-06-05
Fix: exclude host only functions from client

BlaineHeffron:
### What

Exclude functions starting with `__` from the typescript bindings Client methods.  Fix for https://github.com/stellar/stellar-cli/issues/1344

### Why

These functions are intended to only be run by the host, they may contain names like `Context` which cannot be imported by the client. 

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-spec-typescript/src/lib.rs
+++ b/cmd/crates/soroban-spec-typescript/src/lib.rs
@@ -96,13 +96,10 @@ pub fn generate(spec: &[ScSpecEntry]) -> String {
             cases: vec![],
         });
     }
-    // Filter out function entries with names that start with "__"
-    let filtered_collected: Vec<_> = collected
+    // Filter out function entries with names that start with "__" and partition the results
+    let (fns, other): (Vec<_>, Vec<_>) = collected
         .into_iter()
         .filter(|entry| matches!(entry, Entry::Function { name, .. } if !name.starts_with("__")))
-        .collect();
-    let (fns, other): (Vec<_>, Vec<_>) = filtered_collected
-        .into_iter()
         .partition(|entry| matches!(entry, Entry::Function { .. }));
     let top = other.iter().map(entry_to_method_type).join("\n");
     let bottom = generate_class(&fns, spec);
--- a/cmd/crates/soroban-spec-typescript/src/lib.rs
+++ b/cmd/crates/soroban-spec-typescript/src/lib.rs
@@ -96,7 +96,12 @@ pub fn generate(spec: &[ScSpecEntry]) -> String {
             cases: vec![],
         });
     }
-    let (fns, other): (Vec<_>, Vec<_>) = collected
+    // Filter out function entries with names that start with "__"
+    let filtered_collected: Vec<_> = collected
+        .into_iter()
+        .filter(|entry| !matches!(entry, Entry::Function { name, .. } if name.starts_with("__")))
+        .collect();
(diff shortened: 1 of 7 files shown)


cmd/crates/soroban-spec-typescript/src/lib.rs
```diff
@@ -96,7 +96,12 @@ pub fn generate(spec: &[ScSpecEntry]) -> String {
             cases: vec![],
         });
     }
-    let (fns, other): (Vec<_>, Vec<_>) = collected
+    // Filter out function entries with names that start with "__"
+    let filtered_collected: Vec<_> = collected
+        .into_iter()
+        .filter(|entry| !matches!(entry, Entry::Function { name, .. } if name.starts_with("__")))
```
willemneal, line 102:
```suggestion
        .filter(|entry| matches!(entry, Entry::Function { name, .. } if !name.starts_with("__")))
```
Nice! I didn't realize that if gaurds worked with the `matches` macro, which makes sense.

cmd/crates/soroban-spec-typescript/src/lib.rs
```diff
@@ -96,7 +96,12 @@ pub fn generate(spec: &[ScSpecEntry]) -> String {
...
+        .collect();
```
willemneal, line 103: You don't need to collect since you immediately convert back into an iterator.

cmd/crates/soroban-test/tests/it/integration/util.rs
```diff
@@ -33,6 +34,10 @@ pub async fn deploy_swap(sandbox: &TestEnv) -> String {
     deploy_contract(sandbox, SWAP).await
 }
 
+pub async fn deploy_custom_account(sandbox: &TestEnv) -> String {
```
willemneal, line 37: Not for this PR, but a good small one you could do is add `deploy` to `Wasm` that takes a `&TestEnv`.
This we don't need a function per file.  e.g. `HELLO_WORLD.deploy(sandbox)`.

chadoh (approved):
Nice! Once you implement Willem's suggested changes, this looks great 👏🏼

cmd/crates/soroban-spec-typescript/src/lib.rs
BlaineHeffron: good call

cmd/crates/soroban-test/tests/it/integration/util.rs
BlaineHeffron: Okay, noted

BlaineHeffron:
Willem's suggestions have been committed.

cmd/crates/soroban-test/tests/fixtures/test-wasms/custom_account/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+name = "test_custom_account"
+version = "21.0.0-rc.1"
+authors = ["Stellar Development Foundation <info@stellar.org>"]
+license = "Apache-2.0"
+edition = "2021"
+publish = false
+rust-version.workspace = true
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
willemneal, line 11:
```suggestion
crate-type = ["cdylib"]
```
Looks like the hello_world contract had this mistake too.

willemneal (approved):
Yay iterators! Let them flow!

PR #1360, opened 2024-06-06
fix incorrect filter logic

BlaineHeffron:
### What

Fixes a bug introduced in https://github.com/stellar/stellar-cli/pull/1359

### Why

This commit: [8976b6c](https://github.com/stellar/stellar-cli/pull/1359/commits/8976b6cc5f7f91577b487871b2f240169e4eb2a2) needs to be reverted. The original logic was correct. Now all non-functions are being excluded from the typescript generation; I noticed this when working on https://github.com/stellar/stellar-cli/issues/1346 and found that no error enum was being generated. This is because its excluding all function names with __ AND its excluding all Entries that fail the function pattern match. 

### Known limitations

none


willemneal (approved):
Good catch my bad!

PR #1361, opened 2024-06-06
Feat/add container log tailing cmd

elizabethengelman:
### What

* closes https://github.com/stellar/stellar-cli/issues/1246
* adds a command that allows a user to tail the logs of the quickstart container that is started with `network start`: `network container logs <NETWORK`
* This pr also moves the following commands under a new `container` subcommand:
    * `network start <NETWORK>` -> `network container start <NETWORK>`
    * `network stop <NETWORK>` -> `network container stop <NETWORK>` 
* The original `network start` and `network stop` still work, but now have a deprecation warning. We should remove them at the next major release. Here's a ticket to capture that: https://github.com/stellar/stellar-cli/issues/1363

### Why

It would allow users to see what's going on in the container, which will be especially helpful when they are starting up testnet or pubnet, which both take some time to start. 

### Known limitations

N/A


cmd/soroban-cli/src/commands/network/container.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Tail logs of a running network container
+    Logs(logs::Cmd),
+    /// Start network
+    ///
+    /// Start a container running a Stellar node, RPC, API, and friendbot (faucet).
+    ///
+    /// soroban network start <NETWORK> [OPTIONS]
+    ///
+    /// By default, when starting a testnet container, without any optional arguments, it will run the equivalent of the following docker command:
+    /// docker run --rm -p 8000:8000 --name stellar stellar/quickstart:testing --testnet --enable-soroban-rpc
```
leighmcculloch, line 22: This is actually out-of-date, we should probably pass `--enable rpc,horizon` or just `--enable rpc`. The command here still works, but it will run everything and we don't need everything.

cmd/soroban-cli/src/commands/network/container/logs.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub network: Network,
+
+    #[arg(short = 'd', long, help = DOCKER_HOST_HELP, env = "DOCKER_HOST")]
+    pub docker_host: Option<String>,
+}
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
+        let container_name = format!("stellar-{}", self.network);
+        println!("ℹ️  Tailing logs for {container_name}");
```
leighmcculloch, line 29: Since this output is not the logs, I would either remove it, or send it to stderr at least, so that stdout isn't impaired by it. But, given that it looks like we actually log out the stderr and stdout, it might be important we don't add any of our own logs to the output either.
```suggestion
        eprintln!("ℹ️  Tailing logs for {container_name}");
```
or

```suggestion
```

cmd/soroban-cli/src/commands/network/container/logs.rs
```diff
@@ -0,0 +1,52 @@
...
+                tail: "all",
+                ..Default::default()
+            }),
+        );
+
+        while let Some(log) = logs_stream.try_next().await? {
+            print!(
+                "{}: {} {}",
+                container_name,
+                Local::now().format("%Y-%m-%d %H:%M:%S"),
```
leighmcculloch, line 46: I'd lean towards not augmenting the logs that we output. If the YMD is useful, we could add them directly to quickstart.

I'd lean towards leaving the container name out too. Screen real estate is valuable and it adds additional width. As far as I can tell we'll only output logs for a single instance at a time, so we don't have to distinguish between them.

Let me know if I'm missing something.

cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -86,8 +97,18 @@ impl Cmd {
             Cmd::Add(cmd) => cmd.run()?,
             Cmd::Rm(new) => new.run()?,
             Cmd::Ls(cmd) => cmd.run()?,
-            Cmd::Start(cmd) => cmd.run().await?,
-            Cmd::Stop(cmd) => cmd.run().await?,
+            Cmd::Container(cmd) => cmd.run().await?,
+
+            // TODO Remove this once `network start` is removed
+            Cmd::Start(cmd) => {
+                println!("⚠️ Warning: `network start` has been deprecated. Use `network container start` instead");
```
leighmcculloch, line 104: I think this should go to stderr because it's not the main output of the command.
```suggestion
                eprintln!("⚠️ Warning: `network start` has been deprecated. Use `network container start` instead");
```

leighmcculloch (approved):
Couple comments inline, but nice 👏🏻

cmd/soroban-cli/src/commands/network/container.rs
leighmcculloch: We should probably fix this separately though. Apologies, wasn't meaning to imply it needed fixing here.

cmd/soroban-cli/src/commands/network/container/logs.rs
elizabethengelman: Good point! I'll remove it.

cmd/soroban-cli/src/commands/network/container/logs.rs
elizabethengelman: That is a good point. I was trying to match the output I was seeing in Docker Desktop.. but it looks like they are adding YMD themselves, which we don't see when just attaching to a container. I'll remove the YMD, and the container name!

elizabethengelman: Nice! I've add a new issue for this: https://github.com/stellar/stellar-cli/issues/1365

cmd/soroban-cli/src/commands/network/container.rs
elizabethengelman: Nice! I've add a new issue for this: https://github.com/stellar/stellar-cli/issues/1365

PR #1362, opened 2024-06-06
feat: update to newest soroban-rpc and copy over old signing logic

willemneal:
### What

~Add a Stellar trait for signing (up for suggestions for a better name `Signer` seemed overloaded). This will allow for other implementations like the Ledger in a future PR. An initial implementation of the trait `LocalKey` uses a local signing key and has the ability to prompt the user to confirm the signature. Currently this defaults to false to match the current behavior.~

Also updated to use the newest soroban-rpc and used the new signer for all relevant commands. And copied over the signing code from the RPC to make this PR more minimal.

### Why

The soroban-rpc was updated to remove all logic that didn't pertain to core RPC methods. 

### Known limitations

[TODO or N/A]


cmd/soroban-cli/src/commands/config/mod.rs
```diff
@@ -49,6 +60,42 @@ impl Args {
         Ok(key.key_pair(self.hd_path)?)
     }
 
+    pub async fn sign_with_local_key(&self, tx: Transaction) -> Result<TransactionEnvelope, Error> {
+        let signer = LocalKey::new(self.key_pair()?, false);
```
leighmcculloch, line 64: Everywhere I see local key used the prompt is set to false. Do we use the prompting at all, or is that dead code in this change?

leighmcculloch (commented):
I left some comments inline. Mostly minor. A couple areas of commented code and it's unclear if the code is incomplete or what the commented code is for.

There's one area that reads like a bug to me but I might be missing something. (https://github.com/stellar/stellar-cli/pull/1362#discussion_r1630723766)

Can we remove the prompting code from this change since it's dead code? Or comment it. It includes a dependency that won't be used.

This change contains more new code than I think we anticipated this late in the release process that touches a high sensitive area. But I agree it looks like it doesn't change the experience because the new functionality is disabled, but I'm just guessing because the code isn't identical, and I can't tell for sure, especially given the possible bug.

I think this PR is heading in the right direction for signing, and is what we'd want to head towards.

My preference is that we verbatim copy-paste the code that was deleted from https://github.com/stellar/soroban-rpc/commit/587a229541043ca3b2e3e4514296d6c87625b25c and paste it into the stellar-cli, and use that for the final v21 release, then move forward with merging this PR for v22.

(The prompting is so helpful from a security stand point we might actually want to release this in a v21 minor version as a breaking change, so I don't think we'd be delaying merging this till v22.)

cc @fnando @chadoh ☝🏻  if you have alternate opinions.

cmd/soroban-cli/src/signer.rs
```diff
@@ -0,0 +1,260 @@
...
+
+    async fn get_public_key(&self) -> Result<stellar_strkey::ed25519::PublicKey, Error> {
+        Ok(stellar_strkey::ed25519::PublicKey(
+            self.key.verifying_key().to_bytes(),
+        ))
+    }
+}
+
+pub fn read_key() -> char {
+    let tty = get_tty().unwrap();
```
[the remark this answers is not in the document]
willemneal: Yeah in this case it should fail in a script if they didn't include `-y`

willemneal:
@leighmcculloch I think I'll take your idea of just copying the current signing logic from the RPC. It should be pretty easy since the signing now happens through the config, so the callers wouldn't need to be updated. Will work on that now.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -127,7 +127,7 @@ impl NetworkRunnable for Cmd {
     ) -> Result<TxnResult<String>, Error> {
         let config = config.unwrap_or(&self.config);
         let wasm_hash = if let Some(wasm) = &self.wasm {
-            let hash = if self.fee.build_only {
+            let hash = if self.fee.build_only || self.fee.sim_only {
```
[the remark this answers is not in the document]
willemneal: I removed this from this PR. I had changed that in the past for some reason but can't recall, but I think it should be in another PR.

cmd/soroban-cli/src/signer.rs
```diff
@@ -0,0 +1,260 @@
...
+impl LocalKey {
+    pub fn new(key: ed25519_dalek::SigningKey, prompt: bool) -> Self {
+        Self { key, prompt }
+    }
+}
+
+impl Stellar for LocalKey {
+    async fn sign_blob(&self, data: &[u8]) -> Result<Vec<u8>, Error> {
+        if self.prompt {
+            eprintln!("Press 'y' or 'Y' for yes, any other key for no:");
```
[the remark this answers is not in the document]
willemneal: I'll update this in a future PR.

willemneal:
@leighmcculloch This should be good for now and I'll add all the rest in a separate PR.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -381,21 +377,15 @@ impl NetworkRunnable for Cmd {
...
+            }
+            // let txn = self.config.sign_with_local_key(tx)
             let res = client
-                .send_assembled_transaction(
-                    txn,
-                    &unwrap_config.key_pair()?,
-                    &signers,
-                    &network.network_passphrase,
-                    Some(log_events),
-                    (verbose || very_verbose || self.fee.cost).then_some(log_resources),
```
leighmcculloch, line 397: Are we no longer logging events and resources?

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -381,21 +378,20 @@ impl NetworkRunnable for Cmd {
...
+            let mut txn = txn.transaction().clone();
+            for signer in signers.into_iter().map(|key| LocalKey::new(key, false)) {
+                if let Some(tx) = config
+                    .sign_soroban_authorizations(&signer, txn.clone())
+                    .await?
+                {
+                    txn = tx;
+                }
+            }
+            // let txn = self.config.sign_with_local_key(tx)
```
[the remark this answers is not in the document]
leighmcculloch: ?
```suggestion
```

cmd/soroban-cli/src/commands/config/mod.rs
```diff
@@ -49,6 +60,40 @@ impl Args {
...
+    }
+
+    pub async fn sign(&self, mut tx: Transaction) -> Result<TransactionEnvelope, Error> {
+        let key = self.key_pair()?;
+        let verifying_key = key.verifying_key().to_bytes();
+        let account = Strkey::PublicKeyEd25519(stellar_strkey::ed25519::PublicKey(verifying_key));
+        let network = self.get_network()?;
+        let client = Client::new(&network.rpc_url)?;
+        tx.seq_num = SequenceNumber(client.get_account(&account.to_string()).await?.seq_num.0 + 1);
+        tx.source_account = MuxedAccount::Ed25519(Uint256(verifying_key));
```
leighmcculloch, line 74: Why do we overwrite the seq num and source account here during signing? These should have been set earlier in all cases. Signing should just involve attaching a signature of the tx.

leighmcculloch (commented):
Thanks. The change as a whole looks more like a copy-paste from https://github.com/stellar/soroban-rpc/commit/587a229541043ca3b2e3e4514296d6c87625b25c. 

Couple questions inline but I think this is largely ready to merge.

cmd/soroban-cli/src/commands/contract/invoke.rs
willemneal: It was removed from the RPC crate. My thoughts here was that we would move the logging here since the RPC crate shouldn't do any logging imo. I'll add back the event logging. However, the resources should be now part of the action cache.

cmd/soroban-cli/src/commands/config/mod.rs
willemneal: The idea here was that this uses the signing, which does just that, however, all all will need an accurate seq_num and given that the transaction could be out of date, I found it best to set it right before signing so that we won't get an error when sending a transaction with a bad seq_num. So this was more for the `tx sign`.

The source_account should be removed, especially because we should add a `sign_with` argument to allow for signing with a different key than the source account.

Ideally we would have a transaction builder *cough cough* as the logic of fetch and setting the seq_num is repeated over and over.

leighmcculloch: I don't think we should modify the tx during the signing step. Signing in other tools is only a modification of the envelope, not the tx itself, which importantly means the tx hash is consistent, and the CLI magically modifying the seq would be surprising because it's not a different tx.

Agreed, the source account shouldn't be modified either for the reasons you mentioned.

cmd/soroban-cli/src/commands/contract/invoke.rs
leighmcculloch: Can we keep the logs for both the events and resources?

Even though resources are in the action cache, folks could still be invoking with logs enabled to visually see resources.

cmd/soroban-cli/src/commands/config/mod.rs
leighmcculloch:
```suggestion
```

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -446,6 +440,34 @@ impl NetworkRunnable for Cmd {
...
+//         .and_then(|op| match op.body {
+//             OperationBody::InvokeHostFunction(ref body) => (matches!(
+//                 body.auth.first().map(|x| &x.root_invocation.function),
+//                 Some(&SorobanAuthorizedFunction::ContractFn(_))
+//             ))
+//             .then_some(body.auth.clone()),
+//             _ => None,
+//         })
+//         .unwrap_or_default()
+// }
```
chadoh, line 469: is it useful to have these comments here?

chadoh (approved):
Looks good! One small cleanup nit, but no need to address it prior to merging.

PR #1369, opened 2024-06-11
Extract alias logic into its own implementation.

fnando:
### What

Extract alias logic into its own implementation.

### Why

So we can easily reuse it in other commands (e.g. `contract bindings typescript` will also load the contract id alias).

### Known limitations

N/A

#### Storing alias locally

```console
$ soroban contract deploy --wasm ./target/wasm32-unknown-unknown/release/hello_world.wasm --network standalone --source alice --alias alice
CDMCQHHE356XOGBXXUTZ3AVVARRTRYKHFVYAC6UYWM7RYPKSYV2FUPDS

$ cat .soroban/contract-ids/alice.json | jq
{
  "ids": {
    "Standalone Network ; February 2017": "CDMCQHHE356XOGBXXUTZ3AVVARRTRYKHFVYAC6UYWM7RYPKSYV2FUPDS"
  }
}

$ soroban contract invoke --network standalone --source alice --id alice -- hello --to fnando
["Hello","fnando"]
```

#### Storing alias globally

```console
$ soroban contract deploy --wasm ./target/wasm32-unknown-unknown/release/hello_world.wasm --network standalone --source alice --alias alice --global
CDNSYIZ7IUPFGP72DATK4DYOU5ZKPUBBHBTAIFUBTCEDK3RR5WRQOP7F

$ cat ~/.config/soroban/contract-ids/alice.json | jq
{
  "ids": {
    "Standalone Network ; February 2017": "CDNSYIZ7IUPFGP72DATK4DYOU5ZKPUBBHBTAIFUBTCEDK3RR5WRQOP7F"
  }
}

$ soroban contract invoke --network standalone --source alice --id alice --global -- hello --to fnando
["Hello","fnando"]
```


cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -174,3 +180,101 @@ pub enum SpecOutput {
...
+        let alias_data = Self::load(config_dir, Some(alias))?;
+        let Some(alias_data) = alias_data else {
+            return Ok(None);
+        };
+
+        match alias_data.ids.get(network_passphrase) {
+            Some(id) => Ok(Some(id.into())),
+            _ => Ok(None),
+        }
+    }
```
willemneal, line 265:
```suggestion
        let Some(alias_data) = Self::load(config_dir, Some(alias))? else {
            return Ok(None);
        };
        Ok(alias_data.ids.get(network_passphrase).cloned())
    }
```

willemneal:
Looks good so far. One suggestion is to move to it's own file under `config`. Then you can name the types `Error` and `Data`. Then externally, it would be `alias::Error` and `alias::Data`.

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -174,3 +180,101 @@ pub enum SpecOutput {
...
+
+    pub fn load_contract_id_or_default(
+        alias: &str,
```
willemneal, line 268: It is a little confusing that this could also be a contract strkey. So I suggest this just return an `Option` of a contract Strkey. But for now I'd just change the parameter name to make it clearer.

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -174,3 +180,101 @@ pub enum SpecOutput {
...
+        config_dir: &Path,
+        network_passphrase: &str,
+    ) -> Result<[u8; 32], AliasDataError> {
+        let contract_id = match Self::get_contract_id(alias, config_dir, network_passphrase)? {
+            None => alias.to_string(),
+            Some(id) => id,
+        };
```
willemneal, line 275:
```suggestion
        let contract_id  = Self::get_contract_id(alias, config_dir, network_passphrase)?.unwrap_or_else(|| alias.to_string());
```
Pretty much can always get away with a helper method instead of matching on an Option.

cmd/soroban-cli/src/commands/contract/mod.rs
fnando: Can you please explain why returning an `Option` be more clearer? The idea is avoiding this common task on the caller, as defaulting to the input address when no alias can be found will always be the desirable outcome.

willemneal: Personally I think it's a clearer API if the alias tries to resolve the alias and it's None if not there.

Then you could have a method on the config `Arg` that would handle this higher level. TL;DR keep this part simple and then the callers in `invoke` etc can do `config.resolve_contract_id(&self.contract_id)` and it will handle the last part of this method.

cmd/soroban-cli/src/commands/config/alias.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error("cannot access config dir for alias file")]
+    CannotAccessConfigDir,
+    #[error("cannot parse contract ID {0}: {1}")]
+    CannotParseContractId(String, DecodeError),
+}
+
+impl Data {
+    pub fn load(config_dir: &Path, alias: Option<&str>) -> Result<Option<Self>, Error> {
```
willemneal, line 29: Does alias need to an option here?

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -128,7 +120,13 @@ impl Cmd {
         match res {
             TxnEnvelopeResult::TxnEnvelope(tx) => println!("{}", tx.to_xdr_base64(Limits::none())?),
             TxnEnvelopeResult::Res(contract) => {
-                self.save_contract_id(&contract)?;
+                alias::Data::save_contract_id(
+                    &self.config.config_dir()?,
+                    &contract,
+                    self.alias.as_ref(),
```
willemneal, line 126:
```suggestion
                    &self.alias,
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -128,7 +120,13 @@ impl Cmd {
         match res {
             TxnEnvelopeResult::TxnEnvelope(tx) => println!("{}", tx.to_xdr_base64(Limits::none())?),
             TxnEnvelopeResult::Res(contract) => {
-                self.save_contract_id(&contract)?;
+                alias::Data::save_contract_id(
```
willemneal, line 123: Given the dependency on `self.config`, then this could become. `self.config.save_contract_id(&alias, &contract)?;`

willemneal:
Looking good! Just one more suggestion of moving the logic to `config` this way we later if it becomes it's own crate we can use it to load_contract_ids from aliases to import them in smart contracts.

willemneal:
@leighmcculloch @fnando One more thought designwise, is it possibly  confusing to users that `--contract-id` could be an `--contract-alias` and move this into its own `Arg`? This commands that need a contract_id 

```rust
 Cmd {
    #[flatten]
    contract_id: contract_id::Arg,
    /// ...
    }
 ```
 
 and it could be 

```rust
struct Arg {
    #[arg(long = "contract-id", required_unless_present = "contract-alias")]
    pub contract_id: Option<String>,
    #[arg(long = "contract-alias", conflicts_with = "contract-id")]
    pub contract_alias: Option<String>,
}

impl Arg {
    pub fn resolve(&self, config: &config::Args) -> Result<Strkey::Contract, Error>
}
```

Then for all commands that use the contract id it would be `self.contract_id.resolve(&config)?`.

I'm totally fine either way just always am a fan of putting shared logic into flattened Args and not overloading args.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: Tried this, but clippy complains about it, so I applied the suggestion:

```
error[E0308]: mismatched types
   --> cmd/soroban-cli/src/commands/contract/deploy/wasm.rs:126:21
    |
123 |                 alias::Data::save_contract_id(
    |                 ----------------------------- arguments to this function are incorrect
...
126 |                     &self.alias,
    |                     ^^^^^^^^^^^ expected `Option<&String>`, found `&Option<String>`
    |
    = note:   expected enum `std::option::Option<&std::string::String>`
            found reference `&std::option::Option<std::string::String>`
note: associated function defined here
   --> cmd/soroban-cli/src/commands/config/alias.rs:47:12
    |
47  |     pub fn save_contract_id(
    |            ^^^^^^^^^^^^^^^^
...
50  |         alias: Option<&String>,
    |         ----------------------
help: try using `.as_ref()` to convert `&std::option::Option<std::string::String>` to `std::option::Option<&std::string::String>`
    |
126 -                     &self.alias,
126 +                     self.alias.as_ref(),
    |

For more information about this error, try `rustc --explain E0308`.
error: could not compile `soroban-cli` (lib) due to 1 previous error
```

cmd/soroban-cli/src/commands/config/alias.rs
fnando: it doesn't. This was due to porting over the code from the other file.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: I moved everything to self.config. Should I leave all functions in alias, or actually move them to mod.rs?

fnando:
@willemneal as for `--contract-alias`, I think we shouldn't add it, just because is yet another thing to keep in mind when using the CLI. But one I think we should do is improving the documentation for `--contract-id` mentioning the support for aliasing.

As soon as the alias work is done, I plan to update the docs with a section for contract id aliasing, which should also help with adoption.

cmd/soroban-cli/src/commands/config/alias.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn alias_path(&self, alias: &str) -> Result<PathBuf, Error> {
+        let file_name = format!("{alias}.json");
+        let config_dir = self.config_dir()?;
+        Ok(config_dir.join("contract-ids").join(file_name))
+    }
+
+    pub fn save_contract_id(&self, contract_id: &str, alias: Option<&String>) -> Result<(), Error> {
+        let Some(alias) = alias else {
+            return Ok(());
+        };
```
leighmcculloch, line 57: I don't feel strongly about this, but the guard probably belongs elsewhere, if the alias is an option and empty, calling code should probably just not call this code.

cmd/soroban-cli/src/commands/config/alias.rs
```diff
@@ -0,0 +1,107 @@
...
+        network_passphrase: &str,
+    ) -> Result<Option<String>, Error> {
+        let Some(alias_data) = self.load(alias)? else {
+            return Ok(None);
+        };
+
+        Ok(alias_data.ids.get(network_passphrase).cloned())
+    }
+
+    pub fn resolve_contract_id(
```
leighmcculloch, line 95: Nice choice of function name 👏🏻

leighmcculloch (approved):
Just one piece of feedback inline, but regardless looks good to merge.

cmd/soroban-cli/src/commands/config/alias.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[error("cannot access config dir for alias file")]
+    CannotAccessConfigDir,
+    #[error("cannot parse contract ID {0}: {1}")]
+    CannotParseContractId(String, DecodeError),
+    #[error(transparent)]
+    Config(#[from] config::Error),
+}
+
+impl Args {
+    pub fn load(&self, alias: &str) -> Result<Option<Data>, Error> {
```
willemneal, line 35:
```suggestion
    pub fn load_alias_data(&self, alias: &str) -> Result<Option<Data>, Error> {
```
Now this is in `super::Args`, it's too general. Also does this need to be public?

willemneal (changes requested):
@fnando Looking really great! Much cleaner for callers to use it. Just one nit.

cmd/soroban-cli/src/commands/config/alias.rs
fnando: nice catch!

fnando:
@willemneal can i get a 👍 ?

willemneal (approved):
Still think that `load` is too general, but it's now no longer public so we can address in a future PR.  Good work!

PR #1371, opened 2024-06-11
Fix/update network container start command to use updated enable flags

elizabethengelman:
### What

- Update the enable flag that is passed to the quickstart container.
- Default the local network to use the `testing` quickstart image tag instead of `latest`

### Why

- There was a change made to the quickstart container to allow for optionally starting up services, and this change takes advantage of that to only start up the rpc service and horizon by default.

### Known limitations

N/A


cmd/soroban-cli/src/commands/network/container/start.rs
```diff
@@ -135,9 +135,9 @@ fn get_container_args(cmd: &Cmd) -> Vec<String> {
 fn get_image_name(cmd: &Cmd) -> String {
     // this can be overriden with the `-t` flag
     let mut image_tag = match cmd.network {
-        Network::Testnet => "testing",
+        Network::Pubnet => "latest",
         Network::Futurenet => "future",
-        _ => "latest", // default to latest for local and pubnet
+        _ => "testing", // default to testing for local and testnet
```
elizabethengelman, line 140: @chadoh we had talked about changing this so that the local network would default to `testing` instead of `latest`. I forgot to make that change earlier, but here it is now! 😄

elizabethengelman:
@leighmcculloch related to this change, the [dockerhub description](https://hub.docker.com/r/stellar/quickstart/) for quickstart is a bit out-of-date, and doesn't include info about the `--enable rpc,horizon` flag, etc.  I wonder if we could update that description to be a link to the repo's README instead, so we don't have to worry about updating the description in two places.

leighmcculloch (approved):

chadoh (approved):
Thanks! And I like the idea of making the Dockerhub readme a link to the main README. @leighmcculloch do you also like @elizabethengelman's suggestion?

leighmcculloch:
Interesting. The dockerhub description used to update automatically, something must have broken with the linking. We can use a github action to update the readme on each publish, but I think linking to the GitHub repo would also be simpler. I'll open a ticket with the team who can action that.

elizabethengelman:
closes https://github.com/stellar/stellar-cli/pull/1371

Issue #1318, opened 2024-05-08
Contract read errors with the wasm hash param


leighmcculloch:
### What version are you using?

```
$ soroban version
soroban 21.0.0-preview.1 (a643000a6cc15bf4bdf1e358881e5671834806b7)
soroban-env 21.0.1 (f2b734b6360ae04ba730307be3988619dd0f59a3)
soroban-env interface version 90194313216
stellar-xdr 21.0.1 (da0cbc6f75eaf03423810b8e9d730a03dfa8517e)
xdr curr (1a04392432dacc0092caaeae22a600ea1af3c6a5)
```

a643000a6cc15bf4bdf1e358881e5671834806b7

### What did you do?

Setup:
```
$ soroban network add future \
    --rpc-url https://rpc-futurenet.stellar.org:443 \
    --network-passphrase 'Test SDF Future Network ; October 2022'
$ export SOROBAN_NETWORK=future
$ soroban keys generate me
$ SOROBAN_ACCOUNT=me
```
Reading with a wasm hash:
```
$ soroban contract read \
    --wasm-hash 390c0dc0c56b66db3847d1ab75d8bfda164bda5a7032a049921f5529202ddd95 \
    --durability persistent \
    --network future \
    --source me
error: Only contract data and code keys are allowed
```

### What did you expect to see?

The ledger entry for the wasm hash.

### What did you see instead?

Error.

leighmcculloch:
One problem with us allowing the `read` command to output the key and value of contract code, is that the outputs are now of different types. Before the outputs were always keys and values for contract data. Now it's also contract code.

When I opened #1318 I was confused because I was thinking of the read command reading raw ledger entries, but that's not what the read command does, it's a `contract read` command and it is only reading contract data, not all artbirary ledger entries.

There exists `contract fetch` for reading wasm.

It's confusing that the read command has a `--wasm-hash` parameter. I'm not sure why that it needs a parameter for restoring the contract.

I think my expectations in #1318 are incorrect, and the expectation should be probably to:
1. Remove the `--wasm-hash` and `--wasm` options after understanding why they are present.
1. Fix the error to say that only contract data can be read instead of data and code.

_Originally posted by @leighmcculloch in https://github.com/stellar/stellar-cli/pull/1372#discussion_r1692951351_

PR #1372, opened 2024-06-11
feat: contract read can read contract code

willemneal:
### What

fixes: #1318

### Why

### Known limitations

[TODO or N/A]


cmd/soroban-cli/src/commands/contract/read.rs
```diff
@@ -87,7 +88,7 @@ pub enum Error {
     NoContractDataEntryFoundForContractID,
     #[error(transparent)]
     Key(#[from] key::Error),
-    #[error("Only contract data and code keys are allowed")]
+    #[error("Only contract code, data, and code keys are allowed")]
```
leighmcculloch, line 91: It already says "code".
```suggestion
    #[error("Only contract data, and code keys are allowed")]
```

cmd/soroban-cli/src/commands/contract/read.rs
```diff
@@ -112,20 +113,28 @@ impl Cmd {
             last_modified_ledger,
         } in &entries.entries
         {
-            let (
-                LedgerKey::ContractData(LedgerKeyContractData { key, .. }),
-                LedgerEntryData::ContractData(ContractDataEntry { val, .. }),
-            ) = &(key, val)
-            else {
-                return Err(Error::OnlyDataAllowed);
+            let (key, val) = match (key, val) {
```
leighmcculloch, line 116: One problem with us allowing the `read` command to output the key and value of contract code, is that the outputs are now of different types. Before the outputs were always keys and values for contract data. Now it's also contract code.

When I opened #1318 I was confused because I was thinking of the read command reading raw ledger entries, but that's not what the read command does, it's a `contract read` command and it is only reading contract data, not all artbirary ledger entries.

There exists `contract fetch` for reading wasm.

It's confusing that the read command has a `--wasm-hash` parameter. I'm not sure why that it needs a parameter for restoring the contract.

I think my expectations in #1318 are incorrect, and the expectation should be probably to:
1. Remove the `--wasm-hash` and `--wasm` options after understanding why they are present.
1. Fix the error to say that only contract data can be read instead of data and code.

leighmcculloch (changes requested):
I think the issue I opened #1318 suggested we take a path that's not helpful, so I think we should abandon this change, but address the problem another way way, see inline.

Apologies for the wrong steer on this one.

leighmcculloch (changes requested):
I think the issue I opened #1318 suggested we take a path that's not helpful, so I think we should abandon this change, but address the problem another way way, see inline.

Apologies for the wrong steer on this one.

PR #1373, opened 2024-06-13
Fix some renames

janewang:
In the autogenerated help docs


leighmcculloch (commented):
We'll need to update the code spots where these incorrect names are first, then the doc will update automatically.

leighmcculloch (commented):
We'll need to update the code spots where these incorrect names are first, then the doc will update automatically.

PR #1392, opened 2024-06-19
Threshold Payment Proof of Concept Implementation

Fiono11:
### What

This is a proof of concept implementation of the [Olaf](https://eprint.iacr.org/2023/899) protocol made with this [grant](https://communityfund.stellar.org/project/frost-implementation). Olaf is composed of the SimplPedPoP protocol to generate a threshold account, and the FROST protocol to sign from a threshold account.

### Why

It is much more efficient than the current multisignature protocol used, since the signing is done off chain instead of on chain. The advantages of this are:

- Only one signature needs to be verified, which is independent of the number of signers. This is more scalable (up to 1000 signers) and cheaper.
- More private, since the network does not know which individual accounts partially signed the threshold signature.

### Known limitations

Right now this is only implemented for the Payment operation and the communication needs to be done out-of-band, since networking is not implemented yet. In this draft PR there is a tutorial with an example of a 2-of-2 payment.

Let me know if there is interest in continuing this project. Next steps would be:

- Implement the FROST protocol for all Stellar operations, including Soroban contracts, in this codebase.
- Implement networking with the WebRTC protocol in another crate.
- Integrate it with a wallet from the Stellar ecosystem.


willemneal:
I still need to read this in detail, but it seems awesome so far. I suggest, especially since this as a PoC to implement this as a plugin. You create your own binary and prefix it with `soroban-`. We are adding support for a different prefix, e.g. `stellar-plugin-`, but for now that's the way to do it.

Here is a minimal example of a plugin. https://github.com/stellar/stellar-cli/blob/main/cmd/crates/soroban-test/tests/fixtures/hello/Cargo.toml

This way you can do `soroban <your cli> <other commands and args>`. Let me know if you need any help.

Fiono11:
Thanks! That seems to be the right direction to pursue. Will have a look at it.

PR #1394, opened 2024-06-19
Fix/logs

willemneal:
### What

Depends on #1393.

This separates out the contract events, log events, and diagnostic events.

### Why

In the past users have complained about the verbose output of the diagnostic events because they were all considered INFO level. This makes only contract events and log events INFO level which is what users most likely expect and not the noisy other diagnostic events.

### Known limitations

The output is still not the best as it's just the debug output from the generated XDR. However, this should be done in a future PR.


cmd/soroban-cli/src/log/contract_event.rs
```diff
@@ -0,0 +1,19 @@
+pub fn contract_events(events: &[impl std::fmt::Debug], level: tracing::Level) {
```
leighmcculloch, line 1: There seems to be three files in this PR that have the exact same contents as this file, just with a different name. Why is that?

cmd/soroban-cli/src/log.rs
```diff
@@ -1,13 +1,57 @@
...
+fn is_log_event(event: &xdr::DiagnosticEvent) -> bool {
+    match &event.event.body {
+        xdr::ContractEventBody::V0(xdr::ContractEventV0 { topics, .. }) if topics.len() == 1 => {
+            topics[0] == xdr::ScVal::Symbol(str_to_sc_string("log"))
+        }
+        xdr::ContractEventBody::V0(_) => false,
+    }
+}
+
+fn str_to_sc_string(s: &str) -> xdr::ScSymbol {
```
leighmcculloch, line 54: This function creates an ScSymbol, not an ScString.
```suggestion
fn str_to_sc_symbol(s: &str) -> xdr::ScSymbol {
```

cmd/soroban-cli/src/log.rs
```diff
@@ -1,13 +1,57 @@
...
+pub use log_event::*;
+
+pub fn events(events: &[xdr::DiagnosticEvent]) {
+    let (contract_events, other_events): (Vec<_>, Vec<_>) =
+        events.iter().partition(|e| is_contract_event(e));
+    contract_event::contract_events(&contract_events, tracing::Level::INFO);
+    let (log_events, other_events): (Vec<_>, Vec<_>) =
+        other_events.into_iter().partition(|e| is_log_event(e));
+    log_event::log_events(&log_events, tracing::Level::INFO);
+    diagnostic_event::diagnostic_events(&other_events, tracing::Level::DEBUG);
```
leighmcculloch, line 28: This logic results in the events being printed out-of-order. All of the contract events print, then all of the log events, then all of the diagnostic events. It's pretty important that we preserve the order and keep them interleaved otherwise diagnostic events that occurred prior to a contract event won't be observed as occurring prior.

leighmcculloch (changes requested):
This change results in the events being printed out-of-order. All of the contract events print, then all of the log events, then all of the diagnostic events. It's pretty important we preserve the order and keep them interleaved otherwise diagnostic events that occurred prior to a contract event won't be observed as occurring prior but will appear as if they occurred after.

I think a different approach is needed where the log events function simply iterates (e.g. for loop) over the events logging each one, as it did, but instead of using a hardcoded of INFO, determine the level in that moment to whether it should be info, error, or debug.

cmd/soroban-cli/src/log/contract_event.rs
willemneal: It had been so that you could filter the tests, but I'll update it to be functions within the main file and it should still work.

cmd/soroban-cli/src/log/event.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+pub fn events(events: &[xdr::DiagnosticEvent]) {
+    for (i, event) in events.iter().enumerate() {
+        if is_contract_event(event) {
+            contract::contract(i, event);
+        } else if is_log_event(event) {
+            log::log(i, event);
+        } else {
+            diagnostic::diagnostic(i, event);
```
leighmcculloch, line 29: ❗ Can the log events be `debug` and the diagnostic events should be `trace`?

Contract log events are only outputted from contracts build with a debug flag enabled, and system logs are similar debug information.

Other diagnostic logs are really trace level containing information about each invoke and return value and costs.

cmd/soroban-cli/src/log/event.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+fn is_contract_event(event: &xdr::DiagnosticEvent) -> bool {
+    matches!(event.event.type_, xdr::ContractEventType::Contract)
+}
+
+fn is_log_event(event: &xdr::DiagnosticEvent) -> bool {
+    match &event.event.body {
+        xdr::ContractEventBody::V0(xdr::ContractEventV0 { topics, .. }) if topics.len() == 1 => {
```
leighmcculloch, line 40: ❗ We should probably constrain this to the `Diagnostic` event type too. Although I guess it doesn't matter so much because is_contract_event will catch any `log` topic contract event, but with these functions being separate they look like they're able to standalone when in reality they are tightly coupled on being used together and in a specific order.

cmd/soroban-cli/src/log/event.rs
```diff
@@ -0,0 +1,50 @@
+use crate::xdr;
```
leighmcculloch, line 1: 💡 I suggest writing files like this top-to-bottom, as it is it's more difficult for a reader to jump around. I noticed we've used this micro function jump around strategy in some areas of the CLI and it's harder for a reader unfamiliar with the code to jump in. I think we can get the same benefit of the small functions having names by adding a couple comments.

For example, change as it is:
Screenshot 2024-07-26 at 9 25 54 AM

Vs top-to-bottom with comments:
Screenshot 2024-07-26 at 9 45 36 AM

leighmcculloch (commented):
Two asks (❗) and one suggestion (💡).

cmd/soroban-cli/src/log/event.rs
willemneal: Sure that makes sense. My one thought on the logs being debug is that new devs will use the logs and then be confused when they aren't present. So perhaps we should print a message that indicates that the logs are present and require `--verbose` or `RUST_LOGS=soroban_cli::log::event::log=info` or `--filter-logs ...`. `verbose` will generate more output that they probably want. And the other solution is a bit wordy. We could also add a `SOROBAN_LOGS` env that might be better?

leighmcculloch: > My one thought on the logs being debug is that new devs will use the logs and then be confused when they aren't present.

That's a good point. The problem is that entries matching the `is_log_event` function here include system logs, not just contract logs, so we're pushing logs that are system related into info too. Leave the logs as info and we'll see if we tweak it over time.

leighmcculloch (approved):

PR #1401, opened 2024-06-24
Move alias supporting code to locator.

fnando:
### What

Move alias supporting code to locator.

### Why

So it's easier to use from modules that have access to the locator instead.

### Known limitations

N/A


cmd/soroban-cli/src/commands/config/locator.rs
```diff
@@ -211,6 +220,78 @@ impl Args {
...
+        }
+
+        let content = fs::read_to_string(path)?;
+        let data: alias::Data = serde_json::from_str(&content).unwrap_or_default();
+
+        Ok(Some(data))
+    }
+
+    fn alias_path(&self, alias: &str) -> Result<PathBuf, Error> {
+        let file_name = format!("{alias}.json");
```
willemneal, line 238: IMO this is where the validation should happen. 
https://github.com/stellar/stellar-cli/blob/85ec1ee0a3fac79d09111544e894c14741e14e08/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs#L136

Currently we only allow making an alias with deploy, but we could add a way to add one via a `contract alias` command.

willemneal (approved):
I had one comment, but it could be addressed in a future PR.

cmd/soroban-cli/src/commands/config/alias.rs
```diff
@@ -1,103 +1,8 @@
...
 
 use serde::{Deserialize, Serialize};
-use stellar_strkey::DecodeError;
-
-use crate::commands::config;
-
-use super::Args;
 
 #[derive(Serialize, Deserialize, Default)]
 pub struct Data {
```
willemneal, line 6: At this point you could replace this struct with a type alias `type Alias = HashMap<String, String>` and move it to `config/locator.rs`. This type is already serializable and you won't need to `data.ids` anymore.

cmd/soroban-cli/src/commands/config/locator.rs
fnando: As mentioned in the original pr, we can't move the validation to a later step, otherwise the contract could be deployed with an invalid alias.

> Currently we only allow making an alias with deploy, but we could add a way to add one via a contract alias command.

I'm not so sure about adding yet another command to manage aliases right now. Let's see if people ask for it and then we can consider it.

cmd/soroban-cli/src/commands/config/alias.rs
willemneal: Could also call it `AliasMap`?

fnando: The idea is being open for extensions later on. If we move keys to the root not, that's not possible.

cmd/soroban-cli/src/commands/config/locator.rs
willemneal: Yeah I guess for me deploying was still okay, but without another way to add the alias I get it. I guess I would prefer validation via clap: .g.
```rust
/// The alias that will be used to save the contract's id.
#[arg(long, value_parser = clap::builder::ValueParser::new(alias_validator))]
pub alias: Option<String>,
```

``` rust
fn alias_validator(s: &str) -> Result<String, String> {
    let regex = regex::Regex::new(r"^[a-zA-Z0-9_-]+$").unwrap();
    if regex.is_match(s) {
        Ok(s.to_string())
    } else {
        Err(...)
    }
}
```
This way if we moved alias to its own `Arg` struct it could fail at the parsing step. However, again not an issue until or if we add more commands.

willemneal:
Sorry merged on accident! You can add the alias in another PR!

cmd/soroban-cli/src/commands/config/locator.rs
fnando: oh, i think that's a good change. I didn't do it initially because… because I never bothered to look at clap's docs. 😬🫠

PR #1402, opened 2024-06-24
Use clap's validator instead.

fnando:
### What

Use a clap validator to parse the alias argument.

```console
$ stellar contract deploy --wasm ./target/wasm32-unknown-unknown/release/hello_world.wasm --network standalone --source alice --alias '!foo' --global
error: alias must be 1-30 chars long, and have only letters, numbers, underscores and dashes
```

### Why

So we can use a built-in feature.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
@@ -301,4 +301,34 @@ mod tests {
 
         assert!(result.is_ok());
     }
+
+    #[test]
+    fn test_alias_validator_with_valid_inputs() {
+        let valid_inputs = [
+            "hello",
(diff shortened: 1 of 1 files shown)


willemneal (changes requested):
Could we have a test to show the error works?

fnando:
@willemneal can you point me to a test I can borrow?

willemneal:
[cmd/crates/soroban-test/tests/it/config.rs](https://github.com/stellar/stellar-cli/pull/1406/files#diff-f7fce5dbe5b3bdfc63fd5f3e8661920acf93db26642360a14e3ced385d472c82)

willemneal (approved):
We can use the new type solution later. Best to have something now.

PR #1404, opened 2024-06-24
Add support for alias on `stellar contract fetch`.

fnando:
### What

Add support for alias on `stellar contract fetch`.

### Why

So we can use the alias, instead of having to provide the contract id.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/fetch.rs
+++ b/cmd/soroban-cli/src/commands/contract/fetch.rs
@@ -125,9 +125,9 @@ impl Cmd {
 
     fn contract_id(&self) -> Result<[u8; 32], Error> {
         let network = self.network()?;
-        self.locator
-            .resolve_contract_id(&self.contract_id, &network.network_passphrase)
-            .map_err(Error::from)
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/contract/fetch.rs
```diff
@@ -124,8 +124,10 @@ impl Cmd {
     }
 
     fn contract_id(&self) -> Result<[u8; 32], Error> {
-        utils::contract_id_from_str(&self.contract_id)
-            .map_err(|e| Error::CannotParseContractId(self.contract_id.clone(), e))
+        let network = self.network()?;
+        self.locator
+            .resolve_contract_id(&self.contract_id, &network.network_passphrase)
+            .map_err(Error::from)
```
willemneal, line 130:
```suggestion
        let network = self.network()?;
        Ok(self.locator
            .resolve_contract_id(&self.contract_id, &network.network_passphrase)?)
```

willemneal (approved):

PR #1405, opened 2024-06-24
Add support for alias on `stellar events`.

fnando:
### What

Add support for alias on `stellar events`.

### Why

So we can use the alias, instead of having to provide the contract id.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/config/locator.rs
+++ b/cmd/soroban-cli/src/commands/config/locator.rs
@@ -9,7 +9,7 @@ use std::{
     path::{Path, PathBuf},
     str::FromStr,
 };
-use stellar_strkey::DecodeError;
+use stellar_strkey::{Contract, DecodeError};
 
 use crate::{utils::find_config_dir, Pwd};
 
@@ -280,28 +280,19 @@ impl Args {
         Ok(alias_data.ids.get(network_passphrase).cloned())
     }
 
-    pub fn resolve_contract_id_as_string(
-        &self,
-        alias_or_contract_id: &str,
-        network_passphrase: &str,
-    ) -> Result<String, Error> {
-        let contract_id = self
-            .get_contract_id(alias_or_contract_id, network_passphrase)?
-            .unwrap_or_else(|| alias_or_contract_id.to_string());
-
-        Ok(contract_id)
-    }
-
     pub fn resolve_contract_id(
         &self,
         alias_or_contract_id: &str,
         network_passphrase: &str,
-    ) -> Result<[u8; 32], Error> {
-        let contract_id =
-            self.resolve_contract_id_as_string(alias_or_contract_id, network_passphrase)?;
+    ) -> Result<Contract, Error> {
+        let contract_id = self
+            .get_contract_id(alias_or_contract_id, network_passphrase)?
+            .unwrap_or_else(|| alias_or_contract_id.to_string());
 
-        soroban_spec_tools::utils::contract_id_from_str(&contract_id)
-            .map_err(|e| Error::CannotParseContractId(contract_id.clone(), e))
+        Ok(Contract(
+            soroban_spec_tools::utils::contract_id_from_str(&contract_id)
+                .map_err(|e| Error::CannotParseContractId(contract_id.clone(), e))?,
+        ))
     }
 }
 
--- a/cmd/soroban-cli/src/commands/config/locator.rs
+++ b/cmd/soroban-cli/src/commands/config/locator.rs
@@ -280,14 +280,25 @@ impl Args {
         Ok(alias_data.ids.get(network_passphrase).cloned())
     }
 
+    pub fn resolve_contract_id_as_string(
+        &self,
+        alias_or_contract_id: &str,
+        network_passphrase: &str,
+    ) -> Result<String, Error> {
+        let contract_id = self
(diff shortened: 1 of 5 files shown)


cmd/soroban-cli/src/commands/events.rs
```diff
@@ -159,12 +159,12 @@ impl Cmd {
...
+
         // Validate contract_ids
-        for id in &mut self.contract_ids {
-            utils::contract_id_from_str(id).map_err(|e| Error::InvalidContractId {
-                contract_id: id.clone(),
-                error: e,
-            })?;
+        for id in &self.contract_ids {
+            self.locator
+                .resolve_contract_id(id, &network.network_passphrase)?;
```
willemneal, line 167: I haven't read this file in detail, but this wouldn't be needed if we used the strkey type instead then below you could just to `to_string` on it.

cmd/soroban-cli/src/commands/config/locator.rs
```diff
@@ -280,15 +280,26 @@ impl Args {
...
+    }
+
+    pub fn resolve_contract_id(
+        &self,
+        alias_or_contract_id: &str,
+        network_passphrase: &str,
+    ) -> Result<[u8; 32], Error> {
+        let contract_id =
+            self.resolve_contract_id_as_string(alias_or_contract_id, network_passphrase)?;
+
```
willemneal, line 302: We could collapse these two functions into one that returns a `StrKey::Contract` which can then be converted into a string or byte array.

fnando: I wanted to avoid too many changes on that module, as I don't fully grasp it. 

Can `StrKey::Contract` be resolved from just the C-address? The `::from_string` function lacks documentation, but it seems like it waits for a contract payload, right?

* https://github.com/stellar/rs-stellar-strkey/blob/v0.0.8/src/strkey.rs#L204

willemneal: You can either use the from_string method or construct it from the `[u8; 32]`. So it's a small change to `resolve_contract_id`.


```rust
#[derive(Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
pub struct Contract(pub [u8; 32]);
```

Since the inner tuple field is public you can construct it like `stellar_strkey::Contract(<arr>)`. Then from the caller.
```rust
let contract_strkey = resolve_contract_id(...)?;
let contract_id = contract_strkey.0;
let contract_id_str = contract_strkey.to_string();
```

fnando: ah, that makes sense. Let me try it! :)

cmd/soroban-cli/src/commands/events.rs
```diff
@@ -226,11 +218,22 @@ impl NetworkRunnable for Cmd {
...
             .verify_network_passphrase(Some(&network.network_passphrase))
             .await?;
+
+        let contract_ids: Vec<String> = self
+            .contract_ids
+            .iter()
+            .map(|id| {
+                self.locator
+                    .resolve_contract_id(id, &network.network_passphrase)
+                    .map(|contract| contract.to_string())
```
willemneal, line 228:
```suggestion
                Ok(self.locator
                    .resolve_contract_id(id, &network.network_passphrase)?.to_string())
```
I'm fairly confident this should work.

willemneal (approved):
Just one nit.

cmd/soroban-cli/src/commands/events.rs
fnando: Almost:

```diff
diff --git a/cmd/soroban-cli/src/commands/events.rs b/cmd/soroban-cli/src/commands/events.rs
index 2eefb02..139fdc0 100644
--- a/cmd/soroban-cli/src/commands/events.rs
+++ b/cmd/soroban-cli/src/commands/events.rs
@@ -223,9 +223,11 @@ impl NetworkRunnable for Cmd {
             .contract_ids
             .iter()
             .map(|id| {
-                self.locator
-                    .resolve_contract_id(id, &network.network_passphrase)
-                    .map(|contract| contract.to_string())
+                Ok::<String, Error>(
+                    self.locator
+                        .resolve_contract_id(id, &network.network_passphrase)?
+                        .to_string(),
+                )
             })
             .collect::<Result<Vec<_>, _>>()?;
```

cmd/soroban-cli/src/commands/events.rs
```diff
@@ -226,11 +218,24 @@ impl NetworkRunnable for Cmd {
...
+            .contract_ids
+            .iter()
+            .map(|id| {
+                Ok::<String, Error>(
+                    self.locator
+                        .resolve_contract_id(id, &network.network_passphrase)?
+                        .to_string(),
+                )
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
willemneal, line 232:
```suggestion
                Ok(
                    self.locator
                        .resolve_contract_id(id, &network.network_passphrase)?
                        .to_string(),
                )
            })
            .collect::<Result<Vec<_>, Error>>()?;
```
You should be able to do it here.

willemneal: Might need `cargo fmt` unfortunately.

PR #1406, opened 2024-06-25
feat: add tx sign/send and Stellar trait

willemneal:
### What

~Currently relies on #1283~

Adds:
- tx sign - sign just a transaction (envelope), appending the signature
    - Does not have `--source-acount` and requires `--sign-with`
    - In a future PR `tx auth-sign/auth` will be added to handle signing just authorization entries.
- tx send - submit a signed transaction envelope to the network
- `Stellar` trait which handles the signing of transactions and auth-entries
- Updates signing across all commands. 
- For invoke command (currently only command that handles signing auth entries), but these will used with `tx auth`:
    - `--sign-with-key` - takes same input as `--source-account`
    - `--sign-with-lab` - currently does nothing
    - `--yes` - if `--sign-with-key` user verification is required, unless `--yes` is provided
   

### Why

Currently signing happens automatically in the background, which is unsure and doesn't inform users when signing has taken place.

### Known limitations

Doesn't add ledger support. Still need to figure out dep issues. This is to be done in follow up PR.


FULL_HELP_DOCS.md
```diff
@@ -140,6 +140,7 @@ Get Id of builtin Soroban Asset Contract. Deprecated, use `stellar contract id a
 * `--hd-path <HD_PATH>` — If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
 * `--global` — Use global config
 * `--config-dir <CONFIG_DIR>` — Location of config directory, default is "."
+* `--check` — Check with user before signature. Eventually this will be replaced with `--yes`, which does the opposite and will force a check without --yes
```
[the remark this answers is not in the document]
leighmcculloch: I see the term 'prompt' being used internally in the code, and that term here would communicate the intent of the option better I think. e.g. `--prompt`

cmd/soroban-cli/src/signer.rs
```diff
@@ -97,121 +163,118 @@ pub fn sign_soroban_authorizations(
...
+    pub fn new(key: ed25519_dalek::SigningKey, prompt: bool) -> Self {
+        Self { key, prompt }
+    }
+}
+
+#[async_trait::async_trait]
+impl Stellar for LocalKey {
+    async fn sign_blob(&self, data: &[u8]) -> Result<Vec<u8>, Error> {
+        if self.prompt {
+            eprintln!("Press 'y' or 'Y' for yes, any other key for no:");
```
leighmcculloch, line 251: Does the prompt output any information about what is being signed, or just that something is being signed? What if multiple things need signing? Will it output this message a few times without context for each one?

leighmcculloch (commented):
Took a look and I left some thoughts inline.

I'm not completely seeing if what we have here is going to be able to support all of the different ways someone might need to compose signing without a breaking change, especially in the context of:
- https://github.com/stellar/stellar-cli/issues/1264

For example, the following use cases are ideally supported in the future by tx sign:
- Sign a tx with the source account.
- Sign a tx with signers of the source account, all present.
- Sign a tx with one signer of the source account, and then separately on another machine, sign the tx with another signer of the source account.
- Sign a tx with one signer of the source account, then support attaching a signature from a external source.
- Sign a tx with auths.
- Sign only auths.
- Sign only some auths (only auths with the signer I have specified).
- Sign a tx with auths without signing the auths.

Do you think we can add all of them without breaking the API in this PR?

I'm wondering if the call to combine auth with tx signing is too problematic in the context of all these scenarios, and by problematic, I mean it will result in an API where it is too easy for someone to sign something they didn't expect to sign.

I left some comments throughout.

The refactor looks fine, although we don't have much in the way of testing which we should address with signing.

I ran out of time to use the command on the command line, and I'd like to do that tomorrow then bring back more feedback.

cmd/soroban-cli/src/signer.rs
willemneal: It prints each time something is signed, however, the context is currently just the public key and the hash. More context would be great, but as I was currently just following the way we were signing the hash, it was easiest to start with that context. Do you think it's worth adding to this PR? I saw it as a future improvement considering we don't even prompt for signing.

cmd/soroban-cli/src/commands/tx/sign.rs
```diff
@@ -0,0 +1,55 @@
...
+            Ok(self.config.sign(tx).await?)
+        }
+    }
+
+    pub async fn sign_env(
+        &self,
+        tx_env: TransactionEnvelope,
+    ) -> Result<TransactionEnvelope, Error> {
+        self.sign(super::xdr::unwrap_envelope_v1(tx_env)?).await
+    }
```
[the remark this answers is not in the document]
willemneal: True, this was written before we added the `From` trait for `Transaction`. So it was a connivence to allow for both.

cmd/soroban-cli/src/commands/tx/auth.rs
```diff
@@ -0,0 +1,19 @@
+use clap::arg;
+
+#[derive(Debug, clap::Args, Clone)]
+#[group(skip)]
+pub struct Args {
+    /// Number of ledgers from current ledger before the signed auth entry expires. Default 60 ~ 5 minutes.
+    #[arg(
+        long = "ledgers-from-now",
+        visible_alias = "ledgers",
```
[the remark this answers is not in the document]
willemneal: I had been going back and forth on the two commands. I do find it strange when there are incompatible arguments. E.g this doesn't make sense without `--auth` so we might as well break it up into two commands. 

Though perhaps we can just make clap understand that they are incompatible? I will ask for some feedback from users.

cmd/soroban-cli/src/commands/config/secret.rs
```diff
@@ -69,21 +76,21 @@ impl Args {
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(untagged)]
-pub enum Secret {
+pub enum SignerKind {
```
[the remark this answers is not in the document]
willemneal: Yeah I was never happy with this name change. It started because I had the ledger as another case. And it didn't make sense. But it does contain the secret key...

FULL_HELP_DOCS.md
```diff
@@ -1234,6 +1250,45 @@ Calculate the hash of a transaction envelope from stdin
...
+###### **Options:**
+
+* `--rpc-url <RPC_URL>` — RPC server endpoint
+* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
+* `--network <NETWORK>` — Name of network to use from config
+* `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
+* `--hd-path <HD_PATH>` — If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
+* `--global` — Use global config
+* `--config-dir <CONFIG_DIR>` — Location of config directory, default is "."
+* `--check` — Check with user before signature. Eventually this will be replaced with `--yes`, which does the opposite and will force a check without --yes
```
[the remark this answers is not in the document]
willemneal: Yes the idea would be that we would drop `check` (or `prompt` as Leigh suggests). This was to avoid a breaking change.

cmd/soroban-cli/src/commands/config/secret.rs
```diff
@@ -132,14 +145,33 @@ impl Secret {
...
+impl Stellar for StellarSigner {
+    async fn get_public_key(&self) -> Result<stellar_strkey::ed25519::PublicKey, signer::Error> {
+        match self {
+            StellarSigner::Local(signer) => signer.get_public_key().await,
+        }
+    }
+
+    async fn sign_blob(&self, blob: &[u8]) -> Result<Vec<u8>, signer::Error> {
+        match self {
+            StellarSigner::Local(signer) => signer.sign_blob(blob).await,
```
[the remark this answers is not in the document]
willemneal: There is the `sign_transaction` which by default signs the hash, which is what we were currently doing. So signers can overwrite this including hardware wallets.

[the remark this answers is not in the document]
Ifropc: I think the concern was that not every singer method would be able to sign a blob, and that's the only method supported currently by the interface. From what Leigh mentioned, hardware wallet asks for XDR, also not all software wallets support signing arbitrary blob (but support signing XDR)

FULL_HELP_DOCS.md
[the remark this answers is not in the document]
willemneal: Replaced with `--yes` but only in cases of `--sign-with-*`

FULL_HELP_DOCS.md
```diff
@@ -1234,6 +1250,45 @@ Calculate the hash of a transaction envelope from stdin
...
+Sign a transaction
+
+**Usage:** `stellar tx sign [OPTIONS] --source-account <SOURCE_ACCOUNT>`
+
+###### **Options:**
+
+* `--rpc-url <RPC_URL>` — RPC server endpoint
+* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
+* `--network <NETWORK>` — Name of network to use from config
+* `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
```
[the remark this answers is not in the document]
willemneal: Removed!

cmd/soroban-cli/src/commands/config/secret.rs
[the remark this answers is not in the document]
willemneal: `sign_blob` is at the bottom. If you implement it with a panic but then implement `sign_txn`, then `sign_blob` will never be called. Also currently with ledger we only support hash signing, which uses `sign_blob`. There is work to add support to signing transaction, but we are waiting for upstream ledger support.

FULL_HELP_DOCS.md
```diff
@@ -117,10 +117,12 @@ Get Id of builtin Soroban Asset Contract. Deprecated, use `stellar contract id a
 * `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
 * `--network <NETWORK>` — Name of network to use from config
 * `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
-* `--hd-path <HD_PATH>` — If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
 * `--global` — Use global config
 * `--config-dir <CONFIG_DIR>` — Location of config directory, default is "."
-* `--check` — Check with user before signature. Eventually this will be replaced with `--yes`, which does the opposite and will force a check without --yes
+* `--sign-with-key <SIGN_WITH_KEY>` — Sign with secret key
```
Ifropc, line 117: I think we should clarify that it's the "entity" as in `--source-account` (i.e. `Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")`)

cmd/soroban-cli/src/commands/tx/auth.rs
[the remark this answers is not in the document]
willemneal: Added them! Since we are separating out the `tx auth` this currently only applies to the `contract invoke` which previously defaulted to `in 60`.

cmd/soroban-cli/src/commands/tx/sign.rs
```diff
@@ -0,0 +1,55 @@
...
+        } else {
+            Ok(self.config.sign(tx).await?)
+        }
+    }
+
+    pub async fn sign_env(
+        &self,
+        tx_env: TransactionEnvelope,
+    ) -> Result<TransactionEnvelope, Error> {
+        self.sign(super::xdr::unwrap_envelope_v1(tx_env)?).await
```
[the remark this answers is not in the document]
willemneal: Fixed!

willemneal:
> The refactor looks fine, although we don't have much in the way of testing which we should address.

@leighmcculloch I relied on previous tests since this logic touched so many commands. Do you think we need more to approve the PR?

FULL_HELP_DOCS.md
```diff
@@ -1234,6 +1250,45 @@ Calculate the hash of a transaction envelope from stdin
...
+
+Sign a transaction
+
+**Usage:** `stellar tx sign [OPTIONS] --source-account <SOURCE_ACCOUNT>`
+
+###### **Options:**
+
+* `--rpc-url <RPC_URL>` — RPC server endpoint
+* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
+* `--network <NETWORK>` — Name of network to use from config
```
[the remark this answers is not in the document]
willemneal: Looking into this. The `network::Args` struct didn't change and this is true for all commands in the markdown. Checking `--help` they are separated.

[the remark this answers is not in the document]
willemneal: Opened issue about it.

FULL_HELP_DOCS.md
```diff
@@ -1234,6 +1250,45 @@ Calculate the hash of a transaction envelope from stdin
...
+* `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
+* `--hd-path <HD_PATH>` — If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
+* `--global` — Use global config
+* `--config-dir <CONFIG_DIR>` — Location of config directory, default is "."
+* `--check` — Check with user before signature. Eventually this will be replaced with `--yes`, which does the opposite and will force a check without --yes
+* `--ledgers-from-now <FROM_NOW>` — Number of ledgers from current ledger before the signed auth entry expires. Default 60 ~ 5 minutes
```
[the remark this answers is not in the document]
willemneal: Removed and will  be in the `tx auth` command.

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -1,114 +1,102 @@
...
+        Ok(self
+            .sign_with
+            .locator
+            .account(account)?
+            .signer(self.sign_with.hd_path, prompt)?)
     }
 
-    pub async fn sign_with_local_key(&self, tx: Transaction) -> Result<TransactionEnvelope, Error> {
-        self.sign(tx).await
+    pub async fn public_key(&self) -> Result<stellar_strkey::ed25519::PublicKey, Error> {
```
Ifropc, line 58: 1. Doesn't seem to be required to be `async`
2. We want to use `source_account` here, not signer's public key. If user calls a command ` cargo run contract deploy --wasm ~/wasm/hello_world.wasm --source-account alice --sign-with-key me --build-only` transaction should have a source account `alice`, but signed with `me`. Currently, it both originates from `me` and is singed by `me`

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -1,114 +1,102 @@
...
     /// Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…").
     pub source_account: String,
 
-    #[arg(long)]
-    /// If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
-    pub hd_path: Option<usize>,
-
     #[command(flatten)]
-    pub locator: locator::Args,
+    pub sign_with: sign_with::Args,
```
Ifropc, line 42: I'm not convinced that we want to put locator inside `sign_with`. Isn't it a separate logical entity? (As far as flags go)?
We can pass locator into `.signer()` command, as we need it to locate the private key

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -1,114 +1,102 @@
...
     }
 
-    #[allow(clippy::unused_async)]
     pub async fn sign(&self, tx: Transaction) -> Result<TransactionEnvelope, Error> {
-        let key = self.key_pair()?;
-        let Network {
-            network_passphrase, ..
-        } = &self.get_network()?;
-        Ok(signer::sign_tx(&key, &tx, network_passphrase)?)
+        Ok(self
```
Ifropc, line 63: I think `sign(&self, tx: Transaction)` should live in `StellarSigner` (or I guess to be more precise the `Stellar` trait)

cmd/soroban-cli/src/config/secret.rs
```diff
@@ -21,6 +24,10 @@ pub enum Error {
     Ed25519(#[from] ed25519_dalek::SignatureError),
     #[error("Invalid address {0}")]
     InvalidAddress(String),
+    #[error("Ledger does not reveal secret key")]
+    LedgerDoesNotRevealSecretKey,
```
Ifropc, line 28: Let's not mix ledger implementation into this PR (to reduce total number of code to review)

cmd/soroban-cli/src/config/secret.rs
```diff
@@ -140,6 +153,25 @@ impl Secret {
     }
 }
 
+pub enum StellarSigner {
+    Local(LocalKey),
+}
+
+#[async_trait::async_trait]
+impl Stellar for StellarSigner {
+    async fn get_public_key(&self) -> Result<stellar_strkey::ed25519::PublicKey, signer::Error> {
```
Ifropc, line 162: Interestingly, generic `StellarSigner` can't provide public key, nor can it sign a blob. IMO the only thing it should do is just to sign a transaction envelope

cmd/soroban-cli/src/config/secret.rs
```diff
@@ -140,6 +153,25 @@ impl Secret {
     }
 }
 
+pub enum StellarSigner {
```
Ifropc, line 156: I propose to rename `StellarSigner` -> `SingerMethod` and `Stellar` -> `StellarSigner` for clarity

cmd/soroban-cli/src/config/sign_with.rs
```diff
@@ -0,0 +1,120 @@
...
+impl Args {
+    pub fn signer(&self) -> Result<StellarSigner, Error> {
+        let account = self.sign_with_key.as_deref().ok_or(Error::NoSignWithKey)?;
+        Ok(self
+            .locator
+            .account(account)?
+            .signer(self.hd_path, !self.yes)?)
+    }
+
+    pub async fn sign_txn_env(
```
Ifropc, line 74: IMO this 2 methods should go into StellarSigner

cmd/soroban-cli/src/config/sign_with.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub async fn sign_tx_env_with_signer(
+        &self,
+        signer: &(impl Stellar + std::marker::Sync),
+        tx_env: TransactionEnvelope,
+    ) -> Result<TransactionEnvelope, Error> {
+        let network = self.get_network()?;
+        Ok(signer.sign_txn_env(tx_env, &network).await?)
+    }
+
+    pub async fn sign_soroban_authorizations(
```
Ifropc, line 91: Let's remove authorization methods for now and keep it in scope of https://github.com/stellar/stellar-cli/issues/1493

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -202,21 +195,17 @@ impl NetworkRunnable for Cmd {
...
         let (txn, contract_id) = build_create_contract_tx(
             wasm_hash,
             sequence + 1,
             self.fee.fee,
             &network.network_passphrase,
             salt,
-            &key,
+            &public_strkey,
         )?;
         if self.fee.build_only {
```
Ifropc, line 210: Hmmm it looks like currently --`build-only` and --`sim-only` actually conflict with` --sign-with`, because `--sign-with` is getting effectively ignored. I think it makes sense that we always sign the transaction, but we could also make sign-with flags to conflict with `--build-only` and `--sim-only`

leighmcculloch: For --build-only I think we should disallow signing with --sign-with because the tx isn't ready to sign and so we'd just be setting up a user for pain.

For --sim-only the txn is ready to sign, so having --sim-only and then a signer is fine I think, the outputted txn will be ready to submit.

cmd/soroban-cli/src/config/sign_with.rs
willemneal: We can't. We need it for backwards compatibility with the invoke command.

cmd/soroban-cli/src/config/sign_with.rs
willemneal: These methods are so that you don't need to provide the signer. Hence `self.signer`, and then the next one uses `self.get_network`. So in the first case we are looking up which signer to use. And the second which network. This is so the StellarSigner has the context.

cmd/soroban-cli/src/config/secret.rs
willemneal: I'm confused by this comment. Again to maintain backwards compatibility with how we signed previously we signed the hash, which boils down to signing a blob or byte array. Also the `StellarSigner` trait uses default methods on the trait. So you only need to define these two methods and the rest will use these.

cmd/soroban-cli/src/config/secret.rs
willemneal: Well `Stellar` is in the signer module, so it's a little redundant. We could instead do `signer::Stellar` when referencing it so that it's clear. But clippy doesn't like when the module name is used in a name.

But I do agree with `StellarSigner` being a poor name. And `Stellar` for that matter. @leighmcculloch Opinions?  Naming is the worst!

cmd/soroban-cli/src/config/secret.rs
willemneal, line 28:
```suggestion
```

cmd/soroban-cli/src/config/mod.rs
willemneal: As mentioned before we need the network information which `sign_with` has. So we can't move this into the trait since that interface requires the network passphrase.

cmd/soroban-cli/src/config/mod.rs
willemneal: Soon we are going to refactor this out completely since now it is treated as global and we can have the root command pass it. So we will need to update the interface to pass the locator around. However, the locator is flattened into this and then flattened into the general config::Args. So I don't see any harm in it for now and we can refactor it away in a future PR.

cmd/soroban-cli/src/config/mod.rs
willemneal: 1. `get_public_key` needs to be async in the `Stellar` trait to support ledger which has a method. So this is future proofing it.
2. Very true! Will fix it.

FULL_HELP_DOCS.md
```diff
@@ -356,11 +368,14 @@ Deploy builtin Soroban Asset Contract
 ###### **Options:**
 
 * `--asset <ASSET>` — ID of the Stellar classic asset to wrap, e.g. "USDC:G...5"
+* `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
```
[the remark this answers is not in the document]
willemneal: This is autogenerated. So we need to put it with the Arg.

FULL_HELP_DOCS.md
```diff
@@ -356,11 +368,14 @@ Deploy builtin Soroban Asset Contract
...
+* `--sign-with-key <SIGN_WITH_KEY>` — Sign with account. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
+* `--sign-with-laboratory` — Sign with labratory
```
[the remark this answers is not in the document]
willemneal: I can make it hidden, but I thought it would be nice to have for the next PR.

FULL_HELP_DOCS.md
```diff
@@ -1043,6 +1082,42 @@ Calculate the hash of a transaction envelope from stdin
...
+
+**Usage:** `stellar tx sign [OPTIONS]`
+
+###### **Options:**
+
+* `--sign-with-key <SIGN_WITH_KEY>` — Sign with account. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
+* `--sign-with-laboratory` — Sign with labratory
+* `--hd-path <HD_PATH>` — If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
+* `--yes` — If `--sign-with-*` is used this will remove requirement of being prompted to sign transactions
+* `--rpc-url <RPC_URL>` — RPC server endpoint
```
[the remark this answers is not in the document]
willemneal: Yes this is true. Though it would be much work to untangle it since we make use of the network logic and resolution. Perhaps we could do this is a future, PR, but the amount of work to do this would be a bit painful. Though I completely agree it's confusing.

cmd/soroban-cli/src/config/sign_with.rs
Ifropc: Oh sorry I missed that. For some reason my IDE said it was unused

cmd/soroban-cli/src/config/sign_with.rs
Ifropc: But we can move `sign` to signer and add one more argument which is a network?  Don't have a strong preference here, it's just for me it's not clear right now what's the logical separation between `config::Args` and `Signer` is. It feels like `Signer` should do the job, while `config::Args` just provide entities that do the job. But we can always re-visit this later

FULL_HELP_DOCS.md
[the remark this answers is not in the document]
Ifropc: Yeah sorry I was just commenting on docs as it's the first thing I saw in PR

FULL_HELP_DOCS.md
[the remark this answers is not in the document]
Ifropc: Yeah if you could just make it hidden that would be great

cmd/soroban-cli/src/config/secret.rs
willemneal: An implementer of `Stellar` can implement all the methods, instead of relying on the default methods. And if it doesn't support `sign_blob` it can use `unreachable!()` inside the method. Because if it defines it's own version of `sign_txn_env`, `sign_blob` will never be called. The same applies for `get_public_key` if the implementation doesn't need it you don't have to implement it. However, it is used by `keys address ..`. This command retrieves the signer associated  and then resolves the public key if possible.  ledger and we would need to retrieve the key. This  would also require some work to allow this to work with lab as well.

Also which hardware wallets other than ledger support stellar and signing full transactions? Currently the ledger is in works to add that and only signs the hash... Either way I would assume that these wallets provide a way to access the associated public key. So for Lab, we could throw an error in the case of trying to do `keys address`, however, it would be nice to resolve the public key to use locally.

I do agree that the Stellar trait is too big and I did try with ways of inheriting the trait `trait Auth: Stellar`, but ran into some issues. And a lot of the design was guided initially by ledger and the initial implementation. And while it is easier to start with the more modular design, the refactor can be done later without affecting the end user's experience.

I personally don't know how to sign the full transaction with the local key. But it would be better than the hash.

Ifropc: Generally, I agree. We can refactor signing methods later. As far as the public key method goes, I think it was only used to get source account, and as we discussed before it may no longer be needed.

willemneal: > As far as the public key method goes, I think it was only used to get source account

It's not though. It's used in `keys address` and generally in resolving a signer to a public key and will be used for auth signing.

Ifropc: But should it be here, in generic Signer? It can be added separately in `keys address` and auth signing, while generic `--sign-with` doesn't require it

Ifropc: Currently, we have Stellar trait with the following signature:
```rust
#[async_trait::async_trait]
pub trait Stellar {
    /// Currently only supports ed25519 keys
    async fn get_public_key(&self) -> Result<stellar_strkey::ed25519::PublicKey, Error>;

    /// Sign an abritatry byte array
    async fn sign_blob(&self, blob: &[u8]) -> Result<Vec<u8>, Error>;

    // Implementation functions below that use sign_blob (mainly)
    async fn sign_txn_hash(&self, txn: [u8; 32]) -> Result<DecoratedSignature, Error> { // impl }
    async fn sign_txn_env(
        &self,
        txn_env: TransactionEnvelope,
        network: &Network,
    ) -> Result<TransactionEnvelope, Error> { // impl}
    async fn sign_soroban_authorizations(
        &self,
        raw: &Transaction,
        network: &Network,
        expiration_ledger: u32,
    ) -> Result<Option<Transaction>, Error> { // impl }
    // Some extra wrappers over sign_tx_env and sign_soroban_authorizations 
}
```
    
Which I think is a bit too tailored for local signing. 
Generally, I don't think we need to be able to get a public key for signing just transactions (we only need source account, which is passed as an arg and is mandatory in all auto-signing requests). Just for signing we don't need a public key.
Same with signing a blob. We enforce blind signing for the wallet (which we should not do in a generic case, only in specific scenarios where we can't use clear signing).
Instead, I think we should only implement `sign_txn_env`

I understand that we also need to sign Soroban auth entries. For that we can just make a new trait. Not all signing methods might support this trait, and not all commands would need this trait (e.g. tx sign doesn't allow to sign only auth entries, same with most of commands. We only need to sign Soroban auth entries in `tx auth sign` and `tx invoke` -> which only work for the situation when we already have all local signers imported to begin with)

Proposed traits:
```rust
/// A trait for signing Stellar transactions
#[async_trait::async_trait]
pub trait Stellar {
    async fn sign_txn_env(
        &self,
        txn_env: TransactionEnvelope,
        network: &Network,
    ) -> Result<TransactionEnvelope, Error>
}

impl Stellar for LocalKey 
impl Stellar for Ledger
impl Stellar for Laboratory

/// A trait for signing Soroban authorization entries
#[async_trait::async_trait]
pub trait StellarAuth {
    async fn sign_txn_env(
        &self,
        txn_env: TransactionEnvelope,
        network: &Network,
    ) -> Result<TransactionEnvelope, Error> 
}

impl StellarAuth for LocalKey 
```    

WDYT? @willemneal @leighmcculloch

leighmcculloch: I agree the interface is weak and reducing it to the single function would make the interface a better abstraction. Moving the auth into a separate interface too 👍🏻 . [The bigger the interface, the weaker the abstraction.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=5m17s) For better or worse, it's common for Rust interfaces to look like this, similar to Java interfaces, with a mixin of utilities that make assumptions about the common case, but allow an "out" where you reimplement the utilities outside the common case. I think it'd be okay to merge with the large interface and then lets follow up with paring it down.

willemneal: I trimmed it down some, however, I kept it as signing a transaction since `sign_txn_env` can remain as a generic interface that will use a signer. After some thought I think that I had conflated two abstractions, a signer and a wallet. A wallet must have both a signer and a public address. 

I agree with @leighmcculloch that we could do some follow up work to address all this considering I am already working with adding the ledger, so we could meet up to discuss this in detail and how best to abstract this.

FULL_HELP_DOCS.md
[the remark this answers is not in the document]
willemneal: Also the alias should be visible in the markdown so this seems like a bug.

cmd/soroban-cli/src/config/sign_with.rs
willemneal: The idea here is that the `self.signer` for `config.arg` could be from `source_account`, where as for `sign_with` it is the one provided by `--sign-with-*`. So this is just a connivence method that will use the `signer` resolved here instead of the one used in `sign_with::Args`. This is all so that the commands don't need to resolve both the signer and the network to sign. But we could have it that way. Again I think this should be address in a future PR as it wouldn't change the CLI's UX.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
willemneal: Given that we can `stellar ... --sim-only | stellar tx sign` perhaps they should be mutually exclusive as it's a little surprising that it's doing more than signing. We could print a warning if the options are used together.

Regardless I would prefer this in a future PR as it would involve touching each early return for all commands that have `--sim-only`.

Ifropc: Agree, let's do a separate PR, just a warning for now should be good

cmd/soroban-cli/src/config/sign_with.rs
Ifropc: I'd say it's more of a nitpick, so fine to leave as is for now, considering size of PR

cmd/soroban-cli/src/config/secret.rs
Ifropc: I think the main issue is still there, `StellarSigner` must implement `Stellar` trait with `get_public_key` and `sign_blob` methods. Both of them are not required for envelope signing, only for auth signing. 
I don't mind revisiting abstractions later, but it looks like we may need to do it for both ledger and lab signing

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
leighmcculloch: 👍🏻 I think for v22 we should also remove the `--sim-only`. There's multiple ways to achieve it because of the `tx simulate subcommand, and if we remove `--sim-only` it's one less variants of use we need to maintain.

leighmcculloch: Opened this issue to follow up with removing --sim-only closer to v22:
- https://github.com/stellar/stellar-cli/issues/1547

willemneal:
@Ifropc @leighmcculloch 

I wanted to present some thoughts for how this all could be refactored. I have tested it locally.

It's a rather big change, but I propose breaking everything into smaller traits and then make use of blank implementations.  I can make a follow up issue, but wanted to drop it here for thoughts and to see if it's good enough for merging this PR as is. I'm also not married to this idea, but think it is more inline with the smaller traits route. We could also just not worry with blanket impls and just implement them for each type since there weren't be a ton of signers, just thought it could help make it easier to write them.

```rust

/// A trait for signing Stellar transaction Envelopes
#[async_trait::async_trait]
pub trait TxEnv {
    /// Sign a Stellar  with the given source account
    /// This is a default implementation that signs the transaction hash and returns a decorated signature
    ///
    /// Todo: support signing the transaction directly.
    /// # Errors
    /// Returns an error if the source account is not found
    async fn sign_txn_env(
        &self,
        txn_env: TransactionEnvelope,
        network: &Network,
    ) -> Result<TransactionEnvelope, Error>;
    // let hash = transaction_hash(txn, network_passphrase)?;
    // self.sign_txn_hash(hash).await
}

/// A trait for signing Stellar transactions
#[async_trait::async_trait]
pub trait Tx {
    /// Sign a Stellar transaction with the given source account
    /// This is a default implementation that signs the transaction hash and returns a decorated signature
    ///
    /// Todo: support signing the transaction directly.
    /// # Errors
    /// Returns an error if the source account is not found
    async fn sign_txn(
        &self,
        txn: &Transaction,
        network: &Network,
    ) -> Result<DecoratedSignature, Error>;
}

/// A trait for signing Stellar transaction Hashes
#[async_trait::async_trait]
pub trait TxHash {
    /// Sign a transaction hash with the given source account
    /// # Errors
    /// Returns an error if the source account is not found
    async fn sign_txn_hash(&self, txn: [u8; 32]) -> Result<DecoratedSignature, Error>;
}

#[async_trait::async_trait]
impl<T: Blob + HasPublicKey + std::marker::Sync> TxHash for T {
    async fn sign_txn_hash(&self, txn: [u8; 32]) -> Result<DecoratedSignature, Error> {
        let source_account = self.get_public_key().await?;
        eprintln!(
            "{} about to sign hash: {}",
            source_account.to_string(),
            hex::encode(txn)
        );
        let tx_signature = self.sign_blob(&txn).await?;
        Ok(DecoratedSignature {
            // TODO: remove this unwrap. It's safe because we know the length of the array
            hint: SignatureHint(source_account.0[28..].try_into().unwrap()),
            signature: Signature(tx_signature.try_into()?),
        })
    }
}

#[async_trait::async_trait]
impl<T: TxHash + std::marker::Sync> Tx for T {
    async fn sign_txn(
        &self,
        txn: &Transaction,
        Network {
            network_passphrase, ..
        }: &Network,
    ) -> Result<DecoratedSignature, Error> {
        self.sign_txn_hash(transaction_hash(txn, network_passphrase)?)
            .await
    }
}

/// A trait for signing abritrary byte arrays
#[async_trait::async_trait]

pub trait Blob {
    /// Sign an abritatry byte array
    async fn sign_blob(&self, blob: &[u8]) -> Result<Vec<u8>, Error>;
}

#[async_trait::async_trait]
pub trait HasPublicKey {
    /// Get the public key of the signer
    async fn get_public_key(&self) -> Result<stellar_strkey::ed25519::PublicKey, Error>;
}

#[async_trait::async_trait]
impl<T: Tx + std::marker::Sync> TxEnv for T {
    async fn sign_txn_env(
        &self,
        txn_env: TransactionEnvelope,
        network: &Network,
    ) -> Result<TransactionEnvelope, Error> {
        match txn_env {
            TransactionEnvelope::Tx(TransactionV1Envelope { tx, signatures }) => {
                let decorated_signature = self.sign_txn(&tx, network).await?;
                let mut sigs = signatures.to_vec();
                sigs.push(decorated_signature);
                Ok(TransactionEnvelope::Tx(TransactionV1Envelope {
                    tx,
                    signatures: sigs.try_into()?,
                }))
            }
            _ => Err(Error::UnsupportedTransactionEnvelopeType),
        }
    }
}

pub(crate) fn hash(network_passphrase: &str) -> xdr::Hash {
    xdr::Hash(Sha256::digest(network_passphrase.as_bytes()).into())
}

pub struct LocalKey {
    key: ed25519_dalek::SigningKey,
    prompt: bool,
}

impl LocalKey {
    pub fn new(key: ed25519_dalek::SigningKey, prompt: bool) -> Self {
        Self { key, prompt }
    }
}

#[async_trait::async_trait]
impl Blob for LocalKey {
    async fn sign_blob(&self, data: &[u8]) -> Result<Vec<u8>, Error> {
        if self.prompt {
            eprintln!("Press 'y' or 'Y' for yes, any other key for no:");
            match read_key() {
                'y' | 'Y' => {
                    eprintln!("Signing now...");
                }
                _ => return Err(Error::UserCancelledSigning),
            };
        }
        let sig = self.key.sign(data);
        Ok(sig.to_bytes().to_vec())
    }
}
#[async_trait::async_trait]
impl HasPublicKey for LocalKey {
    async fn get_public_key(&self) -> Result<stellar_strkey::ed25519::PublicKey, Error> {
        Ok(stellar_strkey::ed25519::PublicKey(
            self.key.verifying_key().to_bytes(),
        ))
    }
}
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
Ifropc: @leighmcculloch so we want people to: 
`stellar contract <command> | stellar tx simulate | stellar tx sign --withKey alice | stellar tx sign --withKey bob` 
right?
Strongly agree on cutting ways of doing the same thing running different commands

Ifropc:
Hey @willemneal thanks for detailed suggestion, I like the approach of breaking it down into smaller pieces!

Ifropc (approved):

FULL_HELP_DOCS.md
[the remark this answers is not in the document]
Ifropc: Tracking this in https://github.com/stellar/stellar-cli/issues/1545

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -261,6 +221,59 @@ impl Cmd {
...
+        spec: &Spec,
+    ) -> Result<(ScVal, Option<StellarSigner>), Error> {
+        let mut signer: Option<StellarSigner> = None;
+        let name = input.name.to_utf8_string()?;
+        let sc_val = if let Some(mut val) = matches_.get_raw(&name) {
+            let mut s = val.next().unwrap().to_string_lossy().to_string();
+            if matches!(input.type_, ScSpecTypeDef::Address) {
+                if let Ok(signer_) = config
+                    .sign_with
+                    .locator
```
Ifropc, line 238: We assuming here that sign_with have a locator (therefore has an access to a local fs), which is fine by me, as I think we only allow signing soroban authes with local key only (at least for now), but it would be nice to have a comment here

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -211,6 +211,14 @@ impl Args {
         KeyType::Identity.read_with_global(name, &self.local_config()?)
     }
 
+    pub fn account(&self, account_str: &str) -> Result<Secret, Error> {
```
Ifropc, line 214: We have `SOROBAN_SECRET_KEY` variable that we are using in stellar keys add to avoid user passing key directly. Should we support the same functionality in `--sign-with-key` for consistency?

cmd/soroban-cli/src/config/mod.rs
Ifropc: Resolving this conversation, as it can be changed later if needed

leighmcculloch:
@fnando and I are reviewing this PR.

cmd/soroban-cli/src/config/sign_with.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[error(transparent)]
+    Rpc(#[from] soroban_rpc::Error),
+    #[error("No sign with key provided")]
+    NoSignWithKey,
+}
+
+#[derive(Debug, clap::Args, Clone, Default)]
+#[group(skip)]
+pub struct Args {
+    /// Sign with account. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…").
```
Ifropc, line 34:
```suggestion
    /// Sign with a local key. Can be an identity (--sign-with-key alice), a secret key (--sign-with-key SC36…), or a seed phrase (--sign-with-key "kite urban…"). If using seed phrase, `--hd-path` must be specified.
```

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -1,114 +1,109 @@
...
 #[group(skip)]
 pub struct Args {
-    #[command(flatten)]
-    pub network: network::Args,
-
     #[arg(long, visible_alias = "source", env = "STELLAR_ACCOUNT")]
-    /// Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…").
+    /// Account where the final transaction originates from.
+    /// If no `--sign-with-*` flag is passed, passed key will also be used to sign the transaction.
+    /// Can be an identity (`--source alice`), a secret key (`--source SC36…`), or a seed phrase (`--source "kite urban…"`)
```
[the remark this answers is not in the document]
willemneal: In the description? It is in the the decorator macro above and will should up with `--help`.

cmd/soroban-cli/src/config/locator.rs
willemneal: We have `STELLAR_SIGN_WITH_SECRET` is this good enough? I'm confused how we would reuse `STELLAR_SECRET_KEY` and have it know to use `sign-with-key`.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -261,6 +221,59 @@ impl Cmd {
...
+        matches_: &clap::ArgMatches,
+        config: &config::Args,
+        spec: &Spec,
+    ) -> Result<(ScVal, Option<StellarSigner>), Error> {
+        let mut signer: Option<StellarSigner> = None;
+        let name = input.name.to_utf8_string()?;
+        let sc_val = if let Some(mut val) = matches_.get_raw(&name) {
+            let mut s = val.next().unwrap().to_string_lossy().to_string();
+            if matches!(input.type_, ScSpecTypeDef::Address) {
+                if let Ok(signer_) = config
```
willemneal, line 236:
```suggestion
                // Currently we only support local keys, same as input for --sign-with-key`, for signing auth entries.
                if let Ok(signer_) = config
```

cmd/soroban-cli/src/config/locator.rs
Ifropc: I think the point of that env variable is to not provide the secret key directly to the command, at least my understand of it is that. I assume if we will use SOROBAN_SECRET_KEY in --sign-with-key it should be invoked like this
```
# Can also use sign-with-key, but it accepts a mandatory parameter which would need to be changed
SOROBAN_SECRET_KEY=SKEY stellar contract asset deploy --asset <Asset> --source-account GSOURCE --sign-with-secret 
```

I'm not adamant on this, and we can change it later on

willemneal:
closed in favor of: https://github.com/stellar/stellar-cli/issues/1591

PR #1413, opened 2024-06-26
Add support for alias to `stellar contract extend|read|restore`.

fnando:
### What

Add support for alias to `stellar contract extend|read|restore`.

```console
$ stellar contract read --id alice --global --network standalone --source alice --durability persistent
LedgerKeyContractInstance,"{""hash"":""d68f72ee0e95c63263e64e1802d0bb511461e0948da7fe355bcfe16163e78b95""}",360380,2433979

$ stellar contract read --id CAYXES5COIN3UEAQ7QM6CNRKQAXGW45YHRH32GHWAMRLI5AR7PWL7OTS --global --network standalone --source alice --durability persistent
LedgerKeyContractInstance,"{""hash"":""d68f72ee0e95c63263e64e1802d0bb511461e0948da7fe355bcfe16163e78b95""}",360380,2433979
```

### Why

So users can pass in an alias, rather than the contract id.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/extend.rs
+++ b/cmd/soroban-cli/src/commands/contract/extend.rs
@@ -10,7 +10,7 @@ use soroban_env_host::xdr::{
 
 use crate::{
     commands::{
-        config::{self, data},
+        config::{self, data, locator},
         global, network,
         txn_result::{TxnEnvelopeResult, TxnResult},
         NetworkRunnable,
@@ -84,6 +84,8 @@ pub enum Error {
     Data(#[from] data::Error),
     #[error(transparent)]
     Network(#[from] network::Error),
+    #[error(transparent)]
+    Locator(#[from] locator::Error),
 }
 
 impl Cmd {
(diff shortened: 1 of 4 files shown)


cmd/soroban-cli/src/commands/contract/extend.rs
```diff
@@ -84,6 +84,8 @@ pub enum Error {
     Data(#[from] data::Error),
     #[error(transparent)]
     Network(#[from] network::Error),
+    #[error(transparent)]
```
willemneal, line 87: Is this needed? Doesn't it come from the config::Error?

willemneal (approved):
Just one question about the new error case

cmd/soroban-cli/src/commands/contract/extend.rs
fnando: not really! The locator module has its own `Error` enum. See https://github.com/stellar/stellar-cli/blob/41648d99d1714fa1ea0d3788a11508ebcea7904a/cmd/soroban-cli/src/commands/config/locator.rs#L19

willemneal: Yeah but it's added to the config::Error so it will be caught. The quick way to check is if the compiler is made. Either way I think there are Errors all over that aren't needed so a future PR could clean them up but it's not a big issue.

PR #1418, opened 2024-06-28
Regenerate docs on pre-commit.

fnando:
### What

Regenerate docs on pre-commit.

### Why

Right now, the full docs file is only regenerated before pushing changes, which requires you to be aware and do another commit with the changes. We should do it before commiting, so you only have to do it once.

### Known limitations

N/A


willemneal (approved):
I think we should use a docs branch and have CI build this. It's annoying that this is a blocker in CI.

PR #1419, opened 2024-06-28
Update alias help.

fnando:
### What

The original `--alias` implementation had `--force` to overwrite existing contract id aliases, but @leighmcculloch [suggested](https://github.com/stellar/stellar-cli/pull/1356#discussion_r1626762468) that it'd be better to always use the latest id. Now, @willemneal created a [ticket to add that back](https://github.com/stellar/stellar-cli/issues/1417). Until we come to a decision, let's make the flag help explicit.

### Why

So people can know that `--alias` will always store the latest contract id without asking for confirmation.

### Known limitations

N/A

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -460,7 +460,7 @@ Deploy a wasm contract
 
   Possible values: `true`, `false`
 
-* `--alias <ALIAS>` — The alias that will be used to save the contract's id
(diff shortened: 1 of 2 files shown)


willemneal (approved):
Yeah to me we should make the redeploy easier, but devs are currently required to include a redeploy method. But once we have the snapshot stuff we can copy over the state to the newly deployed contract and all should be okay.

chadoh (approved):

PR #1423, opened 2024-07-01
Fix/network container improvements

elizabethengelman:
### What

Improvements to `docker network container` commands. Closes https://github.com/stellar/stellar-cli/issues/1230

- [x] Better error handling when the container doesn't exist (specifically with the stop command)
- [x] Add a flag to change the container name passed to docker

### Why

To make `stellar network container start/stop/logs` a bit easier to use and to update.

### Known limitations

I'm not sure that we have documentation about the `network container` commands at the moment, but it seems like it would be a helpful addition. https://github.com/stellar/stellar-docs/issues/751


cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -31,7 +31,18 @@ pub enum Error {
     UnsupportedURISchemeError { uri: String },
 }
 
-#[derive(ValueEnum, Debug, Clone, PartialEq)]
+#[derive(Debug, clap::Parser, Clone)]
+pub struct ContainerArgs {
+    /// Optional argument to specify the container name
+    #[arg(short = 'c', long, required_unless_present = "network")]
+    pub container_name: Option<String>,
+
```
willemneal, line 39: It seems everywhere you use this struct you also use `Network`, so it would make sense to add it here. Then you can add a `name` method to its `impl`

cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -148,3 +159,11 @@ async fn check_docker_connection(docker: &Docker) -> Result<(), bollard::errors:
         }
     }
 }
+
+pub fn get_container_name(container_name_arg: Option<String>, network: Option<Network>) -> String {
+    if let Some(container_name) = container_name_arg {
+        container_name.to_string()
+    } else {
+        format!("stellar-{}", network.unwrap())
```
willemneal, line 167: This is safe to do because clap will ensure that one of the two are defined. However, I would use `expect` just so that if it does happen we'll get a message and not just a panic.

cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -31,7 +31,18 @@ pub enum Error {
     UnsupportedURISchemeError { uri: String },
 }
 
-#[derive(ValueEnum, Debug, Clone, PartialEq)]
+#[derive(Debug, clap::Parser, Clone)]
+pub struct ContainerArgs {
```
willemneal, line 35:
```suggestion
pub struct Args {
```

cmd/soroban-cli/src/commands/network/container/start.rs
```diff
@@ -8,34 +8,36 @@ use bollard::{
 use futures_util::TryStreamExt;
 
 use crate::commands::network::container::shared::{
-    connect_to_docker, Error as ConnectionError, Network, DOCKER_HOST_HELP,
+    connect_to_docker, get_container_name, Error as ConnectionError, Network,
 };
 
+use super::shared::ContainerArgs;
```
willemneal, line 14:
```suggestion
use super::shared::Args;
```

cmd/soroban-cli/src/commands/network/container/start.rs
```diff
@@ -8,34 +8,36 @@ use bollard::{
...
 
     #[error("⛔ ️Failed to create container: {0}")]
-    BollardErr(#[from] bollard::errors::Error),
+    CreateContainerFailed(#[from] bollard::errors::Error),
 }
 
 #[derive(Debug, clap::Parser, Clone)]
 pub struct Cmd {
+    #[command(flatten)]
+    pub container_args: ContainerArgs,
```
willemneal, line 31:
```suggestion
    pub container_args: Args,
```

cmd/soroban-cli/src/commands/network/container/stop.rs
```diff
@@ -1,28 +1,55 @@
 use crate::commands::network::container::shared::{
-    connect_to_docker, Error as ConnectionError, Network, DOCKER_HOST_HELP,
+    connect_to_docker, get_container_name, Error as BollardConnectionError, Network,
 };
 
+use super::shared::ContainerArgs;
```
willemneal, line 5:
```suggestion
use super::shared::Args;
```

cmd/soroban-cli/src/commands/network/container/stop.rs
```diff
@@ -1,28 +1,55 @@
...
+    #[error("⛔ Failed to stop container: {0}")]
+    ContainerStopFailed(#[from] bollard::errors::Error),
 }
 
 #[derive(Debug, clap::Parser, Clone)]
 pub struct Cmd {
-    /// Network to stop
-    pub network: Network,
+    #[command(flatten)]
+    pub container_args: ContainerArgs,
```
willemneal, line 26:
```suggestion
    pub container_args: Args,
```

cmd/soroban-cli/src/commands/network/container/logs.rs
```diff
@@ -1,9 +1,11 @@
 use futures_util::TryStreamExt;
 
 use crate::commands::network::container::shared::{
-    connect_to_docker, Error as ConnectionError, Network, DOCKER_HOST_HELP,
+    connect_to_docker, Error as ConnectionError, Network,
 };
 
+use super::shared::{get_container_name, ContainerArgs};
```
willemneal, line 7:
```suggestion
use super::shared::{get_container_name, Args};
```

cmd/soroban-cli/src/commands/network/container/logs.rs
```diff
@@ -15,17 +17,19 @@ pub enum Error {
 
 #[derive(Debug, clap::Parser, Clone)]
 pub struct Cmd {
-    /// Network to tail
-    pub network: Network,
+    #[command(flatten)]
+    pub container_args: ContainerArgs,
```
willemneal, line 21:
```suggestion
    pub container_args: Args,
```

cmd/soroban-cli/src/commands/network/container/start.rs
```diff
@@ -105,18 +108,42 @@ async fn run_docker_command(cmd: &Cmd) -> Result<(), Error> {
...
+    let log_message = format!(
+        "ℹ️ To see the logs for this container run: stellar network container logs {arg} {additional_flags}",
+        arg = cmd.container_args.container_name.as_ref().map_or_else(
+            || cmd.network.to_string(),
+            |container_name| format!("--container-name {container_name}")
+        ),
+        additional_flags = cmd.container_args.docker_host.as_ref().map_or_else(
+            String::new,
+            |docker_host| format!("--docker-host {docker_host}")
+        )
```
willemneal, line 126:
```suggestion
        additional_flags = cmd.container_args.docker_host.as_ref().map(
            |docker_host| format!("--docker-host {docker_host}")
        ).unwrap_or_default()
```

willemneal, line 126: Perhaps you could add two methods to the `impl Args` that get the arg strings.

willemneal (changes requested):
Overall looks great! Just some refactoring suggestions to clean it up and remove some duplicated code.

cmd/soroban-cli/src/commands/network/container/start.rs
elizabethengelman: Just pushed an update

cmd/soroban-cli/src/commands/network/container/start.rs
elizabethengelman: Great idea! Just pushed an update

cmd/soroban-cli/src/commands/network/container/shared.rs
elizabethengelman: That's a good call out. I didn't include `Network` in this struct because it is required for the `start` command, but is optional for `stop` and `logs`. Maybe that is confusing and could use a refactor/rethink. 🤔

cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -148,3 +175,14 @@ async fn check_docker_connection(docker: &Docker) -> Result<(), bollard::errors:
         }
     }
 }
+
+pub fn get_container_name(container_name_arg: Option<String>, network: Option<Network>) -> String {
```
willemneal, line 179:
```suggestion
pub fn get_container_name(container_name_arg: Option<&str>, network: Option<Network>) -> String {
```

cmd/soroban-cli/src/commands/network/container/logs.rs
```diff
@@ -15,17 +17,19 @@ pub enum Error {
...
+    #[arg(required_unless_present = "container_name")]
+    pub network: Option<Network>,
 }
 
 impl Cmd {
     pub async fn run(&self) -> Result<(), Error> {
-        let container_name = format!("stellar-{}", self.network);
-        let docker = connect_to_docker(&self.docker_host).await?;
+        let container_name =
+            get_container_name(self.container_args.container_name.clone(), self.network);
```
willemneal, line 31:
```suggestion
            get_container_name(self.container_args.container_name.as_deref(), self.network);
```
This will let you do `Option<String>` --> `Option<&str>`, and remove the need to clone (once the parameter is updated).

cmd/soroban-cli/src/commands/network/container/stop.rs
```diff
@@ -1,28 +1,55 @@
...
 }
 
 impl Cmd {
     pub async fn run(&self) -> Result<(), Error> {
-        let container_name = format!("stellar-{}", self.network);
-        let docker = connect_to_docker(&self.docker_host).await?;
-        println!("ℹ️  Stopping container: {container_name}");
-        docker.stop_container(&container_name, None).await.unwrap();
+        let container_name =
+            get_container_name(self.container_args.container_name.clone(), self.network);
```
willemneal, line 36:
```suggestion
            get_container_name(self.container_args.container_name.as_deref(), self.network);
```

cmd/soroban-cli/src/commands/network/container/shared.rs
willemneal: On second thought, perhaps this method should move into the `impl` since it relies on the internal state.
```rust
  get_container_name(self.container_args.container_name.clone(), self.network)
```
becomes
```rust
  self.container_args.name(self.network)
```

willemneal: Also generally it's better to have `Option<&T>` an the argument type because you can always do `x.as_ref` or `x.as_deref`, and need to have ownership.

elizabethengelman: That makes sense to move `get_container_name` into the impl of `Args` - I pushed up a change for that.

FULL_HELP_DOCS.md
```diff
@@ -1073,6 +1073,7 @@ By default, when starting a testnet container, without any optional arguments, i
 
 ###### **Options:**
 
+* `-c`, `--container-name <CONTAINER_NAME>` — Optional argument to specify the container name
```
leighmcculloch, line 1076: I think we could drop the word container, because most of the options could be prefixed with it or not and mean the same thing. We're also unlikely to have a separate name and container name, so simply name is sufficient.

In the interest of preserving short options for options that become most meaningful I suggest not adding short options when options are first added, and instead add them later when folks get irritated by the long ones.

```suggestion
* `--name <NAME>` — Optional argument to specify the container name
```

FULL_HELP_DOCS.md
```diff
@@ -1089,17 +1090,18 @@ By default, when starting a testnet container, without any optional arguments, i
...
 
-* `<NETWORK>` — Network to stop
+* `<NETWORK>` — Network to stop (used in container name generation)
 
   Possible values: `local`, `testnet`, `futurenet`, `pubnet`
 
 
 ###### **Options:**
 
+* `-c`, `--container-name <CONTAINER_NAME>` — Optional argument to specify the container name
```
leighmcculloch, line 1104: Having both network and name on the creation command makes sense, having both on other commands is confusing. If I create a container like so:

```
$ stellar network start testnet --name amazetestnet
```

Why do I need to specify anything other than the name at shutdown?

```
$ stellar network stop amazetestnet
```

The stop command shouldn't need to know what type of network is being shutdown, just it's identifier/name.

If we're adding configurable names, I think the other commands should have `<NETWORK>` replaced with `<NAME>`. E.g.:

```
stellar network stop [OPTIONS] [NAME]
```

leighmcculloch (commented):
I left a couple comments after reviewing the help documentation changes.

FULL_HELP_DOCS.md
elizabethengelman: Cool, that all makes sense.

FULL_HELP_DOCS.md
elizabethengelman: I agree. I think we originally liked the ability to have `network container start <NETWORK>` and `network container stop <NETWORK>` being the same signature. But now that we're adding the ability to change the container name, that probably makes less sense, and adds confusion.

elizabethengelman: I also think that we were allowing for the user not knowing what the container name was and still being able to stop it. But now that we're printing out the full stop and log commands in the `start` command output, I think this is overkill and not necessary.

elizabethengelman: I just pushed up a commit to simplify this so that it can be

```
stellar network start testnet --name tn1
```
and 
```
$ stellar network logs tn1
```
```
$ stellar network stop tn1
```


However, I'm realizing that this is technically a breaking change since we're changing the way that `stop` and `logs` are called. Is this something we want to wait on?

janewang: Yeah I think we want to hold off all breaking changes to protocol boundaries. Let's wait when CLI upgrades to v22.

cmd/soroban-cli/src/commands/network/container/start.rs
```diff
@@ -56,7 +62,7 @@ impl Cmd {
 }
 
 async fn run_docker_command(cmd: &Cmd) -> Result<(), Error> {
-    let docker = connect_to_docker(&cmd.docker_host).await?;
+    let docker = connect_to_docker(&cmd.container_args.docker_host).await?;
```
willemneal, line 65: Perhaps this should be a method on `container_args`?

cmd/soroban-cli/src/commands/network/container/start.rs
```diff
@@ -87,7 +93,7 @@ async fn run_docker_command(cmd: &Cmd) -> Result<(), Error> {
         ..Default::default()
     };
 
-    let container_name = format!("stellar-{}", cmd.network);
+    let container_name = get_container_name(cmd);
```
willemneal, line 96: Just realized that this function takes the `self` as the only arg. So it seems like it should be in an `impl Cmd`.

cmd/soroban-cli/src/commands/network/container/stop.rs
```diff
@@ -1,28 +1,53 @@
...
+        let docker = connect_to_docker(&self.container_args.docker_host).await?;
+        println!("ℹ️ Stopping container: {container_name}");
+        docker
+            .stop_container(&container_name, None)
+            .await
+            .map_err(|e| {
+                let msg = e.to_string();
+                if msg.contains("No such container") {
+                    Error::ContainerNotFound {
+                        container_name: container_name.clone(),
```
willemneal, line 44:
```suggestion
        let container_name = self.name;
        let docker = connect_to_docker(&self.container_args.docker_host).await?;
        println!("ℹ️ Stopping container: {container_name}");
        docker
            .stop_container(&container_name, None)
            .await
            .map_err(|e| {
                let msg = e.to_string();
                if msg.contains("No such container") {
                    Error::ContainerNotFound {
                        container_name: container_name.clone(),
```
Only need the clone in the error closure.

FULL_HELP_DOCS.md
elizabethengelman: Sounds good - I'll revert this for now. I've also created an issue so that we can handle this at the next protocol release: https://github.com/stellar/stellar-cli/issues/1463

cmd/soroban-cli/src/commands/network/container/start.rs
elizabethengelman: After reverting the change mentioned in https://github.com/stellar/stellar-cli/pull/1423#discussion_r1678472001 `get_container_name` is back implemented on Args with an optional network passed in

cmd/soroban-cli/src/commands/network/container/stop.rs
elizabethengelman: This particular change is no longer relevant after reverting https://github.com/stellar/stellar-cli/commit/d21365a0404be7a37201626dd1ecf073b706e089

cmd/soroban-cli/src/commands/network/container/start.rs
elizabethengelman: I like that idea since it would clean up this line. But on the other hand, would that make Args responsible for too much? LIke it would be handling argument related things, and now also starting a docker connection 🤔

FULL_HELP_DOCS.md
leighmcculloch: I don't think this is a breaking change because today the names are not configurable, and the network type is the name, the commands already appear to accept the names.

Today, where all commands accept the network, this is possible:

```
$ stellar network start testnet
$ stellar network logs testnet
$ stellar network stop testnet
```

Tomorrow, where other commands accept the name, this is possible:

```
$ stellar network start testnet
$ stellar network logs testnet
$ stellar network stop testnet
$ stellar network start testnet --name tn1
$ stellar network logs tn1
$ stellar network stop tn1
```

When no --name parameter is passed, the command sequence is the same and unbroken.

cmd/soroban-cli/src/commands/network/container/start.rs
willemneal: This is common across the crate. Personally, the struct that has the data needed should handle the logic. This way you can share the logic and args across commands.

elizabethengelman: Cool, that makes sense. I'll get that change pushed up shortly

FULL_HELP_DOCS.md
elizabethengelman: In the tomorrow scenario above:

When no `--name` param is passed to `stellar network container start testnet`, the container that is started is currently called `stellar-testnet`.

In order to stop the container or get the logs, the user would have to do:
```
$ stellar network container logs stellar-testnet
$ stellar network container stop stellar-testnet
```

This would be slightly different than their existing workflow... because they'd need to pass in the container name rather than the network.

Perhaps this is identifying that defaulting the container name to just the network is what we're after, instead of prepending it with `stellar-`. That would likely simplify things.

leighmcculloch: Ah, I hadn't noticed the stellar- prefix. I do think the stellar prefix is helpful, and maybe it should always be present even when a custom name is provided?

leighmcculloch: Or remove the stellar- prefix, either approach sounds reasonable to me.

elizabethengelman: Thinking about this more - if I do make the change to default the container to the network name, and remove the optional network arg for `stop` and `logs`, then we would have these scenarios as you mentioned above @leighmcculloch. I like it, and am going to spike it out to see how much it simplifies things.

today:
```
$ stellar network container start testnet
$ stellar network container logs testnet
$ stellar network container stop testnet
```

tomorrow:
```
$ stellar network container start testnet
$ stellar network container logs testnet
$ stellar network container stop testnet
$ stellar network container start testnet --name tn1
$ stellar network container logs tn1
$ stellar network container stop tn1
```

elizabethengelman: I didn't see your message earlier - looks like we posted very close to the same time! 

But yeah, I just pushed a change removing the `stellar-` prefix. I also thought it was helpful to have it... but I am thinking that if a user wants it, they can always use the `--name` flag!

leighmcculloch: Giving this some more thought and I think we might regret removing the prefix. Namespacing the containers the CLI is using is helpful, for examples for folks who use docker for other things. It keeps the containers that the stellar CLI somewhat namespaced. 

Can we keep the prefix and apply it to the --name option too? So if you name a network container `tn1` it's docker container name is `stellar-tn1`.

leighmcculloch: I was imagining with this change that the name of the container from the user pov would be without the stellar- prefix, and the stellar- prefix would be something the CLI added for the docker container.

e.g.:
```
$ stellar network container start testnet
$ docker ps -l
CONTAINER ID   IMAGE                         NAMES
56eb4d93c444   stellar/quickstart:testing    stellar-testnet
$ stellar network container logs testnet
$ stellar network container stop testnet

$ stellar network container start testnet --name tn1
$ docker ps -l
CONTAINER ID   IMAGE                         NAMES
c3f279116899   stellar/quickstart:testing    stellar-tn1
$ stellar network container logs tn1
$ stellar network container stop tn1
```

elizabethengelman: Ah, I gotcha! I just pushed an update.

cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -52,60 +126,25 @@ impl fmt::Display for Network {
...
-        h if h.starts_with("npipe://") => {
-            Docker::connect_with_named_pipe(&h, DEFAULT_TIMEOUT, API_DEFAULT_VERSION)
-        }
-        _ => {
-            return Err(Error::UnsupportedURISchemeError {
-                uri: host.to_string(),
-            });
+pub struct Name(Option<String>, Option<Network>);
+impl Name {
+    pub fn new(name: Option<String>, network: Option<Network>) -> Self {
```
willemneal, line 131: Perhaps the name check should happen here, the inner value could be one string, and this could return a result.

cmd/soroban-cli/src/commands/network/container/start.rs
```diff
@@ -51,137 +55,152 @@ pub struct Cmd {
...
-    if let Some(image_override) = &cmd.image_tag_override {
-        println!("Overriding docker image tag to use '{image_override}' instead of '{image_tag}'");
-        image_tag = image_override;
+        port_mapping_hash
     }
 
-    format!("{DOCKER_IMAGE}:{image_tag}")
-}
+    fn container_name(&self) -> Name {
+        Name::new(self.name.clone(), Some(self.network))
```
willemneal, line 167: This seems like the only place where it could be one or the other.
```suggestion
        Name::new(self.name.cloned().unwrap_or(self.network))
```
This would simplify all the other places where you are currently requiring `None`, and the constructor wouldn't need to return an error no have any other checks.

cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -52,60 +126,25 @@ impl fmt::Display for Network {
...
+        Self(name, network)
+    }
+
+    pub fn get_internal_container_name(&self) -> String {
+        match (&self.0, &self.1) {
+            (Some(name), _) => format!("stellar-{}", name),
+            (None, Some(network)) => format!("stellar-{}", network),
+            (None, None) => panic!("Container name and/or network are required."),
```
leighmcculloch, line 139: We should avoid panics in the CLI.

The `Name` object's structure allows for an invalid state to be represented, but it could be structured as an enum to disallow the illegal state so the panic can't occur when calling functions on what would be assumed to be a valid value.

E.g.:
```rust
pub enum Name {
    Network(Network),
    Custom(String),
}
```

cmd/soroban-cli/src/commands/network/container/shared.rs
willemneal: As I pointed out below, Perhaps just a single `String` is enough. Or if you make the inner type public, you can just use `Name(..)`.

cmd/soroban-cli/src/commands/network/container/shared.rs
willemneal: Hey @leighmcculloch! Funny to be reviewing at the same time!

leighmcculloch: 👋🏻 I'll let you take this, I like the idea to simplify the object and stick with the string.

cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -52,60 +126,25 @@ impl fmt::Display for Network {
...
-            return Err(Error::UnsupportedURISchemeError {
-                uri: host.to_string(),
-            });
+pub struct Name(Option<String>, Option<Network>);
+impl Name {
+    pub fn new(name: Option<String>, network: Option<Network>) -> Self {
+        Self(name, network)
+    }
+
+    pub fn get_internal_container_name(&self) -> String {
```
willemneal, line 135: You can implement `Display` and then you don't need this method since then `format` will called `to_string` on it.

Also if you make the inner type public, you can get rid of the other method and just do `name.0`.

leighmcculloch: Fwiw might not even need a type, or a Display impl:
```rust
fn docker_container_name(name: &str) -> String {
    format!("stellar-{name}")
}
```

willemneal: Very true, but I personally find `Name(s)` with a `Display` cleaner than `docker_container_name(s)`.

cmd/soroban-cli/src/commands/network/container/start.rs
elizabethengelman: Nice call - that simplifies things a bunch!

willemneal (approved):

PR #1424, opened 2024-07-01
fix(ts-bindings): adhere to strict TS settings

chadoh:
@fnando noticed that Astro does a strict TS check before building for production, which means that our current template fails to build packages built with `bindings typescript`-generated modules.

```
$ npm run build

> soroban-hello-world@0.0.1 build
> astro check && astro build

16:27:22 [check] Getting diagnostics for Astro files in /Users/fnando/Projects/stellar/soroban-hello-world...
src/contracts/hello_world.ts:4:34 - error ts(2379): Argument of type '{ rpcUrl: any; allowHttp: true; publicKey: undefined; networkPassphrase: "Test SDF Network ; September 2015"; contractId: "CC24T3LXL5ZGIWXZKPBP5YKCMLK3HO6CWUMZS5AV4NEBNCDKVUHQP4UA"; }' is not assignable to parameter of type 'ClientOptions' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'publicKey' are incompatible.

  4 export default new Client.Client({
                                     ~
  5   ...Client.networks.testnet,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
  8   publicKey: undefined,
    ~~~~~~~~~~~~~~~~~~~~~~~
  9 });
    ~

tmp/src/index.ts:67:15 - error ts(4115): This parameter property must have an 'override' modifier because it overrides a member in base class 'Client'.

67   constructor(public readonly options: ContractClientOptions) {
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tmp/src/index.ts:10:1 - error ts(6192): All imports in import declaration are unused.

 10 import type {
    ~~~~~~~~~~~~~
 11   u32,
    ~~~~~~
...
 21   Duration,
    ~~~~~~~~~~~
 22 } from '@stellar/stellar-sdk/contract';
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tmp/src/index.ts:7:3 - error ts(6133): 'Result' is declared but its value is never read.

7   Result,
    ~~~~~~
tmp/src/index.ts:7:3 - error ts(1484): 'Result' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

7   Result,
    ~~~~~~
tmp/src/index.ts:6:3 - error ts(1484): 'ClientOptions' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

6   ClientOptions as ContractClientOptions,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tmp/src/index.ts:2:1 - error ts(6133): 'Address' is declared but its value is never read.

2 import { Address } from '@stellar/stellar-sdk';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Result (8 files):
- 7 errors
- 0 warnings
- 0 hints
```

One of these needs to be fixed by removing [this line](https://github.com/stellar/soroban-astro-template/blob/3c8e0bed309b466af5d25b69c8fe26f730905bb7/initialize.js#L99) from `soroban-astro-template`s `initialize.js` script.

The others are mostly fixed here. I still haven't figured out how to get `astro check` to chill it with the "X is declared but its value is never read" business, but when I run `astro check` locally these show up as _hints_, not errors.


fnando (approved):
not pretty, but does the trick! Thanks!

chadoh:
Sounds like we don’t need this

PR #1430, opened 2024-07-03
add tx hash command

BlaineHeffron:
### What

Adds a command `stellar tx hash` which accepts a Base64 encoded transaction envelope from the stdin and writes out a hex encoded hash of the transaction that will be submitted to the network.

### Why

Could be useful to see what the hash is being submitted to the network https://github.com/stellar/stellar-cli/issues/1316

### Known limitations

N/A


cmd/soroban-cli/src/commands/tx/mod.rs
```diff
@@ -2,26 +2,33 @@ use clap::Parser;
...
 
+pub mod hash;
 pub mod simulate;
 pub mod xdr;
 
 #[derive(Debug, Parser)]
 pub enum Cmd {
     /// Simulate a transaction envelope from stdin
     Simulate(simulate::Cmd),
+    /// Calculate the hash of a transaction envelope from stdin
```
willemneal, line 13:
```suggestion
    /// Calculate the hash of a transaction envelope from stdin and prints it as a hex string
```

willemneal (changes requested):
Looks good just one nit.

FULL_HELP_DOCS.md
```diff
@@ -1283,6 +1285,24 @@ Simulate a transaction envelope from stdin
...
+
+**Usage:** `stellar tx hash [OPTIONS] --source-account <SOURCE_ACCOUNT>`
+
+###### **Options:**
+
+* `--rpc-url <RPC_URL>` — RPC server endpoint
+* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
+* `--network <NETWORK>` — Name of network to use from config
+* `--source-account <SOURCE_ACCOUNT>` — Account that signs the final transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…")
+* `--hd-path <HD_PATH>` — If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
```
leighmcculloch, line 1300: @BlaineHeffron @willemneal @fnando The rpc url, source account, and hd path shouldn't show up to the user as options and the command shouldn't show the source account being required. The only options that matter for this command are `--network-passphrase` and `--network` for selecting what the network passphrase is for calculating the hash.

I assume they're showing up because of the way we have organised the network args as a single group that are all always required.

I don't think this issue needs to block merging this PR specifically, but we should follow up with telling a clearer story for these commands.

leighmcculloch (approved):

FULL_HELP_DOCS.md
willemneal: There is `commands::network::Args` which should include just the network info required!

willemneal (approved):

PR #1435, opened 2024-07-05
Improve markdown output so we can import it on stellar docs.

fnando:
### What

Improve markdown output so we can import it on stellar docs.

### Why

The idea is importing the `FULL_HELP_DOCS.md` file from the stellar docs build pipeline, so we always have the latest output.

### Known limitations

N/A


chadoh (approved):
Why draft? Looks great!

chadoh:
Oh, but when did we stop putting this in the `docs` directory with the helpful README.md that said "AUTOGENERATED, DON'T EDIT DIRECTLY"?

leighmcculloch (approved):

leighmcculloch:
> Oh, but when did we stop putting this in the `docs` directory with the helpful README.md that said "AUTOGENERATED, DON'T EDIT DIRECTLY"?

I think I moved it out of the directory when I was removing things from the repo not being used. It felt like the type of doc that would live well at the top-level, and the doc already has at the bottom a note saying it's automatically generated. Although it looks like @fnando is removing that footer message.

I wouldn't be too worried about someone accidentally editing it because CI should quickly block someone from merging an out of sync help doc.

fnando:
> Although it looks like @fnando is removing that footer message.

I am, because we're going to use the file directly on stellar/stellar-docs, and it wasn't rendering well.

PR #1444, opened 2024-07-11
chore: ignore vscode settings

willemneal:
This is for developing as it will allow integration tests (hidden behind the feature flag) to be indexed by RA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ captive-core/
 !test.toml
 *.sqlite
 test_snapshots
+.vscode/settings.json
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "rust-analyzer.cargo.features": [
+        "it"
+    ]
+}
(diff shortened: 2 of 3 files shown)


fnando (commented):
Do we need to add editor configs to the repo? I'm always on the fence about this. Maybe we should add the `.vscode` to `.gitignore` instead, then you won't be bugged about having a new file.

willemneal:
I wager most Devs use VSCode and if they are going to work on tests this is a must. Also for me if I move to a new computer I prefer to have this already there. It's a small file that doesn't affect other editors. But if there are strong opinions I'm fine having to copy and paste this to each time I set up.

willemneal:
On second thought devs might want to edit it so I just did `git rm --cached` and added it to the `.gitignore` so that future updates will be ignored.

chadoh (approved):
Thanks

fnando (approved):

leighmcculloch:
I highly recommend the Rust Feature Toggler extension if you aren't using it. It provides a status bar UI for toggling whether a feature is enabled or not in a rust project.
- https://marketplace.visualstudio.com/items?itemName=itsyaasir.rust-feature-toggler

PR #1451, opened 2024-07-15
contract build: fix when used in sub-directories of crates

### LINKED ISSUE #1450: contract build: shouldn't build entire workspace when called within workspace sub-directory

21.1.0

### What did you do?

Change to the directory of a specific crate in a workspace of multiple crates. Ran the contract build command.

```
$ cd contracts/add
$ stellar contract build
```

Just the crate I was in get built, mirroring the behavior of `cargo build`.

```
$ cd contracts/add
$ stellar contract build
cargo rustc --manifest-path=Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
```

### What did you see instead?

Entire workspace get built.

```
$ cd contracts/add
$ stellar contract build
cargo rustc --manifest-path=Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
cargo rustc --manifest-path=../call/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
```

leighmcculloch:
### What
Change contract build behavior when executed in the sub-directory of a workspace that contains a crate of that workspace. Make it build only that crate instead of the entire workspace.

### Why
The current behavior is to build the entire workspace.

The behavior of cargo build is to build only the crate of the sub-directory.

This command is designed to immitate cargo build, and differences in behavior are surprising and unexpected.

This change is a breaking change, but it is fixing a bug.

Close #1450


Cargo.toml
```diff
@@ -12,7 +12,7 @@ exclude = ["cmd/crates/soroban-test/tests/fixtures/hello"]
 
 [workspace.package]
 version = "21.1.0"
-rust-version = "1.74.0"
+rust-version = "1.79.0"
```
leighmcculloch, line 15: Required to use `path::absolute`.

cmd/soroban-cli/src/toid.rs
```diff
@@ -61,9 +63,9 @@ impl From<Toid> for u64 {
     }
 }
 
-impl ToString for Toid {
-    fn to_string(&self) -> String {
+impl Display for Toid {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let u: u64 = (*self).into();
-        u.to_string()
+        write!(f, "{u}")
```
leighmcculloch, line 69: Required by clippy for the MSRV change.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -163,23 +168,43 @@ impl Cmd {
...
                         .iter()
                         .any(|t| t.crate_types.iter().any(|c| c == "cdylib"))
-            })
+                }
+            )
             .cloned()
-            .collect()
+            .collect();
+
+        Ok(package)
```
graydon, line 207: Nomenclature nit: if this is a multi-element vector, I think it would be named something plural? I am actually not sure if it's a single-element vector or multi.

graydon (changes requested):

graydon (approved):

PR #1453, opened 2024-07-15
contract build: fix when used in sub-directories without manifests

### LINKED ISSUE #1452: contract build: should find Cargo.toml in parent directory

21.1.0

### What did you do?

Change to a directory inside a workspace that doesn't contain a Cargo.toml. The workspace should have more than one crate. Ran the contract build command.

```
$ cd contracts/add/src
$ stellar contract build
```

All the crates of the workspace to be built, mirroring the behavior of `cargo build`.

```
$ cd contracts/add/src
$ stellar contract build
cargo rustc --manifest-path=Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
cargo rustc --manifest-path=../call/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
```

### What did you see instead?

Error.

```
$ cd contracts/add/src
$ stellar contract build
error: `cargo metadata` exited with an error: error: manifest path `Cargo.toml` does not exist
```

leighmcculloch:
### What
Change contract build behavior when executed in the sub-directory of a workspace that does not contain a manifest. Make it build the entire workspace.

### Why
The current behavior is to error and build nothing.

The behavior of cargo build is to build the entire workspace.

This command is designed to immitate cargo build, and differences in behavior are surprising and unexpected.

This change is a breaking change, but it is fixing a bug.

Close #1452

### Merging

Intended to be merged to `main` after #1451.


willemneal:
Is this for tests? It is harder to write rust with IDE without a manifest.

graydon (approved):

leighmcculloch:
> Is this for tests? It is harder to write rust with IDE without a manifest.

This is for the situation where a manifest is present, but it is not in the current working directory, such as a workspace sub-directory. It's for behavior consistency with cargo build.

willemneal:
Ahh I read it wrong. Makes sense.

PR #1456, opened 2024-07-15
Fixed default futurenet/testnet configs

Row-Bear:
'testnet' default was using futurenet config and vice versa

Please check if I'm correct on this


leighmcculloch:
Fix is released in stellar-cli `v21.1.1`.

PR #1459, opened 2024-07-15
Allow for retries when checking if the emulator is running

elizabethengelman:
### What

Add a retry when checking to see if the ledger emulator container is fully up and running. 

The reason we were experiencing intermittent failures with these tests is because when checking if the emulator was fully up and running, sometimes the container wouldn't even be started yet, and would panic in the `wait_for_emulator_start_text` fn. This change adds a retry with exponential backoff so if the container isn't up yet, and the request fails, it will try again after waiting.

### Why

These tests were flakey, and it was annoying! 😜 

### Known limitations

N/A


leighmcculloch (approved):

leighmcculloch:
This PR needs updating to merge. I'm unable to update it because the fork doesn't allow changes from maintainers.

willemneal (approved):

PR #1464, opened 2024-07-16
Handle host ports with emulator tests 

elizabethengelman:
Troubleshooting emulator tests

Closes https://github.com/stellar/stellar-cli/issues/1461 🤞 

### What

We were still seeing emulator test failures.

### Why/Issue Description

The speculos emulator dockerfile exposes a bunch of ports that we don't necessarily need for our tests. We relied on testcontainer's port publishing, which publishes all ports by default. I think that there were conflicts in which ports were being used for which tests. Since the tests are run in parallel, sometimes a test would try to access the API/UI endpoint on the GH runner host, and another test container was also using that port. 😵 

This PR explicitly only publishes the two ports that we need, and makes sure to get two open ports per test. 

### Known limitations

We are on an older version of testcontainers, which we can hopefully update as part of https://github.com/stellar/stellar-cli/issues/1471. I'm just pointing this out here, because our current version of testcontainers doesn't make it easy to publish just the ports with want, at dynamic ports on the host. Maybe we can clean this up when testcontainers is updated.

Shout out to @Ifropc for digging into this too!


leighmcculloch (approved):

.github/workflows/temp-ledger-emulator.yml
```diff
@@ -0,0 +1,30 @@
+name: Ledger Emulator Tests (temp)
+on:
+  push:
+    branches:
+      - AhaLabs:fix/troubleshooting-emulator-tests
+  pull_request:
+    branches:
+      - AhaLabs:fix/troubleshooting-emulator-tests
```
leighmcculloch, line 8: Needs updating before merge.

leighmcculloch (approved):

elizabethengelman:
Bummer! Thanks for checking this out - I am struggling to repro this locally, even when running the test a bunch of times in short order. 🤔 

Looking at it again this morning - it looks like we are waiting for the UI port to be running, but are not running for the emulated device itself. I will push an update for that shortly.

Ifropc:
Still get the failures on the latest HEAD...
```
failures:

---- emulator_tests::test_sign_tx::when_the_device_is_nanos stdout ----
Error occurred while exchanging with Ledger device: Error connecting to ledger device
thread 'emulator_tests::test_sign_tx::when_the_device_is_nanos' panicked at cmd/crates/stellar-ledger/tests/test/emulator_tests.rs:190:13:
assertion failed: false


failures:
    emulator_tests::test_sign_tx::when_the_device_is_nanos

test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 41.10s

error: test failed, to rerun pass `--test test`
ifro@nixos:~/projects/stellar-cli/ > git log
commit 0b8850348d168ffac72a44b0d46f5e4b18003413 (HEAD, AhaLabs/fix/troubleshooting-emulator-tests)
Author: Elizabeth Engelman <4752801+elizabethengelman@users.noreply.github.com>
Date:   Wed Jul 31 12:29:21 2024 -0400

    Make sure that ports are not conflicting between test threads
```
Seems to be passing on CI though 😅 
Do you think you could add a dumb of container logs just before the failure? Another idea is that maybe we should restart the test all together if getting `Error connecting to ledger device` 🤔

Ifropc:
Ran the test 30 times locally, the only flaky test is `emulator_tests::test_sign_tx::when_the_device_is_nanos` and it failed 28/30 times. Not sure what makes this test special 🤔

elizabethengelman:
> Ran the test 30 times locally, the only flaky test is emulator_tests::test_sign_tx::when_the_device_is_nanos and it failed 28/30 times. Not sure what makes this test special 🤔

I don't know why I didn't see these failures before, but i am seeing it now! 

I think what's happening now is that even though I was checking if the port is in use, I was checking at 127.0.0.1, but it looks like docker is publishing ports at all network interfaces, not just 127.0.0.1. I'll push up a change shortly to check if the port is open on all interfaces (0.0.0.0).

Ifropc:
I was actually trying to run just this test before (in isolation), and it was still flaky (I was also thinking that maybe there was some environment issue). It was before the changes introduced in this PR though

.github/workflows/temp-ledger-emulator.yml
elizabethengelman: Just reverted the commit that adds this file

leighmcculloch (approved):

PR #1470, opened 2024-07-18
fix durability not setting default

BlaineHeffron:
### What

fixes `--durability` argument so that it properly sets default to `persistent` 

### Why

docs say it should have a default value but it didnt

### Known limitations

N/a


BlaineHeffron:
resolves https://github.com/stellar/stellar-cli/issues/1412

willemneal (approved):
Odd that this is needed but thanks for the fix

PR #1478, opened 2024-07-23
change fund to top off if balance below 10k xlm

BlaineHeffron:
### What

When using `stellar network fund` it will now top off the balance if the account exists and has a balance lower than 10k XLM. 

### Why

This makes `fund` more useful because now one doesn't need to create a temporary account and transfer funds from that temporary account in order to add funds. Before, `fund` only created an account that didn't exist and added funds to it. See https://github.com/stellar/stellar-cli/issues/1389 

### Known limitations

N/A


cmd/soroban-cli/src/commands/keys/fund.rs
```diff
@@ -22,13 +53,174 @@ pub struct Cmd {
...
+            "--network-passphrase",
+            network_passphrase,
+            "--",
+            "transfer",
+            "--to",
+            &to_id,
+            "--from",
+            &from_id,
+            "--amount",
+            &amount.to_string(),
```
willemneal, line 166:
```suggestion
        let (from_id, temp_secret) = self.create_temp_account().await?;
        let cmd = commands::contract::invoke::Cmd::parse_arg_vec(&[
            "--id",
            &id.to_string(),
            "--source-account",
            &temp_secret.to_string(),
            "--rpc-url",
            rpc_url,
            "--network-passphrase",
            network_passphrase,
            "--",
            "transfer",
            "--to",
            &to_id.to_string(),
            "--from",
            &from_id.to_string(),
            "--amount",
            &amount.to_string(),
```

cmd/soroban-cli/src/commands/keys/fund.rs
```diff
@@ -22,13 +53,174 @@ pub struct Cmd {
...
+        ])?;
+        match cmd
+            .run(&global::Args {
+                locator: self.address.locator.clone(),
+                filter_logs: Vec::default(),
+                quiet: false,
+                verbose: false,
+                very_verbose: false,
+                list: false,
+                no_cache: false,
```
willemneal, line 176:
```suggestion
                ..Default::default()
```

cmd/soroban-cli/src/commands/keys/fund.rs
```diff
@@ -22,13 +53,174 @@ pub struct Cmd {
...
+            })
+            .await
+        {
+            Ok(()) => Ok(()),
+            Err(e) => Err(Error::FundTransferError(e.to_string())),
```
willemneal, line 181: You should use `.map_err` instead of the match. If you make the inner value the `invoke::Error` type you can do.

```
.map_err(Error::FundTransferError)
```

cmd/soroban-cli/src/commands/keys/fund.rs
```diff
@@ -22,13 +53,174 @@ pub struct Cmd {
...
+        }
+    }
+
     pub async fn run(&self) -> Result<(), Error> {
         let addr = self.address.public_key()?;
-        self.network
-            .get(&self.address.locator)?
-            .fund_address(&addr)
-            .await?;
+        let balance = match self.check_balance().await {
```
willemneal, line 187: `map_or_else`

cmd/soroban-cli/src/commands/keys/fund.rs
BlaineHeffron: I couldn't map directly to FundTransferError because if I accept the Invoke Error type it created a recursive dependency. Write now  I map it to a string. I think I could also wrap it in a Box if you'd rather go that route.

cmd/soroban-cli/src/commands/keys/fund.rs
```diff
@@ -143,15 +143,13 @@ impl Cmd {
             .network_passphrase
             .as_ref()
             .ok_or(Error::MissingNetworkPassphrase)?;
-        let temp_account = self.create_temp_account().await?;
-        let temp_secret = format!("{}", temp_account.1);
-        let from_id = format!("{}", temp_account.0);
+        let (from_id, temp_secret) = self.create_temp_account().await?;
         let to_id = format!("{addr}");
```
willemneal, line 147:
```suggestion
        let to_id = addr.to_string();
```

cmd/soroban-cli/src/commands/keys/fund.rs
```diff
@@ -10,6 +21,26 @@ pub enum Error {
...
+    #[error(transparent)]
+    Clap(#[from] clap::Error),
+    #[error("RPC URL is missing in the network configuration")]
+    MissingRpcUrl,
+    #[error("Network passphrase is missing in the network configuration")]
+    MissingNetworkPassphrase,
+    #[error("Asset contract could not be deployed")]
+    AssetContractError,
+    #[error("Failed to transfer funds: {0}")]
+    FundTransferError(String),
```
willemneal, line 41:
```suggestion
    FundTransferError(invoke::Error),
```
How does this create an recursive dependency? Does `invoke` use fund?

BlaineHeffron: Config::Error contains Keys::Error which is using Invoke::Error which also contains Config::Error

willemneal: Ah I see. This is definitely fine for now.

willemneal (approved):
Yay easy top ups!

leighmcculloch:
I'd like to roll this change back and move ahead with making this change in friendbot. See my comments on the original issue here:
- https://github.com/stellar/stellar-cli/issues/1389#issuecomment-2246689123

willemneal:
@leighmcculloch Our original thoughts were this was a stop gap measure while it was added to friendbot. That is the definitely the proper solution.

leighmcculloch:
Friendbot has been updated with the following change and that change is rolled out:
- https://github.com/stellar/go/commit/31fc8f4236388f12fc609228b7a7f5494867a1f9

So we should be good to roll this back now. Note that the errors friendbot changed slightly, instead of the created error, it returns the following if the account is already funded at or above the starting balance of 10k:
```
{
  "type": "https://stellar.org/horizon-errors/bad_request",
  "title": "Bad Request",
  "status": 400,
  "detail": "account already funded to starting balance"
}
```

leighmcculloch:
I opened this revert, but I'll hold it until @BlaineHeffron or @willemneal if you can confirm if a complete revert is okay or if any of the change needed keeping. Thanks!
- https://github.com/stellar/stellar-cli/pull/1487

PR #1479, opened 2024-07-23
chore: reorganize config into separate module

willemneal:
Initially the config was a subcommand and so it was mixed in with the commands. This separates it out so that we can begin the process of pulling it out into it's own crate.

This should become it's own crate to allow devs to use accounts and aliases when compiling contract.


leighmcculloch (approved):
Feels right to move logic like the shared args out of the network commands module into a shared place due to the common nature and the args not being scoped to the network command.

The refactor looks good to me in that context 👏🏻 .

> This separates it out so that we can begin the process of pulling it out into it's own crate.

That being said, the refactor groups cli code together, but doesn't separate out config parsing/writing logic from tightly coupled cli logic, so I'm not sure this change sufficiently prepares a module for extraction to another crate.

> This should become it's own crate to allow devs to use accounts and aliases when compiling contract.

Could you share more about this workflow? I wouldn't expect compiling contracts to embed aliases or account data into contracts, and even if the cli needed to do that the cli can already access them.

I am probably missing a critical part of how this will serve folks, but moving config into a separate crate seems very low impact / priority at the moment. What am I missing?

PR #1491, opened 2024-07-25
Clean cargo warnings

Ifropc:
### What

Cleanup cargo warnings

### Why

Cleanup cargo warnings

### Known limitations

N/A


cmd/soroban-cli/Cargo.toml
```diff
@@ -14,7 +14,7 @@ default-run = "soroban"
 
 [[bin]]
 name = "stellar"
-path = "src/bin/main.rs"
+path = "../stellar-cli/src/bin/main.rs"
```
Ifropc, line 17: Not sure if it's the cleanest way, but this way we are ensuring that we are pointing to a correct binary (doubt we would want to change soroban-cli and stellar-cli binaries in the future)

leighmcculloch: We should ignore this warning. It's fine to have two binaries reference the same main. The warning is just to catch people me doing a mistake. We are doing it intentionally.

leighmcculloch: There should be an option we can add to the projects `.cargo/config.toml` file or the the GitHub ci workflow to disable that warning.

leighmcculloch (commented):
Looks good except we don't need to change the binary main references. Can you remove that change then it's good.

cmd/soroban-cli/Cargo.toml
Ifropc: Should we maybe just copy main to silence this warning?

Ifropc: There's only `RUSTFLAGS=-Awarnings` which silences all warnings, which is not ideal. I'd rather either copy file (done in the last commit), or just keep the warning there

leighmcculloch: > Should we maybe just copy main to silence this warning?

Yeah, that would be preferred over referencing the other crate file which is technically outside the project.

The files are small, couple copies is fine.

cmd/soroban-cli/Cargo.toml
```diff
@@ -18,7 +18,7 @@ path = "src/bin/main.rs"
```
leighmcculloch, line 18: Could you name the other file `stellar.rs` for symmetry?

leighmcculloch (approved):
One ask to rename the other file to stellar.rs, but regardless lgtm. 👍🏻

PR #1494, opened 2024-07-27
Add error message on system contract

Ifropc:
### What

Add clear error message when trying to fetch asset contract's WASM
```bash
cargo run --bin stellar -- contract fetch  --rpc-url https://docs-demo.stellar-mainnet.quiknode.pro/ --network-passphrase "Public Global Stellar Network ; September 2015" --id CBBJTE34KZHJACB2A6OXCRTDDLX4LI7GKYYD6XYJSGMX46BLNAG752HS
```
```
error: cannot fetch wasm for contract because the contract is a network built-in asset contract that does not have a downloadable code binary
```
As the result, moved logic of parsing contract data from rpc client to cli

### Why

Previous error message was unclear ([see discussion](https://github.com/stellar/stellar-cli/issues/1488))
### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/fetch.rs
```diff
@@ -150,9 +158,19 @@ impl NetworkRunnable for Cmd {
         client
             .verify_network_passphrase(Some(&network.network_passphrase))
             .await?;
-        Ok(client.get_remote_wasm(&contract_id).await?)
+        let data_entry = client.get_contract_data(&contract_id).await?;
+        if let ScVal::ContractInstance(contract) = &data_entry.val {
+            return match &contract.executable {
+                ContractExecutable::Wasm(hash) => {
+                    Ok(client.get_remote_wasm_from_hash(&hash).await?)
```
leighmcculloch, line 165: Instead of calling this function use the `get_ledger_entries` and assume `get_remote_wasm_...` is deprecated. The rpc-client should be a thin wrapper on the rpc api, and both the `get_remote_wasm_...` functions are logic that should live here.

leighmcculloch (commented):
Thanks for working on this. See my comments on https://github.com/stellar/rs-stellar-rpc-client/pull/9.

cmd/soroban-cli/src/commands/contract/fetch.rs
Ifropc: Makes sense. Perhaps we would want to revisit methods we expose in rpc-client in the future to make sure we don't expose methods we don't want to, before it gets bigger adoption and it's harder to remove/deprecate methods people are using?

leighmcculloch: +1, we should do a pass over the lib before v22. The lib started by extracting logic from the cli, but we extracted a lot more than just the client and the boundary between two had a lot of cli logic on the rpc client side. We've moved a lot over, but there's still things there.

leighmcculloch (approved):
Looks great, thanks! 👏🏻 

One comment inline.

Cargo.toml
```diff
@@ -54,6 +54,8 @@ path = "cmd/soroban-cli"
 [workspace.dependencies.soroban-rpc]
 package = "stellar-rpc-client"
 version = "21.4.0"
+git = "https://github.com/stellar/rs-stellar-rpc-client"
+rev = "495001a0118802fa807714c3ef25f5893bc6a0bb"
```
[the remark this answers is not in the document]
Ifropc: Hmm yes now that we don't change anything only deprecate stuff we can just keep 21.4.0 as is. I kept it in for testing locally, but I agree we can just set it back to 21.4.0

leighmcculloch:
To update the CLI Help Doc, run `make generate-full-help-doc`.

PR #1500, opened 2024-07-29
Remove lock file from template and update sdk version

### LINKED ISSUE #1498: contract init: incompatible versions of soroban-sdk & soroban-token-sdk

`21.2.0`

### What did you do?

```
stellar contract init ohno --with-example token
cd ohno
stellar contract build
```

all contracts build

### What did you see instead?

```
cargo rustc --manifest-path=contracts/hello_world/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished `release` profile [optimized] target(s) in 0.04s
   Compiling soroban-token-contract v0.0.6 (/Users/chadoh/code/s/jsdk/test/e2e/test-contracts/contracts/token)
error[E0152]: duplicate lang item in crate `soroban_sdk` (which `soroban_token_sdk` depends on): `panic_impl`.
  |
  = note: the lang item is first defined in crate `soroban_sdk` (which `soroban_token_contract` depends on)
  = note: first definition in `soroban_sdk` loaded from /Users/chadoh/code/s/jsdk/test/e2e/test-contracts/target/wasm32-unknown-unknown/release/deps/libsoroban_sdk-6a9b2044ff704ebd.rlib
  = note: second definition in `soroban_sdk` loaded from /Users/chadoh/code/s/jsdk/test/e2e/test-contracts/target/wasm32-unknown-unknown/release/deps/libsoroban_sdk-f4651df86ac6e5de.rlib, /Users/chadoh/code/s/jsdk/test/e2e/test-contracts/target/wasm32-unknown-unknown/release/deps/libsoroban_sdk-f4651df86ac6e5de.rmeta
```

### Is there a workaround?

Downgrade token contract's `soroban-token-sdk` dependency to match workspace `soroban-sdk` version.

- in `ohno/contracts/token/Cargo.toml` => `soroban-token-sdk` has version `"21.2.0"`
- in `ohno/Cargo.toml` => `soroban-sdk` has version `"20.3.2"`

The examples were updated to use `21.x` here:

- https://github.com/stellar/soroban-examples/pull/318

Now `contract init` uses an out-of-date `Cargo.toml`.

I tried updating the root workspace's `soroban-sdk` version to `21.4.0` and ended up with irresolvable `ed25519-dalek` dependency resolution 🤔 

### What should we do?

We could fix this the easy way, updating the `Cargo.toml` to use the updated `soroban-sdk` version. I say this is easy in the hopes that I made some dumb mistake with that `ed25519-dalek` dependency resolution issue 😅 

I wonder if it would be better to also add `soroban-token-sdk` to the root workspace dependencies, so that we can update packages to use `soroban-token-sdk = { workspace = true }`, in addition to inheriting `soroban-sdk` from the workspace.

leighmcculloch:
Fixes #1498 

### What
Remove the Cargo.lock file from init template and update sdk version to v21 without specifying a minor/patch version.

### Why
We shouldn't be too prescriptive about the dependencies that folks use, such as providing a Cargo.lock file. Cargo will generate a lock file for them at the point in time they generate their template. This will hopefully prevent us from creating any unexpected inconsistencies such as out of sync ed25519-dalek versions (as was reported).

The CLI should always be encouraging use of the latest SDK, no matter the versions of examples. The CLIs job is to setup a workspace ready to go, and if we've lagged on updating examples we gotta get on that rather than keep the CLI on a past version.

We don't need to specify minor versions for the SDK version. Cargo will select the latest available on the first run.

Close https://github.com/stellar/stellar-cli/issues/1498


cmd/soroban-cli/src/utils/contract-init-template/Cargo.toml.removeextension
```diff
@@ -5,7 +5,7 @@ members = [
 ]
 
 [workspace.dependencies]
-soroban-sdk = "20.3.2"
+soroban-sdk = "21.0.0"
```
Ifropc, line 8: Should it be updated as part of bump-version.yaml?
Alternatively, should we move template to `soroban-templates`? The versions of soroban-sdk in examples are updated as the part of the `soroban-templates` repo, so wouldn't it be easier to bump all versions at once if it's there? We can also pin commit hash in stellar-cli and update the hash when new protocol version is released (and if necessary template change is required)

Ifropc (approved):

cmd/soroban-cli/src/utils/contract-init-template/Cargo.toml.removeextension
leighmcculloch: The bump-version action is shared across repos, so it might be hard to modify it to do this. We could make a CI job that errors if this value is inconsistent with the cli major version, that'd be relatively simple to add.

I'm not sure we should change the overall structure of the examples -> templates since the command advertises that you can use any example.

I'll add the CI check so we don't miss this again.

Ifropc: I also noticed that we have git hash pinned for examples in e2e test
https://github.com/stellar/stellar-cli/blob/main/.github/workflows/e2e.yml#L63-L64
While in `init` we use latest commit. Do you think we should update `system-test` to just use `stellar init` instead? In this case we don't need another CI job, because the e2e test (should) catch a regression

chadoh: Yes, we should update that job. I suspect we were using a pre-released version of the examples because they needed a tweaked version early on when setting them up to use init.

leighmcculloch: I've added a test that checks the value stays updated, and removed the pinning of the examples.

leighmcculloch:
A nodejs test is failing in the e2e tests.

.github/workflows/e2e.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       SYSTEM_TEST_VERBOSE_OUTPUT: "true"
 
       # the soroban test cases will compile various contracts from the examples repo
-      SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_HASH: "782458d31395b0191ac40c9572ac53f13f573aa3"
+      SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_HASH: "main"
```
willemneal, line 63: Is this temporary? Could we move to using tagged releases?

Ifropc: I think we need to use `main`, because that's the behavior of `init` command. If we pin the hash, we need to also do it in the `init` command, otherwise our tests won't catch the regression

willemneal: Oh good point, I feel like main is always dangerous even for init. But since we use it there we should test it.

willemneal (approved):

Ifropc:
@willemneal should this be closed? This PR also includes version bump for `soroban-sdk`

PR #1509, opened 2024-07-31
Add initial implementation of rich output with emojis.

fnando:
### What

Running command with `--quiet`:

```console
$ stellar contract build && stellar --quiet contract deploy --wasm target/wasm32-unknown-unknown/release/hello_world.wasm --source alice --network testnet
cargo rustc --manifest-path=contracts/hello_world/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished `release` profile [optimized] target(s) in 0.17s
CDM53PWFXWQGGHUVST644EHZFEQFPL7S3AXN3Q23NR44OCOPWUE6HCVD

```

Running command and uploading contract for the first time:

```console
$ stellar contract build && stellar contract deploy --wasm target/wasm32-unknown-unknown/release/hello_world.wasm --source alice --network testnet
cargo rustc --manifest-path=contracts/hello_world/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
   Compiling hello-world v0.0.0 (/Users/fnando/Projects/stellar/soroban-hello-world/contracts/hello_world)
    Finished `release` profile [optimized] target(s) in 0.46s
ℹ️ Simulating install transaction…
🌎 Submitting install transaction…
ℹ️ Using wasm hash 48461f23c191d6482cf07d58933642fa9b3daea39dee0b7f7011c55dba2b6694
🔗 https://stellar.expert/explorer/testnet/tx/109e142001245fc7b0465cdb4b16fc26714d1cadc56f4394b4ed9d3412a73bd4
🔗 https://stellar.expert/explorer/testnet/contract/CBUSXDM33IOGNQQFCLC7Q7AALU4OSIVIBFNHBOFRWFRBLPDTRUNIBCII
✅ Deployed!
CBUSXDM33IOGNQQFCLC7Q7AALU4OSIVIBFNHBOFRWFRBLPDTRUNIBCII
```

Reusing contract that hasn't changed:

```console
$ stellar contract build && stellar contract deploy --wasm target/wasm32-unknown-unknown/release/hello_world.wasm --source alice --network testnet
cargo rustc --manifest-path=contracts/hello_world/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished `release` profile [optimized] target(s) in 0.07s
ℹ️ Reusing contract previously uploaded
ℹ️ Using wasm hash 48461f23c191d6482cf07d58933642fa9b3daea39dee0b7f7011c55dba2b6694
🔗 https://stellar.expert/explorer/testnet/tx/7f8bae8b8f47be2a19182060facbeb6f2433aea9950afd503683fde8471d367f
🔗 https://stellar.expert/explorer/testnet/contract/CB7NGFIYUJVECW4ULO4SQFNLO5IYMEYUYZRVWETEWBGI4FHINENCAB2T
✅ Deployed!
CB7NGFIYUJVECW4ULO4SQFNLO5IYMEYUYZRVWETEWBGI4FHINENCAB2T
```

### Why

So users can have more info when commands are being executed.

### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -218,25 +228,43 @@ impl NetworkRunnable for Cmd {
             salt,
             &key,
         )?;
+
         if self.fee.build_only {
+            output.check("Done!");
```
willemneal, line 233:
```suggestion
            output.check("Transaction built!");
```

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -218,25 +228,43 @@ impl NetworkRunnable for Cmd {
...
             return Ok(TxnResult::Txn(txn));
         }
 
+        output.info("Simulating deploy transaction…");
+
         let txn = client.simulate_and_assemble_transaction(&txn).await?;
         let txn = self.fee.apply_to_assembled_txn(txn).transaction().clone();
+
         if self.fee.sim_only {
+            output.check("Done!");
```
willemneal, line 243:
```suggestion
            output.check("Simulated and prepared transaction!");
```
We use "assembled" everywhere else to mean a transaction that has been updated based the info from the simulation. So we could also call it that.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -218,25 +228,43 @@ impl NetworkRunnable for Cmd {
...
+        utils::log_transaction(&output, &txn, &network, true)?;
+
         let get_txn_resp = client
             .send_transaction_polling(&config.sign_with_local_key(txn).await?)
             .await?
             .try_into()?;
+
         if global_args.map_or(true, |a| !a.no_cache) {
             data::write(get_txn_resp, &network.rpc_uri()?)?;
         }
```
willemneal, line 257: This has always had a bad code smell. With a refactor we can ensure that the global is no longer an Option.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -218,25 +228,43 @@ impl NetworkRunnable for Cmd {
...
         let txn = client.simulate_and_assemble_transaction(&txn).await?;
         let txn = self.fee.apply_to_assembled_txn(txn).transaction().clone();
+
         if self.fee.sim_only {
+            output.check("Done!");
             return Ok(TxnResult::Txn(txn));
         }
+
+        output.globe("Submitting deploy transaction…");
+        utils::log_transaction(&output, &txn, &network, true)?;
```
willemneal, line 248: Is the quiet supposed to be the last argument? If so, wouldn't the `Output` print method silence it anyway?

fnando: No, the last argument is whether we want to display the explorer link or not. When only simulating, we won't be submitting transaction, so we won't have a link.

That said, I forgot to send whether is only building/simulating, rather than passing `true`.

fnando: oh, in this case I'm passing `true` because `build_only` and `sim_only` early return, so there's no need to pass those values.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -218,25 +228,43 @@ impl NetworkRunnable for Cmd {
...
```
willemneal, line 248: Given that this function requires `output` does it make sense to make it an instance method?

cmd/soroban-cli/src/utils.rs
```diff
@@ -29,6 +32,44 @@ pub fn transaction_hash(tx: &Transaction, network_passphrase: &str) -> Result<[u
...
+}
+
+/// # Errors
+///
+/// Might return an error
+pub fn log_transaction(
+    output: &Output,
+    tx: &Transaction,
+    network: &Network,
+    show_link: bool,
```
willemneal, line 59: It seems that this function does nothing if this boolean is false. Do we need to call it all in the case where it is false?

cmd/soroban-cli/src/output.rs
```diff
@@ -0,0 +1,31 @@
...
+        self.print("ℹ️", message);
+    }
+
+    pub fn globe(&self, message: &str) {
+        self.print("🌎", message);
+    }
+
+    pub fn link(&self, message: &str) {
+        self.print("🔗", message);
+    }
```
willemneal, line 30:
```suggestion
    fn print<T: Display>(&self, icon: &str, message: T) {
        if !self.quiet {
            eprintln!("{icon} {message}");
        }
    }

    pub fn check<T: Display>(&self, message: T) {
        self.print("✅", message);
    }

    pub fn info<T: Display>(&self, message: T) {
        self.print("ℹ️", message);
    }

    pub fn globe<T: Display>(&self, message: T) {
        self.print("🌎", message);
    }

    pub fn link<T: Display>(&self, message: T) {
        self.print("🔗", message);
    }
```

This will allow passing `String`, or `&str` as well as any type that implements `Display`.

cmd/soroban-cli/src/utils.rs
fnando: Fixed! It should output the transaction hash. 😬

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
fnando: thought about it, but didn't have a strong opinion. I'll move it over to the `Output` struct then.

cmd/soroban-cli/src/commands/contract/install.rs
```diff
@@ -153,6 +163,7 @@ impl NetworkRunnable for Cmd {
                             // Skip reupload if this isn't V0 because V1 extension already
                             // exists.
                             if code.ext.ne(&ContractCodeEntryExt::V0) {
+                                output.info("Reusing contract previously uploaded");
```
leighmcculloch, line 166: The term reusing and contracts make it a little ambiugous as to what is happening. It makes it sound like the contract deployment is being reused and like maybe a deploy isn't taking place. When in fact we're just skipping the code install. Suggest changing to using the 'install' terminology to use consistent vocab and narrative:
```suggestion
                                output.info("Skipping install because wasm already installed");
```

leighmcculloch (approved):
One ask, otherwise 👍🏻.

PR #1510, opened 2024-07-31
Make global arguments... global.

fnando:
### What

Make global arguments global, so we can use it independently of the command order.

```console
$ stellar contract deploy --wasm target/wasm32-unknown-unknown/release/hello_world.wasm --source alice --network testnet --quiet
CBGMG2H756O2DP4BDDOPW6F5S6EJBJIMW3TXLDZ33236XS5KWZFN5PRP

$ echo $?
0
```

^Notice that `--quiet` is being used at the end of the command.

### Why

Better UI/UX.

### Known limitations

N/A


cmd/soroban-cli/src/commands/global.rs
```diff
@@ -11,27 +11,27 @@ pub struct Args {
...
     pub verbose: bool,
 
     /// Log DEBUG and TRACE events
-    #[arg(long, visible_alias = "vv")]
+    #[arg(long, visible_alias = "vv", global = true)]
     pub very_verbose: bool,
 
     /// List installed plugins. E.g. `stellar-hello`
-    #[arg(long)]
+    #[arg(long, global = true)]
```
leighmcculloch, line 30: List doesn't make much sense being global. It's intended for getting a list of plugins, which is something that doesn't make much sense on sub commands.

willemneal, line 30:
```suggestion
    #[arg(long)]
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -76,11 +76,11 @@ pub enum Error {
 #[group(skip)]
 pub struct Args {
     /// Use global config
-    #[arg(long)]
+    #[arg(long, global = true)]
```
willemneal, line 79: Some commands like `keys add` have a `global` arg. So I curious how this would be handled. In those cases we can thread the global through and use this arg.

fnando: I don't think it affects in any way. My understanding is that setting `global = true` only means you can use the switch anywhere in the command.

willemneal: Oh duh it's the same `--global` 🤦

leighmcculloch: Moving this PR to draft until this is addressed.

leighmcculloch: Ah yeah sorry. All good. I misunderstood.

leighmcculloch (approved):

Issue #1380, opened 2024-06-17
`keys generate`: don't overwrite existing account


chadoh:
### What problem does your feature solve?

Right now if you have an existing account named `alice` and you run `keys generate` again, it will generate a new account, overwrite the old `alice.toml` file with the new seed phrase, and fund the new account.

```
$ cat .soroban/identity/alice.toml
───────┬────────────────────────────────────
       │ File: .soroban/identity/alice.toml
───────┼────────────────────────────────────
   1   │ seed_phrase = "zoo solid..."
───────┴────────────────────────────────────
$ soroban keys generate alice --network local
$ cat .soroban/identity/alice.toml
───────┬────────────────────────────────────
       │ File: .soroban/identity/alice.toml
───────┼────────────────────────────────────
   1   │ seed_phrase = "camp endorse..."
───────┴────────────────────────────────────
```

### What would you like to see?

By default

- check if the account exists and is funded above some level
- If it's not sufficiently funded, fund it (see #1389)
- mention that if you want to regenerate the account with a new seed phrase, you can `keys rm` first

Here's an example of what this could look like:

```
$ soroban keys generate alice --network local
alice already exists!
    seed phrase found at: /path/to/.soroban/identity/alice.toml
    current XLM balance: 231
Balance >100 XLM, nothing to do. 
To generate new keys for alice, you can `keys rm alice` first.
```

Or, if the account doesn't have sufficient balance:

```
$ soroban keys generate alice --network local
alice already exists!
    seed phrase found at: /path/to/.soroban/identity/alice.toml
    account XLM balance: 9
Topping off... Balance is now 1009 XLM.
To generate new keys for alice, you can `keys rm alice` first.
```

### What alternatives are there?

With #1389, we're considering adding a `--minimum-balance` option. I think for the sake of clarity & simplicity, we do NOT add that to `keys generate`. But should we mention the existence of `keys fund` in the output here?

leighmcculloch:
Friendbot doesn't support topping off. Friendbot only creates new accounts and doesn't currently have the ability to send funds to existing accounts.

Instead of adding force, we can error and require the explicit removal instead of adding new options. So the way to do a force is just a remove and generate. Wdyt?

willemneal:
You can top off with friendbot. Generate a temporary account then transfer all funds to existing account. But after some reflection, since we have a `keys rm` it's not that hard for users to delete without needing to add another option.

chadoh:
[...]

PR #1513, opened 2024-07-31
Add check to avoid overriding existing keys

BlaineHeffron:
### What

`stellar keys generate` will no longer overwrite an identity if it already exists, unless a different seed is provided.

### Why

Workflows that  utilize `generate` might unintentionally overwrite during contract development. 

Close https://github.com/stellar/stellar-cli/issues/1380

### Known limitations

N/A


cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -60,17 +100,36 @@ impl Cmd {
...
+                } else {
+                    // No seed provided, inform user that identity already exists
+                    self.handle_existing_identity(&existing_secret).await?;
+                    return Ok(());
+                }
+            }
+            Err(_) => {
+                // Identity doesn't exist, create new one
+                self.write_and_fund_identity(&secret).await?;
+            }
```
willemneal, line 132: Too hard to write inline suggestion, but I would use the following since you don't care what the error is.
```suggestion
        let Some(existing_secret) = self.config_locator.read_identity(&self.name) {
        ...
        } else {
             // Identity doesn't exist, create new one
            self.write_and_fund_identity(&secret).await?;
        };
    
```

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -60,17 +100,33 @@ impl Cmd {
...
-            network
-                .fund_address(&addr)
-                .await
-                .map_err(|e| {
-                    tracing::warn!("fund_address failed: {e}");
-                })
-                .unwrap_or_default();
+
+        // Check if identity exists
+        if let Ok(existing_secret) = self.config_locator.read_identity(&self.name) {
```
willemneal, line 105:
```suggestion
        let Ok(existing_secret) = self.config_locator.read_identity(&self.name) else {
            // Identity doesn't exist, create new one
            self.write_and_fund_identity(&secret).await?;
            return Ok(());
        };
```
Then  the stuff currently indented in the `if` can be shifted over for better readability.

leighmcculloch (changes requested):
We can't merge this change at the moment, not until closer to a v22 release, because it breaks existing behavior.

janewang:
Hi @BlaineHeffron Did we agree on this feature? Could you show me the thread of the discussion please

willemneal:
@leighmcculloch How about a flag to prevent the overwrite? Then we can deprecate it in the next breaking release?

BlaineHeffron:
> Hi @BlaineHeffron Did we agree on this feature? Could you show me the thread of the discussion please

There isn't a discussion thread that I'm aware of, it was something we had discussed with other AhaLabs team members.

leighmcculloch:
> @leighmcculloch How about a flag to prevent the overwrite? Then we can deprecate it in the next breaking release?

It's unclear how effective an option would be. I doubt the average CLI user is going to start using an `--prevent-overwrite` option when generating keys, it just seems a bit esoteric.

What use case / workflow does the option serve?

BlaineHeffron:
The app `Loam` which we are developing has a configuration option to auto generate contract wasms and install / deploy them in addition to generating and funding accounts. Currently we have a workaround where we check if the account exists before running `generate`. The problem with overwriting the account is someone might have some contract state tied to a specific account which would get lost if the account is overwritten.

chadoh:
Original discussion thread in #1380 and [in Slack](https://aha-labs.slack.com/archives/C04B02ABF37/p1718654369768779). @janewang, @fnando, @elizabethengelman, and @willemneal were all in strong agreement with "overwrite by default is a bug" back in June.

leighmcculloch:
> ...were all in strong agreement with "overwrite by default is a bug" back in June.

Yup, I think the idea of avoiding overwriting a key, which is a destructive action, is good too, but it needs to be held until the next major version because it is a breaking change. And I was otherwise only pushing back on the suggestion to add a temporary `--no-overwrite` since a major release is planned in the near future.

@Ifropc has labelled the issue as a `breaking change` meaning we'll circle back to it to the period leading up to the major release when main is open for breaking changes.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -60,17 +102,31 @@ impl Cmd {
...
+                existing_secret.private_key(self.hd_path),
+                secret.private_key(self.hd_path),
+            ) {
+                (Ok(existing_pk), Ok(new_pk)) if existing_pk == new_pk => {
+                    self.handle_existing_identity(&existing_secret).await?;
+                }
+                _ => {
+                    // Secrets don't match
+                    eprintln!("An identity with the name {} already exists but has a different secret. Overwriting...", self.name);
+                    self.write_and_fund_identity(&secret).await?;
```
leighmcculloch, line 124: I think the idea that the command will implicitly overwrite in some edge cases is surprising. If we're saying "overwrite by default is a bug" then it's a bug that it still overwrites here.

The only time overwriting should be allowed is if an explicit `--overwrite` is provided so that the overwrite only occurs explicitly.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -49,6 +50,47 @@ pub struct Cmd {
...
+        self.fund_identity(secret).await?;
+        Ok(())
+    }
+
+    async fn handle_existing_identity(&self, existing_secret: &Secret) -> Result<(), Error> {
+        eprintln!("The identity {} already exists!", self.name);
+        if let Ok(root) = self.config_locator.config_dir() {
+            let mut path = root.join("identity").join(&self.name);
+            path.set_extension("toml");
+            eprintln!("    Seed phrase found at: {}", path.display());
```
leighmcculloch, line 88: If someone attempts to generate a key for an alias that already has a key, if the command won't be doing it's primary function of generating a new key, we should have the command error with a non-zero exit status. Otherwise the command appears to have created a new-random-never-seen-before key, when in fact the alias may refer to a key that's not new and not random, which could be a security concern.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -49,6 +50,47 @@ pub struct Cmd {
...
+        }
+        self.fund_identity(existing_secret).await?;
```
leighmcculloch, line 90: I don't think the command should fund an existing account. Unclear the purpose this serves since the generate command is primarily about getting a new key. If someone needs to top up their account they can use the fund command.

There's an open issue to remove funding from the generate command:
- https://github.com/stellar/stellar-cli/issues/1407

leighmcculloch (commented):
Couple more things I noticed that I think we should change before merging this in the next major release.

cmd/soroban-cli/src/commands/keys/generate.rs
BlaineHeffron: I added an `--overwrite` flag which simplified the logic as there was no need to compare existing secrets.

cmd/soroban-cli/src/commands/keys/generate.rs
BlaineHeffron: I now throw an error `Error::IdentityAlreadyExists` which gets thrown if they try to run it without overwrite but the identity is found.

cmd/soroban-cli/src/commands/keys/generate.rs
BlaineHeffron: I removed --no-fund and the funding portions.

BlaineHeffron:
Comments have been addressed. This PR now also resolves https://github.com/stellar/stellar-cli/issues/1407

- I added an `--overwrite` flag to force overwrites. Otherwise it never overwrites. 
- If it finds that it would overwrite but `--overwrite` is not set, it throws an Error.
- It no longer funds the account afterwards. This also means there is no longer a `--no-fund` arg. 

Now I need to fix all the integration tests so that they fund the account after generate

BlaineHeffron:
Actually, does it make sense to just add a --fund so that you could do both at once like the old functionality? (Would also make it easier for me to fix the tests ) @leighmcculloch

BlaineHeffron:
Tests are fixed, should be ready.

fnando:
Closing in favor of #1709

Issue #225, opened 2022-10-20
Bail on submitting if the simulate transaction shows a readonly footprint


kalepail:
### What problem does your feature solve?
Read-only contract functions shouldn't be able to be submitted. Or maybe we allow some function flags at the contract layer like Solidity does for things like `pure`, `view`, `payable`, etc. Being able to submit a transaction that makes no changes and just returns data seems weird.

### What would you like to see?
At the very least there should be a flag that allows the CLI to only call and return the preflight.

See thread on Discord
https://discord.com/channels/897514728459468821/966788672164855829/1032397970911526993

leighmcculloch:
> At the very least there should be a flag that allows the CLI to only call and return the preflight.

I think this is a duplicate of https://github.com/stellar/soroban-cli/issues/221.

---

> Read-only contract functions shouldn't be able to be submitted.

I think it's difficult for the CLI to know for sure if a function is always read-only. It's also difficult for the SDK to guarantee a function will be read-only, so unfortunately if we did add a marker to the function it'd likely be more symbolic, so I'm hesitant to agree we should add that.

As @dmkozh mentioned on Discord the CLI can detect that the footprint in the simulation is read-only, which means that specific invocation was read-only. The CLI could bail early if it detects this.

This behavior might be inconsistent. A function might be read-only 90% of the time, then write sometimes. This could be surprising if a CLI invoke command works 90% of the time then actually submits 10% of the time. I don't think there's anything for us to do about this though. If the simulation is read-only, then the footprint will be read-only, and so even if the tx is submitting to network it will never have any impact, either that or it will fail if it tried to write. So it's probably fine to have it work like this:

The following command would exit early after simulation if the footprint is read-only:
```
$ soroban invoke --id ... --fn sometimes_read_only
stdout: <return value>
```

The following command would only ever do simulation (this is #221):
```
$ soroban invoke --id ... --fn sometimes_read_only --simulate
stdout: <return value>
```

If we discover a need where we want invocation to _always_ submit to network we could add a new option at that time. We wouldn't need to add this until the need presented itself though.
```
$ soroban invoke --id ... --fn sometimes_read_only --force-submit
stdout: <return value>
```

tyvdh:
> The following command would exit early after simulation if the footprint is read-only:
[...]

PR #1518, opened 2024-08-01
feat: add `--send` to `contract invoke` and default to only sending transaction if there are writes in the ledger footprint

willemneal:
### What

See initial discussion here: https://github.com/stellar/stellar-cli/issues/225

### Why

Currently users must opt in to not sending a transaction with `--is-view`. This deprecates `--is-view` and flips the default to only send if there are writes in the ledger footprint returned by the simulation.

It improves UX because now when making a view call a user must wait for the transaction be complete vs viewing the simulation result.

### Known limitations

[TODO or N/A]


FULL_HELP_DOCS.md
```diff
@@ -480,7 +480,7 @@ stellar contract invoke ... -- --help
 ###### **Options:**
 
 * `--id <CONTRACT_ID>` — Contract ID to invoke
-* `--is-view` — View the result simulating and do not sign and submit transaction
+* `--is-view` — View the result simulating and do not sign and submit transaction. Ieprecated use `--send=no`
```
leighmcculloch, line 483:
```suggestion
* `--is-view` — View the result simulating and do not sign and submit transaction. Deprecated use `--send=no`
```

FULL_HELP_DOCS.md
```diff
@@ -495,6 +495,18 @@ stellar contract invoke ... -- --help
...
+
+  Default value: `if-write`
+
+  Possible values:
+  - `if-write`:
+    Only send transaction if there are ledger writes, otherwise return simulation result
+  - `no`:
+    Do not send transaction, return simulation result
+  - `yes`:
+    Always send transaction
```
leighmcculloch, line 508:
```suggestion
  - `if-write`:
    Send the transaction to the network if there are ledger writes or events published, otherwise output the simulation result and do not send to the network
  - `no`:
    Do not send the transaction to the network, and output the simulation result
  - `yes`:
    Send the transaction to the network, and output the transaction result
```

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -561,3 +555,29 @@ Each arg has a corresponding --<arg_name>-file-path which is a path to a file co
...
+
+impl Send {
+    pub fn should_send(self, sim_res: &SimulateTransactionResponse) -> Result<bool, Error> {
+        Ok(match self {
+            Send::IfWrite => !sim_res
+                .transaction_data()?
+                .resources
+                .footprint
+                .read_write
+                .is_empty(),
```
leighmcculloch, line 578: Needs to also take into account if the simulation shows that events will be published. Published events are a write.

leighmcculloch (commented):
Couple requests inline.

I think the concept is still pending feedback from @janewang and others.

But I've been hoping to see this functionality for a couple years (#225) so I think adding this or something like this is 💯.

cmd/soroban-cli/src/commands/contract/invoke.rs
willemneal: Ok thanks!

willemneal:
@leighmcculloch should rename to use your issue's names?

Thanks for the review!

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -44,7 +45,7 @@ pub struct Cmd {
     // For testing only
     #[arg(skip)]
     pub wasm: Option<std::path::PathBuf>,
-    /// View the result simulating and do not sign and submit transaction
+    /// View the result simulating and do not sign and submit transaction. Ieprecated use `--send=no`
```
leighmcculloch, line 48:
```suggestion
    /// View the result simulating and do not sign and submit transaction. Deprecated use `--send=no`
```

leighmcculloch (commented):
Took another look, one fix needed, and otherwise just needs events 👏🏻

FULL_HELP_DOCS.md
janewang, line 483: sp: `Deprecated`

willemneal: Yeah this is why we shouldn't check in this file. I updated the source code where these comments come from by accepting a suggestion from Leigh, but then I need to re generate this file locally...

leighmcculloch:
```suggestion
* `--is-view` — View the result simulating and do not sign and submit transaction. Deprecated use `--send=no`
```

cmd/soroban-cli/src/commands/contract/invoke.rs
leighmcculloch:
```suggestion
    /// View the result simulating and do not sign and submit transaction. Deprecated use `--send=no`
```

FULL_HELP_DOCS.md
leighmcculloch: Git history shows it being fixed in the code, then being changed back to `Ieprecated` in the code. Automation won't prevent us from undoing typo fixes in the code. I don't think we need more automation for this.

I'm going to push back on introducing more complexity for the sake of having a simply accessible help file that we can review in PRs as a way to understanding the UX/UI clearly.

willemneal: Fair enough.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -54,6 +56,9 @@ pub struct Cmd {
     pub config: config::Args,
     #[command(flatten)]
     pub fee: crate::fee::Args,
+    /// Whether or not to send a transaction
+    #[arg(long, value_enum, default_value("if-write"))]
```
leighmcculloch, line 60:
```suggestion
    #[arg(long, value_enum, default_value("if-write"), env = "STELLAR_SEND")]
```

leighmcculloch (commented):
Looks good. One thing missing, an env var for the option, otherwise lgtm.

leighmcculloch (approved):

leighmcculloch:
I think we need a follow up change to this change prior to this change being released. (cc @janewang @fnando).

In this change we didn't take into the account the implication of txs that are read-only (no storage writes, no events), but that require auth. Auth checks are never run, and auth code is not executed.

It's not obvious what the behavior should be.

Simulating and considering the tx as read-only may be appropriate. Or it maybe surprising.

Maybe we at least need a warning outputted that the tx is not being sent because simulation indicated that the tx was read-only.

Or maybe we need to treat any tx with an auth as a read-write. Folks can still force it not to submit with `--send=no`.

Thoughts anyone?

PR #1524, opened 2024-08-06
Move snapshot output to the rich output system.

fnando:
### What

https://github.com/user-attachments/assets/9f6fea54-5e1c-4947-83a8-cdd802493648

### Why

So we have one standardized way of doing it.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/snapshot/create.rs
+++ b/cmd/soroban-cli/src/commands/snapshot/create.rs
@@ -26,17 +26,18 @@ use stellar_xdr::curr::{
 use tokio::fs::OpenOptions;
 
 use crate::{
-    commands::{config::data, HEADING_RPC},
+    commands::{config::data, global, HEADING_RPC},
     config::{self, locator, network::passphrase},
+    output,
     utils::{get_name_from_stellar_asset_contract_storage, parsing::parse_asset},
 };
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, ValueEnum)]
-pub enum Output {
+pub enum OutputFile {
     Json,
 }
 
-impl Default for Output {
+impl Default for OutputFile {
     fn default() -> Self {
         Self::Json
     }
@@ -73,7 +74,7 @@ pub struct Cmd {
     wasm_hashes: Vec<Hash>,
     /// Format of the out file.
     #[arg(long)]
-    output: Output,
+    output: OutputFile,
     /// Out path that the snapshot is written to.
     #[arg(long, default_value=default_out_path().into_os_string())]
     out: PathBuf,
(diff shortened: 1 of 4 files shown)


cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -26,17 +26,18 @@ use stellar_xdr::curr::{
...
 };
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, ValueEnum)]
-pub enum Output {
+pub enum OutputFile {
     Json,
 }
 
-impl Default for Output {
+impl Default for OutputFile {
```
leighmcculloch, line 40: We use the term Output elsewhere for the output type, where the output isn't a file, could we keep it as is, and change the `output` module to `print` since it's for printing specifically?

cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -434,26 +446,33 @@ async fn cache_bucket(
         let bucket_2 = &bucket[4..=5];
         let bucket_url =
             format!("{archive_url}/bucket/{bucket_0}/{bucket_1}/{bucket_2}/bucket-{bucket}.xdr.gz");
-        print!("🪣  Downloading bucket {bucket_index} {bucket}");
+        let message = format!("Downloading bucket {bucket_index} {bucket}");
+
+        output.print("🪣", format!("{message}…"), false);
```
leighmcculloch, line 451: Can we add a `.bucket` function instead of hardcoding this one emoji unlike the others?

cmd/soroban-cli/src/output.rs
```diff
@@ -16,26 +16,44 @@ impl Output {
...
-        if !self.quiet {
+    pub fn print<T: Display>(&self, icon: &str, message: T, new_line: bool) {
+        if self.quiet {
+            return;
+        }
+
+        if new_line {
             eprintln!("{icon} {message}");
+        } else {
+            eprint!("{icon} {message}");
```
leighmcculloch, line 27: For consistency with the Rust stdlib, it'd be helpful if we operated the newline option via different function names, `print` and `println` like:
- https://doc.rust-lang.org/std/macro.print.html
- https://doc.rust-lang.org/std/macro.println.html

Other functions can then also get two variants, `check` and `checkln`.

Then I think we don't need the `icon` parameter to print and it can simply accept a `&str` (or preferably a `Display`). For example, the use would be:
```rust
output.globeln(format!("Downloading history {history_url}"));
...
output.bucket(format!("Downloading bucket {bucket_index} {bucket}…"));
...
output.println("({size})");
```

fnando: I also wonder if we should use a macro to generate these functions, as they're essentially the same thing.

fnando:
Close in favor of #1531

PR #1525, opened 2024-08-06
Remove short name, so it doesn't conflict with the verbose short name.

fnando:
### What

Remove short flag due to a conflict introduced by #1510.

### Why

```console
$ stellar network container start
thread 'main' panicked at /Users/fnando/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.5.8/src/builder/debug_asserts.rs:112:17:
Command start: Short option names must be unique for each argument, but '-v' is in use by both 'protocol_version' and 'verbose'
```

### Known limitations

N/A


leighmcculloch (approved):
Needs help updated otherwise lgtm.

PR #1529, opened 2024-08-06
Fix `contract init` short name conflict.

fnando:
### What

Fix `contract init` short name conflict; `-f` is also used by filtering.

### Why

So users can run the command.

### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -41,8 +41,8 @@ pub struct Cmd {
     pub with_example: Vec<String>,
 
     #[arg(
-        short,
         long,
+        short = 't',
```
leighmcculloch, line 45: I think we should omit the short code rather than add a new one. We can always add one later if this is the wrong call.
```suggestion
```

leighmcculloch (approved):
One request otherwise lgtm.

PR #1532, opened 2024-08-07
Add `contract info`

Ifropc:
### What

Adds `contract info` command following https://github.com/stellar/stellar-cli/issues/1506 spec.

Available flags (for all subcommands)
```
* `--wasm <WASM>` — Wasm file to extract the data from
* `--wasm-hash <WASM_HASH>` — Wasm hash to get the data for
* `--id <CONTRACT_ID>` — Contract id to get the data for
* `--global` — Use global config
* `--config-dir <CONFIG_DIR>` — Location of config directory, default is "."
```
Shared output flag for meta/env-meta:
```
* `--output <OUTPUT>` — Format of the output

  Default value: `text`

  Possible values:
  - `text`:
    Text output of the meta info entry
  - `xdr-base64`:
    XDR output of the info entry
  - `json`:
    JSON output of the info entry (one line, not formatted)
  - `json-formatted`:
    Formatted (multiline) JSON output of the info entry
```
Output flag for interface:
```
* `--output <OUTPUT>` — Format of the output

  Default value: `rust`

  Possible values:
  - `rust`:
    Rust code output of the contract interface
  - `xdr-base64`:
    XDR output of the info entry
  - `json`:
    JSON output of the info entry (one line, not formatted)
  - `json-formatted`:
    Formatted (multiline) JSON output of the info entry
```

Examples of running the command: 
## `stellar contract info env-meta`
```
stellar-cli contract info env-meta --network testnet --wasm ~/wasm/soroban_liquidity_pool_contract.wasm 
```
```
Contract env-meta:
 • Protocol: v21
 • Interface: v0
 • Interface Version: 90194313216
```

## `stellar contract info meta`
```
stellar-cli contract info meta --network testnet --wasm ~/wasm/soroban_liquidity_pool_contract.wasm --output json
```
```
[{"sc_meta_v0":{"key":"Description","val":"Constant product AMM with a .3% swap fee"}},{"sc_meta_v0":{"key":"rsver","val":"1.79.0"}},{"sc_meta_v0":{"key":"rssdkver","val":"21.4.0#d6f5639f643d76e758beecbb0ca391f8cd304c24"}}]
```
_____
```
stellar-cli contract info meta --network testnet --wasm ~/wasm/soroban_liquidity_pool_contract.wasm 
```
```
Contract meta:
 • Description: Constant product AMM with a .3% swap fee
 • rsver: 1.79.0 (Rust version)
 • rssdkver: 21.4.0#d6f5639f643d76e758beecbb0ca391f8cd304c24 (Soroban SDK version and it's commit hash)

```
## `stellar contract info interface`
```
stellar contract info interface --network testnet --wasm-hash \
102431a5a5bc74bdb6bf2803c74080b67d070c522482c278a921563cb71ecc1a
```
```
#[soroban_sdk::contractclient(name = "Client")]
pub trait Contract {
    fn hello(
        env: soroban_sdk::Env,
        to: soroban_sdk::Symbol,
    ) -> soroban_sdk::Vec<soroban_sdk::Symbol>;
}
```
_____
```
stellar-cli contract info interface --network testnet --id CC434MK5OCXTDMZTBZXS7WQP6SOYFWVAZY23PRGAPUQZXARGQRZP3FJH
```
```
#[soroban_sdk::contractclient(name = "Client")]
pub trait Contract {
    fn initialize(
        env: soroban_sdk::Env,
        token_wasm_hash: soroban_sdk::BytesN<32>,
        token_a: soroban_sdk::Address,
        token_b: soroban_sdk::Address,
    );
    fn share_id(env: soroban_sdk::Env) -> soroban_sdk::Address;
    fn deposit(
        env: soroban_sdk::Env,
        to: soroban_sdk::Address,
        desired_a: i128,
        min_a: i128,
        desired_b: i128,
        min_b: i128,
    );
    fn swap(
        env: soroban_sdk::Env,
        to: soroban_sdk::Address,
        buy_a: bool,
        out: i128,
        in_max: i128,
    );
    fn withdraw(
        env: soroban_sdk::Env,
        to: soroban_sdk::Address,
        share_amount: i128,
        min_a: i128,
        min_b: i128,
    ) -> (i128, i128);
    fn get_rsrvs(env: soroban_sdk::Env) -> (i128, i128);
}
```
_____

```
stellar-cli contract info interface --network testnet --wasm ~/wasm/soroban_custom_types_contract.wasm             
```
```
#[soroban_sdk::contractclient(name = "Client")]
pub trait Contract {
    fn hello(env: soroban_sdk::Env, hello: soroban_sdk::Symbol) -> soroban_sdk::Symbol;
    fn auth(
        env: soroban_sdk::Env,
        addr: soroban_sdk::Address,
        world: soroban_sdk::Symbol,
    ) -> soroban_sdk::Address;
    fn get_count(env: soroban_sdk::Env) -> u32;
    fn inc(env: soroban_sdk::Env) -> u32;
    fn woid(env: soroban_sdk::Env);
    fn val(env: soroban_sdk::Env) -> soroban_sdk::Val;
    fn u32_fail_on_even(
        env: soroban_sdk::Env,
        u32_: u32,
    ) -> Result<u32, soroban_sdk::Error>;
    fn u32_(env: soroban_sdk::Env, u32_: u32) -> u32;
    fn i32_(env: soroban_sdk::Env, i32_: i32) -> i32;
    fn i64_(env: soroban_sdk::Env, i64_: i64) -> i64;
    fn strukt_hel(
        env: soroban_sdk::Env,
        strukt: Test,
    ) -> soroban_sdk::Vec<soroban_sdk::Symbol>;
    fn strukt(env: soroban_sdk::Env, strukt: Test) -> Test;
    fn simple(env: soroban_sdk::Env, simple: SimpleEnum) -> SimpleEnum;
    fn complex(env: soroban_sdk::Env, complex: ComplexEnum) -> ComplexEnum;
    fn addresse(
        env: soroban_sdk::Env,
        addresse: soroban_sdk::Address,
    ) -> soroban_sdk::Address;
    fn bytes(env: soroban_sdk::Env, bytes: soroban_sdk::Bytes) -> soroban_sdk::Bytes;
    fn bytes_n(
        env: soroban_sdk::Env,
        bytes_n: soroban_sdk::BytesN<9>,
    ) -> soroban_sdk::BytesN<9>;
    fn card(env: soroban_sdk::Env, card: RoyalCard) -> RoyalCard;
    fn boolean(env: soroban_sdk::Env, boolean: bool) -> bool;
    fn not(env: soroban_sdk::Env, boolean: bool) -> bool;
    fn i128(env: soroban_sdk::Env, i128: i128) -> i128;
    fn u128(env: soroban_sdk::Env, u128: u128) -> u128;
    fn multi_args(env: soroban_sdk::Env, a: u32, b: bool) -> u32;
    fn map(
        env: soroban_sdk::Env,
        map: soroban_sdk::Map<u32, bool>,
    ) -> soroban_sdk::Map<u32, bool>;
    fn vec(env: soroban_sdk::Env, vec: soroban_sdk::Vec<u32>) -> soroban_sdk::Vec<u32>;
    fn tuple(
        env: soroban_sdk::Env,
        tuple: (soroban_sdk::Symbol, u32),
    ) -> (soroban_sdk::Symbol, u32);
    fn option(env: soroban_sdk::Env, option: Option<u32>) -> Option<u32>;
    fn u256(env: soroban_sdk::Env, u256: soroban_sdk::U256) -> soroban_sdk::U256;
    fn i256(env: soroban_sdk::Env, i256: soroban_sdk::I256) -> soroban_sdk::I256;
    fn string(env: soroban_sdk::Env, string: soroban_sdk::String) -> soroban_sdk::String;
    fn tuple_strukt(env: soroban_sdk::Env, tuple_strukt: TupleStruct) -> TupleStruct;
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct Test {
    pub a: u32,
    pub b: bool,
    pub c: soroban_sdk::Symbol,
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct TupleStruct(pub Test, pub SimpleEnum);
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub enum SimpleEnum {
    First,
    Second,
    Third,
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub enum ComplexEnum {
    Struct(Test),
    Tuple(TupleStruct),
    Enum(SimpleEnum),
    Asset(soroban_sdk::Address, i128),
    Void,
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub enum RoyalCard {
    Jack = 11,
    Queen = 12,
    King = 13,
}
#[soroban_sdk::contracterror(export = false)]
#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub enum Error {
    NumberMustBeOdd = 1,
}
```
_____
```
 > stellar-cli contract info interface --network testnet --wasm ~/wasm/soroban_token_contract.wasm --output xdr-base64
```
```
AAAAAAAAAAAAAAAKaW5pdGlhbGl6ZQAAAAAABAAAAAAAAAAFYWRtaW4AAAAAAAATAAAAAAAAAAdkZWNpbWFsAAAAAAQAAAAAAAAABG5hbWUAAAAQAAAAAAAAAAZzeW1ib2wAAAAAABAAAAAAAAAAAAAAAAAAAAAEbWludAAAAAIAAAAAAAAAAnRvAAAAAAATAAAAAAAAAAZhbW91bnQAAAAAAAsAAAAAAAAAAAAAAAAAAAAJc2V0X2FkbWluAAAAAAAAAQAAAAAAAAAJbmV3X2FkbWluAAAAAAAAEwAAAAAAAAAAAAAAAAAAAAlhbGxvd2FuY2UAAAAAAAACAAAAAAAAAARmcm9tAAAAEwAAAAAAAAAHc3BlbmRlcgAAAAATAAAAAQAAAAsAAAAAAAAAAAAAAAdhcHByb3ZlAAAAAAQAAAAAAAAABGZyb20AAAATAAAAAAAAAAdzcGVuZGVyAAAAABMAAAAAAAAABmFtb3VudAAAAAAACwAAAAAAAAARZXhwaXJhdGlvbl9sZWRnZXIAAAAAAAAEAAAAAAAAAAAAAAAAAAAAB2JhbGFuY2UAAAAAAQAAAAAAAAACaWQAAAAAABMAAAABAAAACwAAAAAAAAAAAAAACHRyYW5zZmVyAAAAAwAAAAAAAAAEZnJvbQAAABMAAAAAAAAAAnRvAAAAAAATAAAAAAAAAAZhbW91bnQAAAAAAAsAAAAAAAAAAAAAAAAAAAANdHJhbnNmZXJfZnJvbQAAAAAAAAQAAAAAAAAAB3NwZW5kZXIAAAAAEwAAAAAAAAAEZnJvbQAAABMAAAAAAAAAAnRvAAAAAAATAAAAAAAAAAZhbW91bnQAAAAAAAsAAAAAAAAAAAAAAAAAAAAEYnVybgAAAAIAAAAAAAAABGZyb20AAAATAAAAAAAAAAZhbW91bnQAAAAAAAsAAAAAAAAAAAAAAAAAAAAJYnVybl9mcm9tAAAAAAAAAwAAAAAAAAAHc3BlbmRlcgAAAAATAAAAAAAAAARmcm9tAAAAEwAAAAAAAAAGYW1vdW50AAAAAAALAAAAAAAAAAAAAAAAAAAACGRlY2ltYWxzAAAAAAAAAAEAAAAEAAAAAAAAAAAAAAAEbmFtZQAAAAAAAAABAAAAEAAAAAAAAAAAAAAABnN5bWJvbAAAAAAAAAAAAAEAAAAQAAAAAQAAAAAAAAAAAAAAEEFsbG93YW5jZURhdGFLZXkAAAACAAAAAAAAAARmcm9tAAAAEwAAAAAAAAAHc3BlbmRlcgAAAAATAAAAAQAAAAAAAAAAAAAADkFsbG93YW5jZVZhbHVlAAAAAAACAAAAAAAAAAZhbW91bnQAAAAAAAsAAAAAAAAAEWV4cGlyYXRpb25fbGVkZ2VyAAAAAAAABAAAAAIAAAAAAAAAAAAAAAdEYXRhS2V5AAAAAAQAAAABAAAAAAAAAAlBbGxvd2FuY2UAAAAAAAABAAAH0AAAABBBbGxvd2FuY2VEYXRhS2V5AAAAAQAAAAAAAAAHQmFsYW5jZQAAAAABAAAAEwAAAAEAAAAAAAAABVN0YXRlAAAAAAAAAQAAABMAAAAAAAAAAAAAAAVBZG1pbgAAAAAAAAEAAAAAAAAAAAAAAA1Ub2tlbk1ldGFkYXRhAAAAAAAAAwAAAAAAAAAHZGVjaW1hbAAAAAAEAAAAAAAAAARuYW1lAAAAEAAAAAAAAAAGc3ltYm9sAAAAAAAQ

```
_____
```
 > stellar-cli contract info interface --network testnet --wasm ~/wasm/soroban_token_contract.wasm --output json
```
```
[{"function_v0":{"doc":"","name":"initialize","inputs":[{"doc":"","name":"admin","type_":"address"},{"doc":"","name":"decimal","type_":"u32"},{"doc":"","name":"name","type_":"string"},{"doc":"","name":"symbol","type_":"string"}],"outputs":[]}},{"function_v0":{"doc":"","name":"mint","inputs":[{"doc":"","name":"to","type_":"address"},{"doc":"","name":"amount","type_":"i128"}],"outputs":[]}},{"function_v0":{"doc":"","name":"set_admin","inputs":[{"doc":"","name":"new_admin","type_":"address"}],"outputs":[]}},{"function_v0":{"doc":"","name":"allowance","inputs":[{"doc":"","name":"from","type_":"address"},{"doc":"","name":"spender","type_":"address"}],"outputs":["i128"]}},{"function_v0":{"doc":"","name":"approve","inputs":[{"doc":"","name":"from","type_":"address"},{"doc":"","name":"spender","type_":"address"},{"doc":"","name":"amount","type_":"i128"},{"doc":"","name":"expiration_ledger","type_":"u32"}],"outputs":[]}},{"function_v0":{"doc":"","name":"balance","inputs":[{"doc":"","name":"id","type_":"address"}],"outputs":["i128"]}},{"function_v0":{"doc":"","name":"transfer","inputs":[{"doc":"","name":"from","type_":"address"},{"doc":"","name":"to","type_":"address"},{"doc":"","name":"amount","type_":"i128"}],"outputs":[]}},{"function_v0":{"doc":"","name":"transfer_from","inputs":[{"doc":"","name":"spender","type_":"address"},{"doc":"","name":"from","type_":"address"},{"doc":"","name":"to","type_":"address"},{"doc":"","name":"amount","type_":"i128"}],"outputs":[]}},{"function_v0":{"doc":"","name":"burn","inputs":[{"doc":"","name":"from","type_":"address"},{"doc":"","name":"amount","type_":"i128"}],"outputs":[]}},{"function_v0":{"doc":"","name":"burn_from","inputs":[{"doc":"","name":"spender","type_":"address"},{"doc":"","name":"from","type_":"address"},{"doc":"","name":"amount","type_":"i128"}],"outputs":[]}},{"function_v0":{"doc":"","name":"decimals","inputs":[],"outputs":["u32"]}},{"function_v0":{"doc":"","name":"name","inputs":[],"outputs":["string"]}},{"function_v0":{"doc":"","name":"symbol","inputs":[],"outputs":["string"]}},{"udt_struct_v0":{"doc":"","lib":"","name":"AllowanceDataKey","fields":[{"doc":"","name":"from","type_":"address"},{"doc":"","name":"spender","type_":"address"}]}},{"udt_struct_v0":{"doc":"","lib":"","name":"AllowanceValue","fields":[{"doc":"","name":"amount","type_":"i128"},{"doc":"","name":"expiration_ledger","type_":"u32"}]}},{"udt_union_v0":{"doc":"","lib":"","name":"DataKey","cases":[{"tuple_v0":{"doc":"","name":"Allowance","type_":[{"udt":{"name":"AllowanceDataKey"}}]}},{"tuple_v0":{"doc":"","name":"Balance","type_":["address"]}},{"tuple_v0":{"doc":"","name":"State","type_":["address"]}},{"void_v0":{"doc":"","name":"Admin"}}]}},{"udt_struct_v0":{"doc":"","lib":"","name":"TokenMetadata","fields":[{"doc":"","name":"decimal","type_":"u32"},{"doc":"","name":"name","type_":"string"},{"doc":"","name":"symbol","type_":"string"}]}}]


```
_____
```
 > stellar-cli contract info interface --network testnet --wasm ~/wasm/soroban_token_contract.wasm
```
```
#[soroban_sdk::contractclient(name = "Client")]
pub trait Contract {
    fn initialize(
        env: soroban_sdk::Env,
        admin: soroban_sdk::Address,
        decimal: u32,
        name: soroban_sdk::String,
        symbol: soroban_sdk::String,
    );
    fn mint(env: soroban_sdk::Env, to: soroban_sdk::Address, amount: i128);
    fn set_admin(env: soroban_sdk::Env, new_admin: soroban_sdk::Address);
    fn allowance(
        env: soroban_sdk::Env,
        from: soroban_sdk::Address,
        spender: soroban_sdk::Address,
    ) -> i128;
    fn approve(
        env: soroban_sdk::Env,
        from: soroban_sdk::Address,
        spender: soroban_sdk::Address,
        amount: i128,
        expiration_ledger: u32,
    );
    fn balance(env: soroban_sdk::Env, id: soroban_sdk::Address) -> i128;
    fn transfer(
        env: soroban_sdk::Env,
        from: soroban_sdk::Address,
        to: soroban_sdk::Address,
        amount: i128,
    );
    fn transfer_from(
        env: soroban_sdk::Env,
        spender: soroban_sdk::Address,
        from: soroban_sdk::Address,
        to: soroban_sdk::Address,
        amount: i128,
    );
    fn burn(env: soroban_sdk::Env, from: soroban_sdk::Address, amount: i128);
    fn burn_from(
        env: soroban_sdk::Env,
        spender: soroban_sdk::Address,
        from: soroban_sdk::Address,
        amount: i128,
    );
    fn decimals(env: soroban_sdk::Env) -> u32;
    fn name(env: soroban_sdk::Env) -> soroban_sdk::String;
    fn symbol(env: soroban_sdk::Env) -> soroban_sdk::String;
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct AllowanceDataKey {
    pub from: soroban_sdk::Address,
    pub spender: soroban_sdk::Address,
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct AllowanceValue {
    pub amount: i128,
    pub expiration_ledger: u32,
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct TokenMetadata {
    pub decimal: u32,
    pub name: soroban_sdk::String,
    pub symbol: soroban_sdk::String,
}
#[soroban_sdk::contracttype(export = false)]
#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub enum DataKey {
    Allowance(AllowanceDataKey),
    Balance(soroban_sdk::Address),
    State(soroban_sdk::Address),
    Admin,
}



```

### Why

See https://github.com/stellar/stellar-cli/issues/1506

### Known limitations

Network configuration is mandatory, even when fetching from local. This is due to group can't be marked as optional with clap's derive API. We can fix that in the future, potentially using more flexible builder API instead.


cmd/crates/soroban-spec-tools/src/contract.rs
```diff
@@ -276,3 +285,214 @@ fn format_name(lib: &StringM<80>, name: &StringM<60>) -> String {
         name.to_utf8_string_lossy()
     }
 }
+
+pub fn pretty_spec(spec: &Vec<ScSpecEntry>) -> String {
```
leighmcculloch, line 289: I think we should use the implementation that is available at the following link to generate the interface, because it generates verified valid rust code that gets used in contracts:
https://docs.rs/soroban-spec-rust

I apologise for not pointing out the existence of that crate earlier as it would have saved you some time.

The existing `generate` functions located in that crate will generate code that looks like this:
```
$ stellar contract bindings rust --wasm test_empty.wasm
pub const WASM: &[u8] = soroban_sdk::contractfile!(
    file = "../rs-soroban-sdk/target/wasm32-unknown-unknown/release/test_empty.wasm",
    sha256 = "ec784e29ec4b9dcceb014302805d6a85485b71b06438f6ce69e6baafc18997ff"
);
#[soroban_sdk::contractclient(name = "Client")]
pub trait Contract {
    fn empty(env: soroban_sdk::Env);
}
```

That output is currently available on this command:
```
stellar contract bindings rust
```

So I guess, there's some overlap between the `contract info interface` command and this command. However, the purposes of the two commands are a little different.

The `bindings generate`'s goal is to generate client code to use for calling a contract, which is why the generated code includes the `contractclient` attribute, and the `WASM` bytes.

The `contract info interface`'s goal is to really just report the interface itself.

So I think what we should do is in this PR you could move forward with generating the same output as `bindings generate`, and then as a next step we can make a change to `soroban-spec-rust` to expose a function for generating only the interface. It'll just require a small refactor of the crate so that we can get at the interface without the other noise.

cmd/soroban-cli/src/commands/contract/info.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// that the contract was built for. Env-meta allows the Soroban Env
+    /// to know whether the contract is compatibility with the network in
+    /// its current configuration.
+    ///
+    /// The data outputted by this command is a stream of `SCEnvMetaEntry` XDR values.
+    /// See the type definitions in [stellar-xdr](https://github.com/stellar/stellar-xdr).
+    /// [See also XDR data format](https://developers.stellar.org/docs/learn/encyclopedia/data-format/xdr).
+    ///
+    /// Outputs no data when no data is present in the contract.
+    EnvMeta(env_meta::Cmd),
```
leighmcculloch, line 23: So that these commands are ordered by importantance to the average dev, I'd put Interface first, then Meta, then EnvMeta.

cmd/soroban-cli/src/commands/contract/info/meta.rs
```diff
@@ -0,0 +1,76 @@
...
+                            let key = key.to_string();
+                            let (key, val) = match key.as_str() {
+                                "rsver" => ("Rust version", val.to_string()),
+                                "rssdkver" => (
+                                    "Soroban SDK version",
+                                    format!("{})", val.to_string().replace('#', " (commit hash: ")),
+                                ),
+                                _ => (key.as_str(), val.to_string()),
+                            };
+                            meta_str.push_str(&format!(" • {key}: {val}\n"));
```
leighmcculloch, line 65: I like the UX the mapping creates, but I think we should remove the mapping and output the raw keys and values, mostly because devs will be able to figure it out, and it gives them the same 1-1 experience of the meta that they'll have if they write any tools to process the values.

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -165,3 +174,16 @@ pub enum SpecOutput {
...
+    /// Pretty-printed output of the info entry
+    #[default]
+    Pretty,
+    /// XDR output of the info entry
+    XdrBase64,
+    /// JSON output of the info entry (one line, not formatted)
+    Json,
+    /// Formatted (multiline) JSON output of the info entry
+    JsonFormatted,
+}
```
leighmcculloch, line 189: Can we keep this type in the info shared module? It seems used primarily there. If it ever gets used elsewhere we can move it to a suitable place to be shared with more things then.

cmd/soroban-cli/src/commands/contract/info/interface.rs
```diff
@@ -0,0 +1,55 @@
...
+                return Err(NoInterfacePresent());
+            }
+
+            (spec.spec_base64.unwrap(), spec.spec)
+        };
+
+        let res = match self.common.output {
+            InfoOutput::XdrBase64 => base64,
+            InfoOutput::Json => serde_json::to_string(&spec)?,
+            InfoOutput::JsonFormatted => serde_json::to_string_pretty(&spec)?,
```
leighmcculloch, line 49: I like the idea of using the raw XDR<>JSON for the interface format.

There is actually another JSON format we invented at some point, that is visible with this command:
```
❯ stellar contract bindings json --wasm test_empty.wasm
[
  {
    "type": "function",
    "doc": "",
    "name": "empty",
    "inputs": [],
    "outputs": []
  }
]
```

The json command really doesn't belong in the bindings command, because bindings is for generating actual clients, and the JSON is clearly not a client. So I think we can deprecate the `bindings json` command.

The question becomes should we keep this JSON interface format around, or abandon it for the XDR/JSON.

Realistically the two formats don't look _that_ different, and so the value in maintaining the other format that I'm not sure many people know exists has that much value.

```
$ wasm-cs test_empty.wasm read contractspecv0 -f binary | stellar xdr decode --input stream --type ScSpecEntry --output json-formatted
{
  "function_v0": {
    "doc": "",
    "name": "empty",
    "inputs": [],
    "outputs": []
  }
}
```

cmd/soroban-cli/src/commands/contract/info/shared.rs
```diff
@@ -0,0 +1,82 @@
...
+#[group(skip)]
+pub struct Args {
+    /// Wasm file to extract the data from
+    #[arg(long, group = "src")]
+    pub wasm: Option<PathBuf>,
+    /// Wasm hash to get the data for
+    #[arg(long = "wasm-hash", group = "src")]
+    pub wasm_hash: Option<String>,
+    /// Contract id to get the data for
+    #[arg(long = "id", env = "STELLAR_CONTRACT_ID", group = "src")]
```
leighmcculloch, line 29: I'm inclined to leave off the env for this command. Envs are useful for setting for use across many commands, where-as this command is quite specific.
```suggestion
    #[arg(long = "id", group = "src")]
```

cmd/soroban-cli/src/commands/contract/info/shared.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::commands::contract::info::shared::Error::InvalidWasmHash;
+use crate::commands::contract::InfoOutput;
+use crate::config::{locator, network};
+use crate::utils::rpc::get_remote_wasm_from_hash;
+use crate::wasm;
+use crate::wasm::Error::ContractIsStellarAsset;
+
+#[derive(Debug, clap::Args, Clone, Default)]
+#[command(group(
+    clap::ArgGroup::new("src")
```
leighmcculloch, line 16: I suggest naming the group a full word captialized because I'm pretty sure they get rendered alongside other group names that are by default cased, e.g. `Options`.
```suggestion
    clap::ArgGroup::new("Source")
```

Unclear to me if the source grouping is of significant value. The source grouping is not something the cli uses on other similar commands. But we could update those other places to use it. If we're going to do it we should make sure it works well consistently in all similar commands, so I suggest leaving off the group in this PR, and following up after this PR with a holistic update of all similar commands so we can validate that a holistic story for it is good.

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -165,3 +174,16 @@ pub enum SpecOutput {
     /// Pretty print of contract spec entries
     Docs,
 }
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, clap::ValueEnum, Default)]
+pub enum InfoOutput {
+    /// Pretty-printed output of the info entry
+    #[default]
+    Pretty,
```
leighmcculloch, line 182: `pretty` is ambiguous. For the `interface` subcommand it is Rust code. For the other subcommands it is human readable unstructured text. I think we should use a different enum for each of the three commands and use more specific names for the formats that align 1-1 with each.

So for the `interface` subcommand accept the value, and use as default, `rust`.

For the other subcommands I'd probably use the term `text`.

leighmcculloch (commented):
Nice work. I left some comments inline.

cmd/crates/soroban-spec-tools/src/contract.rs
leighmcculloch: I've opened a PR on the `soroban-spec-rust` crate adding a function to get at just the interface:
- https://github.com/stellar/rs-soroban-sdk/pull/1311

Ifropc: Oh got it sorry didn't do enough DD before spending some time on the prettyfying it. I'll use the new function from soroban-spec-rust crate for pretty printing.

cmd/soroban-cli/src/commands/contract/info/meta.rs
Ifropc: Don't mind just outputting the raw keys, but I'd say we should aim towards people not using pretty output in any tools. We have json format for that (that's much easier to use in tools), that will print the same output in between minor versions. For pretty printing, I think we can (and should be able to) break it without a notice in a minor version

leighmcculloch: I'm less concerned with breaking the output. I agree the pretty / text output shouldn't provide many guarantees on content or structure.

I'm more concerned with creating a consistent narrative. For example, someone views a contract with a mix of custom keys plus the common rsver and rssdk. They'd see output like this:
```
$ stellar-cli contract info meta --wasm contract.wasm 
Contract meta:
 • Rust version: 1.79.0
 • Soroban SDK version: 21.4.0 (commit hash: d6f5639f643d76e758beecbb0ca391f8cd304c24)
 • customfield: myvalue
```

A developer may be mistaken to think they can lookup the outputted values using the keys `customfield`, `Rust version`, `Soroban SDK Version`. When they actually need to lookup using keys `customfield`, `rsver`, `rssdk`.

We could achieve key consistency and educate the user with less pretty but just as functional output:

```
$ stellar-cli contract info meta --wasm contract.wasm 
Contract meta:
 • rsver: 1.79.0 (Rust Version)
 • rssdk: 21.4.0#d6f5639f643d76e758beecbb0ca391f8cd304c24 (Rust SDK Version)
 • customfield: myvalue
```

cmd/soroban-cli/src/commands/contract/info/shared.rs
leighmcculloch: Oh right, we already use this env consistently across the cli. Please ignore the comment above.

cmd/soroban-cli/src/commands/contract/info/meta.rs
Ifropc: Oh yeah I like that idea!

cmd/crates/soroban-spec-tools/src/contract.rs
Ifropc: Left one comment in the PR, replaced the usage of this method with `generate_without_file`

cmd/soroban-cli/src/commands/contract/info/interface.rs
Ifropc: Not sure which wasm file was it, but shouldn't you be able to do the same thing as in 2nd command with `info interface --output json-formatted` ?
Also agree on keeping 1 unified standard, if anybody is using outputs JSONs and parsing it, IMO we 100% want to keep it same across the commands (unless we really need to make it different) 

https://github.com/stellar/stellar-cli/issues/1535

cmd/soroban-cli/src/commands/contract/mod.rs
Ifropc: I was not sure if it needs to be in this particular place or not and followed how another selection enums are structured. We should probably move them out as well in a separate PR

cmd/soroban-cli/src/commands/contract/info/shared.rs
Ifropc: Oh I was using the group to make only one selection option to be valid. I guess we can just use conflicts-with and required instead? Personally, groups are more readable to me than mix of `conflicts-with` + `required_unless_present`, but I can stick to just that parameters combination instead ^

leighmcculloch: Ah, I didn't realise this was another way to do that. I think that's fine, would be good to maintain some consistency so maybe an issue to follow up changing other commands to do same.

Ifropc: Unfortunately, not very flexible when using derive API 😔 

> Network configuration is mandatory, even when fetching from local. This is due to group can't be marked as optional with clap's derive API. We can fix that in the future, potentially using more flexible builder API instead.

fnando:
Can we make the json output be prettified?

Ifropc:
@fnando pretty json is `--output json-formatted`. Regular json is if you want to use something like jq or process it programmatically (and we don't care about formatting)

cmd/crates/soroban-spec-tools/src/utils.rs
```diff
@@ -1,268 +1,4 @@
...
-    fmt::Display,
-    io::{self, Cursor},
-};
-
-use stellar_xdr::curr::{
-    Limited, Limits, ReadXdr, ScEnvMetaEntry, ScMetaEntry, ScMetaV0, ScSpecEntry, ScSpecFunctionV0,
-    ScSpecUdtEnumV0, ScSpecUdtErrorEnumV0, ScSpecUdtStructV0, ScSpecUdtUnionV0, StringM,
-};
-
-pub struct ContractSpec {
```
leighmcculloch, line 13: Is the code in this file being deleted because it was already unused? We'll need to keep that deletion for a major version. Can you open an issue for it and label the issue with `breaking` so we know to come back and do it before then?

cmd/soroban-cli/src/commands/contract/info/env_meta.rs
```diff
@@ -0,0 +1,72 @@
...
+            return Err(NoSACEnvMeta());
+        }
+        let spec = Spec::new(&bytes.unwrap())?;
+
+        if spec.env_meta_base64.is_none() {
+            return Err(NoEnvMetaPresent());
+        }
+
+        let res = match self.output {
+            MetasInfoOutput::XdrBase64 => spec.env_meta_base64.unwrap(),
```
leighmcculloch, line 50: Where possible it is best in Rust to avoid the use of .unwrap and instead safely unwrap using pattern matching.

Using is_none/is_some and unwrap means there will be two separate lines of code that must be logically synchronized, and if they ever get out of sync the unwrap will panic.

For example:
```rust
let Some(bytes) = bytes else {
    return Err(NoSACEnvMeta());
};
```

cmd/soroban-cli/src/wasm.rs
```diff
@@ -28,6 +35,18 @@ pub enum Error {
...
+    Locator(#[from] locator::Error),
+    #[error(transparent)]
+    Rpc(#[from] soroban_rpc::Error),
+    #[error("unexpected contract data {0:?}")]
+    UnexpectedContractToken(ContractDataEntry),
+    #[error(
+        "cannot fetch wasm for contract because the contract is \
+    a network built-in asset contract that does not have a downloadable code binary"
+    )]
+    ContractIsStellarAsset(),
```
leighmcculloch, line 49: Enum values with no arguments don't need to be a tuple enum type.
```suggestion
    ContractIsStellarAsset,
```

cmd/crates/soroban-spec-tools/src/utils.rs
Ifropc: Yeah it was unused, it's copy-paste of [contract-rs ](https://github.com/stellar/stellar-cli/blob/56a81bedbbc33300fa5e2d933808a5bbd7b7b875/cmd/crates/soroban-spec-tools/src/contract.rs)

leighmcculloch (commented):
There's some code being deleted that is a breaking change, and one other ask, otherwise looks good.

cmd/crates/soroban-spec-tools/src/utils.rs
leighmcculloch: Ah, it isn't even exported from the crate? Then it is fine to delete.

Ifropc: > pub mod utils;
> pub struct ContractSpec

I think it's exported 🤔

cmd/soroban-cli/src/commands/contract/info/env_meta.rs
Ifropc: Good point. I was trying to reduce amount of nested layers

leighmcculloch: That resonates with me. It's sort of something I found I had to let go of coming into Rust. I was used to writing Go / Java / Ruby that favors early exit patterns rather than nested if..else, but in Rust the latter just works better with unwrapping and pattern matching most of the time. 🤷🏻

cmd/crates/soroban-spec-tools/src/utils.rs
Ifropc: https://github.com/stellar/stellar-cli/issues/1544

leighmcculloch (approved):

PR #1534, opened 2024-08-08
Add dev shell environment

Ifropc:
### What

This PR adds a [devshell environment](https://nix.dev/tutorials/first-steps/declarative-shell.html) using [nix](https://nixos.org/guides/how-nix-works/) 
### Why

Currently, in order to develop stellar-cli locally you need to install some dependencies, and sometimes you also need to modify your `.<shell>rc` file to make your environment easy to use for development.
However, this makes switching between different branches/versions or switching between development/production builds more complicated.
This change adds a configuration for a development shell, which automatically does the following:
1. Installs all necessary dependencies (only accessible in the shell) -> rust, make, libudev, etc. without affecting your global environment. Dependencies are pinned into the lockfile. We can always update rust version to latest stable or change it to nightly with re-generating lockfile/changing nix config. Dependencies are installed once and cached. Can clean them out of your system with `nix collect-garbage`.
2. Setup `stellar` alias, configure development key and environment (set `STELLAR_NETWORK` and `STELLAR_ACCOUNT` variables) 
3. Add up-to-date autocomplete in the shell

Developer simply needs to install nix and run `nix develop`

This should work both on any Linux distro, Mac (Darwin builds are available for all packages in nixpkgs) and Windows WSL 

### Known limitations

[nix](https://nixos.org/download/) is required to be installed in the system, with v2.20+
Currently `bash` is used as a default shell. It's possible to use other shells with `nix develop -c zsh`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ There are 2 ways to being developing stellar-cli:
 
 ### Installing all required dependencies 
 
-You may want to install all required dependencies locally. This includes installing `rustup`, `make`, `libudev`, `jq` from your package manager. After all dependencies are installed, you can start with running `make build` to build `stellar-cli` on your machine! 
+You may want to install all required dependencies locally. This includes installing `rustup`, `make`, `libudev`, `jq` from your package manager. After all dependencies are installed, you can start with running `make install` to build `stellar-cli` and install it! 
 
 ### Using `nix`
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,31 @@ changes quickly.
   * Clearly describe the issue including steps to reproduce if it is a bug.
 * Fork the repository on GitHub.
 
+## Setting up development environment
+
+There are 2 ways to being developing stellar-cli:
+
+### Installing all required dependencies 
+
+You may want to install all required dependencies locally. This includes installing `rustup`, `make`, `libudev`, `jq` from your package manager. After all dependencies are installed, you can start with running `make build` to build `stellar-cli` on your machine! 
+
+### Using `nix`
+
+If you don't want to install necessary dependencies from above, you can run development shell using [nix](https://nixos.org/guides/how-nix-works/) (make sure to [install](https://nixos.org/download/) version above 2.20). After installing `nix`, simply run `nix develop` that will start new `bash` session in your current terminal. If you want to use different shell (e.g. `zsh`) you can run `nix develop -c zsh`
+
+This session will have:
+1. All required dependencies installed
+2. `stellar` alias (overwriting your existing `stellar` installed via cargo, if any)
+3. Configured auto-complete for the working git branch
+
+You can add extra configuration in your `local.sh` file (for example, if you want to export some extra variables for your devshell you can put following in your `local.sh`:
+```shell
+#!/usr/bin/env bash
+export STELLAR_NETWORK=testnet
+```
+Note that all of dependencies and configurations mentioned above is available only in your local development shell, not outside of it.
+
+
 ### Minor Changes
 
 #### Documentation
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,45 @@
+{
+  description = "stellar-cli development shell";
+
+  inputs = {
+    nixpkgs.url      = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    flake-utils.url  = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, rust-overlay, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        overlays = [ (import rust-overlay) ];
+        pkgs = import nixpkgs {
+          inherit system overlays;
+        };
+      in
+      with pkgs;
+      {
+        devShells.default = mkShell {
+          buildInputs = [
+            openssl
+            pkg-config
+            libudev-zero
+            jq
+            (rust-bin.stable.latest.default.override {
+              extensions = [ "rust-src" ];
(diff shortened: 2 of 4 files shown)


Ifropc:
Examples:
```bash
# No rust toolchain is installed in my system, and I have `stellar` installed in ~/.cargo/bin/stellar
ifro@nixos:~/projects/stellar-cli/ > which cargo
cargo not found
ifro@nixos:~/projects/stellar-cli/ > stellar --version
stellar 21.2.0 (v20.0.0-248-g96e819c949089a6744621164bb406fc411de676e-dirty)
soroban-env 21.2.0 (8809852dcf8489f99407a5ceac12625ee3d14693)
soroban-env interface version 90194313216
stellar-xdr 21.2.0 (9bea881f2057e412fdbb98875841626bf77b4b88)
xdr curr (70180d5e8d9caee9e8645ed8a38c36a8cf403cd9)
ifro@nixos:~/projects/stellar-cli/ > nix develop
Using nix (Nix) 2.20.8
     Running `target/debug/stellar keys add dev_key`
     Running `target/debug/stellar completion --shell bash`

# Have rust toolchain in devshell, and `stellar` points to the freshly built binary 
[ifro@nixos:~/projects/stellar-cli]$ which cargo
/nix/store/i223gfb0a86hxywdsqxmyy291z511056-rust-default-1.80.0/bin/cargo

[ifro@nixos:~/projects/stellar-cli]$ stellar --version
     Running `target/debug/stellar --version`
stellar 21.3.0 (v20.0.0-257-g0e615b6c61d03ab294fd6ccab42b1b08f317a437)
soroban-env 21.2.0 (8809852dcf8489f99407a5ceac12625ee3d14693)
soroban-env interface version 90194313216
stellar-xdr 21.2.0 (9bea881f2057e412fdbb98875841626bf77b4b88)
xdr curr (70180d5e8d9caee9e8645ed8a38c36a8cf403cd9)
```

Ifropc:
Also no documentation on it at the moment, open to suggestions on where to add it (README.md or CONTRIBUTING.md)

leighmcculloch:
Documentation should be in CONTRIBUTING.md I think.

flake.nix
```diff
@@ -0,0 +1,45 @@
...
+              targets = [ "wasm32-unknown-unknown" ];
+            })
+          ];
+          shellHook =
+          ''
+            echo "Using `nix --version`"
+            alias stellar="cargo run --bin stellar --"
+            SOROBAN_SECRET_KEY=SBPXU3OP65M2DQPRM4YFZ5NME66NWWH5DHN44IBDYAN4TFKKY7CS4OHY stellar keys add dev_key
+            export STELLAR_NETWORK=testnet
+            export STELLAR_ACCOUNT=dev_key
```
leighmcculloch, line 37: I don't think we can make assumptions about the config, key setup, or default network.
```suggestion
```

leighmcculloch (approved):
I don't know if many folks deving on the cli use nixos, but sounds like you do @Ifropc. Can you remove the secret key and setup assumptions so that the flake is only the tool setup and doesn't make other assumptions? Otherwise lgtm.

Ifropc:
I added secret key just cause I thought it'd be a good idea to easily test things locally (so shared dev secret key is fine) but I now I think it's better if i remove exporting the variable when key is not defined instead

leighmcculloch:
> I added secret key just cause I thought it'd be a good idea to easily test things locally (so shared dev secret key is fine) but I now I think it's better if i remove exporting the variable when key is not defined instead

It's very easy for devs to create a key, it's a single command, `stellar keys generate me`.

flake.nix
```diff
@@ -0,0 +1,43 @@
...
+            (rust-bin.stable.latest.default.override {
+              extensions = [ "rust-src" ];
+              targets = [ "wasm32-unknown-unknown" ];
+            })
+          ];
+          shellHook =
+          ''
+            echo "Using `nix --version`"
+            alias stellar="cargo run --bin stellar --"
+            [ -f ./local.sh ] && source ./local.sh
```
leighmcculloch, line 35: Executing local.sh seems surprising, and user specific. Can we leave that out? It seems simple enough for folks to run any setup they need.

Fwiw folks can simply put STELLAR_ env vars into their shell file if they want default values to be used.

CONTRIBUTING.md
```diff
@@ -14,6 +14,31 @@ changes quickly.
   * Clearly describe the issue including steps to reproduce if it is a bug.
 * Fork the repository on GitHub.
 
+## Setting up development environment
+
+There are 2 ways to being developing stellar-cli:
+
+### Installing all required dependencies 
+
+You may want to install all required dependencies locally. This includes installing `rustup`, `make`, `libudev`, `jq` from your package manager. After all dependencies are installed, you can start with running `make build` to build `stellar-cli` on your machine!
```
leighmcculloch, line 23:
```suggestion
You may want to install all required dependencies locally. This includes installing `rustup`, `make`, `libudev`, `jq` from your package manager. After all dependencies are installed, you can start with running `make install` to build `stellar-cli` and install it! 
```

leighmcculloch (approved):

flake.nix
Ifropc: >Fwiw folks can simply put STELLAR_ env vars into their shell file if they want default values to be used.

You mean shell config file (`bashrc`/`zshrc`)?

I originally added this `local.sh` with the idea of developers being able to put something extra in there easily (like importing a private key or exporting extra variables), which is probably not something super useful.

As for env vars (if we are talking about shell config files), the idea is that we want to keep default shell "clean" from anything related to cli development. 
For example, I may want to use my regular cli with a different account or network exported in the env var defined in my .rc file.
In reality, I'm sure most people already just store their env vars there and re-define in the shell when necessary

So I can just remove it before the merge 👍🏼

Ifropc:
Hey @TomMD answered in #1661 thanks for raising this up

PR #1536, opened 2024-08-09
Fix invoke send default behavior

leighmcculloch:
### What
Change the recently merged (and not yet released) invoke `--send` default behavior so that by default send occurs if there are any contract events, send occurs if there is any soroban auth involved, and change the name of the default enum value to `default`.

Also add output that tells the user simulation-only has occurred automatically.

### Why
There's a bug in the code that checks if there are any events, a not that shouldn't be present, so the default behavior is sending when there are no events, rather when there are events.

Also, if auth is required it's more ambiguous to if a user would expect or not a transaction to be sent. Since read-only operations are rarely gated by auth, we can probably assume auth means the transaction should be sent, even if we can't otherwise detect a write to storage or a published event. It could be that the auth itself will do one of these two things and sending to the network could be desirable because of that.

The default behavior of sending should for the most part be as least surprising as possible. It's meant to be an optimisation for when folks are truly calling those read-only operations where sending makes little sense, but if there's any doubt it should just send.

I renamed the enum value because this isn't released yet and because `if-write` doesn't feel like it fully captures the default behavior given the addition of the auth logic. Given the default behavior is somewhat complex I named it `default` influenced by some of the git cli's options where it uses that term for default behavior that isn't easy to capture in another word.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/invoke.rs
+++ b/cmd/soroban-cli/src/commands/contract/invoke.rs
@@ -388,7 +388,7 @@ impl NetworkRunnable for Cmd {
             (res.return_value()?, events)
         } else {
             let print = print::Print::new(global_args.map_or(false, |g| g.quiet));
-            print.infoln("Invoke simulated only because simulation identified a read-only transaction. To send invoke to network in a transaction use --send=yes.");
+            print.infoln("Send skipped because simulation identified a read-only invoke. Send invoke to network with `--send=yes`.");
             (sim_res.results()?[0].xdr.clone(), sim_res.events()?)
         };
         crate::log::events(&events);
--- a/cmd/soroban-cli/src/commands/contract/invoke.rs
+++ b/cmd/soroban-cli/src/commands/contract/invoke.rs
@@ -58,7 +58,7 @@ pub struct Cmd {
     #[command(flatten)]
     pub fee: crate::fee::Args,
     /// Whether or not to send a transaction
-    #[arg(long, value_enum, env = "STELLAR_SEND")]
+    #[arg(long, value_enum, default_value_t, env = "STELLAR_SEND")]
     pub send: Send,
 }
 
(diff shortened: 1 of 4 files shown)


cmd/soroban-cli/src/print.rs
```diff
@@ -18,21 +18,21 @@ impl Print {
 
     pub fn print<T: Display + Sized>(&self, message: T) {
         if !self.quiet {
-            print!("{message}");
+            eprint!("{message}");
```
leighmcculloch, line 21: @fnando I slipped this fix into this PR as well, because it was causing a test to fail where I was adding new usage of the print and that print value was showing up in stdout instead of stderr. I think all this human readable text should be stderr.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -590,11 +596,19 @@ fn has_write(sim_res: &SimulateTransactionResponse) -> Result<bool, Error> {
         .read_write
         .is_empty())
 }
+
 fn has_published_event(sim_res: &SimulateTransactionResponse) -> Result<bool, Error> {
-    Ok(!sim_res.events()?.iter().any(
+    Ok(sim_res.events()?.iter().any(
```
leighmcculloch, line 601: @willemneal This is the fix for events influencing send.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -293,6 +291,23 @@ impl Cmd {
             })
             .transpose()
     }
+
+    fn send(&self, sim_res: &SimulateTransactionResponse) -> Result<ShouldSend, Error> {
+        Ok(match self.send {
+            Send::Default => {
+                if self.is_view {
+                    ShouldSend::No
```
leighmcculloch, line 299: `--is-view` is basically ignored if a non-default send mode is set, since `--send` is the new option it should take precedence. This code got reorganized so that the combination of the two options can set the `ShouldSend` enum value which was easier to do when the mapping isn't happening in the `Send` enum's fn.

leighmcculloch:
E2e tests should be fixed by:
- https://github.com/stellar/system-test/pull/102

cmd/soroban-cli/src/print.rs
fnando: nice catch!

fnando (approved):

janewang:
Thank you for fixing @leighmcculloch

PR #1548, opened 2024-08-16
Feat/use rich printer for init

### LINKED ISSUE #1553: Use rich CLI printer in `contract init`

elizabethengelman:
### What

Refactor `contract init` command to use`soroban-cli/src/print.rs`.

Closes #1553 

### Why

To be more consistent without our output, and to reduce duplication.

### Known limitations
n/a


cmd/soroban-cli/src/print.rs
```diff
@@ -86,6 +86,8 @@ create_print_functions!(error, errorln, "❌");
 create_print_functions!(globe, globeln, "🌎");
 create_print_functions!(info, infoln, "ℹ️");
 create_print_functions!(link, linkln, "🔗");
+create_print_functions!(over, overln, "🔄");
```
elizabethengelman, line 89: I'm using this specifically when we're overwriting a file in init. I'd love some other ideas for what to call this!

janewang: `over` or `overwrite` sounds fine

leighmcculloch: +1 for `overwrite`.

leighmcculloch: The icon doesn't speak overwrite to me though. The icon feels like "swap".

What if for overwrites we wrote two lines, in much the same way git might communicate this. E.g.:
```
➕ Writing file1...
➕ Writing file2...
➖ Removing existing file3...
➕ Writing file3...
```

It might also be reasonable to use the same icon
```
➕ Writing file1...
➕ Writing file2...
➕ Writing file3 (overwriting existing file)...
```

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -87,380 +89,412 @@ pub enum Error {
...
-        clone_repo(SOROBAN_EXAMPLES_URL, examples_dir.path())?;
+        if !Self::check_internet_connection() {
+            self.print.warnln("It doesn't look like you're connected to the internet. We're still able to initialize a new project, but additional examples and the frontend template will not be included.");
+            return Ok(());
+        }
 
-        // copy the example contracts into the project
-        copy_example_contracts(examples_dir.path(), project_path, with_examples, overwrite)?;
-    }
+        if !&self.args.frontend_template.is_empty() {
```
willemneal, line 129: is the `&` needed?

willemneal (approved):
I have some nits, but in general I think that we should have a the root command handle the `print.errorln` of the errors. The map_err's with `print.errorln` here are providing context to the error and should be an error case so that the caller would have that context.

cmd/soroban-cli/src/print.rs
chadoh: I like @leighmcculloch's one-liner suggestion

```
➕ Writing file3 (overwriting existing file)...
```

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -87,380 +89,412 @@ pub enum Error {
...
-
-        // clone the template repo into the temp dir
-        clone_repo(frontend_template, fe_template_dir.path())?;
-
-        // copy the frontend template files into the project
-        copy_frontend_files(fe_template_dir.path(), project_path, overwrite)?;
-    }
+impl Runner {
+    fn run(&self) -> Result<(), Error> {
+        let project_path = Path::new(&self.args.project_path);
```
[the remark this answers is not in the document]
elizabethengelman: just made the change! 

curious how much of an impact this would have in this particular instance since we are creating the directories recursively for the path the user passes in.

cmd/soroban-cli/src/print.rs
elizabethengelman: I like that suggestion!

cmd/soroban-cli/src/commands/contract/init.rs
elizabethengelman: nice catch!

elizabethengelman:
> I have some nits, but in general I think that we should have a the root command handle the print.errorln of the errors. The map_err's with print.errorln here are providing context to the error and should be an error case so that the caller would have that context.

@willemneal that's a good point - instead of printing the error messages and then swallowing the errors, we should return the error for those cases. I can work on that change in a separate PR.

PR #1551, opened 2024-08-19
feat: add `tx::builder` module and initial classic commands

willemneal:
### What

Add a transaction builder and operation builder trait. Also adds the first two stellar classic commands under the subcommand `tx new`:

- [x] Create account 
- [x] Payment
- [x] Set options
- [x] Bump sequence
- [x] Account Merge
- [x] Manage data
- [x] InvokeFunction 
- [x] Extend footprint ttl 
- [x] Restore footprint 
- [x] Allow trust
- [x] Change trust
- [x] Set trustline flags
- [x] Doc review

Todo:
- [x] Tests

### Why

Currently all transactions are built by hand and require a lot of knowledge around the internal types. Furthermore, there is a lot of redundant code around creating, submitting transaction.

These examples should make it easy to have a quick design discussion before adding the rest of the operations.

### Known limitations

[TODO or N/A]


cmd/soroban-cli/src/tx/builder/bytesm.rs
```diff
@@ -0,0 +1,28 @@
...
+    Xdr(#[from] xdr::Error),
+    #[error(transparent)]
+    Hex(#[from] hex::FromHexError),
+}
+
+impl FromStr for Bytes64 {
+    type Err = Error;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        Ok(Self(xdr::BytesM::<64>::try_from(&hex::decode(value)?)?))
```
leighmcculloch, line 20: The `xdr::BytesM` type should already support from_str of hex as we implement in stellar-xdr. What was the reason for needing to add that here?

cmd/soroban-cli/src/tx/builder/muxed_account.rs
```diff
@@ -0,0 +1,70 @@
+use soroban_sdk::xdr::MuxedAccountMed25519;
+
+use crate::xdr::{self, Uint256};
+
+pub struct MuxedAccount(pub xdr::MuxedAccount);
```
leighmcculloch, line 5: I think we can eliminate this type, and most of the conversions between the different things by using the xdr::MuxedAccount after this PR merges:
- https://github.com/stellar/rs-stellar-xdr/pull/396

cmd/soroban-cli/src/tx/builder/transaction.rs
```diff
@@ -0,0 +1,66 @@
+use soroban_env_host::xdr::{self, Memo, SequenceNumber, TransactionExt};
+
+use super::{Error, MuxedAccount, Operation};
+
+#[derive(Debug, Clone)]
+pub struct Transaction {
```
leighmcculloch, line 6: We made a conscious decision, lost to tribal knowledge, with the Rust stellar-xdr lib to attempt to go as far as possible with using XDR types, and filling gaps in their interface as we go, so that we didn't end up with a similar burden in the future that we have discovered on other SDKs where these layers end up diverging from the networks definition.

It's part of the reason the XDR types support conversion to and from things like strkeys (in addition to supporting the JSON encoding).

It looks like in this PR we're introducing a layer that looks much like other Stellar SDKs, similar to but different to the underlying XDR layer, that'll need to be maintained. Maybe that's no bigger than if the layer didn't exist, but once https://github.com/stellar/rs-stellar-xdr/pull/396 is adopted I think we could chisel this change back and remove the need for most conversions, and maybe the layer should get a lot thinner.

cmd/soroban-cli/src/commands/tx/args.rs
```diff
@@ -0,0 +1,92 @@
...
+    #[clap(flatten)]
+    pub config: config::Args,
+    //// The source account for the operation, Public key or Muxxed Account
+    /// e.g. `GA3D5...` or `MA3D5...`
+    #[arg(
+        long,
+        visible_alias = "with_source",
+        env = "STELLAR_WITH_SOURCE_ACCOUNT"
+    )]
+    pub with_source_account: Option<address::Address>,
```
leighmcculloch, line 25: Can we make this option named the same as the option is named in other commands? i.e. `--source-account` with alias `--source`?

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
```diff
@@ -0,0 +1,414 @@
...
+            "--set-authorize",
+            "--source",
+            "test1",
+        ])
+        .assert()
+        .success()
+        .stdout_as_str();
+    let after = client.get_account(&test).await.unwrap();
+    println!("{before:#?}\n{after_issue:#?}\n{after:#?}");
+    println!("{res}");
```
leighmcculloch, line 188: This test needs assertions.

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
```diff
@@ -0,0 +1,414 @@
...
+            "set-options",
+            "--signer",
+            test1.as_str(),
+            "--signer-weight",
+            "0",
+        ])
+        .assert()
+        .success();
+    let after = client.get_account(&test).await.unwrap();
+    assert_eq!(before.signers.len(), after.signers.len());
```
leighmcculloch, line 225: Requesting the appropriate ledger entry it should be possible to retrieve the added signer too and use that in an assertion.

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
```diff
@@ -0,0 +1,414 @@
...
+}
+
+#[tokio::test]
+async fn set_options() {
```
leighmcculloch, line 229: I think we need a much larger variety of set options tests, where we test the different variations of its use. For example, that fields aren't set when optional and no value.

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
```diff
@@ -0,0 +1,414 @@
...
+            "--set-required",
+            "--set-revocable",
+            "--set-clawback-enabled",
+        ])
+        .assert()
+        .failure();
+}
+
+#[tokio::test]
+async fn change_trust() {
```
leighmcculloch, line 298: I don't see this test doing anything with the change trust op. What am I missing?

leighmcculloch (commented):
I still need to do a full pass but a cursory look and this is very elegant. I appreciate that we're consolidating the building of some existing txs in the CLI along the way. Some questions and thoughts inline, specifically to avoid recreating another Stellar SDK-like layer, but that might be necessary at this point.

I've got this in my review queue to do a full review, just ran out of time to go deep today.

cmd/soroban-cli/src/commands/tx/args.rs
willemneal: This is the source account for the operation. Should it be renamed to `op_source_account`/`op_source`?

cmd/soroban-cli/src/tx/builder/transaction.rs
willemneal: I removed it in favor an extension trait.

cmd/soroban-cli/src/tx/builder/muxed_account.rs
willemneal: This still requires a FromStr implementation.

cmd/soroban-cli/src/tx/builder/bytesm.rs
willemneal: Good point! Removed!

willemneal:
@leighmcculloch Updated to reflect feedback. Still need to update tests.

cmd/soroban-cli/src/commands/tx/args.rs
leighmcculloch: Ah. Given these commands are a convenience for single operation transactions, I don't think they need to support setting the operation source. There's few cases where in a single op transaction that a user plans to submit right then that they need a different source account as on the transaction. The only reason would be to disconnect the sequence number, which is a valid thing to need to do, but I suspect rare in this use case.

We can add it later if we need to.

cmd/soroban-cli/src/tx/builder/muxed_account.rs
leighmcculloch: `xdr::MuxedAccount` already has a `FromStr` impl for strkeys:
https://github.com/stellar/rs-stellar-xdr/blob/67be5955a15f1d3a4df83fe86e6ae107f687141b/src/curr/str.rs#L90-L100

cmd/soroban-cli/src/tx/builder/transaction.rs
```diff
@@ -0,0 +1,54 @@
...
+        operation: xdr::Operation,
+    ) -> xdr::Transaction {
+        xdr::Transaction {
+            source_account: source_account.into().into(),
+            fee,
+            seq_num: seq_num.into(),
+            cond: soroban_env_host::xdr::Preconditions::None,
+            memo: Memo::None,
+            // Safe because we know we have only one operation
+            operations: unsafe { vec![operation].try_into().unwrap_unchecked() },
```
Ifropc, line 34: I don't think we need to optimize cli for performance that much, we can take a runtime check and simplify to
 ```suggestion
            operations: vec![operation].try_into().unwrap(),
```

File: cmd/soroban-cli/src/commands/tx/new/change_trust.rs [hunk 10]
```diff
@@ -0,0 +1,47 @@
...
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub tx: tx::args::Args,
+    #[arg(long)]
+    pub line: builder::Asset,
+    /// Limit for the trust line
+    #[arg(long)]
```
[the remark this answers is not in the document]
leighmcculloch: I think the default value should be u64::MAX for consistency with other developer touch points.

cmd/soroban-cli/src/commands/tx/new/manage_data.rs
```diff
@@ -0,0 +1,45 @@
...
+    #[command(flatten)]
+    pub tx: tx::args::Args,
+    /// Line to change, either 4 or 12 alphanumeric characters, or "native" if not specified
+    #[arg(long)]
+    pub data_name: xdr::StringM<64>,
+    /// Up to 64 bytes long hex string
+    /// If not present then the existing Name will be deleted.
+    /// If present then this value will be set in the `DataEntry`.
+    #[arg(long)]
+    pub data_value: Option<xdr::BytesM<64>>,
```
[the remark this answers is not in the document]
leighmcculloch: That does seem a little more elegant, but balancing the impact / benefit with maintaining, and setting user expectations, it may also be beneficial if users can rely on these tx building commands to reflect the networks structure/design for the operations, so that devs don't have to remember / reconcile the differences between the network and the CLI interface.

For e.g. if I read stellar docs I learn that a manage-data op with key and no value deletes the key. Then I use the CLI, but now I learn that there's two different manage-data ops, an add and a delete, but then I look at an SDK, and I don't find an add and delete op. Confusion.

Vs I read stellar docs I learn that a manage-data op with key and no value deletes the key. Then I use the CLI, and I do same, a manage-data op with key and no value when I need to delete. Then I look at an SDK, and I do same, a manage-data op with key and no value when I need to delete.

cmd/soroban-cli/src/tx/builder/transaction.rs
leighmcculloch: The `vec![...]` should be unnecessary, I believe you can pass an array:

 ```suggestion
            operations: [operation].try_into().unwrap(),
```

File: cmd/soroban-cli/src/commands/tx/new/change_trust.rs (hunk 10 above)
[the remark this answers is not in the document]
Comment (Ifropc) [reply cmd/soroban-cli/src/commands/tx/new/change_trust.rs:17]: (We also should mention that limit set to 0 deletes trustline)

File: cmd/soroban-cli/src/commands/tx/new/manage_data.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (Ifropc) [reply cmd/soroban-cli/src/commands/tx/new/manage_data.rs:21]: Hmm, agree, when coming from this angle it's better to align the tooling we have

File: cmd/crates/soroban-test/tests/it/integration/tx/operations.rs (hunk 8 above)
Comment (willemneal) [reply cmd/crates/soroban-test/tests/it/integration/tx/operations.rs:294]: It's used in the issue_asset function. But I can add an additional check for a third account.

File: cmd/crates/soroban-test/tests/it/integration/tx/operations.rs [hunk 12]
```diff
@@ -0,0 +1,410 @@
...
+            asset,
+            "--amount",
+            initial_balance.to_string().as_str(),
+            "--source=test1",
+        ])
+        .assert()
+        .success();
+    let issuer_after = client.get_account(issuer).await.unwrap();
+    println!("{issuer_after:#?}, {issuer_before:#?}");
+}
```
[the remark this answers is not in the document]
willemneal: This is just setup for other tests which make assertions.

willemneal:
@leighmcculloch, I've removed all of the builder types that I could. One that still needs to be added to the xdr crate is the FromStr for `Asset`, which I will open another PR for soon. Also I removed the unneeded Operation trait and just use `From` for Operation Body, all of which simplified the code a good deal.

cmd/soroban-cli/src/commands/tx/args.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[error(transparent)]
+    Xdr(#[from] xdr::Error),
+}
+
+impl Args {
+    pub async fn tx(&self, body: impl Into<xdr::OperationBody>) -> Result<xdr::Transaction, Error> {
+        let source_account = self.source_account()?;
+        let seq_num = self
+            .config
+            .next_sequence_number(&source_account.to_string())
```
leighmcculloch, line 42: We don't have to do this in this PR, but we should make sequence number an option, that if isn't set uses the nxt seq num lookup, otherwise uses the value passed in. A good follow up to add later.

cmd/soroban-cli/src/commands/tx/new/set_trustline_flags.rs
```diff
@@ -0,0 +1,71 @@
+use clap::{command, Parser};
+
+use soroban_sdk::xdr::{self};
+
+use crate::{commands::tx, tx::builder};
+
+#[allow(clippy::struct_excessive_bools)]
```
leighmcculloch, line 7: Fix the test:
```suggestion
#[allow(clippy::struct_excessive_bools, clippy::doc_markdown)]
```

cmd/soroban-cli/src/commands/contract/id/wasm.rs
```diff
@@ -19,35 +19,40 @@ pub struct Cmd {
...
     }
 }
 
-pub fn contract_preimage(key: &ed25519_dalek::VerifyingKey, salt: [u8; 32]) -> ContractIdPreimage {
-    let source_account = AccountId(PublicKey::PublicKeyTypeEd25519(key.to_bytes().into()));
+pub fn contract_preimage(
+    key: &stellar_strkey::ed25519::PublicKey,
+    salt: [u8; 32],
+) -> ContractIdPreimage {
+    let source_account = AccountId(PublicKey::PublicKeyTypeEd25519(key.0.into()));
```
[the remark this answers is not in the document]
leighmcculloch: Ah, actually you can't do this yet. This function exists in the xdr lib that was just released and isn't imported here.

cmd/soroban-cli/src/config/address.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::xdr;
+
+use super::{locator, secret};
+
+/// Address can be either a public key or eventually an alias of a address.
+#[derive(Clone, Debug)]
+pub enum Address {
+    MuxedAccount(xdr::MuxedAccount),
+    AliasOrSecret(String),
+}
```
leighmcculloch, line 12: Nice.

leighmcculloch (approved):
👏🏻

leighmcculloch:
@janewang @stellar/devx There's a large amount of functionality being introduced with this PR. We should plan to do some dog fooding of it for ergonomics in the next couple weeks, maybe get some others to dog food it too.

PR #1569, opened 2024-08-28
chore: move custom arg parsing to separate module for reuse

willemneal:
Technically this is a breaking change because of the Error type change, but not from CLI perspective


leighmcculloch (approved):

leighmcculloch:
> Technically this is a breaking change because of the Error type change, but not from CLI perspective

With the CLI I think we can keep all breaking changes limited to the CLI user interface. It is not a priority to serve the CLI as a library.

PR #1570, opened 2024-08-28
feat: add arg parsing to deploy command to support new constructor feature

### LINKED ISSUE #1561: CAP-58: Support deploy with constructor arguments

From the CLI's perspective this means that the deploy can have required parameters same to how an invoke can have required parameters. So the ability to use the contract interface to determine which parameters are required is needed on the deploy command.

Ideally the deploy command works the exact same way the invoke command does. Using the `-- -h` syntax should show the help for the constructor, and the parameters for the constructor should be passed after the `--`.

willemneal:
~Depends on #1569.~

If a contract has a method called `__contructor` then it parses the slop, e.i. anything after `--`. So after adding one to the hello world contract you would use it like:

```
cargo s contract deploy --build-only --wasm ./target/wasm32-unknown-unknown/test-wasms/test_hello_world.wasm --source default --network testnet -- --help
ℹ️ Using wasm hash 75eca9c92c997f6b1b1c7caba32311e81077f29c2f11eb3be6d453406e939a34
Example constructor
Usage Notes:
Each arg has a corresponding --<arg_name>-file-path which is a path to a file containing the corresponding JSON argument.
Note: The only types which aren't JSON are Bytes and BytesN, which are raw bytes

Usage: __constructor [OPTIONS]

Options:
      --world <String>
          Example:
            --world '"hello world"'

  -h, --help
```

For now this is a noop. And adding the new operation is still needed but hard to test now.

Close #1561

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -367,7 +367,11 @@ If no keys are specified the contract itself is extended.
 
 Deploy a wasm contract
 
-**Usage:** `stellar contract deploy [OPTIONS] --source-account <SOURCE_ACCOUNT> <--wasm <WASM>|--wasm-hash <WASM_HASH>>`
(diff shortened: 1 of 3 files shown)


willemneal:
Closed in favor of #1574 as I am moving away from using AhaLabs as a remote.

PR #1571, opened 2024-08-28
Provide a prompt when the new version is available.

### LINKED ISSUE #1162: soroban-cli: add update check

People install soroban-cli and then don't upgrade until a problem is encountered.

For example: https://discord.com/channels/897514728459468821/966788672164855829/1194759622339342588

The CLI checking if a new version is available and informing the user they could update to it.

overcat:
### What

Provide a prompt when the new version is available.

Snipaste_2024-08-28_22-47-05

### Why

Closes #1162

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-cli/src/cli.rs
+++ b/cmd/soroban-cli/src/cli.rs
@@ -75,7 +75,9 @@ pub async fn main() {
     // Spawn a thread to check if a new version exists.
     // It depends on logger, so we need to place it after
     // the code block that initializes the logger.
-    thread::spawn(upgrade_check);
+    thread::spawn(move || {
+        upgrade_check(root.global_args.quiet);
+    });
 
     if let Err(e) = root.run().await {
         eprintln!("error: {e}");
--- a/cmd/soroban-cli/src/cli.rs
+++ b/cmd/soroban-cli/src/cli.rs
@@ -1,5 +1,6 @@
 use clap::CommandFactory;
 use dotenvy::dotenv;
+use std::thread;
 use tracing_subscriber::{fmt, EnvFilter};
 
 use crate::self_outdated_check::print_upgrade_prompt;
@@ -7,6 +8,9 @@ use crate::{commands, Root};
 
 #[tokio::main]
 pub async fn main() {
+    // Spawn a thread to print the upgrade prompt in the background
+    thread::spawn(print_upgrade_prompt);
+
     let _ = dotenv().unwrap_or_default();
 
     // Map SOROBAN_ env vars to STELLAR_ env vars for backwards compatibility
@@ -35,16 +39,13 @@ pub async fn main() {
     let mut root = Root::new().unwrap_or_else(|e| match e {
         commands::Error::Clap(e) => {
             let mut cmd = Root::command();
-            let e = e.format(&mut cmd);
-            let _ = e.print();
-            exit(e.exit_code());
+            e.format(&mut cmd).exit();
         }
         e => {
             eprintln!("{e}");
-            exit(1);
+            std::process::exit(1);
         }
     });
-
     // Now use root to setup the logger
     if let Some(level) = root.global_args.log_level() {
         let mut e_filter = EnvFilter::from_default_env()
@@ -58,7 +59,7 @@ pub async fn main() {
                     .parse()
                     .map_err(|e| {
                         eprintln!("{e}: {filter}");
-                        exit(1);
+                        std::process::exit(1);
                     })
                     .unwrap(),
             );
@@ -76,13 +77,6 @@ pub async fn main() {
 
     if let Err(e) = root.run().await {
         eprintln!("error: {e}");
-        exit(1);
+        std::process::exit(1);
     }
-
-    print_upgrade_prompt(root.global_args.quiet);
-}
-
-fn exit(exit_code: i32) -> ! {
-    print_upgrade_prompt(false);
-    std::process::exit(exit_code);
 }
--- a/cmd/soroban-cli/src/cli.rs
+++ b/cmd/soroban-cli/src/cli.rs
@@ -2,6 +2,7 @@ use clap::CommandFactory;
 use dotenvy::dotenv;
 use tracing_subscriber::{fmt, EnvFilter};
 
+use crate::self_outdated_check::print_upgrade_prompt;
 use crate::{commands, Root};
 
 #[tokio::main]
@@ -34,13 +35,16 @@ pub async fn main() {
     let mut root = Root::new().unwrap_or_else(|e| match e {
         commands::Error::Clap(e) => {
             let mut cmd = Root::command();
-            e.format(&mut cmd).exit();
+            let e = e.format(&mut cmd);
+            let _ = e.print();
+            exit(e.exit_code());
         }
         e => {
             eprintln!("{e}");
-            std::process::exit(1);
+            exit(1);
         }
     });
+
     // Now use root to setup the logger
     if let Some(level) = root.global_args.log_level() {
         let mut e_filter = EnvFilter::from_default_env()
@@ -54,7 +58,7 @@ pub async fn main() {
                     .parse()
                     .map_err(|e| {
                         eprintln!("{e}: {filter}");
-                        std::process::exit(1);
+                        exit(1);
                     })
                     .unwrap(),
             );
@@ -72,6 +76,13 @@ pub async fn main() {
 
     if let Err(e) = root.run().await {
         eprintln!("error: {e}");
-        std::process::exit(1);
+        exit(1);
     }
+
+    print_upgrade_prompt(root.global_args.quiet);
+}
+
+fn exit(exit_code: i32) -> ! {
+    print_upgrade_prompt(false);
+    std::process::exit(exit_code);
 }
--- a/cmd/soroban-cli/src/cli.rs
+++ b/cmd/soroban-cli/src/cli.rs
@@ -3,13 +3,13 @@ use dotenvy::dotenv;
 use std::thread;
 use tracing_subscriber::{fmt, EnvFilter};
 
-use crate::self_outdated_check::print_upgrade_prompt;
+use crate::upgrade_check::upgrade_check;
 use crate::{commands, Root};
 
 #[tokio::main]
 pub async fn main() {
     // Spawn a thread to print the upgrade prompt in the background
-    thread::spawn(print_upgrade_prompt);
+    thread::spawn(upgrade_check);
 
     let _ = dotenv().unwrap_or_default();
 
--- a/cmd/soroban-cli/src/cli.rs
+++ b/cmd/soroban-cli/src/cli.rs
@@ -8,9 +8,6 @@ use crate::{commands, Root};
 
 #[tokio::main]
 pub async fn main() {
-    // Spawn a thread to print the upgrade prompt in the background
-    thread::spawn(upgrade_check);
-
     let _ = dotenv().unwrap_or_default();
 
     // Map SOROBAN_ env vars to STELLAR_ env vars for backwards compatibility
@@ -74,6 +71,11 @@ pub async fn main() {
         tracing::subscriber::set_global_default(subscriber)
             .expect("Failed to set the global tracing subscriber");
     }
+
+    // Spawn a thread to check if a new version exists.
+    // It depends on logger, so we need to place it after
+    // the code block that initializes the logger.
+    thread::spawn(upgrade_check);
 
     if let Err(e) = root.run().await {
         eprintln!("error: {e}");
--- a/cmd/soroban-cli/src/config/self_outdated_check.rs
+++ b/cmd/soroban-cli/src/config/self_outdated_check.rs
@@ -2,6 +2,7 @@ use crate::config::locator;
 use jsonrpsee_core::Serialize;
 use serde::Deserialize;
 use std::fs;
+use semver::Version;
 
 const FILE_NAME: &str = "self_outdated_check.toml";
 
@@ -13,17 +14,17 @@ pub struct SelfOutdatedCheck {
     /// The timestamp of the latest check for a new version of the CLI.
     pub latest_check_time: u64,
     /// The latest stable version of the CLI available on crates.io.
-    pub max_stable_version: String,
+    pub max_stable_version: Version,
     /// The latest version of the CLI available on crates.io, including pre-releases.
-    pub max_version: String,
+    pub max_version: Version,
 }
 
 impl Default for SelfOutdatedCheck {
     fn default() -> Self {
         Self {
             latest_check_time: 0,
-            max_stable_version: "0.0.0".to_string(),
-            max_version: "0.0.0".to_string(),
+            max_stable_version: Version::new(0, 0, 0),
+            max_version: Version::new(0, 0, 0),
         }
     }
 }
@@ -67,13 +68,13 @@ mod tests {
         let default_check = SelfOutdatedCheck::load().unwrap();
         assert_eq!(default_check, SelfOutdatedCheck::default());
         assert_eq!(default_check.latest_check_time, 0);
-        assert_eq!(default_check.max_stable_version, "0.0.0");
+        assert_eq!(default_check.max_stable_version, Version::new(0, 0, 0));
 
         // Test saving and loading
         let saved_check = SelfOutdatedCheck {
             latest_check_time: 1_234_567_890,
-            max_stable_version: "1.2.3".to_string(),
-            max_version: "1.2.4-rc.1".to_string(),
+            max_stable_version: Version::new(1, 2, 3),
+            max_version: Version::parse("1.2.4-rc.1").unwrap()
         };
         saved_check.save().unwrap();
         let loaded_check = SelfOutdatedCheck::load().unwrap();
--- a/cmd/soroban-cli/src/config/self_outdated_check.rs
+++ b/cmd/soroban-cli/src/config/self_outdated_check.rs
@@ -39,8 +39,7 @@ impl SelfOutdatedCheck {
         }
         let data = fs::read(&path)
             .map_err(|error| locator::Error::SelfOutdatedCheckReadFailed { path, error })?;
-        let res = toml::from_slice(data.as_slice());
-        Ok(res?)
+        Ok(toml::from_slice(data.as_slice())?)
     }
 
     /// Saves the state of the self-outdated check to the `self_outdated_check.toml` file in the global configuration directory.
--- a/cmd/soroban-cli/src/config/self_outdated_check.rs
+++ b/cmd/soroban-cli/src/config/self_outdated_check.rs
@@ -1,8 +1,8 @@
 use crate::config::locator;
 use jsonrpsee_core::Serialize;
+use semver::Version;
 use serde::Deserialize;
 use std::fs;
-use semver::Version;
 
 const FILE_NAME: &str = "self_outdated_check.toml";
 
@@ -73,7 +73,7 @@ mod tests {
         let saved_check = SelfOutdatedCheck {
             latest_check_time: 1_234_567_890,
             max_stable_version: Version::new(1, 2, 3),
-            max_version: Version::parse("1.2.4-rc.1").unwrap()
+            max_version: Version::parse("1.2.4-rc.1").unwrap(),
         };
         saved_check.save().unwrap();
         let loaded_check = SelfOutdatedCheck::load().unwrap();
--- a/cmd/soroban-cli/src/config/self_outdated_check.rs
+++ b/cmd/soroban-cli/src/config/self_outdated_check.rs
@@ -10,15 +10,20 @@ const FILE_NAME: &str = "self_outdated_check.toml";
 /// the global configuration directory.
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct SelfOutdatedCheck {
+    /// The timestamp of the latest check for a new version of the CLI.
     pub latest_check_time: u64,
-    pub latest_version: String,
+    /// The latest stable version of the CLI available on crates.io.
+    pub max_stable_version: String,
+    /// The latest version of the CLI available on crates.io, including pre-releases.
+    pub max_version: String,
 }
 
 impl Default for SelfOutdatedCheck {
     fn default() -> Self {
         Self {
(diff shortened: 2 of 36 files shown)


cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,77 @@
...
+                latest_version: latest_stable_version,
+            };
+            stats.save().unwrap_or_default();
+        }
+    }
+
+    let current_version = Version::parse(current_version).unwrap();
+    let latest_version = Version::parse(&stats.latest_version).unwrap();
+
+    if latest_version > current_version {
```
leighmcculloch, line 57: Some folks install from source from git with a dev release. I guess it's fine to keep this.

Some folks also install rc releases, and it'd be nice to offer them a reminder too when a new rc release comes out, not just a new stable release, but I think that can be an improvement for another day.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,77 @@
...
+    let crate_name = env!("CARGO_PKG_NAME");
+    let url = format!("{CRATES_IO_API_URL}{crate_name}");
+    let response = ureq::get(&url).timeout(REQUEST_TIMEOUT).call()?;
+    let crate_data: CrateResponse = response.into_json()?;
+    Ok(crate_data.crate_.max_stable_version)
+}
+
+/// Print a warning if a new version of the CLI is available
+pub fn print_upgrade_prompt(quiet: bool) {
+    let current_version = env!("CARGO_PKG_VERSION");
```
leighmcculloch, line 35: I'm inclined to get this from the function that we use elsewhere to contain where to get the version from, but I don't feel strongly the env var is fine.
```suggestion
    let current_version = crate::version::pkg();
```

leighmcculloch (commented):
👏🏻 Looks great. A couple asks inline.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,77 @@
...
+fn fetch_latest_stable_version() -> Result<String, Box<dyn Error>> {
+    let crate_name = env!("CARGO_PKG_NAME");
+    let url = format!("{CRATES_IO_API_URL}{crate_name}");
+    let response = ureq::get(&url).timeout(REQUEST_TIMEOUT).call()?;
+    let crate_data: CrateResponse = response.into_json()?;
+    Ok(crate_data.crate_.max_stable_version)
+}
+
+/// Print a warning if a new version of the CLI is available
+pub fn print_upgrade_prompt(quiet: bool) {
```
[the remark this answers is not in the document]
overcat: 9fece38ebd1fa3b30f0713c00ee2e141e34bd7f5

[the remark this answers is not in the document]
overcat: 9fece38ebd1fa3b30f0713c00ee2e141e34bd7f5

cmd/soroban-cli/src/self_outdated_check.rs
overcat: aeff349d18c88edb58bdd288abfe8c53696e19d6

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,77 @@
...
+    let latest_version = Version::parse(&stats.latest_version).unwrap();
+
+    if latest_version > current_version {
+        print.println("");
+        print.warnln(format!(
+            "A new release of stellar-cli is available: {current_version} -> {latest_version}",
+        ));
+        print.warnln(
+            "We recommend upgrading to the latest version to get the latest features and bug fixes."
+        );
```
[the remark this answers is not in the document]
overcat: 29a74a18bb4a1b6f69fcb29e72a59282be56840b

cmd/soroban-cli/src/self_outdated_check.rs
overcat: I have added support for pre-release here(c1e8bbcc6bc875505a53b1ba1757b7ac69c12ceb). If the current user is using a pre-release version, they will be prompted to upgrade to the stable version first. If the stable version has not been released yet, they will be prompted to upgrade to the latest pre-release version.

cmd/soroban-cli/src/cli.rs
```diff
@@ -72,6 +76,13 @@ pub async fn main() {
 
     if let Err(e) = root.run().await {
         eprintln!("error: {e}");
-        std::process::exit(1);
+        exit(1);
     }
+
+    print_upgrade_prompt(root.global_args.quiet);
```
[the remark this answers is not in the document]
overcat: a3d935966aa0aeee3b5fa1b234290a680a50c3c8

Hi @leighmcculloch, currently, I have still modified it to run in a separate thread. But I still doubt whether this is better than the previous design.

The previous version was intentional by me, because I wanted this check to always be executed, so that users can see this prompt when encountering issues due to using a lower version of the CLI. Also, I think it should only make one HTTP request per day and have a timeout configuration of 30 seconds, which should be an acceptable resource consumption. Additionally, users can disable it by configuring environment variables.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    // Skip fetch from crates.io if we've checked recently
+    if now - stats.latest_check_time >= MINIMUM_CHECK_INTERVAL.as_secs() {
+        if let Ok(c) = fetch_latest_crate_info() {
+            stats = SelfOutdatedCheck {
+                latest_check_time: now,
+                max_stable_version: c.max_stable_version,
+                max_version: c.max_version,
+            };
+            stats.save().unwrap_or_default();
```
willemneal, line 66: This will suppress the error. I suggest creating a new function that does return an error and in the error case print a warning. It would be good for users to report if they have any issues.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,147 @@
...
+    #[serde(rename = "crate")]
+    crate_: Crate,
+}
+
+#[derive(Deserialize)]
+struct Crate {
+    #[serde(rename = "max_stable_version")]
+    max_stable_version: String,
+    #[serde(rename = "max_version")]
+    max_version: String, // This is the latest version, including pre-releases
```
willemneal, line 25:
```suggestion
    #[serde(rename = "max_stable_version")]
    max_stable_version: Version,
    #[serde(rename = "max_version")]
    max_version: Version, // This is the latest version, including pre-releases
```
Does this work?

cmd/soroban-cli/src/config/self_outdated_check.rs
```diff
@@ -0,0 +1,82 @@
...
+/// This state is global and stored in the `self_outdated_check.toml` file in
+/// the global configuration directory.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SelfOutdatedCheck {
+    /// The timestamp of the latest check for a new version of the CLI.
+    pub latest_check_time: u64,
+    /// The latest stable version of the CLI available on crates.io.
+    pub max_stable_version: String,
+    /// The latest version of the CLI available on crates.io, including pre-releases.
+    pub max_version: String,
```
willemneal, line 18: Could also use `Version` here. This will prevent needing to parse every time when it's used.

willemneal (changes requested):
Agree this is awesome! Just one nit about Error handling and another about using `Version` directly in the new types.

cmd/soroban-cli/src/config/self_outdated_check.rs
```diff
@@ -0,0 +1,82 @@
...
+        let data = fs::read(&path)
+            .map_err(|error| locator::Error::SelfOutdatedCheckReadFailed { path, error })?;
+        Ok(toml::from_slice(data.as_slice())?)
+    }
+
+    /// Saves the state of the self-outdated check to the `self_outdated_check.toml` file in the global configuration directory.
+    pub fn save(&self) -> Result<(), locator::Error> {
+        let path = locator::global_config_path()?.join(FILE_NAME);
+        let path = locator::ensure_directory(path)?;
+        let data = toml::to_string(self).map_err(|_| locator::Error::ConfigSerialization)?;
```
leighmcculloch, line 49: Can we use JSON for this file? The CLI uses JSON for most of the config files that get written, except for one, the key file that for some historical reason uses TOML.

cmd/soroban-cli/src/config/self_outdated_check.rs
```diff
@@ -0,0 +1,82 @@
+use crate::config::locator;
+use jsonrpsee_core::Serialize;
+use semver::Version;
+use serde::Deserialize;
+use std::fs;
+
+const FILE_NAME: &str = "self_outdated_check.toml";
```
leighmcculloch, line 7: This is really minor, but I think we should call this check the 'update_check'. The phrase rolls off the tongue a little easier.
```suggestion
const FILE_NAME: &str = "upgrade_check.toml";
```

cmd/soroban-cli/src/config/self_outdated_check.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// If the file doesn't exist, returns a default instance of `SelfOutdatedCheck`.
+    pub fn load() -> Result<Self, locator::Error> {
+        let path = locator::global_config_path()?.join(FILE_NAME);
+        if !path.exists() {
+            return Ok(Self::default());
+        }
+        let data = fs::read(&path)
+            .map_err(|error| locator::Error::SelfOutdatedCheckReadFailed { path, error })?;
+        let res = toml::from_slice(data.as_slice());
+        Ok(res?)
```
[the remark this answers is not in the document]
overcat: Changed.

cmd/soroban-cli/src/config/self_outdated_check.rs
overcat: Changed.

cmd/soroban-cli/src/self_outdated_check.rs
overcat: I made modifications, now it prints an alert when encountering issues.

overcat: I made modifications, now it prints an alert when encountering issues.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,159 @@
...
+        return;
+    }
+
+    let current_version = crate::commands::version::pkg();
+    let print = Print::new(false);
+
+    let mut stats = match SelfOutdatedCheck::load() {
+        Ok(stats) => stats,
+        Err(e) => {
+            print.warnln(format!("Failed to load self outdated check data: {}", e));
```
leighmcculloch, line 56: I think we should send this line to the logs instead of to stderr. The update check is supporting functionality, and if it fails, but the user is still able to move ahead with the action they're performing we shouldn't get in the users way with info about the update check imo.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,159 @@
...
+        match fetch_latest_crate_info() {
+            Ok(c) => {
+                stats = SelfOutdatedCheck {
+                    latest_check_time: now,
+                    max_stable_version: c.max_stable_version,
+                    max_version: c.max_version,
+                };
+            }
+            Err(e) => {
+                print.warnln(format!("Failed to fetch stellar-cli info from crates.io: {}", e));
```
leighmcculloch: Same with this line, we should send this to the log instead of to stderr.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,159 @@
...
+                // Only update the latest check time if the fetch failed
+                // This way we don't spam the user with errors
+                stats.latest_check_time = now;
+            }
+        }
+
+        if let Err(e) = stats.save() {
+            print.warnln(format!("Failed to save self outdated check data: {}", e));
```
leighmcculloch, line 86: Same with this line, we should send this to the log instead of to stderr.

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,159 @@
...
+        // If we are currently using a non-preview version
+        &stats.max_stable_version
+    } else {
+        // If we are currently using a preview version
+        if stats.max_stable_version > *current_version {
+            // If there is a new stable version available, we should use that instead
+            &stats.max_stable_version
+        } else {
+            &stats.max_version
+        }
```
leighmcculloch, line 111: Nice! 👏🏻

cmd/soroban-cli/src/cli.rs
```diff
@@ -1,11 +1,16 @@
...
+use std::thread;
 use tracing_subscriber::{fmt, EnvFilter};
 
+use crate::self_outdated_check::print_upgrade_prompt;
 use crate::{commands, Root};
 
 #[tokio::main]
 pub async fn main() {
+    // Spawn a thread to print the upgrade prompt in the background
+    thread::spawn(print_upgrade_prompt);
```
leighmcculloch, line 12:
```suggestion
    thread::spawn(upgrade_check);
```

cmd/soroban-cli/src/cli.rs
```diff
@@ -1,11 +1,16 @@
 use clap::CommandFactory;
 use dotenvy::dotenv;
+use std::thread;
 use tracing_subscriber::{fmt, EnvFilter};
 
+use crate::self_outdated_check::print_upgrade_prompt;
```
leighmcculloch, line 6:
```suggestion
use crate::self_outdated_check::upgrade_check;
```

cmd/soroban-cli/src/self_outdated_check.rs
```diff
@@ -0,0 +1,162 @@
...
+fn fetch_latest_crate_info() -> Result<Crate, Box<dyn Error>> {
+    let crate_name = env!("CARGO_PKG_NAME");
+    let url = format!("{CRATES_IO_API_URL}{crate_name}");
+    let response = ureq::get(&url).timeout(REQUEST_TIMEOUT).call()?;
+    let crate_data: CrateResponse = response.into_json()?;
+    Ok(crate_data.crate_)
+}
+
+/// Print a warning if a new version of the CLI is available
+pub fn print_upgrade_prompt() {
```
leighmcculloch, line 38:
```suggestion
pub fn upgrade_check() {
```

cmd/soroban-cli/src/cli.rs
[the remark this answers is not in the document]
leighmcculloch: > I wanted this check to always be executed

I understand the motivation here, but the upgrade check is a convenience and secondary feature. I think it's important that it nevers reduces the experience of any commands primary functionality. It's important that folks don't see 30 second delays (the timeout period) if they're on spotty wifi, or have slow internet. Especially when running commands that otherwise don't make any connection to the outside world.

I understand this will mean some fast commands won't see the message until a slower command has picked it up, then fast commands will display the warning. I think this is an appropriate trade off.

cmd/soroban-cli/src/config/self_outdated_check.rs
```diff
@@ -0,0 +1,82 @@
...
+use serde::Deserialize;
+use std::fs;
+
+const FILE_NAME: &str = "self_outdated_check.toml";
+
+/// The `SelfOutdatedCheck` struct represents the state of the self-outdated check.
+/// This state is global and stored in the `self_outdated_check.toml` file in
+/// the global configuration directory.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct SelfOutdatedCheck {
```
leighmcculloch, line 13: Could we rename this type `UpdateCheck`? I think the name communicates the intent a little clearer.

cmd/soroban-cli/src/config/self_outdated_check.rs
willemneal: Why JSON over TOML?

overcat: We should decide on a format, to maintain consistency in the project.

leighmcculloch: We've used a mix of JSON and TOML for configs. Most of the "config" configs (internal, not in the user directory) are JSON. The update check config is another internal (not in the user directory).

In general, I think we should pick JSON unless there's value in picking something else. JSON is more accessible with a wide array of tooling without having to navigate the world of toml. Toml is popular, but the tooling isn't anywhere near as solid as JSON. (I had a case earlier today where a popular TOML lib couldn't render a JS Object, even though JSON could.)

There are some advantages to using TOML for very simple (single level) user facing config files, but this isn't a user facing config file.

Even for user facing config files, TOML can quickly become unintuitive for nested tables.

I'm not against toml in general, and they are easier on the eyes for simple flat config files. But JSON is sufficient, especially for internal config files.

cmd/soroban-cli/src/self_outdated_check.rs
overcat: Changed to use `println!` for output.

leighmcculloch: Sorry, I should have been more specific, we need to use `tracing::warn` to write to the logs. The logs are a separate stream that can be enabled/disabled separately to the main application output.

overcat: Thank you for pointing this out. I initially thought there was a log library here, but when I searched for `log`, I didn't find it. 🥲

overcat: Fixed in 5e3c073a206688e42fbcb049ab12499d2e120993.
It seems that we are currently unable to place `thread::spawn(upgrade_check);` at the very beginning.

willemneal:
This is because of my update. Trying to merge PR that needed the update

cmd/soroban-cli/src/config/upgrade_check.rs
```diff
@@ -0,0 +1,83 @@
...
+
+const FILE_NAME: &str = "upgrade_check.json";
+
+/// The `UpgradeCheck` struct represents the state of the upgrade check.
+/// This state is global and stored in the `upgrade_check.json` file in
+/// the global configuration directory.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct UpgradeCheck {
+    /// The timestamp of the latest check for a new version of the CLI.
+    pub latest_check_time: u64,
```
willemneal, line 16: One last idea. Use `chrono::DateTime<chrono::offset::Utc>` directly. This way again don't have to deal with converting to a primitive and can compare directly with `chrono::Utc::now()`

`now - MINIMUM_CHECK_INTERVAL >= stats.latest_check_time`. 

Small nit but makes it a little cleaner.

willemneal (approved):
Small nit, but otherwise looking really great

cmd/soroban-cli/src/config/upgrade_check.rs
overcat: Thanks for spotting this, changed in c967a6102d07ed6160dd3bbfbdffa2ab160c9ba4.

willemneal:
@overcat the e2e issue is fixed. Just need to update your branch (I don't have permission)

cmd/soroban-cli/src/upgrade_check.rs
```diff
@@ -0,0 +1,156 @@
...
+        match fetch_latest_crate_info() {
+            Ok(c) => {
+                stats = UpgradeCheck {
+                    latest_check_time: now,
+                    max_stable_version: c.max_stable_version,
+                    max_version: c.max_version,
+                };
+            }
+            Err(e) => {
+                tracing::error!("⚠️ Failed to fetch stellar-cli info from crates.io: {e}");
```
leighmcculloch, line 70:
```suggestion
                tracing::error!("Failed to fetch stellar-cli info from crates.io: {e}");
```

cmd/soroban-cli/src/upgrade_check.rs
```diff
@@ -0,0 +1,156 @@
...
+                // Only update the latest check time if the fetch failed
+                // This way we don't spam the user with errors
+                stats.latest_check_time = now;
+            }
+        }
+
+        if let Err(e) = stats.save() {
+            tracing::error!("⚠️ Failed to save upgrade check data: {e}");
```
leighmcculloch, line 78:
```suggestion
            tracing::error!("Failed to save upgrade check data: {e}");
```

cmd/soroban-cli/src/upgrade_check.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    let current_version = Version::parse(current_version).unwrap();
+    let latest_version = get_latest_version(&current_version, &stats);
+
+    if *latest_version > current_version {
+        tracing::warn!(
+            "⚠️ A new release of stellar-cli is available: {current_version} -> {latest_version}",
```
leighmcculloch, line 87: This line was changed to a log file output, but this line should stay a print to stderr using the Print type.

The goal of sending errors to logs was to only send the errors to logs, but the message to users to upgrade should be a user facing text string, not a log.

leighmcculloch (changes requested):
One line got changed to logs that should have started a print line. Otherwise this looks good.

cmd/soroban-cli/src/upgrade_check.rs
overcat: Fixed in 523256860c71ff446a242fbfa439971d268996f2

If I understand correctly, for it we should follow the setting of `root.global_args.quiet`, right?

overcat: Fixed in 523256860c71ff446a242fbfa439971d268996f2

If I understand correctly, for it we should follow the setting of `root.global_args.quiet`, right?

leighmcculloch (approved):

cmd/soroban-cli/src/upgrade_check.rs
leighmcculloch: Yup!

overcat:
Thank you both for your patient guidance 😊

Issue #1420, opened 2024-06-28
Add `stellar contract alias` subcommand


janewang:
- [ ] `stellar contract alias rm [ALIAS] --network NETWORK`
- [ ] `stellar contract alias add [ALIAS] --network NETWORK`
- [ ] `stellar contract alias ls --network NETWORK`
- [ ] `stellar contract alias show [ALIAS] --network NETWORK`

Maybe add `--all` to list all networks.

fnando:
Suggestion: `stellar contract alias list|rm`, because we may want to introduce other scoped aliasing (e.g. assets)

willemneal:
Could add `add` while you're at it.

willemneal:
Added `show` !

leighmcculloch:
Why does the alias command require the network? I thought there was only a single alias.

fnando:
> Why does the alias command require the network? I thought there was only a single alias.

It's scoped by network. I personally would just recycle commands on the terminal and change just the network.

leighmcculloch:
Did this issue get closed because we resolved to no longer do it? As far as I can tell this has not been done yet.

PR #1572, opened 2024-08-28
Add `contract alias ls`.

fnando:
### What

```console
$ cargo run contract alias ls --global
   Compiling soroban-cli v21.4.1 (/Users/fnando/Projects/stellar/stellar-cli/cmd/soroban-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.08s
     Running `target/debug/soroban contract alias list --global`
ℹ️ Aliases available for network 'Test SDF Network ; September 2015'
hello_world: CAXUMKGXAOY6IAO36XJV6DSOQBMBANSUJ6NR5EXIF27ZWNPW7OOTTHPC
increment: CALLOOCCCJRZ6AZAJSKNCTDZ766N4ANVRKRJPHC65KDV7RJCS37PS4XY

ℹ️ Aliases available for network 'Standalone Network ; February 2017'
hello_world: CC47LWR7ZJH2QQTFUUNRERMBPBAUHIVLKHIKACV6VVZEEQJIHJOJNGPL


$ echo $?
0

$ cargo run contract alias ls
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/soroban contract alias list`
⚠️ No aliases defined for network

$ echo $?
1
```

### Why

#1420 

### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/alias/list.rs
```diff
@@ -0,0 +1,88 @@
...
+        #[allow(unused_variables)]
+        let network_passphrase = &network.network_passphrase;
+        let config_dir = self.config_locator.config_dir()?;
+        let pattern = config_dir
+            .join("contract-ids")
+            .join("*.json")
+            .to_string_lossy()
+            .into_owned();
+
+        let paths = glob::glob(&pattern)?;
```
Ifropc, line 55: If refactoring slightly we don't need to add a new dependency :)
```suggestion
        let paths = fs::read_dir(config_dir.join("contract-ids"))?
            .map(|res| res.map(|e| e.path()))
            .collect::<Result<Vec<_>, io::Error>>()?
            .into_iter()
            .filter(|e| e.extension() == Some("json".as_ref()));
```

cmd/soroban-cli/src/commands/contract/alias/list.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Locator(#[from] locator::Error),
+
+    #[error(transparent)]
+    Network(#[from] network::Error),
+
+    #[error("no contract found with alias '{alias}' for network '{network_passphrase}'")]
+    NoContract {
```
Ifropc, line 29: Seems to be unused 🤔

FULL_HELP_DOCS.md
```diff
@@ -163,6 +163,7 @@ Utilities to manage contract aliases
 * `remove` — Remove contract alias
 * `add` — Add contract alias
 * `show` — Show the contract id associated with a given alias
+* `list` — List all stored aliases for a given network
```
Ifropc, line 166: Should it be `ls` instead? Also maybe `remove` -> `rm`?
To use the same style as `stellar keys`/`stellar network`

cmd/soroban-cli/src/commands/contract/alias/list.rs
fnando: The dependency was already there (by another crate). I just made it explicit. :)

FULL_HELP_DOCS.md
fnando: Good catch. I think so. I'll rename it in a separate pr.

willemneal (commented):
My one thought is that it would be good to have a way to list all across all networks. Like -l on other list commands.

fnando:
@willemneal I think that's a good idea (#1420 even mentions it), but I wonder how we can achieve that by using `command(flatten)`. 🤔

I tried to make it work, but failed to make everything work.

Ifropc:
As an idea we can add list `--all` that conflicts with `--network`
The only thing is that our current flags are already pretty complicated, not sure if it makes sense to complicate it even further

fnando:
@Ifropc I did that, but network options are required, and simply using `Option<network::Args>` doesn't work (it complains about `#[group(skip)]`, and even without it doesn't work).

Ifropc:
Yeah, I was looking into groups, unfortunately we can't make it work easily while using derive API, this is a known issue in clap (and I don't think it will be changed, based on the discussion in Clap repo). So our only option is to migrate to builder for networks and maybe other flags as well, which we probably don't want to spend our time on right now.  
What I suggest is to not use `Option<network::Args>` and instead only use network name as a string (testnet/mainnet/etc) -> as we don't need to know RPC url to fetch aliases.
But it will conflict with how we are using network everywhere else, so maybe not a very good idea

fnando:
I guess we should ignore `--all` for now.

leighmcculloch:
I think we can keep this command simpler and output all aliases for all networks, and provide no filtering options.

fnando:
@leighmcculloch works for me!

fnando:
The list output is currently this:

```console
$ cargo run contract alias list --global
   Compiling soroban-cli v21.4.1 (/Users/fnando/Projects/stellar/stellar-cli/cmd/soroban-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.08s
     Running `target/debug/soroban contract alias list --global`
ℹ️ Aliases available for network 'Test SDF Network ; September 2015'
hello_world: CAXUMKGXAOY6IAO36XJV6DSOQBMBANSUJ6NR5EXIF27ZWNPW7OOTTHPC
increment: CALLOOCCCJRZ6AZAJSKNCTDZ766N4ANVRKRJPHC65KDV7RJCS37PS4XY

ℹ️ Aliases available for network 'Standalone Network ; February 2017'
hello_world: CC47LWR7ZJH2QQTFUUNRERMBPBAUHIVLKHIKACV6VVZEEQJIHJOJNGPL


$ echo $?
0

$ cargo run contract alias list
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/soroban contract alias list`
⚠️ No aliases defined for network

$ echo $?
1
```

chadoh (approved):
Thanks!

PR #1573, opened 2024-08-28
chore: remove unused `enum FrontendTemplate`

chadoh:
It was marked as `pub` so the checker didn't tell us it was unused.

But it's not used anywhere else in the codebase.


leighmcculloch (approved):

leighmcculloch:
@chadoh The change causes an unused import. Needs fixing before merge.

PR #1574, opened 2024-08-29
feat: Add parsing to deploy to support new constructor feature

### LINKED ISSUE #1561: CAP-58: Support deploy with constructor arguments

willemneal:
If a contract has a method called `__contructor` then it parses the slop, e.i. anything after `--`. So after adding one to the hello world contract you would use it like:

```
cargo s contract deploy --build-only --wasm ./target/wasm32-unknown-unknown/test-wasms/test_hello_world.wasm --source default --network testnet -- --help
ℹ️ Using wasm hash 75eca9c92c997f6b1b1c7caba32311e81077f29c2f11eb3be6d453406e939a34
Example constructor
Usage Notes:
Each arg has a corresponding --<arg_name>-file-path which is a path to a file containing the corresponding JSON argument.
Note: The only types which aren't JSON are Bytes and BytesN, which are raw bytes

Usage: __constructor [OPTIONS]

Options:
      --world <String>
          Example:
            --world '"hello world"'

  -h, --help
```

For now this is a noop. And adding the new operation is still needed but hard to test now.

Close #1561


willemneal:
Requires RPC updates and release https://github.com/stellar/rs-stellar-rpc-client/pull/12

Cargo.lock
```diff
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.2"
```
leighmcculloch, line 7: It's surprising so many versions are changing. Is that required by this change?

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -215,17 +228,54 @@ impl NetworkRunnable for Cmd {
...
+                )
+            }
+        } else {
+            None
+        };
+
         // Get the account sequence number
-        let account_details = client.get_account(&key.to_string()).await?;
+        let account_details = client
+            .get_account(&config.source_account()?.to_string())
```
leighmcculloch, line 269: Why did this change to accessing config.source_account(), when higher up the source account is defined as coming from the bytes? Iiuc both would need to be the same, what's the reason to change this location?

Cargo.lock
willemneal: Not sure why they got included.

leighmcculloch: There's a huge number of deps changing. When I reverted the dep changes locally, it seemed none of them were required by this change. I pushed the revert in eb87e6be6865b0663ab61b575d74ad815865d38f.

leighmcculloch (approved):

leighmcculloch:
Pushed a few changes, mostly removing deps, a couple other minor things. Approved. Set to merge.

PR #1584, opened 2024-09-05
Fix/update init error handling

elizabethengelman:
### What

Return custom errors with context from the init command. This PR comes from a comment on a previous init PR: https://github.com/stellar/stellar-cli/pull/1548#pullrequestreview-2262486637

### Why

We can reduce the error printing we're doing in individual commands by letting those errors bubble up to the root command. This will also allow the callers of the functions have some more context about the error.

### Known limitations

I'm not sure how I feel about the io error handling. Some of these errors were already printing out messages, so I used those messages for the custom error text. Some of the other io errors just returned as is - for those I created a `StdIo` error - partially because it may be overkill to `map_err` for each of those errors. Happy to rethink this piece if folks have other ideas.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/init.rs
+++ b/cmd/soroban-cli/src/commands/contract/init.rs
@@ -73,10 +73,10 @@ pub enum Error {
     ConvertBytesToString(#[from] str::Utf8Error),
 
     #[error("Error preparing fetch repository: {0}")]
-    PrepareFetch(#[from] clone::Error),
+    PrepareFetch(Box<clone::Error>),
 
     #[error("Failed to fetch repository: {0}")]
-    Fetch(clone::fetch::Error),
+    Fetch(Box<clone::fetch::Error>),
 
     #[error("Failed to checkout main worktree: {0}")]
     Checkout(#[from] clone::checkout::main_worktree::Error),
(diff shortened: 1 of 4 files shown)


cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -387,73 +344,64 @@ impl Runner {
...
-            ));
-            e
-        })?;
+        Self::write(&cargo_path, &doc.to_string())?;
 
         Ok(())
     }
 
     fn copy_frontend_files(&self, from: &Path, to: &Path) -> Result<(), Error> {
-        self.print.infoln("ℹ️  Initializing with frontend template");
```
elizabethengelman, line 409: this was a bug introduced in a previous PR - it was printing an extra ℹ️

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -219,16 +206,11 @@ impl Runner {
...
-                .errorln(format!("Error reading directory: {from:?}"));
-            e
-        })? {
-            let entry = entry.map_err(|e| {
-                self.print
-                    .errorln(format!("Error reading entry in directory: {from:?}"));
-                e
-            })?;
+        for entry in read_dir(from)
+            .map_err(|e| Error::Io(format!("Error reading directory: {from:?}"), e))?
```
leighmcculloch, line 210: Error strings usually start with a lower case letter I think. We might have some inconsistency, but I think that's what I saw last I checked.

leighmcculloch: Also we can drop the prefix "Error " because when outputted errors are proceeded with "error: " iirc.

elizabethengelman: Good call to both of those - I'll make an update!

leighmcculloch (approved):

PR #1590, opened 2024-09-10
feat: `tx sign` and new signer

willemneal:
The new `sign` subcommand allows for signing a passed transaction envelope.

Additionally, `signer::{Transaction, TransactionHash}` traits which simplifies the interface. `Transaction` has just a single method which returns a `DecoratedSignature`. `TransactionHash` breaks this up into two methods to get a `Signature` and a `SignatureHint`.


cmd/soroban-cli/src/signer/types.rs
```diff
@@ -0,0 +1,183 @@
...
+                }
+                _ => return Err(Error::UserCancelledSigning),
+            };
+        }
+        let sig = self.key.sign(data);
+        Ok(sig.to_bytes().to_vec())
+    }
+}
+
+pub fn read_key() -> char {
```
elizabethengelman, line 173: I noticed a potential issue with this functionality while trying to add it to the `sign_with_lab` [branch](https://github.com/AhaLabs/stellar-cli/pull/25#discussion_r1755504162). Also possible that I am missing something or calling the command incorrectly.

When I am piping my tx xdr into the cli, the program seems to not have access to the stdin reader when it gets to trying to read the user's y response. In other words, when I run this command:
```
echo AAAAAgAAAADcOHnq5sGLOngOCEMyLqqn5CvFV2HGbOSjJAIzhqBdkAAAAGQAEb7UAAAAAQAAAAAAAAAAAAAAAQAAAAAAAAAYAAAAAQAAAAAAAAAAAAAAANw4eermwYs6eA4IQzIuqqfkK8VXYcZs5KMkAjOGoF2Q04SY/PMseMCgfbEFkX5gTr9qD/d0mvD9H3acFOz/X5oAAAAALtGgcgsAibk5q8VOLY2R1G1NdGX7fLzlz7iBFBy64JIAAAAAAAAAAAAAAAA= \
| stellar tx sign --network testnet --sign-with-key testnet-alice
```

I'm getting the same error that I'm seeing over on the `sign_with_lab` PR:
```
thread 'main' panicked at cmd/soroban-cli/src/signer.rs:164:41:
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Failed to initialize input reader" }
```

cmd/soroban-cli/src/signer/types.rs
```diff
@@ -0,0 +1,183 @@
...
+    async fn sign_txn_hash(
+        &self,
+        source_account: &stellar_strkey::ed25519::PublicKey,
+        txn: [u8; 32],
+    ) -> Result<DecoratedSignature, Error> {
+        eprintln!(
+            "{} about to sign hash: {}",
+            source_account.to_string(),
+            hex::encode(txn)
+        );
```
leighmcculloch, line 77: This log line feels out of place now that we've moved to using the Print object. There'd need to be someway to pass the print object down to print here, or better move this print to a higher place. Printing here is probably not the right layer.

cmd/soroban-cli/src/signer/types.rs
```diff
@@ -0,0 +1,183 @@
...
+        let tx_signature = self.sign_blob(&txn).await?;
+        Ok(DecoratedSignature {
+            // TODO: remove this unwrap. It's safe because we know the length of the array
+            hint: SignatureHint(source_account.0[28..].try_into().unwrap()),
```
leighmcculloch, line 81: I think the trait mechanics need tweaking:

The signer defines both the signature, and the hint included in the decorated signature. The hint is not the source account, it's the signer.

So we don't need source account passed down.

The signer implementing this trait will be able to specify its own hint.

We should also remove the Blob trait, because it offers little value. Any signer to produce DecoratedSignatures would need to implement TransactionHash so they could also define the hint. No blanket impl of Blob -> TransactionHash could guess the hint.

leighmcculloch (changes requested):
A few issues with this command, elaborated inline:
- Signing shouldn't require specifying a source account.
- Signing shouldn't be able to alter the source account.
- Signing shouldn't require additional user confirmation, given the command's only purpose is to sign.
- Trait mechanics. See inline. I think @Ifropc might have commented on this in the past too.

cmd/soroban-cli/src/signer/types.rs
elizabethengelman: > The nice thing about the design of the tx sign UI is there is no need to ask for y. The command already contains a very clear authorization to sign. The user is literally stating --sign-with-key and so additional confirmation is not required imo.

This makes sense to me. 👍  I think this could be the case for most of (all of?) the `sign-with-*` options.

cmd/soroban-cli/Cargo.toml
```diff
@@ -123,6 +123,7 @@ humantime = "2.1.0"
 phf = { version = "0.11.2", features = ["macros"] }
 semver = "1.0.0"
 glob = "0.3.1"
+crossterm = "0.28.1"
```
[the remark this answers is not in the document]
willemneal: Removed it from this PR since as you said it doesn't make sense to approve if we are using the sign command.

cmd/soroban-cli/src/config/sign_with.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// If using a seed phrase to sign, sets which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
+    pub hd_path: Option<usize>,
+
+    /// If one of `--sign-with-*` flags is provided, don't ask to confirm to sign a transaction
+    #[arg(long)]
+    pub yes: bool,
+
+    /// Account that signs the transaction. Alias `source`. Can be an identity (--source alice), a secret key (--source SC36…), or a seed phrase (--source "kite urban…").
+    #[arg(long, visible_alias = "source", env = "STELLAR_ACCOUNT")]
+    pub source_account: String,
```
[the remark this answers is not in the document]
willemneal: This was a mistake on my part it's been removed.

cmd/soroban-cli/src/signer/types.rs
willemneal: This has also been moved to the top level signer and not in the default implementation. In fact there is no longer a default implementation rather a blanket impl.

cmd/soroban-cli/src/signer/types.rs
willemneal: Removed 👍

leighmcculloch (approved):
I've pushed a series of changes that I think simplify the type system around the new logic being added. The new logic is largely unchanged, just moved out of the trait system.

I was going to leave a comment with the suggestion, but wanted to try it out before suggesting, and then discovered some things along the way.

The changes involved in brief:
- Removing traits `Stellar`, `TransactionHash`, `Transaction`. Part of the design of the `StellarSigner` that was already in this PR was an enum containing the different keys, and it then called each function on its key type anyway. So we actually don't need the implementations of different keys to sit behind a common trait, at least not yet with what's presented in this PR. Also the traits were implemented at two layers, which was confusing. I think we can bring back traits when they add value, but it seems like the design of StellarSigner already satisfies.
- Removing terms like "accounts" in any of the signing logic, since these aren't accounts, just keys.
- Updated the signing code used by other fns to call through to the StellarSigner. Doing this only required ~4 lines of code because of how the StellarSigner was already designed, and it means after this PR merges we still only have one signing implementation for txs.

I'm approving this PR, but now that I've contributed changes, it would be ideal if @fnando you could 👀 the PR as a whole and sign off before we merge.

cmd/soroban-cli/src/config/sign_with.rs
```diff
@@ -0,0 +1,61 @@
...
+    #[error(transparent)]
+    StrKey(#[from] stellar_strkey::DecodeError),
+}
+
+#[derive(Debug, clap::Args, Clone, Default)]
+#[group(skip)]
+pub struct Args {
+    /// Sign with a local key. Can be an identity (--sign-with-key alice), a secret key (--sign-with-key SC36…), or a seed phrase (--sign-with-key "kite urban…"). If using seed phrase, `--hd-path` defaults to the `0` path.
+    #[arg(long, conflicts_with = "sign_with_lab", env = "STELLAR_SIGN_WITH_KEY")]
+    pub sign_with_key: Option<String>,
```
fnando, line 33: this is quite verbose. Should we introduce a short flag like `-s`?

cmd/soroban-cli/src/config/sign_with.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Sign with <https://lab.stellar.org>
+    #[arg(
+        long,
+        conflicts_with = "sign_with_key",
+        env = "STELLAR_SIGN_WITH_LAB",
+        hide = true
+    )]
+    pub sign_with_lab: bool,
```
fnando, line 41: also verbose. Should we add `-l` for signing with the lab?

cmd/soroban-cli/src/config/sign_with.rs
willemneal: We have prematurely added short flags in the past and who is to say that `-s` could mean sign with ledger?

cmd/crates/soroban-test/tests/it/integration/tx.rs
```diff
@@ -49,3 +50,47 @@ async fn txn_hash() {
...
+    let sandbox = &TestEnv::new();
+    sandbox
+        .new_assert_cmd("contract")
+        .arg("install")
+        .args(["--wasm", HELLO_WORLD.path().as_os_str().to_str().unwrap()])
+        .assert()
+        .success();
+
+    let xdr_base64 = deploy_contract(sandbox, HELLO_WORLD, DeployKind::SimOnly).await;
+    println!("{xdr_base64}");
```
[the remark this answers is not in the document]
willemneal: By default prints are suppressed so this is just if you are interested in the output if these test. At the time it was just for debugging purposes.

cmd/soroban-cli/src/config/sign_with.rs
willemneal: As I said it the other comment we can always add this later if it is requested. We could also add an alias `--with-lab`. I do think it's okay to be a little verbose as it makes the scripts more readable and you don't have to go look up the short flags if you forget.

willemneal:
@leighmcculloch Thanks for the clean up. I do think it's ready to be merged and we can iterate when the demands change.

cmd/soroban-cli/src/config/sign_with.rs
leighmcculloch: We can always add short flags later if it turns out folks really need it. Lets start with self describing commands and see how they are adopted.

Alternatively we could reduce it to `--key`.

Originally when we were adding the `--sign-with-*` we were thinking about how to introduce the sign options on other commands. But now it seems we are only doing custom signing on this command so `--sign-with-*` does seem repetitive given the command is called `tx sign`.

leighmcculloch: I'm removing the suggestion to go with `--key`. Going with `--sign-with-key` gives us the option to add that same optino to other commands in the future, so I think we should stick with it. We can add aliases and shorten it later if we decide this was the wrong call.

cmd/soroban-cli/src/signer.rs
```diff
@@ -189,29 +192,72 @@ fn sign_soroban_authorization_entry(
...
-pub fn hash(tx: &Transaction, network_passphrase: &str) -> Result<[u8; 32], xdr::Error> {
-    let signature_payload = TransactionSignaturePayload {
-        network_id: Hash(Sha256::digest(network_passphrase).into()),
-        tagged_transaction: TransactionSignaturePayloadTaggedTransaction::Tx(tx.clone()),
-    };
-    Ok(Sha256::digest(signature_payload.to_xdr(Limits::none())?).into())
+pub struct LocalKey {
+    key: ed25519_dalek::SigningKey,
+    #[allow(dead_code)]
+    prompt: bool,
```
Ifropc, line 248: Do we not intend to use prompt property here for now?

Ifropc (approved):
Like the new PR, very compact and simple!

cmd/soroban-cli/src/signer.rs
leighmcculloch: It'll be used in a future change. We could remove it now though. I'll remove itand we can bring it back.

PR #1599, opened 2024-09-11
Fix/cookbook no fund

BlaineHeffron:
### What

`generate` command should have `--no-fund` since we will be moving to that being the default in v22

### Why

Make documentation consistent with intended usage of the CLI

### Known limitations

N/A

### DIFF
--- a/cookbook/contract-lifecycle.mdx
+++ b/cookbook/contract-lifecycle.mdx
@@ -1,7 +1,7 @@
 ---
 title: Contract Lifecycle
 hide_table_of_contents: true
-description: Manage the lifecycle of a Stellar smart contract using the CLI.
+description: Manage the lifecycle of a Stellar smart contract using the CLI
 ---
 
 To manage the lifecycle of a Stellar smart contract using the CLI, follow these steps:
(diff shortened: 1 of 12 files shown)


cookbook/contract-lifecycle.mdx
```diff
@@ -1,15 +1,15 @@
...
+description: Manage the lifecycle of a Stellar smart contract using the CLI
 ---
 
 To manage the lifecycle of a Stellar smart contract using the CLI, follow these steps:
 
 1. Create an identity for Alice:
 
 ```bash
-stellar keys generate alice
+stellar keys generate alice --no-fund
```
leighmcculloch, line 12: Seems a bit odd to me to push this change where we no fund, then fund immediately after. Nothing appears to be broken in the cookbook without this change, what's the motivation for pushing it out now?

BlaineHeffron: No it works fine, I had originally been asked to add `--no-fund` and explicitly fund it (when I first wrote the guide I just called `generate`) but when I was testing I had to remove `--no-fund` because it wasn't working. Anyway, I added it back because Nando was confused as to why we were calling `fund` afterwards, so this was just adding it back.. but you are right its a bit silly to put it in if this will be the default behavior soon and its already working without it.

PR #1616, opened 2024-09-18
Drop `--rpc-url` and `--network-passphrase`.

### LINKED ISSUE #1610: Use alias for network
Simplification of the UI to use network alias everywhere

Simplify network config on stellar-cli: drop --rpc-url and --network-passphrase and require a network alias everywhere

fnando:
### What

Close #1610

### Why

To simplify our codebase and improve the CLI usability by having only one way of setting the network.

### Known limitations

N/A


fnando:
@leighmcculloch only soroban e2e tests are failing. I believe this requires changing https://github.com/stellar/system-test, right? What's the process? First I update stellar/system-test then merge this, or is it the other way around?

leighmcculloch:
> @leighmcculloch only soroban e2e tests are failing. I believe this requires changing [stellar/system-test](https://github.com/stellar/system-test), right? What's the process? First I update stellar/system-test then merge this, or is it the other way around?

We don't have a process. We haven't broken it much. 🫢

FULL_HELP_DOCS.md
```diff
@@ -116,8 +116,6 @@ Get Id of builtin Soroban Asset Contract. Deprecated, use `stellar contract id a
 ###### **Options:**
 
 * `--asset <ASSET>` — ID of the Stellar classic asset to wrap, e.g. "USDC:G...5"
-* `--rpc-url <RPC_URL>` — RPC server endpoint
-* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
```
leighmcculloch, line 120: Something I'm wondering about is folks might be using these in scripts, or with env vars, so that they can interact with specific RPC in a way that doesn't rely on stateful changes.

fnando: fair point. I wonder if it's worth all the complexity on our codebase due to how command flattening happens with clap though. 🤔 

Any suggestions? Should we abandon this change?

cmd/soroban-cli/src/config/network.rs
```diff
@@ -178,31 +140,50 @@ impl Network {
...
+                rpc_url: "https://soroban-testnet.stellar.org".to_string(),
+                network_passphrase: passphrase::TESTNET.to_string(),
+            },
+        );
+
+        map.insert(
+            "mainnet".to_string(),
+            Network {
+                name: "mainnet".to_string(),
+                rpc_url:
```
Ifropc, line 180: Can we print RPC url in the network resolver to get a better error message?
```
$ stellar contract info interface --id CBBJTE34KZHJACB2A6OXCRTDDLX4LI7GKYYD6XYJSGMX46BLNAG752HS --network mainnet
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/stellar contract info interface --id CBBJTE34KZHJACB2A6OXCRTDDLX4LI7GKYYD6XYJSGMX46BLNAG752HS --network mainnet`
❌ error: invalid rpc url: invalid uri character
# should be 
error: invalid rpc url (Bring Your Own: https://developers.stellar.org/docs/data/rpc/rpc-providers): invalid uri character 
```

Ifropc (commented):
Not related to the PR, but should we add 
`network info <name>` command to list network passhprase and RPC? Currently, the flow is to do `rm mainnet && add mainnet --rpc-url <my-url> --passphrase` and it'd be easier to see the passphrase of mainnet in the CLI

cmd/soroban-cli/src/config/network.rs
fnando: This error comes from soroban rpc client, so we'd need to change the upstream dependency.

PR #1618, opened 2024-09-24
Feat/add rpc header arg

elizabethengelman:
### What

Adds a `--rpc-headers` argument to allow users to pass in an auth header, so that RPC providers that require API keys can be used with the CLI.

Companion rpc-client PR: https://github.com/stellar/rs-stellar-rpc-client/pull/13

Here are some screenshots showing that this change works for multiple headers, with both adding a new network, and using a configured rpc provider:

#### using multiple `—rpc-headers` args for adding a network
Screenshot 2024-09-30 at 10 08 47 AM

#### using env vars for multiple headers for adding a network
Screenshot 2024-09-30 at 10 09 53 AM

#### using multiple —rpc-headers args for using the rpc provider
Screenshot 2024-09-30 at 10 14 52 AM

#### using the env var for multiple headers for using the rpc provider
Screenshot 2024-09-30 at 10 18 04 AM

### Why

This allows users to use one of the rpc providers that require an API passed in via a header.

Closes https://github.com/stellar/stellar-cli/issues/1583

### Known limitations

This change requires this change from the rpc-client: https://github.com/stellar/rs-stellar-rpc-client/pull/13


cmd/soroban-cli/src/config/network.rs
```diff
@@ -47,6 +47,14 @@ pub struct Args {
...
     )]
     pub rpc_url: Option<String>,
+    /// Optional RPC provider api key headers
+    #[arg(
+            long = "rpc-header",
+            requires = "rpc_url",
+            env = "STELLAR_RPC_HEADER",
+            help_heading = HEADING_RPC,
+        )]
+    pub rpc_header: Option<String>,
```
leighmcculloch, line 57: I think this should probably be a multi-value key, therefore be a `Vec<String>`, but as far as I know the clap-rs `env` feature doesn't work well with it. We'd have to try it out.

Given all use cases we have only need a single header at this time, then as it is looks good, but if you wanted to do a mini timeboxed exploration on whether `Vec<String>` is supported with `env` in some way, and it turned out it worked, then I'd change it to:

```suggestion
    /// Optional RPC provider api key headers
    #[arg(
            long = "rpc-header",
            num_args(0..),
            requires = "rpc_url",
            env = "STELLAR_RPC_HEADERS",
            help_heading = HEADING_RPC,
        )]
    pub rpc_header: Vec<String>,
```

cmd/soroban-cli/src/config/network.rs
```diff
@@ -47,6 +47,14 @@ pub struct Args {
         help_heading = HEADING_RPC,
     )]
     pub rpc_url: Option<String>,
+    /// Optional RPC provider api key headers
```
leighmcculloch, line 50:
```suggestion
    /// Optional header (e.g. API Key) to include in requests to the RPC
```

cmd/soroban-cli/src/config/network.rs
```diff
@@ -97,6 +106,13 @@ pub struct Network {
         help_heading = HEADING_RPC,
     )]
     pub rpc_url: String,
+    /// Optional RPC provider api key header
```
leighmcculloch, line 109:
```suggestion
    /// Optional header (e.g. API Key) to include in requests to the RPC
```

leighmcculloch (commented):
Couple thoughts.

cmd/soroban-cli/src/config/network.rs
elizabethengelman: Just re-reading your comment, and I missed the `env` part the first time. I'll timebox looking into this today 👌

cmd/soroban-cli/src/config/network.rs
```diff
@@ -47,6 +47,14 @@ pub struct Args {
...
     )]
     pub rpc_url: Option<String>,
+    /// Optional header (e.g. API Key) to include in requests to the RPC
+    #[arg(
+            long = "rpc-header",
+            requires = "rpc_url",
+            env = "STELLAR_RPC_HEADER",
+            help_heading = HEADING_RPC,
+        )]
+    pub rpc_header: Option<String>,
```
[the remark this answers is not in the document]
elizabethengelman: Thanks @overcat! I agree that `rpc_header` should be a vec, and I like that you made it a vec of tuples instead! 👌 That definitely makes it easier to handle `STELLAR_RPC_HEADER`. 🎉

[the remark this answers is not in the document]
leighmcculloch: Does that work with env vars though? I don't think the clap env feature knows what to do with multiple values inside the env. Or can you share an example of what hte cli and env UX becomes with this?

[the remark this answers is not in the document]
overcat: Hi @leighmcculloch, I think it is supported, please check the screenshot.

image

cmd/soroban-cli/src/config/network.rs
```diff
@@ -97,6 +116,17 @@ pub struct Network {
...
     )]
     pub rpc_url: String,
+    /// Optional header (e.g. API Key) to include in requests to the RPC
+    #[arg(
+        long = "rpc-header",
+        env = "STELLAR_RPC_HEADER",
+        help_heading = HEADING_RPC,
+        num_args = 1,
+        action = clap::ArgAction::Append,
+        value_delimiter = ',',
```
overcat, line 126: For more general purposes, we may need to consider whether we should use `,` as a separator, because , is likely to appear in the value. For example, my User-Agent contains it: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Safari/537.36`

leighmcculloch: I like the idea of using new line as you did in the other example, given that new lines are the natural delimiter. New lines may create issues in some environments for env vars, but also, we expect only a single header to be used so I'm not sure it's worth dwelling on that.

cmd/soroban-cli/src/config/network.rs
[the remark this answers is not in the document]
elizabethengelman: Adding to this, I've put screenshots in the pr description showing that multiple headers work for adding a network and using the configured rpc provider.

cmd/soroban-cli/src/config/network.rs
```diff
@@ -106,6 +136,24 @@ pub struct Network {
...
+fn parse_http_header(header: &str) -> Result<(String, String), Error> {
+    let header_components = header.split(':').collect::<Vec<&str>>();
+    if header_components.len() != 2 {
+        return Err(Error::InvalidHeader(format!(
+            "Missing a header name and/or value: {header}"
+        )));
+    }
+
+    let key = header_components[0].trim().to_string();
+    let value = header_components[1].trim().to_string();
```
leighmcculloch, line 148: I think technically `:` can appear in the header value, so splitting and exploding that into a Vec on every `:` is probably not a resilient, as splitting on the first `:` and otherwise keeping the rest of the header intact.

It should be a small change to something like:

```suggestion
    let header_components = header.split(':');
    let Some(key) = header_components.next() else {
        return Err(Error::InvalidHeader(format!(
            "Missing a header name and/or value: {header}"
        )));
    };
    let value = header_components.remaining();
```

cmd/soroban-cli/src/rpc_client.rs
```diff
@@ -0,0 +1,131 @@
...
+        &self.client
+    }
+}
+
+// implementing DerefMut for mutable access
+impl DerefMut for RpcClient {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.client
+    }
+}
```
leighmcculloch, line 52: The use of Deref to sub one type another one is misusing the Deref trait a bit. It is a pattern used at times, but also considered an antipattern if the benefits don't truly outweigh the cons. There's some conversation about it here: https://rust-unofficial.github.io/patterns/anti_patterns/deref.html

I don't think we need to use this pattern here. Can I suggest we instead do something a little more boring, and replace this:
```diff
-    let client = Client::new(&network.rpc_url)?;
+    let client = RpcClient::new(network)?;
```
with:
```diff
-    let client = Client::new(&network.rpc_url)?;
+    let client = network.rpc_client()?;
```
where `rpc_client` is a function on `Network` that returns an `rpc::Client` and constructs it from the necessary parameters?

leighmcculloch (commented):
Couple requests inline.

cmd/soroban-cli/src/rpc_client.rs
elizabethengelman: Good to know that using Deref like this is an antipattern. Thanks for the resource!

And I like this refactor - definitely simpler. I'll get a change pushed up shortly.

cmd/soroban-cli/src/config/network.rs
elizabethengelman: That makes sense to me - the one piece that I'm noticing is that `remainder` is only available with the unstable nightly build at the moment. Though, I think we can get a similar effect by just using `next` again. 
https://doc.rust-lang.org/std/str/struct.Split.html#impl-Split%3C'a,+P%3E

cmd/soroban-cli/src/config/network.rs
```diff
@@ -44,6 +53,17 @@ pub struct Args {
         help_heading = HEADING_RPC,
     )]
     pub rpc_url: Option<String>,
+    /// RPC Header(s) to include in requests to the RPC provider
+    #[arg(
+        long = "rpc-header",
```
leighmcculloch, line 58: @elizabethengelman while trying out the pr locally I made this tweak, I renamed the option `--rpc-header` (no plural) so that at the command line it's more obvious that using the option is intended to be a single arg per use, so it looks like this:

```
stellar ... --rpc-header A:B --rpc-header C:D
```

So that folks don't think they can do:

```
stellar ... --rpc-headers A:B C:D
```

leighmcculloch (approved):
I think this is good 👍🏻. Thanks for navigating both the rpc and cli changes. I've merged the rpc change, and bumped the ref in this pr to match it so this should be good to merge.

@elizabethengelman a couple minor asks for a follow up:

1️⃣ The error messages when passing invalid values of keys / values to the header could do a better job of helping someone know what to correct.

For example, it's easy to naively think you can just plop in an api key:

```
$ stellar network add lfg --rpc-header myapikeyyay
error: invalid value 'myapikeyyay' for '--rpc-header <RPC_HEADERS>': invalid header: Missing header value: myapikeyyay

```

I understand what the error above needs of me, but I think devs without context won't. 

I think we should tell them that the header should be in the format `key: value`.

2️⃣ White space could be ignored.

For example, it's common for folks to state headers with whitespace around the `:`:

```
$ stellar network add lfg \
    --rpc-url http://example.org \
    --network-passphrase "Example Network" \
    --rpc-header 'Authorization: myapikeyyay'
```

Which renders later with the whitespace retained:

```
❯ stellar network ls -l
Local "/Users/lfg/.soroban/network/lfg.toml"
Name: lfg
Network {
    rpc_url: "http://example.org",
    rpc_headers: [
        (
            "Authorization",
            " myapikeyyay",
        ),
    ],
    network_passphrase: "Example Network",
}
```

cmd/soroban-cli/src/config/network.rs
leighmcculloch: Ah perfect, that's great 👏🏻.

cmd/soroban-cli/src/config/network.rs
```diff
@@ -103,6 +136,21 @@ pub struct Network {
     pub network_passphrase: String,
 }
 
+fn parse_http_header(header: &str) -> Result<(String, String), Error> {
+    let header_components = header.splitn(2, ':');
+
+    let (key, value) = header_components
+        .map(str::trim)
+        .next_tuple()
+        .ok_or_else(|| Error::InvalidHeader)?;
```
leighmcculloch, line 145: I pushed a change to trim values, and to tweak the error message. The error generated by this line looks like this now:
```
$ stellar network add lfg \
    --rpc-url http://example.org \
    --network-passphrase "Example Network" \
    --rpc-header 'Authorization'
error: invalid value 'Authorization' for '--rpc-header <RPC_HEADERS>': invalid HTTP header: must be in the form 'key:value'
```

I removed passing in the string into the error, because the clap error handler will print out the value they passed in at the start of the error message anyway.

cmd/soroban-cli/src/config/network.rs
elizabethengelman: sounds good, thanks for making that update!

PR #1619, opened 2024-09-24
feat: allow source account to be a public key

willemneal:
### What

Allow public keys to be used with `--source`.

### Why

This is will be useful for `--sign-with-lab` which users will likely copy their public keys from their wallet to use on the CLI. It will also be useful when we expand `--sign-with-*` more generally allowing a different key than the source account to sign the transaction.

### Known limitations

Currently this would fail if not `--build-only` since we rely on source account being a reference to a private key or seed


leighmcculloch (approved):
Suggestion inline, otherwise lgtm.

PR #1620, opened 2024-09-24
fix: bug in tx sign and send test

willemneal:
### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/crates/soroban-test/tests/it/integration/tx.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/tx.rs
@@ -74,12 +74,11 @@ async fn build_simulate_sign_send() {
         .stdout_as_str();
     dbg!("{tx_signed}");
 
-    let output = sandbox
+    sandbox
         .new_assert_cmd("tx")
         .arg("send")
         .write_stdin(tx_signed.as_bytes())
         .assert()
         .success()
-        .stdout_as_str();
-    assert_eq!(output, "SUCCESS");
+        .stdout(predicates::str::contains("SUCCESS"));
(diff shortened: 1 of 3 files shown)


leighmcculloch (commented):
There's one issue with tests it appears. We need to change the min rust version.

cmd/crates/soroban-test/tests/it/integration/tx.rs
```diff
@@ -74,12 +74,11 @@ async fn build_simulate_sign_send() {
...
-    let output = sandbox
+    sandbox
         .new_assert_cmd("tx")
         .arg("send")
         .write_stdin(tx_signed.as_bytes())
         .assert()
         .success()
-        .stdout_as_str();
-    assert_eq!(output, "SUCCESS");
+        .stdout(predicates::str::contains("SUCCESS"));
```
leighmcculloch, line 83: I'm not even sure this works, does SUCCESS get outputted? I added the other assertion as a placeholder and never came back to update it. We should probably make the tx send output more useful than just SUCCESS, such as outputting some print emoji information.

If this works though we can merge.

leighmcculloch (approved):

cmd/crates/soroban-test/tests/it/integration/tx.rs
willemneal: Yeah as is the `send` returns the `GetTransactionResponse` which should definitely include `SUCCESS`, however, I agree that it should do more to check that the transaction had the effect it was supposed to have.

PR #1625, opened 2024-09-25
Update contract init

Ifropc:
### What

Update `contract init` implementation. 
Now it uses workspace template as the project root and also adds `--name` for easier contract initialization. It's now possible to initialize multiple contracts from a hello world template with passing different `--name`'s

### Why

Changes according to https://github.com/stellar/stellar-cli/issues/1586

### Known limitations

N/A


FULL_HELP_DOCS.md
```diff
@@ -611,7 +611,9 @@ Outputs no data when no data is present in the contract.
 
 ## `stellar contract init`
 
-Initialize a Soroban project with an example contract
+Initialize a Soroban contract.
+
+When running with empty or non-existent `--project-path`, this command will generate a template Cargo workspace project and add a sample contract package. When running in the existing Cargo project, it will add a new package for a sample contract with a given `--name`.
```
fnando, line 616: isn't the project path positional? You don't actually have `--project-path`.

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -1,85 +1,76 @@
...
-    "An optional flag to specify Soroban example contracts to include. A hello-world contract will be included by default.";
+use crate::commands::contract::init::Error::{
+    AlreadyExists, PathExistsNotCargoProject, PathExistsNotDir,
+};
+use crate::{commands::global, print};
 
 #[derive(Parser, Debug, Clone)]
 #[group(skip)]
 pub struct Cmd {
     pub project_path: String,
```
fnando, line 19: I think this needs to be updated to `Option<project_path>`. This will require either `project_path` or `--name` to be provided; if you run the command like `stellar contract init`, it should error.

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -1,85 +1,76 @@
...
+    pub name: String,
 
+    // TODO: remove in 23.0
     #[arg(
+        short,
         long,
-        default_value = "",
-        long_help = "An optional flag to pass in a url for a frontend template repository."
+        action = clap::ArgAction::HelpLong,
+        long_help = "This argument has been deprecated and will be removed in the future versions of CLI. You can still clone examples from the repo https://github.com/stellar/soroban-examples",
```
fnando, line 33: Given that these deprecated switches are in fact noop, I think we should mention that the switch doesn't do anything anymore. The way is phrased right now gives the impression it worked, but won't in the future.

cmd/soroban-cli/src/commands/contract/init.rs
```diff
@@ -105,656 +101,146 @@ struct Runner {
...
-            let contract_path = Path::new(&contract_as_string);
-            let from_contract_path = from.join(contract_path);
-            let to_contract_path = project_contracts_path.join(contract_path);
-            Self::create_dir_all(&to_contract_path)?;
-
-            self.copy_contents(&from_contract_path, &to_contract_path)?;
-            Self::edit_contract_cargo_file(&to_contract_path)?;
+        if project_path.exists() {
+            return Err(AlreadyExists(self.args.name.clone()));
         }
```
fnando, line 163: I think we should keep allowing users to overwrite a project. It's a great way to upgrade all your project, because you can undo undesired changes with git, while keeping everything else.

Ifropc: The behavior has changed, so I don't think this flag is needed anymore

cmd/soroban-cli/src/commands/contract/init.rs
leighmcculloch: Isn't project path always required? Name is just specifying the name of the contract in the workspace, but presumably we still need to tell the tool where the workspace is.

cmd/soroban-cli/src/commands/contract/init.rs
leighmcculloch: If the switches don't do anything anymore, we should remove them so folks get a hard error if they try to use it.

Ifropc:
I'm converting this PR back to draft, I think my original implementation is not what we would like to move forward with

Ifropc:
After internal discussion I changed the implementation to keep the current behavior that switches template to the workspace, and make name of generated contract customizable via `--name`

leighmcculloch (approved):

PR #1628, opened 2024-09-25
Contract init: add flag removal message

Ifropc:
### What

We plan to remove some flags from `contract init` as part of breaking changes (see https://github.com/stellar/stellar-cli/issues/1586). This PR does just that.
Removed argument are hidden from help:
out

And  when trying to use them there's an error that explains what are the alternatives.
```
error: invalid value 'test' for '--with-example <WITH_EXAMPLE>': This argument has been removed and will be not be recognized by the future versions of CLI: Adding examples via cli is no longer supported. You can still clone examples from the repo https://github.com/stellar/soroban-examples

```

### Why

See above

### Known limitations

N/A


cmd/soroban-cli/src/utils.rs
```diff
@@ -161,6 +161,41 @@ pub fn get_name_from_stellar_asset_contract_storage(storage: &ScMap) -> Option<S
     }
 }
 
+pub mod http {
```
fnando, line 164: Was this added by accident?

Ifropc: Hmm no, I think you are looking at the merge commit. It's not in the [diff](https://github.com/stellar/stellar-cli/pull/1628/files#diff-643bf04585e9188dfeb0f09377be2943d1b4ed23326b6ff31344035a9e7ef592)

cmd/soroban-cli/src/commands/network/container/shared.rs
```diff
@@ -157,7 +157,7 @@ fn try_docker_desktop_socket(
         &default_docker_desktop_host,
         DEFAULT_TIMEOUT,
         API_DEFAULT_VERSION,
-    ).map_err(|e| {
+    ).inspect_err(|_| {
```
elizabethengelman, line 160: 👍

elizabethengelman (approved):
This lgtm! 👍 

One additional thing that we can do as part of this PR is to remove the `gix` dependency - I think it was only being used for this command.

elizabethengelman:
Also, I changed it in https://github.com/stellar/stellar-cli/pull/1660 (which will not be necessary after this PR), but I think we'll want to move the `init` integration test into the integration dir. I think as it is now, it is not running.

Cargo.lock
Ifropc, line 1: There are a lot of changes coming from re-generated lockfile, but I assume it should not be a problem (especially considering we are releasing 22.0 🤔 )

cmd/soroban-cli/src/utils.rs
```diff
@@ -196,6 +196,26 @@ pub mod http {
...
+pub mod args {
+    #[derive(thiserror::Error, Debug)]
+    pub enum DeprecatedError<'a> {
+        #[error("This argument has been removed and will be not be recognized by the future versions of CLI: {0}"
+        )]
+        RemovedArgument(&'a str),
+    }
+
+    #[macro_export]
+    macro_rules! removed_arg {
```
leighmcculloch, line 208: I think we should make this function a little more explicit as to it's intended effect, so it's clear it isn't going to pass the value, but it's a way to display a meaningful message and then error.
```suggestion
    macro_rules! error_on_use_of_removed_arg {
```

leighmcculloch:
> There are a lot of changes coming from re-generated lockfile, but I assume it should not be a problem (especially considering we are releasing 22.0 🤔 )

@Ifropc I pushed a change reverting the Cargo.lock file, and then applying the minimal changes. It reduced the changes to the lock file from 1871 lines, to 899 lines. You may have done an update of all crates accidentally.

leighmcculloch (approved):

leighmcculloch:
This change appears to be missing the `--name` option that's mentioned in https://github.com/stellar/stellar-cli/issues/1586. Is that coming in a separate PR?

Ifropc:
> It reduced the changes to the lock file from 1871 lines, to 899 lines. You may have done an update of all crates accidentally.

Thank you! This is strange, because I got lockfile from main and then just ran `cargo build` (nothing else 🤔 )

PR #1636, opened 2024-09-27
Deprecate contract inspect

Ifropc:
### What

Add deprecate notice to `contract inspect` command

### Why

https://github.com/stellar/stellar-cli/issues/1528

We should add deprecate notice first before committing to removing the command

### Known limitations

`info` technically doesn't support `xdr-base64-array` which `inspect` does. This shouldn't be an issue once we bring support of XDR streams to our JS tooling (this is already supported by RS tools)
Alternatively, we can bring  `xdr-base64-array` to `info` command


willemneal (approved):

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -56,7 +56,8 @@ pub enum Cmd {
     /// Initialize a Soroban project with an example contract
     Init(init::Cmd),
 
-    /// Inspect a WASM file listing contract functions, meta, etc
+    /// (Deprecated) Inspect a WASM file listing contract functions, meta, etc
```
willemneal, line 59:
```suggestion
    /// (Deprecated in favor of `contract info` subcommands) Inspect a WASM file listing contract functions, meta, etc
```

fnando (approved):

PR #1637, opened 2024-09-27
Add warning on no fund flag

Ifropc:
### What

Add deprecation warning when running `keys generate` with no `--fund` flag

### Why

We plan to flip the logic of this command in https://github.com/stellar/stellar-cli/issues/1407 
As part of this change, we should add a warning for people to add the new flag, if it's necessary for them to fund the key.  

### Known limitations

N/A


willemneal (approved):
LGTM! Just one small nit!

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -1,3 +1,5 @@
+use crate::commands::global;
+use crate::print::Print;
 use clap::{arg, command};
```
willemneal, line 4: We try to keep external deps at the top.

```suggestion
use clap::{arg, command};

use crate::{commands::global, print::Print};
```

leighmcculloch (approved):

PR #1638, opened 2024-09-30
Chore/use rich printer for network container

elizabethengelman:
### What

Uses the rich printer for `network container` commands.

### Why

Closes https://github.com/stellar/stellar-cli/issues/1617
### Known limitations
 
N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/network/container/logs.rs
+++ b/cmd/soroban-cli/src/commands/network/container/logs.rs
@@ -27,9 +27,9 @@ pub struct Cmd {
 
 impl Cmd {
     pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
-        let printer = print::Print::new(global_args.quiet);
+        let print = print::Print::new(global_args.quiet);
         let container_name = Name(self.name.clone()).get_internal_container_name();
-        let docker = self.container_args.connect_to_docker(&printer).await?;
+        let docker = self.container_args.connect_to_docker(&print).await?;
         let logs_stream = &mut docker.logs(
             &container_name,
             Some(bollard::container::LogsOptions {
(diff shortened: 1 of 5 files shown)


cmd/soroban-cli/src/commands/network/container/logs.rs
```diff
@@ -23,9 +26,10 @@ pub struct Cmd {
 }
 
 impl Cmd {
-    pub async fn run(&self) -> Result<(), Error> {
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let printer = print::Print::new(global_args.quiet);
```
fnando, line 30: Can we name this `print` instead, because I've seen that's how you named it in another file (`cmd/soroban-cli/src/commands/network/container/stop.rs`).

I think `print` was also the name I've used in other places.

elizabethengelman: Yeah for sure! I'll get a change pushed up shortly.

fnando, line 30: one last occurrence. :)

```suggestion
        let print = print::Print::new(global_args.quiet);
```

cmd/soroban-cli/src/commands/network/container/logs.rs
```diff
@@ -23,9 +26,10 @@ pub struct Cmd {
...
         let container_name = Name(self.name.clone()).get_internal_container_name();
-        let docker = self.container_args.connect_to_docker().await?;
+        let docker = self.container_args.connect_to_docker(&printer).await?;
```
fnando, line 32:
```suggestion
        let docker = self.container_args.connect_to_docker(&print).await?;
```

cmd/soroban-cli/src/commands/network/container/logs.rs
elizabethengelman: Ah, how did i miss that?! Thanks! :)

fnando (approved):

PR #1639, opened 2024-09-30
Durability bump issue  # ODHACK

KoxyG:
### What
The issue was related to the required durability when bumping/extending a contract. I removed the durability field from the argument struct in the "soroban-cli/command/key" file. As a result, the durability flag will no longer be required when running the command.

### Why
This change was made to resolve the issue of requiring the durability flag during contract extension. Additionally, I implemented default persistent storage, so the durability flag is no longer necessary.

### Known limitations
There are currently no known limitations.


KoxyG:
Hi @janewang i saw you  reviewed my  pr. I also saw one check failed. 

Am i missing an installation process?. Also, is there any issue with my pr?

janewang:
@KoxyG I just kicked off the tests. Some integration tests are failing

leighmcculloch (changes requested):
Hi @KoxyG, thanks for contributing. 

The durability option needs to be retained. The issue is that it's default value is not functioning.

willemneal:
I would change this from `clap::ValueEnum` to `Default`, with `#[default]` above `Persistent`. Then change the arg to be an option and not use the clap's default value. Then you can use `.unwrap_or_default()`.  TL;DR don't rely on clap for the default value.

https://github.com/stellar/stellar-cli/blob/aad9be9708037e122e8e9ff09a67a09c26168513/cmd/soroban-cli/src/commands/contract/mod.rs#L162-L168

leighmcculloch:
@willemneal If the enum is derive `Default`, then it won't be necessary to make the value an `Option<>`.

KoxyG:
> Hi @KoxyG, thanks for contributing.
> 
> The durability option needs to be retained. The issue is that it's default value is not functioning.

So, should i change it back to the initial state?. The  default value is persistent

KoxyG:
The initial issue is that, when the command `soroban contract bump --id C... --ledgers-to-expire 1000` is used. It doesn't work unless the --durability persistent is used along side with the command.

willemneal:
@leighmcculloch @KoxyG, can either of you reproduce the original issue? I updated the tests to rely on a default durability and it is passing: https://github.com/stellar/stellar-cli/pull/1644

KoxyG:
> @leighmcculloch @KoxyG, can either of you reproduce the original issue? I updated the tests to rely on a default durability and it is passing: #1644

You updated the default durability as  only "persistent"?

willemneal:
No just updated the tests to test if the default was working: https://github.com/stellar/stellar-cli/pull/1644/files

So I'm questioning if this PR is needed.  If you deploy a contract to testnet and then use extend are you required to use the `--durability`? When I try it locally and in the test linked above the default seems to be working so perhaps this issue was fixed a while ago. It is from last November so my guess is that the default didn't exist then and was added without closing that issue.

KoxyG:
Hi @janewang I noticed you reviewed my pull request. Do i need to work on anything froom my end?

Issue #1615, opened 2024-09-18
Render events as json in stellar events pretty output instead of as rust debug


leighmcculloch:
Events in pretty print mode print as Rust debug format.

We should print as JSON for the same reasons as, because developers are more familiar with the JSON format than the Rust debug format:
- https://github.com/stellar/stellar-cli/pull/1600

Example:
```
$ stellar contract event --ledger 3423
...
Event 0000014701673062400-0000000000 [DIAGNOSTIC]:
  Ledger:   3423 (closed at 2024-09-17T22:20:30Z)
  Contract: 
  Topics:
            Symbol(ScSymbol(StringM(fn_call)))
            Bytes(ScBytes(BytesM(2ab7f467d192fcb2aefa138e0aae1f596216692d5e3d631a2c2ee81e263ab83f)))
            Symbol(ScSymbol(StringM(set_price)))
  Value: Vec(Some(ScVec(VecM([Vec(Some(ScVec(VecM([I128(Int128Parts { hi: 0, lo: 0 }), I128(Int128Parts { hi: 0, lo: 62622635909 }), I128(Int128Parts { hi: 0, lo: 99750087794911 }), I128(Int128Parts { hi: 0, lo: 1498490882301 }), I128(Int128Parts { hi: 0, lo: 199315054592 }), I128(Int128Parts { hi: 0, lo: 80315405446 }), I128(Int128Parts { hi: 0, lo: 110532281494896 })])))), U64(1726611600000)]))))
...
```

Gerson2102:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/stellar-cli).

### My background and how it can be leveraged
---

### Hey! I'm Gerson. 👋

Member of **Dojo Coding**. I have contributed before to other Rust projects like:

- `Cairo Native`
- `kakarot-rpc`
- `ef-tests`
- `Madara`

Can I work on this issue?

Check my OnlyDust profile, I've contributed to many projects: [Profile](https://app.onlydust.com/u/Gerson2102)

---



### How I plan on tackling this issue
### Plan to Solve the Issue:

1. **Understand the task** by carefully reviewing the issue details.
2. If I encounter new concepts or things I don't know, I will **consult the Rust book** for guidance.
3. I will also **conduct my own research** to find the best way to implement the solution.
4. If I get stuck, I will **ask questions ASAP** to solve the issue efficiently.
5. Once the work is completed, I'll **open a PR** and **wait for your feedback**.

---

[...]

PR #1642, opened 2024-10-02
Render Stellar Events as JSON Instead of Rust Debug Format and Update WASM Path

juandiegocv27:
### What

This PR modifies the Stellar events output to render as JSON instead of using the Rust debug format. Additionally, it updates the path in `boilerplate.rs` for including the WASM file.

### Why

The change is being made because developers are more familiar with the JSON format, which provides a standardized way of viewing event data. It aligns with Render events as json in stellar events pretty output instead of as rust debug issue, aiming to improve the readability and usability of the event logs for developers.

- closes: #1615

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -85,7 +85,7 @@ Tools for smart contract developers
 * `id` — Generate the contract id for a given contract or asset
 * `info` — Access info about contracts
 * `init` — Initialize a Soroban project with an example contract
-* `inspect` — Inspect a WASM file listing contract functions, meta, etc
+* `inspect` — (Deprecated in favor of `contract info` subcommands) Inspect a WASM file listing contract functions, meta, etc
 * `install` — Install a WASM file to the ledger without creating a contract instance
 * `invoke` — Invoke a contract function
 * `optimize` — Optimize a WASM file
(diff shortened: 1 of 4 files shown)


juandiegocv27:
ready for review: @leighmcculloch

cmd/soroban-cli/src/log/event.rs
```diff
@@ -15,12 +13,13 @@ pub fn events(events: &[xdr::DiagnosticEvent]) {
...
-        let xdr = event.to_xdr_base64(xdr::Limits::none()).unwrap();
+        // Convert the event to JSON
         let json = serde_json::to_string(event).unwrap();
+
         if span.metadata().unwrap().level() == &Level::INFO {
-            info!("{i}: {xdr} {json}");
+            info!("{i}: {json}");
         } else {
-            debug!("{i}: {xdr} {json}");
+            debug!("{i}: {json}");
```
leighmcculloch, line 22: This output is part of the logging and it was just recently written to output both the XDR and the JSON in:
- #1600

The issue being addressed by this PR is to change the stdout output of the `stellar contract events` subcommand, not the log output. If you see the original issue the original issue has an example of what the command is to run, and what the output looks like today. It's that output that the issue describes being changed from Rust debug format to JSON.

leighmcculloch (changes requested):
Hi @juandiegocv27, thanks for contributing.

See my comments inline.

juandiegocv27:
> Hi @juandiegocv27, thanks for contributing.
> 
> 
> 
> See my comments inline.

thank you for the feedback, I'll fix the issues 🫡

juandiegocv27:
@leighmcculloch make the changes and implement some new code, please review and let me know

juandiegocv27:
tell what to fix to do solve the problem as fast as possible or if its ready to merge

PR #1646, opened 2024-10-02
Update xdr, env, sdk, strkey deps to compat with 22.0.0-rc.1 stream

leighmcculloch:
### What
Update xdr, env, sdk, strkey deps to compat with 22.0.0-rc.1 stream.

Also remove the env as a direct dependency.

### Why
Routine, so we can bring in more upstream work on protocol 22 to start testing, etc.

The additional change to remove env as a direct dependency came about because I realised that it and the sdk are in many places an artificial dependency that don't actually serve a purpose. Once upon a time the env was a critical dependency because simulation actually occurred within the cli itself, but that was a long time ago and is no longer.


fnando (approved):

leighmcculloch:
@fnando The rpc tests are failing because the rpc isn't updated yet, but all the test-wasms in this repo build with the same protocol as the cli, rather than a fixed protocol defined separately for them.

I'm thinking of taking a route similar to the env repo, where the dep graph of the test wasms is independent of the crates in the repo, and so we can build the test wasms with the previous protocol version, and the cli with the current protocol version. Whenever the rpc is released we can update the test-wasms to use the new protocol version.

If we use this pattern now and in the future it'll help us be able to update to the new protocol faster than the rpc can or in parallel with it.

cmd/crates/soroban-test/tests/fixtures/test-wasms/custom_account/Cargo.toml
```diff
@@ -12,7 +12,7 @@ crate-type = ["cdylib"]
 doctest = false
 
 [dependencies]
-soroban-sdk = { workspace = true }
+soroban-sdk = { version = "=21.7.2" }
```
leighmcculloch, line 15: fyi @fnando @willemneal @Ifropc when we get a v22 rpc setup in tests, then we can switch this back to the workspace version which is currently v22. This is just so that contracts being deployed will deploy to the v21 core successfully.

In the future we should break test-wasms into their own workspace so it's easy to set them to a different sdk version when needed.

Issue #1545, opened 2024-08-13
Aliases should be visible in the auto generated markdown


willemneal:
Also the alias should be visible in the markdown so this seems like a bug.

_Originally posted by @willemneal in https://github.com/stellar/stellar-cli/pull/1406#discussion_r1715856681_

janewang:
Clap markdown bug. we should open an issue on the clap reop @willemneal

leighmcculloch:
This issue has already been requested on the clap-markdown repository:
- https://github.com/ConnorGray/clap-markdown/issues/32

thesledge04:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/stellar-cli).

### My background and how it can be leveraged
I have experience in software development, particularly with solidity and Rust, and I am familiar with GitHub workflows. My background enables me to efficiently identify and fix bugs in code while ensuring adherence to best practices and user needs.

### How I plan on tackling this issue
Review the code to locate alias processing.
Identify the rendering issue for the alias.
Modify the logic to include the alias in Markdown.
Test the changes to ensure correct output.
Document the fix for clarity.

josephchimebuka:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/stellar-cli).

### My background and how it can be leveraged
Hello, I am Joseph I am a frontend developer and blockchain developer and I am also an active contributor here on only dust here is my profile https://app.onlydust.com/u/josephchimebuka. This is my first time to contribute to this repo ill appreciate the opportunity to contribute. 

### How I plan on tackling this issue
I would like to fix this

devcollinss:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/stellar-cli).

### My background and how it can be leveraged
My name is Collins Ikechukwu. I'm a full stack blockchain developer developer.

### How I plan on tackling this issue
### Steps to Implement:

1. **Identify the Issue Context**
   - Review the PR referenced [here](https://github.com/stellar/stellar-cli/pull/1406#discussion_r1715856681) to understand the alias visibility issue in markdown.
   - Confirm how the alias is currently being rendered and why it may not be displaying as expected.

2. **Check Markdown Syntax**
   - Ensure that the alias is correctly formatted in the markdown file. Aliases should be displayed using proper syntax (backticks for inline code or fenced code blocks if necessary):
     ```markdown
     `alias_name`
     ```
[...]

PR #1649, opened 2024-10-03
fix: Ensure aliases are visible in auto-generated markdown (#1545)

devcollinss:
### What

This change updates the markdown generation process to include aliases for both commands and arguments in the generated documentation.

### Why

This change updates the markdown generation process to include aliases for both commands and arguments in the generated documentation.

### Known limitations

N\A

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -1625,3 +1625,75 @@ Print version information
 
 
 
+
+
+**Aliases for argument `filter_logs`**: f
+
+
+**Aliases for argument `quiet`**: q
+
+
+**Aliases for argument `verbose`**: v
+
+
+**Aliases for argument `ignore_checks`**: i
+
+
+**Aliases for argument `out_file`**: o
+
+
+**Aliases for argument `with_example`**: w
+
+
+**Aliases for argument `ignore_checks`**: i
+
+
+**Aliases for argument `count`**: c
+
+
+**Aliases for argument `as_secret`**: s
+
+
+**Aliases for argument `default_seed`**: d
+
+
+**Aliases for argument `long`**: l
+
+
+**Aliases for argument `long`**: l
+
+
+**Aliases for argument `docker_host`**: d
+
+
+**Aliases for argument `limits`**: l
+
+
+**Aliases for argument `ports_mapping`**: p
+
+
+**Aliases for argument `image_tag_override`**: t
+
+
+**Aliases for argument `docker_host`**: d
+
+
+**Aliases for argument `docker_host`**: d
+
+
+**Aliases for argument `docker_host`**: d
+
+
+**Aliases for argument `limits`**: l
+
+
+**Aliases for argument `ports_mapping`**: p
+
+
+**Aliases for argument `image_tag_override`**: t
+
+
+**Aliases for argument `docker_host`**: d
+
+
+**Aliases for argument `long`**: l
--- a/cmd/soroban-cli/src/bin/doc-gen.rs
+++ b/cmd/soroban-cli/src/bin/doc-gen.rs
@@ -2,7 +2,8 @@ use std::{
     env,
     path::{Path, PathBuf},
 };
-
+// use clap::{Command, Arg};
+// use clap_markdown::MarkdownOptions;
 type DynError = Box<dyn std::error::Error>;
 
 fn main() -> Result<(), DynError> {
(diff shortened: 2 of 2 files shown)


FULL_HELP_DOCS.md
```diff
@@ -1625,3 +1625,75 @@ Print version information
...
+**Aliases for argument `image_tag_override`**: t
+
+
+**Aliases for argument `docker_host`**: d
+
+
+**Aliases for argument `docker_host`**: d
+
+
+**Aliases for argument `docker_host`**: d
```
willemneal, line 1684: There seem to be multiples.

devcollinss: Okay. will update it ASAP

devcollinss:
@willemneal, kindly review the PR. Also, could you please share your preferred means of communication—Discord or Telegram?

willemneal:
Darn! I wrote a longer response but it didn't get commented!

The summary is that you need to use [get_visible_aliases](https://docs.rs/clap/latest/clap/struct.Arg.html#method.get_visible_aliases) instead.  Using short aliases was a mistake we made early on, but for better readability in scripts it's better to use full words as it can be confusing reading it later which arg the short corresponds to. e.g. if two args start with s, what is `-s`? . So the short aliases can be for power users and they can find it with `--help`. But for the markdown I suggest we just highlight the normal aliases as these save space but are readable.

Also the aliases should appear inline with their argument.
e.g.
```
* --source-account, --source
```
Having them at the bottom will not help devs to discover them and this is more inline with the `--help`.

Note that rust field names become kebab case when converted to be a CLI arg. So you should be able to take your current setup and insert aliases by finding the corresponding arguments after converting the name kebab-case and adding the alias after like the example above.

Lastly, this bug seems like an issue for upstream clap-markdown itself since this would be the case for any markdown generated by it not just ours. However, given the scope of this work I think this is a fine compromise and won't rely on waiting for upstream releases.

As for communication Discord is preferred, but perhaps best to open an issue on the discord Dev channel so that others can also reply. And feel free to tag me (sir -  willem), just the letters, didn't want the full thing in plain text (although an LLM could still figure it out lol).

devcollinss:
@willemneal i just updated the repo. please review it

cmd/soroban-cli/src/bin/doc-gen.rs
```diff
@@ -17,17 +17,72 @@ fn doc_gen() -> std::io::Result<()> {
...
+            .get_visible_aliases()
+            .into_iter()
+            .flatten()
+            .map(|alias| format!("--{}", alias.to_kebab_case()))
+            .collect();
+
+            alias_list.push(arg_name.clone());
+        alias_list.extend(visible_aliases);
+
+        content.push_str(&format!(
```
willemneal, line 53: This is still just adding the aliases to the end of the document. And it seems that you include the arg_name in the alias list resulting in:
```
**Argument --shell**: --shell
```

Rather you should first check if the arg has an alias and then if so find  the locations of `arg_name` in content and insert the alias after it. `--source-account, --source`. I would also suggest splitting `content` into a vec of lines and passing that in. Then in the caller join the string again. Inserting into a large string can be very expensive.  When inserting in the middle of a string, a new string is created and then first part is copied over, the new string is appended and then the end of the of the original string. So by splitting the string once you pay once for the copying and then editing each line is much cheaper. Then pay once more for the joining.

An interesting article about how Zed handles string insertions for editing files: https://zed.dev/blog/zed-decoded-rope-sumtree

Since our use case is non-interactive and a one time build it's not so important to be as efficient as possible, but I still think it's worth splitting into lines.

0xdevcollins:
Hello @willemneal, just a friendly reminder that my latest changes are still awaiting your review. Let me know if you need any additional information. Thank you!

0xdevcollins (commented):
@willemneal

cmd/soroban-cli/src/bin/doc-gen.rs
```diff
@@ -17,17 +17,86 @@ fn doc_gen() -> std::io::Result<()> {
...
+        let arg_name = format!("--{}", arg.get_id().as_str().to_kebab_case());
+
+        // Collect aliases and deduplicate
+        let mut alias_list = vec![];
+        let visible_aliases: Vec<String> = arg
+            .get_visible_aliases()
+            .into_iter()
+            .flatten()
+            .map(|alias| format!("--{}", alias.to_kebab_case()))
+            .collect();
```
willemneal, line 53:
```suggestion
        for alias in arg
            .get_visible_aliases()
            .into_iter()
            .flatten()
            .map(|alias| format!("--{}", alias.to_kebab_case())) {
```
Why collect? and why is the arg_name added to the list of aliases?

Are are there commands with duplicate aliases?

cmd/soroban-cli/src/bin/doc-gen.rs
```diff
@@ -17,17 +17,86 @@ fn doc_gen() -> std::io::Result<()> {
...
+        alias_list.extend(visible_aliases);
+
+        // Sort and deduplicate to avoid repetitive aliases
+        alias_list.sort();
+        alias_list.dedup();
+
+        if alias_list.len() > 1 {
+            // Insert alias after arg_name, ensuring no duplicate insertion
+            if let Some(pos) = lines.iter().position(|line| line.contains(&arg_name)) {
+                if !lines[pos + 1].contains("**Alias**") {
```
willemneal, line 66: The idea was to update inline rather than insert a new one. E.g.
```
* `--very-verbose (--vv)` — Log DEBUG and TRACE events
```
Also it seems that your current match checks for any where in the string, which results in this:
```
**Usage:** `stellar contract asset deploy [OPTIONS] --asset <ASSET> --source-account <SOURCE_ACCOUNT>`
**Alias**: --source-account
```
Instead you should check that it starts like `* `--very-verbose`, then insert the alias as shown above.

willemneal:
There is something wrong which results in only a few aliases being found. The more I look into this the more I'm convinced that upstream is where this should be fixed. It's much easier to insert as the line is being produced than to ad hoc find it again. The current solution checks over the whole document for every arg. Which I'm sure is much slower too. @leighmcculloch What are your thoughts can this been salvaged?

0xdevcollins:
Thank you @willemneal @leighmcculloch

PR #1650, opened 2024-10-03
feat: Bring colors to help message

### LINKED ISSUE #1624: Bring colors to help message

Currently our help input is black and white

It would be great to add color theme, just it's in cargo and other commands
Ref: https://github.com/clap-rs/clap/issues/3234

N/A

Abeeujah:
### Bring colors to help message

Adds a color theme to the help message printed on the terminal

### Current color theme is black and white

- Improved Readability
- More engaging presentation

### closes #1624

### DIFF
--- a/cmd/soroban-cli/src/commands/global.rs
+++ b/cmd/soroban-cli/src/commands/global.rs
@@ -7,9 +7,9 @@ use std::path::PathBuf;
 use super::config;
 
 const USAGE_STYLES: Styles = Styles::styled()
-    .header(AnsiColor::Red.on_default().bold())
-    .usage(AnsiColor::Red.on_default().bold())
-    .literal(AnsiColor::Blue.on_default().bold())
+    .header(AnsiColor::Yellow.on_default())
+    .usage(AnsiColor::Green.on_default())
+    .literal(AnsiColor::Green.on_default())
     .placeholder(AnsiColor::Green.on_default());
 
 #[derive(Debug, clap::Args, Clone, Default)]
--- a/cmd/soroban-cli/src/commands/global.rs
+++ b/cmd/soroban-cli/src/commands/global.rs
@@ -1,16 +1,19 @@
 use clap::{
     arg,
-    builder::styling::{AnsiColor, Styles},
+    builder::styling::{AnsiColor, Effects, Styles},
 };
 use std::path::PathBuf;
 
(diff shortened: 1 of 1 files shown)


leighmcculloch:
This change has the appearance:
Screenshot 2024-10-04 at 10 06 24 PM

cmd/soroban-cli/src/commands/global.rs
```diff
@@ -1,11 +1,21 @@
...
+};
 use std::path::PathBuf;
 
 use super::config;
 
+const USAGE_STYLES: Styles = Styles::styled()
+    .header(AnsiColor::Red.on_default().bold())
+    .usage(AnsiColor::Red.on_default().bold())
+    .literal(AnsiColor::Blue.on_default().bold())
+    .placeholder(AnsiColor::Green.on_default());
```
leighmcculloch, line 13: I think we should probably go with the clap v3 default colours, which are more green yellow focused. The red blue is very red, and red is usually associated with errors, at least in a lot of CLIs and interfaces. I realise red means a variety of things across different cultures.

```suggestion
    .header(AnsiColor::Yellow.on_default())
    .usage(AnsiColor::Green.on_default())
    .literal(AnsiColor::Green.on_default())
    .placeholder(AnsiColor::Green.on_default());
```

leighmcculloch:
I pushed a change to change the colours to match what they used to look like when we were using clap v3, because that was the default back then.
Screenshot 2024-10-04 at 10 08 39 PM

leighmcculloch:
Actually, just tweaked them to exactly match cargo, since the issue @Ifropc opened was requesting alignment with cargo.

The styles were pulled from:
- https://github.com/rust-lang/cargo/blob/0473ee8b87dc7dbee53d13065c204ae63a0a2a9e/src/cargo/util/style.rs#L3-L13

Screenshot 2024-10-04 at 10 23 14 PM

leighmcculloch (approved):

cmd/soroban-cli/src/commands/global.rs
Abeeujah: I agree. The green-yellow color scheme in clap v3 is a more familiar to developers and less alarming choice, especially given the common association of red with errors. It's a good balance between readability and has stellar color brand (yellow) as well.

PR #1652, opened 2024-10-04
TS Bindings: Minify generated files

### LINKED ISSUE #1154: TS Bindings: Minify generated files
It'd be best practice to minify the output files in the `dist` files

t0fudev:
### What

- Add `terser` package to minify ES6 files.
- Add `minify` script to run `terser`.
- Modify `build` script to run `minify` after transpilation.

### Why

Closes #1154.

I was wondering if the output file should be named `index.min.js` or should I name it `index.js` to replace the original file. Also, I would like to know how can I test my changes to ensure everything is working properly. Thanks in advance!


chadoh (changes requested):
> I was wondering if the output file should be named `index.min.js` or should I name it `index.js`

I think it's probably fine to only ship it with a single, minified `index.js`.

There are tests in `cmd/crates/soroban-spec-typescript/ts-tests`. You can use these to make sure it's working!

It looks like you only found the built artifacts in the `fixtures` folder. If you check the GitHub PR files, you'll see that we've set all of these to be hidden, because they're a transient artifact, and will be wiped if you run `make build-snapshot`! You'll want to make your changes in `cmd/crates/soroban-spec-typescript/src/project_template`. Then you can run `make build-snapshot` to replace what's in that `fixtures` directory you edited.

janewang:
Closing PR due to no activity

PR #1657, opened 2024-10-06
contract read doesn't use default for durability like help said it does

gregemax:
updated contract to  use default for durability


cmd/soroban-cli/src/commands/contract/restore.rs
```diff
@@ -211,6 +211,7 @@ impl NetworkRunnable for Cmd {
             parse_operations(operations).ok_or(Error::MissingOperationResult)?,
         ))
     }
+    //
```
willemneal, line 214: ?

gregemax: Hi @willemneal 
Can u explain more

PR #1659, opened 2024-10-08
Render Stellar Events as JSON Instead of Rust Debug Forma

juandiegocv27:
### What

This PR modifies the Stellar events output to render as JSON instead of using the Rust debug format. Additionally, it updates the path in `boilerplate.rs` for including the WASM file.

### Why

The change is being made because developers are more familiar with the JSON format, which provides a standardized way of viewing event data. It aligns with Render events as json in stellar events pretty output instead of as rust debug issue, aiming to improve the readability and usability of the event logs for developers.

- closes: #1615 
- and reference to: #1642

### DIFF
--- a/cmd/soroban-cli/src/commands/events.rs
+++ b/cmd/soroban-cli/src/commands/events.rs
@@ -156,26 +156,16 @@ impl Cmd {
 
         let response = self.run_against_rpc_server(None, None).await?;
 
-        for event in &response.events {
-            match self.output {
-                // Should we pretty-print the JSON like we're doing here or just
-                // dump an event in raw JSON on each line? The latter is easier
-                // to consume programmatically.
-                OutputFormat::Json => {
-                    println!(
-                        "{}",
-                        serde_json::to_string_pretty(&event).map_err(|e| {
-                            Error::InvalidJson {
-                                debug: format!("{event:#?}"),
-                                error: e,
-                            }
-                        })?,
-                    );
-                }
-                OutputFormat::Plain => println!("{event}"),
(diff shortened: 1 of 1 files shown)


juandiegocv27:
@leighmcculloch I missclick and closed my PR, this is a new one with the changes that you requested🫡🫶🏻

juandiegocv27:
ready to review

juandiegocv27:
@willemneal let me know what you think please

janewang:
@juandiegocv27 could you look into the tests

cmd/soroban-cli/src/commands/events.rs
```diff
@@ -156,26 +156,16 @@ impl Cmd {
...
-                        serde_json::to_string_pretty(&event).map_err(|e| {
-                            Error::InvalidJson {
-                                debug: format!("{event:#?}"),
-                                error: e,
-                            }
-                        })?,
-                    );
-                }
-                OutputFormat::Plain => println!("{event}"),
-                OutputFormat::Pretty => event.pretty_print()?,
```
leighmcculloch, line 176: The hope of the issue was to address that "Events in pretty print mode print as Rust debug format," and same in the plain mode too I think. The intent is to fix that, rather than return only JSON. So each topic and the value should print as JSON, but the rest of the format should stay the same.

See:
- https://github.com/stellar/stellar-cli/issues/1615

juandiegocv27: ok sure, thank you for letting me know, i'll fix that as soon as possible

PR #1660, opened 2024-10-08
Fix/init blocking

elizabethengelman:
### What

- moves the init integration test into the integration dir
- makes the init command async, and then uses the async reqwest client. 

### Why

I was seeing this error when running `contract init` on `main`:
```
thread 'main' panicked at .cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.
```
This happens because we were trying to use the blocking reqwest client inside of an async runtime, since all of the cli is in an tokio runtime. This doesn't work well together, so I've changed the `init` command to be async and then we're able to `await` the HTTP calls.

### Known limitations

This won't be an issue once https://github.com/stellar/stellar-cli/pull/1628/files is merged - but we'll still want to move the `init` test into the integration dir - I think that this test was previously not being run.

### DIFF
--- a/cmd/crates/soroban-test/tests/it/integration.rs
+++ b/cmd/crates/soroban-test/tests/it/integration.rs
@@ -4,6 +4,7 @@ mod custom_types;
 mod dotenv;
 mod fund;
 mod hello_world;
+mod init;
 mod snapshot;
 mod tx;
 mod util;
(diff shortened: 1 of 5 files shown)


willemneal (approved):
Good catch

leighmcculloch:
Because #1628 is merged, does anything from this change still need to be merged?

elizabethengelman:
> Because https://github.com/stellar/stellar-cli/pull/1628 is merged, does anything from this change still need to be merged?

Sorry for missing this question earlier! You're right, since #1628 was merged, this can just be closed.

elizabethengelman:
I should also note that I think this error wasn't caught in the init integration test because of https://github.com/stellar/stellar-cli/issues/1682

PR #1663, opened 2024-10-08
feat: add `tx op add`

willemneal:
This allows chaining `tx op add` commands to build up a transaction with multiple operations.


willemneal:
Still need more tests.

FULL_HELP_DOCS.md
```diff
@@ -1291,6 +1292,206 @@ Sign, Simulate, and Send transactions
 
 
 
+## `stellar tx add-op`
```
leighmcculloch, line 1295: I think we should make this command `stellar tx op add`. The command will grow a little more gracefully if we want to add remove op, update op, and typing hyphens on the terminal is annoying in commands and it's a graceful way to remove the hyphen.

cmd/soroban-cli/src/commands/tx/xdr.rs
```diff
@@ -18,6 +19,8 @@ pub enum Error {
     Io(#[from] std::io::Error),
     #[error("only transaction v1 is supported")]
     OnlyTransactionV1Supported,
+    #[error("Only 100 Operations are allowed in a transaction")]
```
leighmcculloch, line 22:
```suggestion
    #[error("too many operations, limited to 100 operations in a transaction")]
```

cmd/soroban-cli/src/commands/tx/add_op/args.rs
```diff
@@ -0,0 +1,46 @@
...
+pub struct Args {
+    #[clap(flatten)]
+    pub locator: locator::Args,
+    /// Source account used for the operation
+    #[arg(
+        long,
+        visible_alias = "op-source",
+        env = "STELLAR_OPERATION_SOURCE_ACCOUNT"
+    )]
+    pub operation_source_account: Option<address::Address>,
```
leighmcculloch, line 26: Do you think we need a separate option for choosing the source account of operations? When adding an operation, there's no new tx being built, so the `--source-account` option should be available to use for this here, and the consistency will be more intuitive.

leighmcculloch (commented):
Couple minor things, otherwise lgtm 👏🏻

sagpatil:
https://github.com/stellar/stellar-cli/issues/1566 Linked

leighmcculloch (approved):

PR #1665, opened 2024-10-09
Fix flake for nixpkgs on darwin

TomMD:
Three issues addressed.

1. On darwin we need SystemConfig as a native build input.
2. Darwin doesn't have libudev-zero
3. Most darwin users will have the default zsh shell. When invoking `nix develop` they'll get a bash shell. The shell script setting up autocompletion of commands needs to not use the SHELL variable because that will be zsh and not the correct value of bash. See below for `nix develop` with this PR (without this PR the first result is zsh and thus the completion script fails).

```
$ echo $shell
bash
$ echo $SHELL
/bin/zsh
$ echo $0
bash
```


flake.nix
```diff
@@ -18,23 +18,24 @@
...
             (rust-bin.stable.latest.default.override {
               extensions = [ "rust-src" ];
               targets = [ "wasm32-unknown-unknown" ];
             })
-          ];
+          ] ++ lib.optionals (stdenv.isLinux) [libudev-zero];
           shellHook =
           ''
             echo "Using `nix --version`"
             alias stellar="cargo run --bin stellar --"
```
leighmcculloch, line 36: The fact the flake modifies the local shell's aliases is surprising and I think an anti-pattern. Aliases are something personal, something the terminal user should be in control of, and dev environments shouldn't fiddle with them imo.
```suggestion
```

flake.nix
```diff
@@ -18,23 +18,24 @@
...
-            [ -f ./local.sh ] && source ./local.sh
```
leighmcculloch, line 35: This change seems unrelated, but good. In general it's surprising when shells automatically execute scripts and push changes into the current environment.

flake.nix
```diff
@@ -18,23 +18,24 @@
...
             shell=$0
-            shell=`basename $SHELL`
+            shell=`basename $shell`
             source <(stellar completion --shell $shell)
```
leighmcculloch, line 39: I'm also unsure why or if we should have this line. This means that when you use the flake in this directory you automatically start compiling and setting up completions for the project, which seems surprising, and like the completions would get out of date if I'm working on multiple branches and swapping back and forth.

leighmcculloch (commented):
I defer to @Ifropc that knows more about flakes and nix than I do, but there is one thing being removed as part of this PR that I think is unrelated, but good. And there are two other things I think we should also remove.

flake.nix
Ifropc: We can remove it, I don't have any particular preferences. I mostly was writing it for myself, as there wasn't a demand for it. Now that we see people actually using it, I'm all down to make it user-friendly. So we can remove shell completions as well

Ifropc (approved):
Thanks for the fix @TomMD

PR #1675, opened 2024-10-15
Use attestation in install action

leighmcculloch:
### What
Verify GitHub attestation in install action for the cli and fail if the verification fails, or the verified subject does not match the binaries.yml build for the tag of the release.

### Why
The GitHub Attestation allows us to see who produced the binary. We may as well at install points check that the binary being installed from the GitHub releases was built by the appropriate job, and not modified at some point.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -16,11 +16,14 @@ runs:
       mkdir -p $HOME/.local/bin
       echo "$HOME/.local/bin" >> $GITHUB_PATH
   - name: Copy binary to install location
+    id: version
     shell: bash
-    env:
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -15,12 +15,15 @@ runs:
     run: |
       mkdir -p $HOME/.local/bin
       echo "$HOME/.local/bin" >> $GITHUB_PATH
+  - name: Copy binary to install location
```
leighmcculloch, line 18:
```suggestion
  - name: Determine version to install
```

fnando (approved):

Issue #1653, opened 2024-10-05
stellar contract info * commands require network when network not required


leighmcculloch:
```
❯ stellar contract info interface --wasm target/wasm32-unknown-unknown/release/test_empty.wasm
error: the following required arguments were not provided:
  --rpc-url <RPC_URL>
  --network-passphrase <NETWORK_PASSPHRASE>
  --network <NETWORK>

Usage: stellar contract info interface --rpc-url <RPC_URL> --network-passphrase <NETWORK_PASSPHRASE> --network <NETWORK> <--wasm <WASM>|--wasm-hash <WASM_HASH>|--id <CONTRACT_ID>>
```

cc @Ifropc

Ifropc:
@janewang it was fixed in https://github.com/stellar/stellar-cli/pull/1698

PR #1676, opened 2024-10-15
Fix: stellar contract info * commands require network when network no…

Dhanraj30:
…t required
#1653 
### What

 Additional logic for checking network requirement
 If WASM file is provided, don't require network arguments

### Why

The current implementation of the stellar contract info commands requires network parameters (e.g., --rpc-url, --network-passphrase, --network) even when users are working with local files such as WebAssembly (.wasm) files. This requirement can lead to unnecessary errors and confusion for users who are only looking to interact with local resources.

This change introduces conditional checks in the command handling logic to ensure that network parameters are only enforced when they are actually needed

### Known limitations

User Documentation: Existing documentation may need updates to clarify the new behavior of the commands concerning local .wasm file interactions. It is essential to ensure users are aware that network parameters are optional in these cases.


cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -97,6 +97,17 @@ impl Cmd {
...
+        // Additional logic for checking network requirement
+        // If WASM file is provided, don't require network arguments
+        if global_args.wasm.is_some() {
+            return Ok(());
+        }
+
+        // If no wasm is provided, network arguments are required
+        if global_args.rpc_url.is_none() || global_args.network_passphrase.is_none() || global_args.network.is_none() {
+            return Err(Error::Network);
+        }
```
willemneal, line 110: This code will run after any of the `stellar network` subcommands, not changing the behavior or arguments of `stellar contract info`. Furthermore, it uses fields that don't exist.
```
error[E0609]: no field `wasm` on type `&commands::global::Args`
   --> cmd\soroban-cli\src\commands\network\mod.rs:103:24
    |
103 |         if global_args.wasm.is_some() {
    |                        ^^^^ unknown field
    |
    = note: available fields are: `locator`, `filter_logs`, `quiet`, `verbose`, `very_verbose` ... and 2 others

error[E0609]: no field `rpc_url` on type `&commands::global::Args`
   --> cmd\soroban-cli\src\commands\network\mod.rs:108:24
    |
108 |         if global_args.rpc_url.is_none() || global_args.network_passphrase.is_none() || global_args.network.is_none() {
    |                        ^^^^^^^ unknown field
    |
    = note: available fields are: `locator`, `filter_logs`, `quiet`, `verbose`, `very_verbose` ... and 2 others

error[E0609]: no field `network_passphrase` on type `&commands::global::Args`
   --> cmd\soroban-cli\src\commands\network\mod.rs:108:57
    |
108 |         if global_args.rpc_url.is_none() || global_args.network_passphrase.is_none() || global_args.network.is_none() {
    |                                                         ^^^^^^^^^^^^^^^^^^ unknown field
    |
    = note: available fields are: `locator`, `filter_logs`, `quiet`, `verbose`, `very_verbose` ... and 2 others

error[E0609]: no field `network` on type `&commands::global::Args`
   --> cmd\soroban-cli\src\commands\network\mod.rs:108:101
    |
108 |         if global_args.rpc_url.is_none() || global_args.network_passphrase.is_none() || global_args.network.is_none() {
    |                                                                                                     ^^^^^^^ unknown field
    |
    = note: available fields are: `locator`, `filter_logs`, `quiet`, `verbose`, `very_verbose` ... and 2 others
    ```

Comment (Ifropc) [reply cmd/soroban-cli/src/commands/network/mod.rs:110]: Would like to add that the fixes for this issue is not very obvious, due to clap limitations. We might want to hold this one off until we do some code reorganization

Comment (Dhanraj30) [reply cmd/soroban-cli/src/commands/network/mod.rs]: I add fixes for this issue

File: cmd/soroban-cli/src/commands/global.rs [hunk 2]
```diff
@@ -46,6 +46,22 @@ pub struct Args {
...
+    #[arg(long, env = "STELLAR_NETWORK_PASSPHRASE")]
+    pub network_passphrase: Option<String>,
+ 
+    /// Network name (e.g., 'testnet', 'mainnet')
+    #[arg(long, env = "STELLAR_NETWORK")]
+    pub network: Option<String>,
+
+    /// Path to the WebAssembly file
+    #[arg(long, env = "STELLAR_WASM", global = true)]
+    pub wasm: Option<PathBuf>,
```
willemneal, line 64: This does the opposite of what this PR is trying to solve. The issue is that the subcommand of `contract info`, require `network` which doesn't make sense since they all are reading a local file. This change adds all these arguments to the global args, e.g. `stellar --network ... contract info`, which is not what we want.

cmd/soroban-cli/src/commands/global.rs
```diff
@@ -46,6 +46,7 @@ pub struct Args {
     /// Do not cache your simulations and transactions
     #[arg(long, env = "STELLAR_NO_CACHE", global = true)]
     pub no_cache: bool,
+
```
willemneal, line 49:
```suggestion
```

cmd/soroban-cli/src/commands/global.rs
```diff
@@ -61,6 +62,7 @@ pub enum Error {
         filepath: PathBuf,
         error: soroban_ledger_snapshot::Error,
     },
+
```
willemneal, line 65:
```suggestion
```

cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -97,6 +97,7 @@ impl Cmd {
                 cmd.run(global_args).await?;
             }
         };
+
```
willemneal, line 100:
```suggestion
```

cmd/soroban-cli/src/commands/contract/info/shared.rs
```diff
@@ -59,11 +59,17 @@ pub enum Error {
...
+        let wasm_bytes = wasm::Args { wasm: path.clone() }.read()?;
+        return Ok(Some(wasm_bytes));
+    }
 
-    let wasm = if let Some(path) = &args.wasm {
-        wasm::Args { wasm: path.clone() }.read()?
-    } else if let Some(wasm_hash) = &args.wasm_hash {
+    // If no local wasm, then check for wasm_hash and fetch from the network
+    let network = &args.network.get(&args.locator)?;
+
```
willemneal, line 71:
```suggestion
```

willemneal (approved):
LGTM just some small nits for removing extra new lines

Ifropc:
I'm pretty sure this won't work unless arguments are changed. From what I know Clap will parse arguments first, and we have network as one of the arguments, so I don't think this issue will be resolved with this PR
```
cargo run contract info interface --wasm ~/wasm/hello_world.wasm
     Running `target/debug/soroban contract info interface --wasm /home/ifro/wasm/hello_world.wasm`
error: the following required arguments were not provided:
  --rpc-url <RPC_URL>
  --network-passphrase <NETWORK_PASSPHRASE>
  --network <NETWORK>
```

willemneal:
@Ifropc I fixed it so that clap doesn't require network options at all. This should fix this and make commands like `keys generate` only complain if the RPC is required, e.g. `--fund`.

Ifropc:
@willemneal I think we can go with that approach, the only downside is that now arguments are no longer validated by clap. So we now will need to make sure we do the proper validation and give a good error message (e.g. current error message is not descriptive enough)
```
$ cargo run keys generate foo
   Compiling soroban-cli v21.5.0 (/home/ifro/projects/stellar-cli/cmd/soroban-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.34s
     Running `target/debug/soroban keys generate foo`
⚠️ Behavior of `generate` will change in the future, and it will no longer fund by default. If you want to fund please provide `--fund` flag. If you don't need to fund your keys in the future, ignore this warning. It can be suppressed with -q flag.
❌ error: network arg or rpc url and network passphrase are required if using the network
```

willemneal:
@Ifropc FYI you can do `cargo s` which is the same as `cargo run --quiet`.

Ifropc:
Fixed in https://github.com/stellar/stellar-cli/pull/1698

willemneal:
@Ifropc We still need this PR because it moves the network resolution after checking if there is a wasm arg.

willemneal (approved):

Ifropc:
Oh you are right sorry missed that

Ifropc (approved):

PR #1680, opened 2024-10-16
Remove the `soroban` binary from prebuilt binaries in the release

leighmcculloch:
### What
Remove the `soroban` binary from the prebuilt binaries in the release folder.

### Why
The `stellar` binary has been published and released for sometime now. We no longer need to be providing the `soroban` binary for folks or expecting folks to be using the `soroban-cli` crate.

This doesn't remove the soroban binary entirely, but it takes one small step towards doing so.


Ifropc:
> This doesn't remove the soroban and stellar binaries from the soroban-cli crate, because we seem to have a web of dependencies in the system-test repo that are using it

Looks like system tests needs to be changed either way 🤔

leighmcculloch:
> Looks like system tests needs to be changed either way 🤔

Yeah this seems much more complicated than I expected. I started working on the system-test repo to change it, and I'm just digging myself a hole at this point without my visibility to what's going on. I'm going to pare this change back to just not publishing binaries.

Ifropc:
Tests still are be failing but it seems to be unrelated

Ifropc (approved):

PR #1681, opened 2024-10-18
Favor stellar config dir.

fnando:
### What

Favor `~/.config/stellar` and `pwd/.stellar` config directories. This will create missing config dirs by using the stellar name, instead of soroban.

### Why

Given the push to the stellar-cli, we're should match config directories as well. Right now, we already have checks for local .stellar config directories, even though we generate a `.soroban` in case it doesn't exist.

### Known limitations

N/A


cmd/soroban-cli/src/config_dir_check.rs
```diff
@@ -0,0 +1,35 @@
...
+        parent_dir.join(".stellar")
+    };
+
+    let message = format!(
+        "The config directory {config_dir:?} is deprecated and should be \
+        renamed to {new_config_dir:?}.",
+    );
+
+    if (locator.global && dirname == "soroban") || (!locator.global && dirname == ".soroban") {
+        printer.warnln(message);
```
Ifropc, line 33: Should we do it automatically? This config directory is created and managed by the cli

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -468,14 +468,22 @@ impl KeyType {
...
     } else {
         dirs::home_dir()
             .ok_or(Error::HomeDirNotFound)?
             .join(".config")
+    };
+
+    let soroban_dir = config_dir.join("soroban");
+    let stellar_dir = config_dir.join("stellar");
+
+    if soroban_dir.exists() {
```
Ifropc, line 482: If both exists, IMO we should either use `stellar_dir` first or return an error

cmd/crates/soroban-test/tests/it/integration/hello_world.rs
```diff
@@ -113,7 +113,7 @@ async fn invoke() {
             },
         )
         .unwrap();
-    let sk_from_file = std::fs::read_to_string(dir.join(".soroban/identity/testone.toml")).unwrap();
+    let sk_from_file = std::fs::read_to_string(dir.join(".stellar/identity/testone.toml")).unwrap();
```
willemneal, line 116: Perhaps add a test to show that the stellar get's priority?

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -468,14 +468,22 @@ impl KeyType {
 }
 
 pub fn global_config_path() -> Result<PathBuf, Error> {
```
willemneal, line 470: I think this should return `[Option<PathBuf>; 2]`. Same with the local config. We already have the notion of prioritizing local over global. And so then we can add additional dimension to each. You would need to update `get_local_and_global` but the other functions expect locations in priority order so everything should work.

cmd/soroban-cli/src/config_dir_check.rs
willemneal: All the logic after `dirname`'s extraction is only needed in the true case.

cmd/soroban-cli/src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub use cli::main;
 pub mod assembled;
 pub mod commands;
 pub mod config;
```
willemneal, line 16: Why not move this into `config`? Then you can do `use crate::config::dir_check;`

cmd/soroban-cli/src/config_dir_check.rs
fnando: Leigh commented on Slack that we should support both directories indefinitely, so this check has been removed.

cmd/soroban-cli/src/config/locator.rs
fnando: Good catch. Also, we should log a warning message in case both directories exist, just like the local config is doing right now.

cmd/soroban-cli/src/config/locator.rs
fnando: All places that actually need both global and local config are calling `local_and_global`, which is returning `Result<[Location; 2], Error>`, so they can display all configs. I fail to see why changing the return type like you suggested is something we need to do.

willemneal: The point is that you want to collect all keys from all configs. For example you have a key for Alice in .stellar, and Bob in .soroban. Then others in the other two global configs. 

So you need to look for the keys that prioritize local then global as before then also stellar then soroban for each.

fnando: I don't think that's true. Right now, we prioritize .stellar over .soroban, and show an warning in case both exists at the same time: https://github.com/stellar/stellar-cli/blob/main/cmd/soroban-cli/src/utils.rs#L87-L89

I followed the same idea here, so it works similarly for both local and global configs. https://github.com/stellar/stellar-cli/pull/1681/files#diff-09da31e3b060c449882b17935698b71f2d9a1e1cd5a8198c135b594da32a4788R484-R486

willemneal: I am not at my computer but if Bob didn't exist in stellar but did in sororban this would find it? When I looked before it seemed to pick stellar and ignore soroban

willemneal: I am not at my computer but if Bob didn't exist in stellar but did in sororban this would find it? When I looked before it seemed to pick stellar and ignore soroban

fnando: > I am not at my computer but if Bob didn't exist in stellar but did in sororban this would find it?

No, it wouldn't, but this is already the case. You would get a warning like `tracing::warn!("the .stellar and .soroban config directories exist at path {pwd:?}, using the .stellar");` though.

fnando: We're always favoring `.stellar`, and not fetching both directories. If we decide we should, then that's a change for another pr, because it affects the current behavior.

willemneal: Then I think we should add a migration option. As @leighmcculloch said users could already have keys in their old soroban configs and currently not being able to resolve them is a bug IMO. And this seems like the perfect PR to fix that or add a migration option because we don't want users to ever have to manually inspect these directories.

fnando: We discussed this on slack and decided that it's not worth supporting the edge case where both directories can exist; the warning should be enough to give users a hint on what's being used.

fnando:
@leighmcculloch a 👍 here, please?

willemneal (approved):

PR #1684, opened 2024-10-23
Feat/add meta on build

### LINKED ISSUE #1605: Add meta during stellar contract build

Build tooling, CI environments, and other things outside the code want to conveniently add meta data to the contract as part of the build process.

The main use case is to support setting meta during a verified build process that @orbitlens is building.

A new option on the `stellar contract build` command that includes additional metadata to be written into the `contractmetav0` custom section as `SCMetaEntry` XDR entities.

```
stellar contract build --meta key1=val1 --meta key2=val2 ...
```

Contract can already embed meta in themselves by making the following call in the code:
```rust
soroban_sdk::contractmeta!(key = "...", val = "...");
```

However, sometimes build systems would like to add meta information, such as the source code repo the build is occurring from, or the commit being built, and that information may not be accessible to the code to insert itself.

It may be possible to structure that stuff with a build.rs script, but adding the ability to pass meta at the tooling level means that contracts don't need to write any code for build tooling to augment the meta.

### Thanks

This idea was shared by and requested by @orbitlens in Discord:
- https://discord.com/channels/897514728459468821/1218934322745315389/1285154853752012852

cc @janewang

elizabethengelman:
### What

Closes #1605 

### Why

Allows users to add additional meta to their contract wasm on build.

### Known limitations

n/a


cmd/crates/soroban-spec-tools/src/contract.rs
```diff
@@ -113,6 +120,115 @@ impl Spec {
...
+        // module.section(&RawSection {
+        //     id: SectionId::Custom as u8,
+        //     data: new_data,
+        // });
+
+        // then add the new custom section
+        module.section(&CustomSection {
+            name: Cow::Borrowed("contractmetav0"),
+            data: Cow::Borrowed(new_data),
+        });
```
leighmcculloch, line 217: If the plan is to leave the existing custom section intact, and add a second custom section with the same name, can I suggest we use the wasmgen crate, which has the ability to simply add a new section without needing to parse or know anything about the existing sections?

An example of using it is here:
https://github.com/leighmcculloch/wasm-cs/blob/488eb67ef04f750ffb4ef1f2030d57cb5964af45/src/write.rs#L35

There might be a way to do the same with wasm-encoder. I'm just thinking it looks like we don't actually need to rewrite the existing module which seems like a risk if we don't support a section or there's a bug in that rewriting logic.

elizabethengelman: Thank you for the example of wasmgen! 

I agree that rewriting the existing module does seem risky! Doing it that way in the PR was really helpful in understanding the tooling better, but I'll work on just appending a new section today.

elizabethengelman: Just pushed up a change to use wasmgen, so much nicer! 🙏 

I'll plan to work on allowing a user to pass in meta as args to build, and testing next.

cmd/crates/soroban-spec-tools/src/contract.rs
```diff
@@ -52,7 +51,13 @@ impl Spec {
                     "contractspecv0" => &mut spec,
                     _ => continue,
                 };
-                *out = Some(section.data());
+
+                if let Some(existing_data) = out.as_mut() {
+                    let combined_data = [existing_data, section.data()].concat();
+                    *out = Some(Box::leak(combined_data.into_boxed_slice()));
```
leighmcculloch, line 57: Does this mean if a .wasm has meta with field a in it, that the new custom section written will have the first custom section copied into it then append onto?

That'll mean the meta for a will appear twice, in both custom sections.

It should be okay to not join them, and just append the new custom section with the additional fields.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -145,6 +147,28 @@ impl Cmd {
                     return Err(Error::Exit(status));
                 }
 
+                let file = format!("{}.wasm", p.name.replace('-', "_"));
+                let target_file_path = Path::new(target_dir)
+                    .join("wasm32-unknown-unknown")
+                    .join(&self.profile)
+                    .join(&file);
```
leighmcculloch, line 154: Since `file` and `target_file_path` are now defined here, they no longer need to be defined on lines 175-179. The same definition appears there and is a duplicate.

elizabethengelman: updated in [344378a](https://github.com/stellar/stellar-cli/pull/1684/commits/344378a3f68ec0adf39de08906391bf046ea5b56)

cmd/crates/soroban-spec-tools/src/contract.rs
elizabethengelman: 🤔 I may be misunderstanding how this `Spec` struct is being used - I'm interpreting it as a reader here, since it parses the wasm bytes into an easier-to-manage type.

Without joining the existing data with the new custom section here, I'm seeing that even though the new custom sections were added to the wasm properly, when I print out the contract's meta (via `stellar contract info meta`) only the last meta passed in on build was being displayed.

For example, if I run:
```
stellar contract build --metadata new-meta-key=new-meta-value \ 
    --metadata another-new-meta-key=another-new-meta-value
```
I would expect that my contract should have the following metadata:
```
- Description: A test hello world contract (added with contractmeta macro in the contract)
- rsver: 1.81.0 (builtin meta)
- rssdkver: 21.7.3#1f81... (builtin meta) 
- new-meta-key: new-meta-value (added on build)
- another-new-meta-key: another-new-meta-value (added on build)
```

But when I run `stellar contract info meta`, I only see this:
```
Contract meta:
 • another-new-meta-key: another-new-meta-value
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -145,14 +156,32 @@ impl Cmd {
...
+
+                let mut wasm_bytes = fs::read(&target_file_path).unwrap();
```
willemneal, line 165: This should use an Error.

elizabethengelman: nice catch!

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -145,14 +156,32 @@ impl Cmd {
...
+                    let xdr: Vec<u8> = meta_entry.to_xdr(Limits::none()).unwrap();
+
+                    wasm_gen::write_custom_section(&mut wasm_bytes, "contractmetav0", &xdr);
+                });
+
+                let updated_spec = Spec::new(&wasm_bytes).unwrap();
+                println!("Updated meta!: {:?}", updated_spec.meta);
+
+                let str_path: &str = target_file_path.to_str().unwrap();
+                fs::write(str_path, wasm_bytes).unwrap();
```
willemneal, line 181: You don't need to convert to a string. See the signature:
`pub fn write<P: AsRef<Path>, C: AsRef<[u8]>>(path: P, contents: C) -> Result<()> `
```suggestion
                fs::write(&target_file_path, wasm_bytes).unwrap();
```
Also should use an Error.

cmd/crates/soroban-spec-tools/src/contract.rs
```diff
@@ -52,7 +52,13 @@ impl Spec {
...
```
leighmcculloch, line 58: We should avoid the use of the leak here, as this memory will never be recollected iiuc. We can make the meta a `Vec<u8>` so that we can concat and simply store a Vec. Would that work?

elizabethengelman: Ah, good catch! It looks like changing `meta` to a `Vec<u8>` should work! I've also changed `env_meta` and `spec` as well, so we can continue handling the payload sections similarly. I get it pushed up shortly.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -62,6 +62,19 @@ pub struct Cmd {
     /// Print commands to build without executing them
     #[arg(long, conflicts_with = "out_dir", help_heading = "Other")]
     pub print_commands_only: bool,
+    #[arg(long, num_args=1, value_parser=parse_metadata, action=clap::ArgAction::Append, help_heading = "Metadata")]
```
willemneal, line 65:
```suggestion
    /// Add key-value data to the `contractmetav0` custom section
    #[arg(long, num_args=1, value_parser=parse_metadata, action=clap::ArgAction::Append, help_heading = "Metadata")]
```

leighmcculloch:
```suggestion
    /// Add key-value to contract meta (adds the meta to the `contractmetav0` custom section)
    #[arg(long, num_args=1, value_parser=parse_metadata, action=clap::ArgAction::Append, help_heading = "Metadata")]
```

Cargo.lock
```diff
@@ -4991,7 +5004,7 @@ version = "21.5.0"
 dependencies = [
  "async-trait",
  "bollard",
- "byteorder",
+ "byteorder 1.5.0",
```
leighmcculloch, line 5007: This means that there are now two versions of byteorder in use. That might be unnecessary.

leighmcculloch (commented):
Looks good, three minor tweaks otherwise 👍🏻.

Have you tried deploying a contract built with the additional meta option?

Cargo.lock
elizabethengelman: nice catch!

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -62,6 +62,20 @@ pub struct Cmd {
     /// Print commands to build without executing them
     #[arg(long, conflicts_with = "out_dir", help_heading = "Other")]
     pub print_commands_only: bool,
+    /// Add key-value to contract meta (adds the meta to the `contractmetav0` custom section)
+    #[arg(long, num_args=1, value_parser=parse_meta_arg, action=clap::ArgAction::Append, help_heading = "Metadata")]
+    pub meta: Vec<(String, String)>,
+}
+
+fn parse_meta_arg(s: &str) -> Result<(String, String), Error> {
+    let parts = s.splitn(2, '=');
```
Ifropc, line 71: Should probably also check that there's only one `=` in the string

leighmcculloch: `=` is a valid character in a header value, so it's allowed to occur more than once since only the first one we use as a delimiter on the key-value.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -82,8 +96,17 @@ pub enum Error {
     CopyingWasmFile(io::Error),
     #[error("getting the current directory: {0}")]
     GettingCurrentDir(io::Error),
+    #[error("reading wasm file: {0}")]
+    ReadingWasmFile(io::Error),
+    #[error("writing wasm file: {0}")]
+    WritingWasmFile(io::Error),
+    #[error("invalid metadata entry: {0}")]
+    MetadataArg(String),
```
leighmcculloch, line 104: Prefer the term "meta" over "metadata" for consistency with existing vocab. Same elsewhere in the change where "metadata" is being used to refer to the contract meta.
```suggestion
    #[error("invalid meta entry: {0}")]
    MetaArg(String),
```

leighmcculloch (approved):
👏🏻 

This could impact others, so we should do an announcement 📣 in Discord so folks know that it's appending a second custom section with the same name.

cmd/soroban-cli/src/commands/contract/build.rs
elizabethengelman: good catch, i missed that one!

elizabethengelman:
> This could impact others, so we should do an announcement 📣 in Discord so folks know that it's appending a second custom section with the same name. @elizabethengelman can you do that?

@leighmcculloch sure thing! does posting in #announcements make sense, or is there another channel?

cmd/soroban-cli/src/commands/contract/build.rs
leighmcculloch: Oops, thought this was about the rpc headers PR 😓. There's little reason to block people from using `=` in a meta value too.

leighmcculloch:
> #announcements make sense, or is there another channel?

It's a bit low level and won't be important for contract developers likely to know, so in the `soroban-itself` channel, starting a thread, and then we can mention folks we expect will be of interest. And anyone else can see thet hread and hopefully tag others if they know of anyone.

PR #1685, opened 2024-10-24
Add cargo-deny for license checks and dep duplicate avoidance

leighmcculloch:
### What
Add cargo-deny.

### Why
To perform license checks and identify unnecessarily dependency duplication that increases build time and binary size.

The config was taken from the `rs-soroban-sdk` repo with minor changes made to fit.

We should have added this a while ago.


.github/workflows/rust.yml
```diff
@@ -30,6 +30,21 @@ jobs:
     - run: rustup update
     - run: cargo fmt --all --check
 
+  cargo-deny:
```
Ifropc, line 33: Should we run it every time on PR? Maybe we can skip running it if `Cargo.lock` was unchanged?

leighmcculloch: It's cheap to run, so I don't think it matters much. We run it on much larger repos on each commit.

Ifropc (dismissed):

.github/workflows/rust.yml
Ifropc: That makes sense it also detects new vulnerabilities which is very useful: should we also update our dependencies highlighted by cargo-deny?

fnando (approved):

PR #1686, opened 2024-10-25
Make flake install work

TomMD:
### What

I wanted to `nix profile install github:stellar/stellar-cli`. I couldn't.

### Why

Dev convenience. Also being able to write in docs something like `nix run "github:tommd/stellar-cli?ref=make-flake-install-work" -- --help` and eventually `nix run github:stellar/stellar-cli` is fun.


Makefile
```diff
@@ -2,17 +2,17 @@ all: check build test
 
 export RUSTFLAGS=-Dwarnings -Dclippy::all -Dclippy::pedantic
 
-REPOSITORY_COMMIT_HASH := "$(shell git rev-parse HEAD)"
+REPOSITORY_COMMIT_HASH := "$(shell git rev-parse HEAD || echo NOREPO)"
```
Ifropc, line 5: Why do we need this addition?

Ifropc (approved):
LGMT with one question

README.md
```diff
@@ -37,6 +37,13 @@ Install the latest version from source:
 cargo install --locked stellar-cli --features opt
 ```
 
+Run directly with nix, ex:
+```
+% nix run 'github:stellar/stellar-cli' -- --help
```
leighmcculloch, line 42: Will this run the latest release / tag, or the main branch?

flake.nix
```diff
@@ -31,6 +30,30 @@
...
           ] ++ lib.optionals (stdenv.isLinux) [libudev-zero];
         };
+        stellarcli = stellardev // {
+          cargoLock = {
+            lockFile = ./Cargo.lock;
+          };
+
+          cargoLock.outputHashes = {
+              # The only git+https dependency in Cargo.lock
+              "stellar-rpc-client-21.4.0" = "sha256-ue7Ynux9YaDP3f/XkHz2OPd2g0iCX5R0yS5SaVHEYxQ";
```
leighmcculloch, line 40: The used of a git dep is temporary until its released. Do we need to maintain this list whenever temporarily using git deps?

flake.nix
```diff
@@ -31,6 +30,30 @@
...
+          };
+
+          # As of writing 'cargo test' fails
+          doCheck = false;
```
leighmcculloch, line 44: Let's fix it separate to this PR so this isn't required.

leighmcculloch (commented):
Thanks for improving the flake and finding ways folks can run the stellar-cli! Couple questions inline.

flake.nix
Ifropc: I think the idea is to fix tests first and then merge this PR with `doCheck = false;` removed, is it correct?

leighmcculloch: It doesn't need to block this PR assuming nothing is actually broken, but once it is fixed we should circle back and remove the doCheck = false. I'm also a little confused as to why it's broken. The tests appear to pass for me outside of nix. Are they just failing only within nix?

Ifropc: I actually think we should keep `doCheck = false;`. Our current tests are doing too much at the moment and have some dependencies (e.g. when you simply run tests they will fail, unless specifics steps are done) -- this makes it harder for is to maintain the flake. Second, this tests are unstable. If someone want to just try out the cli and uses `nix run` I think it'll be quite frustrating to not be able to run the cli (it already takes a lot of time to compile everything and it also takes a lot of time to run all the tests -- which should be overkill as we _should_ have al itl tested in our CI/Ci)
BTW, I also have tests failing in `nix run` but it works just fine in `nix develop -> cargo test`

@leighmcculloch thoughts?

leighmcculloch: > Oh it works for you?

No sorry, I'm not using nix. I just meant the tests work when I run `cargo test`.



> I actually think we should keep `doCheck = false;`. Our current tests are doing too much

👍🏻

README.md
```diff
@@ -37,6 +37,13 @@ Install the latest version from source:
 cargo install --locked stellar-cli --features opt
 ```
 
+Run directly with nix, ex:
```
leighmcculloch, line 40:
```suggestion
Install or run the unreleased main branch with nix:
```

flake.nix
```diff
@@ -30,7 +34,33 @@
...
+        rustPlatformMod = pkgs.makeRustPlatform {
+          cargo = pkgs.rust-bin.stable.latest.default;
+          rustc = pkgs.rust-bin.stable.latest.default;
         };
+      in
+      with pkgs;
+      {
+        devShells.default = mkShell stellardev;
+        packages.default = rustPlatformMod.buildRustPackage stellarcli;
       }
```
leighmcculloch, line 64: This is outside the scope of the current PR really, but is there a way to specify `--features opt` for the build?

flake.nix
Ifropc: @TomMD I think we are on agreement that we should revert latest changes made for `doCheck = true;` support, I think the revision before enabling tests was perfect

Ifropc:
Works perfectly, thank you so much @TomMD !

PR #1689, opened 2024-10-28
Enable unit tests for all platforms

Ifropc:
### What

Enables unit tests for all platforms for both latest and msrv rust

### Why

Previously unit tests were running only on `x86_64-pc-windows` (for example see https://github.com/stellar/stellar-cli/actions/runs/11563040391/job/32185499011?pr=1686) 

### Known limitations

N/A


.github/workflows/rust.yml
```diff
@@ -96,10 +96,11 @@ jobs:
           target: x86_64-unknown-linux-gnu
           cargo-hack-feature-options: --feature-powerset
           additional-deb-packages: libudev-dev
-        # - os: ubuntu-latest-16-cores
-        #   target: aarch64-unknown-linux-gnu
-        #   cargo-hack-feature-options: --feature-powerset
-        #   additional-deb-packages: libudev-dev libssl-dev
+        # TODO: add back ARM support
```
fnando, line 99: should we try this and see if it works?

Ifropc: I wanted to test it, but currently it's blocked by us having git dependency in the soroban client 
> it has a git rev dep on the rpc client, which is incompatible with releasing.

(From DMs with Leigh) 
I will check back on it once we release the new client version

willemneal (approved):

.github/workflows/rust.yml
leighmcculloch: Hmm, this shouldn't be blocked by a git dep. Git deps should work just fine for the rust workflow. It's the publish dry run workflow that we should expect to error when a git dep is present since we can't publish with git deps.

leighmcculloch: 🤦🏻 Oh right, this is the publish dry run. Either way, this should be fine to add back in. The publish dry run is failing for a different reason, but it should pass when the git dep is removed. Fair to keep it commented for now until we can confirm.

.github/workflows/rust.yml
```diff
@@ -73,17 +73,17 @@ jobs:
...
     - run: rustup target add wasm32-unknown-unknown
-    - if: matrix.target == 'aarch64-unknown-linux-gnu'
-      run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
-    - run: cargo clippy --all-targets --target ${{ matrix.target }}
+    - if: matrix.sys.os == 'ubuntu-latest-16-cores'
+      run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev
+    - run: cargo clippy --all-targets --target ${{ matrix.sys.target }}
     - run: make build-test
-    - if: startsWith(matrix.target, 'x86_64')
+    - if: startsWith(matrix.sys.target, 'x86_64')
```
leighmcculloch, line 82: This guard is wrong really. macos-latest is arm64 and we can run tests on it. We just can't run tests on arm64 on linux, without emulation and we don't have emulation setup. We have access to arm64 linux runners too so we could run tests if we wanted to now. Mentioning for a follow up PR.

leighmcculloch (approved):

leighmcculloch (approved):
I'm trying to make sure I understand this diff. The material part of this diff is just one line, adding `libudev-dev` to the apt-get install list? Everything else looks like a no-op. 👍🏻

.github/workflows/rust.yml
```diff
@@ -44,16 +44,16 @@ jobs:
...
-            target: x86_64-unknown-linux-gnu
-          - os: ubuntu-latest-16-cores
-            target: aarch64-unknown-linux-gnu
-          - os: macos-latest
-            target: x86_64-apple-darwin
-          - os: macos-latest
-            target: aarch64-apple-darwin
-          - os: windows-latest-8-cores
-            target: x86_64-pc-windows-msvc
+        - os: ubuntu-latest-16-cores
```
Ifropc, line 47: @leighmcculloch I know we don't really follow this style, but isn't proper yaml style to add indent on new object start?
So it should be
```
sys:
__-os: ...
```
not 
```
sys:
- os: <...>
```
Asking because I personally find it more readable + my IDE auto-formats it like that 😆

Ifropc:
> The material part of this diff is just one line, adding libudev-dev to the apt-get install list

Also changed `includes` -> `sys` in matrix. The way `includes` was used is incorrect, as it adds one more permutation to already existing martix. So we weren't testing on all OSes, just windows (you can check that it's how it worked in older builds, it had only 2 Rust / build-and-test jobs, while it should have 2 (rust toolchains) * 5 (supported OSes))

.github/workflows/rust.yml
Ifropc: > We just can't run tests on arm64 on linux, without emulation

Could you elaborate more on this please?

.github/workflows/rust.yml
Ifropc: BTW I think this should be moved out to a different workflow 🤔

Issue #1396, opened 2024-06-20
Defaults for global config


JakeUrban:
### What problem does your feature solve?

If you are using a local project-based config `.soroban` directory, you can set environment variables in a `.env` file that specify the network and source account that should be used if omitted from a command.

note: being able to set these environment variables is undocumented other than in the `--help` message for some commands

However, if you're using global config, you have to set the environment variables in your shell.

### What would you like to see?

It would be great if I could configure which network and source account to use via a `use` command or something similar.

```bash
stellar network add testnet --rpc-url ... --network-passphrase ...
stellar network use testnet
stellar keys generate primary  # no --network flag needed
stellar keys use primary
stellar contract id asset --asset native  # no --network or --source-account flags needed
```

### What alternatives are there?

Using environment variables.

willemneal:
Great idea! We are planning on refactoring the config so that you could for example take advantage of the network you are compiling to so you could get the id of the native SAC like in your final example. I had been wondering how you would handle checking which network is currently in use when compiling and this seems like a good way to handle it.

leighmcculloch:
It would be helpful if we add a command that shows you the current state, based off all the inputs that can affect state such as env vars, the `use` commands, etc.

I think we need to explore a bit to what that could look like, but as an example:

```
$ export STELLAR_NETWORK=testnet
$ stellar keys use primary
$ stellar config
network=testnet (env)
key=primary (use)
```

leighmcculloch:
> ```
> $ export STELLAR_NETWORK=testnet
> $ stellar keys use primary
> $ stellar env
> network=testnet (env)
> key=primary (use)
> ```

Floating this idea: think we should make this operate around env vars, since every config has an env var, instead of mapping to some third concept (the `network`, `key` above).

This is something other tools like `go env` do:

```
$ export STELLAR_NETWORK=testnet
$ stellar keys use primary
$ stellar env
STELLAR_NETWORK=testnet # env
STELLAR_ACCOUNT=primary # env
```

We can still add annotations, except use `#` so it's a comment.

If you want to save the current env in a .env file, it's essentially `stellar env > .env`. If we're planning to get rid of the local .stellar folder, this may play into a story that folks use for project based config.

leighmcculloch:
[...]

PR #1690, opened 2024-10-29
Add global config for default network and identity.

fnando:
### What

This pr introduces a global config file called `~/.stellar/config.toml`. Right now, it contains a new section called `defaults`, which can hold the values set by the new commands `stellar keys use NAME` and `stellar network use NAME`. An upcoming pr will use this values by default, when available.

```console
$ stellar network use testnet
ℹ️ The default network is set to `testnet`

$ stellar keys use alice
ℹ️ The default source account is set to `alice`

$ stellar env
STELLAR_NETWORK=testnet # default
STELLAR_ACCOUNT=alice   # default

$ cat ~/.config/stellar/config.toml
[defaults]
network = "testnet"
identity = "alice"
```

### Why

This is the first step to implement #1396.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/env/mod.rs
+++ b/cmd/soroban-cli/src/commands/env/mod.rs
@@ -4,6 +4,7 @@ use crate::{
         locator::{self},
         Config,
     },
+    print::Print,
 };
 use clap::Parser;
 
@@ -17,13 +18,11 @@ pub struct Cmd {
 pub enum Error {
     #[error(transparent)]
     Locator(#[from] locator::Error),
-
-    #[error("no defaults or environment variables set")]
-    NoEnv,
 }
 
 impl Cmd {
-    pub fn run(&self, _global_args: &global::Args) -> Result<(), Error> {
+    pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let print = Print::new(global_args.quiet);
         let config = Config::new()?;
         let mut lines: Vec<(String, String)> = Vec::new();
 
@@ -36,10 +35,13 @@ impl Cmd {
         }
 
         if lines.is_empty() {
-            return Err(Error::NoEnv);
+            print.warnln("No defaults or environment variables set".to_string());
+            return Ok(());
         }
 
         let max_len = lines.iter().map(|l| l.0.len()).max().unwrap_or(0);
+
+        lines.sort();
 
         for (value, source) in lines {
             println!("{value:max_len$} # {source}");
--- a/cmd/soroban-cli/src/commands/env/mod.rs
+++ b/cmd/soroban-cli/src/commands/env/mod.rs
@@ -1,6 +1,6 @@
 use crate::{
     commands::global,
-    config::locator::{self, config, config_file},
+    config::locator::{self, config},
 };
 use clap::Parser;
 
@@ -14,24 +14,46 @@ pub struct Cmd {
 pub enum Error {
     #[error(transparent)]
     Locator(#[from] locator::Error),
+
+    #[error("no defaults or environment variables set")]
+    NoEnv,
 }
 
 impl Cmd {
     pub fn run(&self, _global_args: &global::Args) -> Result<(), Error> {
         let config = config()?;
+        let mut lines: Vec<(String, String)> = Vec::new();
 
-        println!("config_file={}", config_file()?.to_string_lossy());
+        if let Some(data) = get("STELLAR_NETWORK", config.defaults.network) {
+            lines.push(data);
+        }
 
-        println!(
-            "network={}",
-            config.defaults.network.unwrap_or("(unset)".to_string())
-        );
+        if let Some(data) = get("STELLAR_ACCOUNT", config.defaults.identity) {
+            lines.push(data);
+        }
 
-        println!(
-            "identity={}",
-            config.defaults.identity.unwrap_or("(unset)".to_string())
-        );
+        if lines.is_empty() {
+            return Err(Error::NoEnv);
+        }
+
+        let max_len = lines.iter().map(|l| l.0.len()).max().unwrap_or(0);
+
+        for (value, source) in lines {
+            println!("{value:max_len$} # {source}");
+        }
 
         Ok(())
     }
+}
+
+fn get(env_var: &str, default_value: Option<String>) -> Option<(String, String)> {
+    if let Ok(value) = std::env::var(env_var) {
+        return Some((format!("{env_var}={value}"), "env".to_string()));
+    }
+
+    if let Some(value) = default_value {
+        return Some((format!("{env_var}={value}"), "default".to_string()));
+    }
+
+    None
 }
--- a/cmd/soroban-cli/src/commands/env/mod.rs
+++ b/cmd/soroban-cli/src/commands/env/mod.rs
@@ -1,6 +1,9 @@
 use crate::{
     commands::global,
-    config::locator::{self, config},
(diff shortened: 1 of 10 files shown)


cmd/soroban-cli/src/config/mod.rs
```diff
@@ -143,3 +140,31 @@ impl ArgsLocatorAndNetwork {
         Ok(self.network.get(&self.locator)?)
     }
 }
+
+#[derive(Serialize, Deserialize, Debug)]
```
willemneal, line 144:
```suggestion
#[derive(Serialize, Deserialize, Debug, Default)]
```

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -143,3 +140,31 @@ impl ArgsLocatorAndNetwork {
...
+            },
+        }
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
willemneal, line 170:
```suggestion
```

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -143,3 +140,31 @@ impl ArgsLocatorAndNetwork {
         Ok(self.network.get(&self.locator)?)
     }
 }
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Config {
+    pub defaults: Defaults,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
```
willemneal, line 149:
```suggestion
#[derive(Serialize, Deserialize, Debug, Default)]
```

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -143,3 +140,31 @@ impl ArgsLocatorAndNetwork {
...
+impl Config {
+    pub fn new() -> Config {
+        Config {
+            defaults: Defaults {
+                network: None,
+                identity: None,
+            },
+        }
+    }
+}
```
willemneal, line 164:
```suggestion
    pub fn new() -> Result<Config, locator::Error> {
        let path = config_file()?;
        if path.exists() {
            let data = fs::read(&path).map_err(|_| locator::Error::FileRead { path })?;
            Ok(toml::from_slice(&data)?)
        } else {
            Ok(Config::default())
        }
    }
```

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -143,3 +140,31 @@ impl ArgsLocatorAndNetwork {
...
+    pub defaults: Defaults,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Defaults {
+    pub network: Option<String>,
+    pub identity: Option<String>,
+}
+
+impl Config {
```
willemneal, line 155:
```suggestion
impl Config {
    pub fn network(mut self, s: &str) -> Self {
        self.defaults.network = Some(s.to_string())
        self
    }

    pub fn identity(mut self, s: &str) -> Self {
        self.defaults.identity = Some(s.to_string())
        self
    }
    
    pub fn save(&self) -> Result<(), locactor::Error> {
        let toml_string = toml::to_string(&self)?;
        let mut file = File::create(locator::config_file()?)?;
        file.write_all(toml_string.as_bytes())?;
        Ok(())
    }
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -163,6 +167,30 @@ impl Args {
...
+        let mut config = config()?;
+
+        config.defaults.identity = Some(name.to_string());
+
+        let toml_string = toml::to_string(&config)?;
+        let mut file = File::create(config_file()?)?;
+        file.write_all(toml_string.as_bytes())?;
+
+        Ok(())
+    }
```
willemneal, line 192:
```suggestion
        Config::new()?.set_identity(name).save()
    }
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -163,6 +167,30 @@ impl Args {
...
+        let mut config = config()?;
+
+        config.defaults.network = Some(name.to_string());
+
+        let toml_string = toml::to_string(&config)?;
+        let mut file = File::create(config_file()?)?;
+        file.write_all(toml_string.as_bytes())?;
+
+        Ok(())
+    }
```
willemneal, line 180:
```suggestion
        Config::new()?.set_network(name).save()
    }
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -496,8 +514,19 @@ pub fn global_config_path() -> Result<PathBuf, Error> {
...
+pub fn config() -> Result<Config, Error> {
+    let path = config_file()?;
+
+    if path.exists() {
+        let data = fs::read(&path).map_err(|_| Error::FileRead { path })?;
+        let config: Config = toml::from_slice(data.as_slice())?;
+
+        Ok(config)
+    } else {
+        Ok(Config::new()?)
```
willemneal, line 530: Is this still needed?

fnando: We do not.

cmd/soroban-cli/src/commands/env/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+            return Err(Error::NoEnv);
+        }
+
+        let max_len = lines.iter().map(|l| l.0.len()).max().unwrap_or(0);
+
+        for (value, source) in lines {
+            println!("{value:max_len$} # {source}");
+        }
+
+        Ok(())
```
willemneal, line 48: This could all go in a `Display` impl for `Config`. Also is an empty config an error or just an empty string?

cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -6,17 +6,21 @@ use super::{config::locator, global};
...
 pub mod container;
+pub mod default;
 pub mod ls;
 pub mod rm;
 
 #[derive(Debug, Parser)]
 pub enum Cmd {
     /// Add a new network
     Add(add::Cmd),
+
```
willemneal, line 17: Why so many new empty lines? Is this personal preference?

fnando: Because it's very hard to quickly read the file when you have one giant blob without breaklines around specific concerns (the effort is too big).

But yeah, def a personal preference.

cmd/soroban-cli/src/commands/env/mod.rs
fnando: Maybe we can use the rich printer to say that instead of failing, and returning an empty string. 🤔

fnando: As for moving implementing this as a `Display` trait, I think it makes more sense to keep it here, because it also requires env vars, and the conditional to display the value. Wdyt?

willemneal: The printer is a good point. But I do think that `config::new()` maybe should use envs? Currently I guess we are relying on clap to handle it, but it could be handy for other plugins?

fnando: > Currently I guess we are relying on clap to handle it, but it could be handy for other plugins?

That's fair, but I'd defer the work to a separate pr, because it's out of the scope of this one.

willemneal (approved):

PR #1691, opened 2024-10-29
Cleanup unit testing/add ARM support

Ifropc:
### What

Update GH workflow to use Makefile and re-enabled ARM tests. Makefile was also updated to run all workspace tests, that included 4 tests that were previously not run with the job's for loop. Switched instance for ARM Linux tests from X86-64 Ubuntu to ARM Ubuntu. 
Note: binaries for distribution are still build on x86-64 (Ubuntu 20) in order to use an older glibc

### Why

Pipeline improvements 

### Known limitations

N/A

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,16 +45,16 @@ jobs:
       matrix:
         rust: [msrv, latest]
         sys:
-        - os: ubuntu-latest-16-cores
-          target: x86_64-unknown-linux-gnu
+#        - os: ubuntu-latest-16-cores
+#          target: x86_64-unknown-linux-gnu
         - os: ubuntu-latest-16-cores
           target: aarch64-unknown-linux-gnu
-        - os: macos-latest
-          target: x86_64-apple-darwin
-        - os: macos-latest
-          target: aarch64-apple-darwin
-        - os: windows-latest-8-cores
-          target: x86_64-pc-windows-msvc
+#        - os: macos-latest
+#          target: x86_64-apple-darwin
+#        - os: macos-latest
+#          target: aarch64-apple-darwin
+#        - os: windows-latest-8-cores
+#          target: x86_64-pc-windows-msvc
     runs-on: ${{ matrix.sys.os }}
     env:
       CI_TESTS: true
@@ -78,7 +78,9 @@ jobs:
     - run: rustup target add wasm32-unknown-unknown
     - if: runner.os == 'Linux'
       run: |
-        sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev pkg-config libssl-dev:armhf
+        cat /etc/apt/sources.list
+        sudo dpkg --add-architecture arm64
+        sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev pkg-config libssl-dev:arm64
         cfg_path=`pkg-config --variable pc_path pkg-config`
         echo "PKG_CONFIG_PATH=${cfg_path}" >>  $GITHUB_ENV
         echo "PKG_CONFIG_SYSROOT_DIR=/home/runner/.local/sysroot" >>  $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -79,14 +79,7 @@ jobs:
...
     - run: cargo clippy --all-targets --target ${{ matrix.sys.target }}
-    - run: make build-test
-      # TODO: enable ARM linux
-    - if: startsWith(matrix.sys.target, 'x86_64') || runner.os == 'macOS'
-      # specify directories explicitly (otherwise it will fail with missing symbols)
-      run: |
-        for I in cmd/soroban-cli cmd/crates/* cmd/crates/soroban-test/tests/fixtures/test-wasms/hello_world ; do
-          cargo test --target ${{ matrix.sys.target }} --manifest-path $I/Cargo.toml
-        done
+    - run: make test
```
leighmcculloch, line 82: The change to call `make test` instead of `cargo test --target ...` will mean when we run the tests for
```yaml
        - os: macos-latest
          target: x86_64-apple-darwin
        - os: macos-latest
          target: aarch64-apple-darwin
```
We'll actually be running the tests twice for:
```
        - os: macos-latest
          target: aarch64-apple-darwin
        - os: macos-latest
          target: aarch64-apple-darwin
```
I think we still need to specify the target.

I think we also still need to call `make build-test` to build any test wasms.

Maybe a more appropriate thing to do would be to introduce the `--workspace` in the rust.yml, like the change does in the Makefile:
```suggestion
    - run: make build-test
    - run: cargo test --workspace --target ${{ matrix.sys.target }}
```

Ifropc: Hmm I wonder why cargo is not already using an appropriate default target 🤔 is it because runners for x86-64 and ARM are cached and have both 86-64 and ARM targets?
Thanks for the catch, I fixed it with just using env variable. I think it's generally a good idea to keep logic of how we run tests in Makefile so it's easier to run it locally in the same configuration as in CI. 
> I think we also still need to call `make build-test` to build any test wasms.

Should already be called as first step of `make test` 👍🏼

.github/workflows/rust.yml
```diff
@@ -77,16 +77,12 @@ jobs:
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
     - if: runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev
+      run: sudo apt update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev
```
leighmcculloch, line 80: I believe we should be using `apt-get` because it has a stable API for scripting. What's the reason we need to use `apt`?
```suggestion
      run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev
```

.github/workflows/rust.yml
leighmcculloch: Love it 👏🏻

leighmcculloch (approved):

.github/workflows/rust.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         sys:
         - os: ubuntu-latest-16-cores
           target: x86_64-unknown-linux-gnu
-        - os: ubuntu-latest-16-cores
+        - os: ubuntu-jammy-16-cores-arm64
           target: aarch64-unknown-linux-gnu
         - os: macos-latest
           target: x86_64-apple-darwin
```
leighmcculloch, line 53: On GitHub macos-latest is arm64 (equiv to macos-14), whilst macos-12 was the last stable x86 harware. Technically there is also macos-13, but on GitHub macos-13 was never in the "latest" chain.

We could run the x86 build on macos-12, we do that for the `binaries.yml` workflow already. Then the build willrun on native hardware. Functionally it shouldn't matter which we use because macos-14 can run x86 binaries. The only reason to do it would be if we saw better performance by not emulating.

Taking a look at the builds, it doesn't look like we need it. At least the builds both take the same amount of time.

Screenshot 2024-11-06 at 10 59 27 PM

PR #1692, opened 2024-10-30
fix: Add ContractAddress type to consolidate contract id parsing with aliases

willemneal:
New type is similar to Address (which later should be combined since contract addresses can be used with `contract invoke` contract method args.


leighmcculloch (approved):

willemneal:
It seems that this is not needed since it is resolved in the function that fetches the wasm. However, given the work that @fnando is doing with the config it might be worth adding a similar `ContractAddress` type.

Ifropc:
Change seems to be super simple but looks like there are some test failures cause of error types

cmd/soroban-cli/src/config/alias.rs
```diff
@@ -1,8 +1,49 @@
...
 }
+
+/// Address can be either a contract address, C.. or eventually an alias of a contract address.
+#[derive(Clone, Debug)]
+pub enum ContractAddress {
+    ContractId(stellar_strkey::Contract),
+    Alias(String),
+}
+
+impl Default for ContractAddress {
```
Ifropc, line 19: Does ContractAddress need to have a default? 🤔

Ifropc (commented):
Love the addition of the new enum, one small question and LGTM

cmd/soroban-cli/src/config/alias.rs
willemneal: Good question I copy/pasted from address::Address. I'll look into removing it.

leighmcculloch:
Converted to draft because builds seem to be broken.

cmd/soroban-cli/src/config/alias.rs
willemneal: Checked and some of the commands that use it require it.

willemneal:
@Ifropc

Ifropc (approved):
LGTM!

PR #1695, opened 2024-11-01
fix: update stellar dependencies

willemneal:
Still need to update soroban client to a released version. Otherwise just updating to newest release candidates

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,12 +242,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d151e35f61089500b617991b791fc8bfd237ae50cd5950803758a179b41e67a"
 
 [[package]]
-name = "arrayvec"
-version = "0.7.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


leighmcculloch (approved):

Cargo.lock
```diff
@@ -4735,18 +4720,19 @@ dependencies = [
...
  "serde",
  "serde_json",
- "soroban-env-guest 22.0.0-rc.1.1",
- "soroban-env-host 22.0.0-rc.1.1",
- "soroban-ledger-snapshot 22.0.0-rc.1.1",
- "soroban-sdk-macros 22.0.0-rc.1.1",
+ "soroban-env-guest 22.0.0-rc.3",
+ "soroban-env-host 22.0.0-rc.3",
+ "soroban-ledger-snapshot 22.0.0-rc.3",
+ "soroban-sdk-macros 22.0.0-rc.3",
```
leighmcculloch, line 4735: @willemneal The fact versions are included for these soroban deps, means that somehow we're pulling in multiple versions of each. We should be only using one version of these across the whole repo, so some investigation is needed to figure out what is importing the old version and remove it.

PR #1697, opened 2024-11-01
Fix/invoke view only

elizabethengelman:
### What

This allows users to invoke a contract with a view-only transaction to use a default account instead of erroring if the account doesn't exist on the passed network. 

Closes https://github.com/stellar/stellar-cli/issues/1454

### Why
Currently in the CLI, if you pass the `--help` flag for example, but pass an account that doesn't exist on the configured network, you would see an error. With this change, you will not see an error, just the help menu.

### Known limitations

n/a


cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -62,6 +62,15 @@ pub fn build_host_function_parameters(
     }
     cmd.build();
     let long_help = cmd.render_long_help();
+
+    match cmd.clone().try_get_matches_from(slop) {
```
elizabethengelman, line 66: I added this change to point something out/remind myself on Monday - I don't think this should be merged.

It did help me see that when we pass `help` or `version` in the slop, the following `cmd.get_matches_from` (on line 74) will fail and exit the process. This is because in the source code, `get_matches_from` calls `try_get_matches_from` and does an `uwrap_or_else` and exits the process. So I don't think we ever get to `&matches_.remove_subcommand()...` in this case.



This is not a big deal in itself, because the default behavior for this to to then print the help string, which is what we want! But this was a bit challenging to troubleshoot things related to invoke, because I wasn't seeing printlns that I was expecting.

I am leaving this here for now, because I wonder if we should use `cmd.try_get_matches_from` directly and handle the `None` case more explicitly for ✨ _future us_ ✨

elizabethengelman: after thinking about it... i think this is probably good how it works now, i just added a comment so that i'd be able to track things down in the future!

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -163,7 +163,7 @@ impl Cmd {
             .transpose()
     }
 
-    fn send(&self, sim_res: &SimulateTransactionResponse) -> Result<ShouldSend, Error> {
+    fn should_send_tx(&self, sim_res: &SimulateTransactionResponse) -> Result<ShouldSend, Error> {
```
elizabethengelman, line 166: I don't love the names of these fns, but wanted to get this PR up so I can get some additional eyes on this approach

willemneal: Yeah something like should_send_after_sim? names are hard!

willemneal (approved):
Makes sense to me to move the send check to after parsing the arguments!

PR #1698, opened 2024-11-02
fix: remove clap from enforcing network args and update error message

willemneal:
Now the error of missing network args is deferred to when the network is resolved.


Ifropc (approved):
I like this change, I think there shouldn't be any issues with error messages as well, as we usually access the network (where needed) right away

PR #1700, opened 2024-11-05
feat: Add new `Key` type; allowing public keys to be named keys

willemneal:
This will allow users to add public keys with `keys add` so that they can be referenced in commands that need public addresses and contract invoke's that take address arguments.

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #1769 
- <kbd>&nbsp;1&nbsp;</kbd> #1700 👈


cmd/soroban-cli/src/config/key.rs
```diff
@@ -0,0 +1,148 @@
...
+    #[error("failed to parse key {0}")]
+    Parse(String),
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum Key {
+    Secret(Secret),
+    PublicKey { public_key: Public },
+    MuxedAccount { muxed_account: MuxedAccount },
```
leighmcculloch, line 25: Why the different patterns here where enum tuple variant is used sometime and struct variant is used in the other times?

cmd/soroban-cli/src/config/key.rs
```diff
@@ -0,0 +1,148 @@
+use std::{fmt::Display, str::FromStr};
+
+use serde::{Deserialize, Serialize};
+
+use super::secret::{self, Secret};
+use crate::xdr;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Failed to extract public key from secret")]
```
leighmcculloch, line 10:
```suggestion
    #[error("failed to extract public key from secret")]
```

cmd/soroban-cli/src/config/key.rs
```diff
@@ -0,0 +1,148 @@
...
+    SecretPublicKey,
+    #[error(transparent)]
+    Secret(#[from] secret::Error),
+    #[error(transparent)]
+    StrKey(#[from] stellar_strkey::DecodeError),
+    #[error("failed to parse key {0}")]
```
leighmcculloch, line 16: We shouldn't output a key into an error message, we don't know where that is possibly being leaked. Even if it didn't parse.

cmd/soroban-cli/src/config/key.rs
```diff
@@ -0,0 +1,148 @@
...
+        Ok(xdr::MuxedAccount::Ed25519(xdr::Uint256(bytes)))
+    }
+
+    pub fn private_key(
+        &self,
+        hd_path: Option<usize>,
+    ) -> Result<stellar_strkey::ed25519::PrivateKey, Error> {
+        match self {
+            Key::Secret(secret) => Ok(secret.private_key(hd_path)?),
+            _ => Err(Error::SecretPublicKey),
```
leighmcculloch, line 53: It appears this function returns a private key, but when no private key occurs, it returns an error saying that "Failed to extract public key from secret".

cmd/soroban-cli/src/config/key.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn from(Public(stellar_strkey::ed25519::PublicKey(key)): &Public) -> Self {
+        stellar_strkey::ed25519::MuxedAccount {
+            id: 0,
+            ed25519: *key,
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, serde_with::SerializeDisplay, serde_with::DeserializeFromStr)]
+pub struct MuxedAccount(pub stellar_strkey::ed25519::MuxedAccount);
```
leighmcculloch, line 116: I don't think this type is required. FromStr and Display are implemented for MuxedAccount:

https://github.com/stellar/rs-stellar-xdr/blob/715003372ea6380044b5a4a02907ff73e56ae9e7/src/curr/str.rs#L102-L120

Same with Public.

cmd/soroban-cli/src/config/key.rs
willemneal: It is because of how I'm using serde for serialization, e.i. the field names (e.g. `public_key`) determine the case. `Secret` is itself an enum of structs. So if you prefer I could make an enum `Public` which serves the same.

cmd/soroban-cli/src/config/key.rs
willemneal: It's the `serde_with` part which requires the type, allowing serialization via those traits. We could remove this once this is released: https://github.com/stellar/rs-stellar-strkey/pull/72

cmd/soroban-cli/src/config/key.rs
willemneal: Oops!

FULL_HELP_DOCS.md
```diff
@@ -958,6 +958,7 @@ Add a new identity (keypair, ledger, macOS keychain)
 
 * `--secret-key` — Add using `secret_key` Can provide with `SOROBAN_SECRET_KEY`
 * `--seed-phrase` — Add using 12 word seed phrase to generate `secret_key`
+* `--public-key <PUBLIC_KEY>` — Add a public key, ed25519, or muxed account, e.g. G1.., M2..
```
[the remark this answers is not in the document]
willemneal: Yeah we need to also make `install` an alias of upload. So we could do both in one PR.

cmd/soroban-cli/src/config/key.rs
leighmcculloch: That's because we're using `#[serde(untagged)]`. If we used the default, they should be unambiguously tagged which would be the more common way to deal with this issue I think?

willemneal: Fixed it! Should change `Secret` too. But could be in separate PR.

willemneal:
@leighmcculloch I fixed your final issue!

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -228,20 +245,33 @@ impl Args {
...
+    pub fn read_key(&self, key_or_name: &str) -> Result<Key, Error> {
+        if let Ok(key) = self.read_identity(key_or_name) {
+            Ok(key)
         } else {
-            self.read_identity(key_or_name)
+            Ok(key_or_name.parse()?)
+        }
+    }
+
+    pub fn get_private_key(&self, key_or_name: &str) -> Result<Secret, Error> {
```
elizabethengelman, line 260: nit: I wonder about maybe changing this to `get_secret_key`, since it returns a `Secret` and not the private key directly? I don't feel strongly about this, just an idea!

```suggestion
    pub fn get_secret_key(&self, key_or_name: &str) -> Result<Secret, Error> {
```

elizabethengelman (commented):
This looks great!

One thing I'm thinking of is if it would make sense to merge these two PRs before or after the PRs that add the new signers (ledger & keychain), to reduce conflicts. 🤔

cmd/soroban-cli/src/config/locator.rs
leighmcculloch: I generally prefer the term secret over private, because it's harder to mistake or mixup public vs secret, than public vs private which can look like similar words. Also abbreviations also differ conveniently, pk vs sk.

cmd/soroban-cli/src/config/secret.rs
```diff
@@ -39,19 +41,26 @@ pub struct Args {
...
     pub seed_phrase: bool,
+
+    /// Add a public key, ed25519, or muxed account, e.g. G1.., M2..
+    #[arg(long, conflicts_with = "seed_phrase", conflicts_with = "secret_key")]
+    pub public_key: Option<String>,
 }
 
 impl Args {
-    pub fn read_secret(&self) -> Result<Secret, Error> {
+    pub fn read_key(&self) -> Result<Key, key::Error> {
```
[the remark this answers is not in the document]
willemneal: Here it is just for parsing the Arg struct. So it only should be used with the `keys add` command. Though then maybe it should just go into that command? Since there is no reuse of this function.

cmd/soroban-cli/src/config/locator.rs
willemneal: Good points!

cmd/soroban-cli/src/config/secret.rs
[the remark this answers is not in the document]
elizabethengelman: I kind of like the idea of moving it to the `add` command, since that is where it's being used. But not merge blocking!

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -228,20 +245,33 @@ impl Args {
...
 
-    pub fn key(&self, key_or_name: &str) -> Result<Secret, Error> {
-        if let Ok(signer) = key_or_name.parse::<Secret>() {
-            Ok(signer)
+    pub fn read_key(&self, key_or_name: &str) -> Result<Key, Error> {
+        if let Ok(key) = self.read_identity(key_or_name) {
+            Ok(key)
         } else {
-            self.read_identity(key_or_name)
+            Ok(key_or_name.parse()?)
```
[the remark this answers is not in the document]
elizabethengelman: Ah, gotcha. Thanks, that is helpful to understand!

elizabethengelman (approved):
Looks good! Just a couple of small comments/questions.

Cargo.lock
```diff
@@ -4,19 +4,13 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.2"
```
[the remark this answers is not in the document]
elizabethengelman: i just pushed a change to undo some of these changes so that the `cargo deny bans` check passes. i figure we can do dep updates in a follow up PR if needed

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -173,6 +178,18 @@ impl Args {
         KeyType::Identity.write(name, secret, &self.config_dir()?)
     }
 
+    pub fn write_public_key(
+        &self,
+        name: &str,
+        public_key: &stellar_strkey::ed25519::PublicKey,
+    ) -> Result<PathBuf, Error> {
+        self.write_key(name, &public_key.into())
+    }
```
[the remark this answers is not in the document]
elizabethengelman: resolving this so we can get this merged in - we can address this in a follow up pr if needed

willemneal:
Yay! Sorry meant to return to this! I was working on #1774 which is also ready for a review!

PR #1702, opened 2024-11-05
feat: add seq_num as an option for transactions

willemneal:


cmd/soroban-cli/src/fee.rs
```diff
@@ -24,6 +24,8 @@ pub struct Args {
     /// Build the transaction and only write the base64 xdr to stdout
     #[arg(long, help_heading = HEADING_RPC)]
     pub build_only: bool,
+    #[arg(long, visible_alias = "seq-num", help_heading = HEADING_RPC)]
```
Ifropc, line 27: Missing description on the sequence number (should also mention that it's optional)

Ifropc (approved):

FULL_HELP_DOCS.md
```diff
@@ -148,6 +148,7 @@ Deploy builtin Soroban Asset Contract
 * `--cost` — Output the cost execution to stderr
 * `--instructions <INSTRUCTIONS>` — Number of instructions to simulate
 * `--build-only` — Build the transaction and only write the base64 xdr to stdout
+* `--sequence-number <SEQUENCE_NUMBER>` — Optionally set the sequence number for the transaction; default is the current sequence number of the source account incremented by one
```
leighmcculloch, line 151: After playing with this new option, I think we should change the design.

The new option is not really usable where it is in a meaningful way without being coupled with `--build-only`. The reason being is that any use where the tx will be sent immediately really just needs to use the current sequence number plus one. The only need for setting a different sequence number occurs when building and later sending.

Since the option is only usable in that advanced build-only case, I think we should make this something they can do with the `stellar tx ...` commands. It will keep the options advertised smaller for each of the subcommands being changed here, and keeps the advanced functionality for tx modifications in the same place. Maybe as a way to simply change a sequence number `stellar tx set seq`, where those commands could grow to support others like `stellar tx set source-account`. Or maybe by using the `stellar tx edit`.

leighmcculloch (changes requested):
The impl is 💯, but I think we should shift this functionality into another place. See inline for the comment.

willemneal:
I totally agree!

FULL_HELP_DOCS.md
Ifropc: I think `stellar tx edit` is a good place for this ^ 
We can also allow to modify the fee in `tx edit`

leighmcculloch: Maybe. `stellar tx edit` was envisioned as a interactive editor, where I think the thing @willemneal is adding here is more of a non-interactive modification of the transaction. Both things have a place, and for the non-interactive modification I think it's helpful to think about the user api being modifying a single thing, similar to how `stellar tx op add` modifies one thing by adding a single tx and that's why I suggest a `stellar tx set`, something that is separate to the interactive modification.

willemneal:
in favor of https://github.com/stellar/stellar-cli/pull/1760

PR #1707, opened 2024-11-06
Check meta before adding new custom section

elizabethengelman:
### What

Closes https://github.com/stellar/stellar-cli/issues/1694

### Why

### Known limitations

[TODO or N/A]


Ifropc:
Update: added a test to reproduce, have been trying to delete a wasm file in `release/<contract>.wasm` and `release/deps/<contract>.wasm` for metadata to still persists in between builds 🤔 
Will dig deeper into that. Using rustc + cargo 1.82 on 2 different Linux distros (not sure if that matters)

elizabethengelman:
Thank you for adding that test @Ifropc!

leighmcculloch:
This fix appears to address the symptom, avoid adding new meta when the meta is already, rather than addressing the issue that when building on linux without changes the .wasm file doesn't seem to be rewritten. I think we need to further understand what's going on and why before we patch this at another level.

It's also occurred to me while rereading this code that the cli is writing a whole new section per entry, but it should be joining the new entries together and writing them in one new entry, otherwise the wasm is unnecessarily larger due to managing one section per meta.

As an optimisation I think this change would be worthwhile, but after addressing the second paragraph above because as part of addressing it, it'll need to be solved a different way.

Ifropc:
BTW the current fix doesn't work because it only checks for meta in the wasm file with the same name. Ideally it shouldn't even be there in a first place, and I'm looking into why it's there, though it seems that it might be coming somewhere from cargo/rustc itself

elizabethengelman:
> This fix appears to address the symptom, avoid adding new meta when the meta is already, rather than addressing the issue that when building on linux without changes the .wasm file doesn't seem to be rewritten. I think we need to further understand what's going on and why before we patch this at another level.

Good call - that makes sense. I'll close this PR as we understand the root cause.

> It's also occurred to me while rereading this code that the cli is writing a whole new section per entry, but it should be joining the new entries together and writing them in one new entry, otherwise the wasm is unnecessarily larger due to managing one section per meta.

Good point! Does it make sense to create a new issue for this?

leighmcculloch:
Opened a new issue:
- https://github.com/stellar/stellar-cli/issues/1721

PR #1709, opened 2024-11-06
Do not overwrite existing keys when generating new ones.

### LINKED ISSUE #1380: `keys generate`: don't overwrite existing account

fnando:
### What

`stellar keys generate` will no longer overwrite an identity if it already exists, unless `--overwrite` is specified.

Supersedes #1513. Close #1380.

It differs from #1513 in the sense that it doesn't make any additional checks other than requiring the identity to not exist.


cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -52,29 +57,47 @@ pub struct Cmd {
...
 impl Cmd {
     pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let print = Print::new(global_args.quiet);
+
+        if self.config_locator.read_identity(&self.name).is_ok() {
+            if !self.overwrite {
+                return Err(Error::IdentityAlreadyExists(self.name.clone()));
+            }
+
+            print.infoln(format!("Overwriting identity '{}'", &self.name));
```
leighmcculloch, line 75: Instead of the info symbol, maybe we need an ❗ to signify destruction. Wdyt?

leighmcculloch (approved):
One suggestion inline, approved regardless and defer to you on it.

cmd/soroban-cli/src/commands/keys/generate.rs
fnando: I think it's overkill given that you already provided `--overwrite` explicitly. But I'll do it anyway.

leighmcculloch: That's a good point. And we should probably reserve red icons like ❗ for errors given the red => error signal.

fnando: We do have `error` and `errorln` with ❌ though (so we're good).

PR #1720, opened 2024-11-11
fix: migrate from slip10 to slipped10 crate

willemneal:
Fix advisory of older version of ed25519 dalek being used by slip10.


willemneal:
in favor of #1722

PR #1722, opened 2024-11-12
fix/add-exceptions-to-cargo-deny

willemneal:
This pull request addresses issues related to duplicate crates in the `deny.toml` configuration. The following changes have been made:

-  Temporarily allowed duplicates to facilitate the ongoing process of updating upstream crates. This change is necessary as the removal of duplicated crates will be handled in separate pull requests.
  
- Made updates to various crates where it was possible to remove duplicates

- Still need to remove warnings of shared crates not  used as workspace dep


deny.toml
```diff
@@ -252,6 +252,63 @@ deny = [
 
 # Certain crates/versions that will be skipped when doing duplicate detection.
 skip = [
+
+    # Requires updating slipped10 to newest sha2 others are dependents that will be updated
+    { crate = "sha2", reason = "temporary duplicate until upstream updates" },
```
Ifropc, line 257: Do we plan to clean up all of this before the release? Or will some of this dependencies stay past the release?

willemneal: There are a lot of third party dependencies that we need to advocate to update. So my guess is that this would be an on going effort if the deduplication of crates is a priority.

cmd/crates/soroban-test/tests/it/integration/hello_world.rs
```diff
@@ -96,7 +96,7 @@ async fn invoke() {
         .arg("--id")
         .arg(id)
         .assert()
-        .stdout(predicates::str::contains(id).not())
+        .stdout(predicates::str::contains(id))
```
leighmcculloch, line 99: This change is surprising. Wonder why it was 'not' before?

leighmcculloch (approved):

PR #1726, opened 2024-11-12
Generate windows installer.

fnando:
### What

Create installer for Windows.

CleanShot 2024-11-12 at 16 07 35

CleanShot 2024-11-12 at 16 09 35

CleanShot 2024-11-13 at 10 32 47

### Why

This is a first step on making stellar-cli available on winget.

### Known limitations

N/A


.github/workflows/binaries.yml
```diff
@@ -92,3 +101,53 @@ jobs:
...
+        with:
+          name: stellar-cli-${{ env.VERSION }}-x86_64-pc-windows-msvc.tar.gz
+          path: stellar-cli.tar.gz
+
+      - name: Uncompress Artifact
+        run: tar xvf stellar-cli.tar.gz
+
+      - name: Build Installer
+        run: |
+          makensis -V4 -DSTELLAR_CLI_INSTALLER="${{ env.STELLAR_CLI_INSTALLER }}" -DSTELLAR_CLI_VERSION="${{ env.VERSION }}" installer.nsi
```
leighmcculloch, line 132: Can we add in a use of `actions/upload-artifact@v4` to upload the artifact to the GitHub build? It allows us to download it and use it in PRs to either debug it or use it, even if a release hasn't occurred.

leighmcculloch (commented):
One ask otherwise lgtm.

leighmcculloch (commented):
Three more questions.

Did the design folks provide the ico? Looks like the stellar.ico is a white on black icon. It's more common for the black on white icon to be used, so we should probably use it, or check in with the design folks. 

The EnVar plugin is embedded as a binary and binaries are always a risk. Given the cli handles keys, we should be particularly careful about embedding the binary especially if we haven't built it from source ourselves. Even if we have built it ourselves isn't great. What do we need EnVar for? Can we get away without it?

The license for EnVar looks custom, or maybe based off the zlib/libpng licenses. The license looks straightforward to me personally, but we need the legal team to approve it before embedding.

.github/workflows/binaries.yml
fnando: Done!

fnando:
@leighmcculloch this is good to go! Just tested [this installer](https://github.com/stellar/stellar-cli/actions/runs/11823111174) and it worked! 🎉

installer.iss
```diff
@@ -0,0 +1,94 @@
...
+Source: "stellar.exe"; DestDir: "{app}"
+Source: "stellar.ico"; DestDir: "{app}"
+Source: "LICENSE"; DestDir: "{app}"; Flags: ignoreversion
+
+[Icons]
+; Windows optimizes start menu, and removes the uninstall entry. Unless we
+; specify it twice. 🫠
+Name: "{group}\Uninstall Stellar CLI"; Filename: "{uninstallexe}"
+Name: "{group}\Uninstall Stellar CLI"; Filename: "{uninstallexe}"
+Name: "{group}\Stellar Docs"; Filename: "https://stellar.org/docs"
```
leighmcculloch, line 27:
```suggestion
Name: "{group}\Stellar Developer Docs"; Filename: "https://stellar.org/docs"
```

leighmcculloch (approved):
One suggestion, regardless lgtm. 👏🏻

installer.iss
```diff
@@ -0,0 +1,94 @@
...
+    then EnvRemovePath(ExpandConstant('{app}'));
+end;
+
+[UninstallDelete]
+; Remove the Start Menu group
+Type: filesandordirs; Name: "{group}"
+
+; Remove installed files and directory
+Type: files; Name: "{app}\stellar.exe"
+Type: files; Name: "{app}\unins000.exe"
```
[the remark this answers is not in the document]
fnando: Correct, but we can use the placeholder instead. I'll update this.

Issue #1694, opened 2024-10-31
`stellar tx build --meta` creates multiple entries


Ifropc:
### What version are you using?

Latest main 

### What did you do?

Run the same command multiple times:
```
stellar contract build --meta example-key=example-value --out-dir /tmp
```

### What did you expect to see?

Only one meta with `example-key: example-value` is present in the contract

### What did you see instead?

Every time command is executed new meta is appended: 

```
[ifro@nixos:~/projects/soroban-examples/hello_world]$ stellar contract build --meta example-key=example-value --out-dir /tmp
⚠️ A new release of stellar-cli is available: 21.5.0 -> 21.5.3
cargo rustc --manifest-path=Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished `release` profile [optimized] target(s) in 0.09s

[ifro@nixos:~/projects/soroban-examples/hello_world]$ stellar contract deploy --source-account alice --network testnet --wasm /tmp/soroban_hello_world_contract.wasm
⚠️ A new release of stellar-cli is available: 21.5.0 -> 21.5.3
ℹ️ Simulating install transaction…
🌎 Submitting install transaction…
ℹ️ Signing transaction: 54fb1841df752dcc278eea7408763d4cf1e2c7f061f6462542e27bf8c7e22fb1
ℹ️ Using wasm hash 51090ba2e814d119bccdd78e81f454aa694bff5c7ab15945cac65d8425369780
ℹ️ Simulating deploy transaction…
🌎 Submitting deploy transaction…
ℹ️ Transaction hash is 5005c2b7027f67a059c45e7e37feac8fffec2327f0197b2af42d5c6b18a35132
🔗 https://stellar.expert/explorer/testnet/tx/5005c2b7027f67a059c45e7e37feac8fffec2327f0197b2af42d5c6b18a35132
ℹ️ Signing transaction: 5005c2b7027f67a059c45e7e37feac8fffec2327f0197b2af42d5c6b18a35132
🔗 https://stellar.expert/explorer/testnet/contract/CDHJ7TEZPRB2QOOJFGUN32PRZAGGZSP7C2BMW5OFZCS6AYBP63RAPJ7H
✅ Deployed!
CDHJ7TEZPRB2QOOJFGUN32PRZAGGZSP7C2BMW5OFZCS6AYBP63RAPJ7H

[...]

PR #1727, opened 2024-11-13
Add test to reproduce #1694

Ifropc:
### What

Add test to reproduce #1694

### Why

Reproducing test, originally a part of #1707

### Known limitations

N/A


leighmcculloch (approved):
I made a couple tweaks to the way this test was written, so that the stellar-xdr cli is not required and functionality available in the soroban-spec-tools is used instead.

leighmcculloch:
Going to merge this PR since the test passes and it's marked as ready.

PR #1729, opened 2024-11-15
fix(change trust): default limit to i64::Max

willemneal:


leighmcculloch (commented):
Is this good to go?

Ifropc (approved):
Looks like a straightforward change, should we merge?

PR #1732, opened 2024-11-18
Change default network in typescript bindings to testnet

leighmcculloch:
### What
Change default network in typescript bindings to testnet

### Why
Testnet is the most likely network to be using when developing and testing. Futurenet is unstable.

Futurenet was only used here because back when this functionality was first added Soroban was not yet on Testnet.


leighmcculloch:
Superseded by:
- https://github.com/stellar/stellar-cli/pull/1733

Issue #1682, opened 2024-10-21
Fix soroban-test integration tests to make sure the full suite is being run in CI/CD


elizabethengelman:
I noticed that when running `cargo test --package soroban-test --test it` locally, or in a GH Action, not all of the tests seem to run. The output looks like this:
```
running 50 tests
test arg_parsing::parse_bool ... ok
test arg_parsing::parse_i32 ... ok
...
test arg_parsing::parse_u32 ... ok
test arg_parsing::parse_u64 ... ok
test config::generate_key_on_testnet ... ok
test arg_parsing::parse_enum_const ... ok
test arg_parsing::parse_enum ... ok
test arg_parsing::parse_i256 ... ok
test arg_parsing::parse_obj ... ok
test config::seed_phrase ... ok
test config::read_key ... ok
Example contract method which takes a struct
Usage Notes:
Each arg has a corresponding --<arg_name>-file-path which is a path to a file containing the corresponding JSON argument.
Note: The only types which aren't JSON are Bytes and BytesN, which are raw bytes

Usage: strukt_hel [OPTIONS]

Options:
      --strukt <{ a: u32, b: bool, c: Symbol }>
          This is from the rust doc above the struct Test
          Example:
            --strukt '{ "a": 1, "b": true, "c": "hello" }'

  -h, --help
          Print help (see a summary with '-h')
```

The output doesn't include the final test result including the number of failed and passed tests (i.e. `test result: FAILED, 31 passed; 10 failed;...`), nor does it include the status of all of the config tests, or any of the build, init, plugin, or version tests.

After digging a little bit, it seems like something in the `help.rs` file is causing this, though I'm not 100% sure what is going on. It looks like `help.rs` is potentially capturing the stdout (and maybe stderr too), so that the rest of the test output doesn't make it to the command line.

I discovered this because there are a couple of tests that should be failing now, but they aren't. 

### Repo steps
To see this, comment out the tests in `help.rs` and you'll see that there are some failing tests. I'm not sure if all of them are valid failures, but the error in`version.rs` looks legit.

```
➜  stellar-cli git:(main) cargo test --package soroban-test --test it
    Finished `test` profile [unoptimized + debuginfo] target(s) in 17.30s
     Running tests/it/main.rs (target/debug/deps/it-d93d586aa933d606)

running 41 tests
test arg_parsing::parse_bool ... ok
test arg_parsing::parse_i128 ... ok
test arg_parsing::parse_i32 ... ok
[...]

PR #1734, opened 2024-11-18
#1682: fix unit tests

Ifropc:
Part of #1682 

Fixes unit tests, including restoring `help` tests
 
### Why

Previously, we were using clap's `get_matches_from` that exits with code 0 (printing help message). This made our tests exit with code 0 as well (when calling `--help`). 
As part of the change we handle `DisplayError` returning help message, and keeping existing behavior for all other clap errors.   

### Known limitations

Some win tests are failing


leighmcculloch:
> This tests are fixed in other PRs (see #1682 )

What needs to merge before this PR can be merged?

Ifropc:
@leighmcculloch we can either merge all listed PRs at the same time or merge all of them here first
* #1734
* #1736 
* #1737 
* #1738

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -204,13 +217,13 @@ impl Cmd {
 #[async_trait::async_trait]
 impl NetworkRunnable for Cmd {
     type Error = Error;
-    type Result = TxnResult<String>;
+    type Result = Either<String, TxnResult<String>>;
```
elizabethengelman, line 220: I don't know a lot about the `either` crate, but what is the benefit of using it over creating our own enum?

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -63,12 +68,20 @@ pub fn build_host_function_parameters(
     cmd.build();
     let long_help = cmd.render_long_help();
 
-    // get_matches_from exits the process if `help`, `--help` or `-h`are passed in the slop
+    // try_get_matches_from returns an error if `help`, `--help` or `-h`are passed in the slop
     // see clap documentation for more info: https://github.com/clap-rs/clap/blob/v4.1.8/src/builder/command.rs#L631
```
elizabethengelman, line 72:
```suggestion
    // see clap documentation for more info: https://github.com/clap-rs/clap/blob/v4.1.8/src/builder/command.rs#L586
```

elizabethengelman: this is just a change to the line number where it mentions that `try_get_matches_from` doesn't exit the process, but returns an error

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -248,15 +254,18 @@ impl NetworkRunnable for Cmd {
...
-                    .2,
-                )
+                let params = arg_parsing::build_host_function_parameters(
+                    &stellar_strkey::Contract(contract_id.0),
+                    &slop,
+                    &entries,
+                    config,
+                )?;
+                match params {
+                    HostFunctionParameters::Params(p) => Some(p.2),
```
elizabethengelman, line 264: is it possible the `p.2` will ever be `None`? I wonder if it's worth having a default arm in this match in case that happens.

elizabethengelman:
This is looking great! Thanks for digging into it - it was driving me crazy ;p 

> we can either merge all listed PRs at the same time or merge all of them here first

W/r/t when to merge this and the other PRs, i like the idea of merging those test fixing into this PR first.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
Ifropc: No, it shouldn't be (it will indicate contract's --help somehow passed into deploy)

cmd/soroban-cli/src/commands/contract/invoke.rs
Ifropc: We can create a separate enum I originally though it'll only be used in one place and there's much reason of doing so

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -63,12 +68,20 @@ pub fn build_host_function_parameters(
...
+        Ok(mut matches) => &matches.remove_subcommand(),
+        Err(e) => {
+            // to not exit immediately (to be able to fetch help message in tests), check for an error
+            if e.kind() == DisplayHelp {
+                return Ok(HelpMessage(e.to_string()));
+            }
+            e.exit();
+        }
+    }) else {
+        return Ok(HelpMessage(format!("{long_help}")));
```
leighmcculloch, line 84: Instead of passing the Help message up through the stack as an "Ok", can we wrap it in an error, and then unpack that error at the top level before exit is called? That's the pattern clap uses, see https://docs.rs/clap/latest/clap/error/enum.ErrorKind.html#variant.DisplayHelp in its own errors, and is much less confusing than adding a third channel of communication because we'd still have `(Ok, Err)` instead of `((Left, Right), Err)`.

leighmcculloch (commented):
Thanks for figuring out why these tests were preventing from running. Great find and learning for us all.

For the solution, I'd like us to reconsider the approach here, see the inline comment. Please lmk if there's a reason we can't do the error approach.

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
Ifropc: My thoughts were to indicate that it can either return help message or some result (both should be handled to not return error code) -- but I agree that following error pattern is better readability

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -128,26 +128,32 @@ pub enum Error {
...
+                            &contract,
+                            &alias,
+                        )?;
+                    }
+
+                    println!("{contract}");
+                }
+            },
+            Err(ArgParse(HelpMessage(help))) => {
+                println!("{help}");
```
leighmcculloch, line 152: I think we need to do this at the top level, rather than inside the arg parsing, otherwise the exit codes will be incorrect when generating the help when no `--help` is passed. It's a small thing, but also I believe a small fix to keep passing the error as is up the chain, and at the top level doing as you do here.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -131,12 +132,20 @@ impl From<Infallible> for Error {
...
+                TxnEnvelopeResult::TxnEnvelope(tx) => {
+                    println!("{}", tx.to_xdr_base64(Limits::none())?);
+                }
+                TxnEnvelopeResult::Res(output) => {
+                    println!("{output}");
+                }
+            },
+            Err(ArgParsing(HelpMessage(help))) => {
+                println!("{help}");
             }
```
leighmcculloch, line 147: Same comment about doing this at the top level.

leighmcculloch (commented):
One ask, otherwise this looks good, very easy to follow.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
Ifropc: Do you mean we should handle HelpMessage error at the very [top of the chain?](https://github.com/stellar/stellar-cli/blob/61550ec2157e5550dfaa967fa20c2ca9e2de4df2/cmd/soroban-cli/src/cli.rs#L87)

cmd/crates/soroban-test/tests/it/config.rs
```diff
@@ -19,58 +19,35 @@ fn ls(sandbox: &TestEnv) -> Vec<String> {
         .collect::<Vec<_>>()
 }
 
+pub const NETWORKS: &str = "\
+local
+futurenet
+mainnet
+testnet
+";
```
willemneal, line 27:
```suggestion
pub const NETWORKS: &str = r#"
local
futurenet
mainnet
testnet
"#;
```

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -194,7 +194,13 @@ impl Config {
...
     pub fn save(&self) -> Result<(), locator::Error> {
         let toml_string = toml::to_string(&self)?;
-        let mut file = File::create(locator::config_file()?)?;
+        let path = locator::config_file()?;
+        let parent = path.parent();
+        if let Some(parent) = parent {
+            fs::create_dir_all(parent)?;
+        }
+        // Depending on the platform, this function may fail if the full directory path does not exist
+        let mut file = File::create(path)?;
```
willemneal, line 203:
```suggestion
        let path = locator::config_file()?;
        // Depending on the platform, this function may fail if the full directory path does not exist
        let mut file = File::create(locator::ensure_directory(path)?)?;
```

elizabethengelman (approved):
This is looking good to me, besides @leighmcculloch 's comment about moving the HelpMessage switching to the top level. I wonder this PR can go in as-is, and that can be a follow up PR?

I'm also curious what the exit code should be when no `--help` flag is passed. I am currently seeing that scenario, as well as when a `--help` is passing both returning a 0 error code.

PR #1735, opened 2024-11-18
Rollback rust toolchain to 1.81.0

Ifropc:
### What

Rolling back toolchain to 1.81.0

### Why

Clippy emits false positives on 1.82.0

### Known limitations

N/A

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,7 +24,7 @@
             openssl
             pkg-config
             jq
-            (rust-bin.stable.latest.default.override {
+            (rust-bin.stable."1.81.0".default.override {
               extensions = [ "rust-src" ];
               targets = [ "wasm32-unknown-unknown" ];
             })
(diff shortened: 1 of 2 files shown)


leighmcculloch:
We should fix the issues imo rather than pin to the past.

leighmcculloch:
> We also [target](https://github.com/stellar/stellar-cli/blob/main/Cargo.toml#L23) 1.81.0

That is the minimum supported rust version, not the target.

Ifropc:
> We should fix the issues imo rather than pin to the past.

We do use older version in the CI (not sure why) -- as there aren't any clippy failures.
We can pin to 1.81 for now (so clippy at least works locally), and then investigate on both upgrading to the latest in CI and investigating clippy issues (it could be a regression in clippy based on the open issues), WDYT?

Ifropc:
Closing in favor of https://github.com/stellar/stellar-cli/pull/1741

PR #1736, opened 2024-11-18
#1682: fix build tests

Ifropc:
### What

Part of #1682 test fixes. This PR fixes build tests.

### Why

As part of #1594 `CARGO_BUILD_RUSTFLAGS` were added, making existing tests to fail (if `RUSTFLAGS` are not set).
In addition, according to the [documentation](https://doc.rust-lang.org/rustc/command-line-arguments.html#--remap-path-prefix-remap-source-names-in-output) `--remap-path-prefix` should be in format `--remap-path-prefix<SPACE>FROM=TO` while in #1594 it uses format `--remap-path-prefix=FROM=TO`

### Known limitations

Some unrelated tests fail, see #1682


brson:
I just discovered all these failing tests. Thanks for fixing them @Ifropc.

cc @leighmcculloch oops...

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -378,9 +379,11 @@ fn make_rustflags_to_remap_absolute_paths(print: &Print) -> Result<Option<String
     }
 
     let registry_prefix = format!("{cargo_home}/registry/src/");
-    let new_rustflag = format!("--remap-path-prefix={registry_prefix}=");
+    let new_rustflag = format!("{registry_prefix}=");
 
     let mut rustflags = get_rustflags().unwrap_or_default();
+    // --remap-path-prefix is separated by space with the arguments
+    rustflags.push("--remap-path-prefix".to_string());
```
leighmcculloch, line 386: It's completely okay for the option and value to be separated by a `=` instead. This is very commonly supported by most CLI tool frameworks, and it supported by rustc. Conveniently by doing so the key and value can be combined into a single non-space-separated string.

Both work, but I don't see a reason to change the behaviour.

Note that if you pass rustflags to rustc that are incorrect, it'll hard error on the unrecognised option or incorrectly provided option.

So unless there's something broken by the current logic here I don't think we should change it unnecessarily.

Ifropc: Hmm I was checking the documentation and noticed the misalignment between how we do it and how documentation describes the usage, that's it. I can rollback this change and adjust tests.

leighmcculloch (approved):

PR #1738, opened 2024-11-18
1682 fix init test

Ifropc:
### What

As part of #1682 fix init test.
Moves `init_and_deploy` into `./integration` as it uses quickstart image to deploy.
For existing `init` test allow to use `-rc` SDK if our current version is `-rc` itself. 
Update template soroban SDK version to `22-rc3`

### Why

Test fixes

### Known limitations

Some unrelated tests fail, see https://github.com/stellar/stellar-cli/issues/1682


elizabethengelman (approved):
:shipit:

willemneal (approved):

cmd/soroban-cli/src/utils/contract-workspace-template/Cargo.toml.removeextension
```diff
@@ -5,7 +5,7 @@ members = [
 ]
 
 [workspace.dependencies]
-soroban-sdk = "21.0.0"
+soroban-sdk = "22.0.0-rc.3"
```
leighmcculloch, line 8: The CLI should probably only ever generate templates for stable releases, not release candidates. Even if folks are using an RC of the CLI it's unclear to me it should encourages folks to adopt other unstable software releases in other areas. Can we keep this as the stable?
```suggestion
soroban-sdk = "22.0.0"
```

cmd/crates/soroban-test/tests/it/init.rs
```diff
@@ -20,68 +19,12 @@ fn init() {
...
         .assert(predicate::function(|c: &str| {
             let table = toml::from_str::<toml::Table>(c).unwrap();
-            table["workspace"]["dependencies"]["soroban-sdk"].as_str()
-                == Some(&format!("{major}.0.0"))
+            let sdk_version = table["workspace"]["dependencies"]["soroban-sdk"].as_str();
+            println!("Check expected version {major}.0.0 matches template's {sdk_version:?}");
+            if is_rc {
+                sdk_version.and_then(|x| x.split('-').next()) == Some(&format!("{major}.0.0"))
+            } else {
+                sdk_version == Some(&format!("{major}.0.0"))
```
leighmcculloch, line 27: I think we should write an assertion that the generated template is always a stable, non-rc version. See other comment.

cmd/soroban-cli/src/utils/contract-workspace-template/Cargo.toml.removeextension
Ifropc: I think this check was more for our development process, but if we are fine with tests temporarily failing in dev once we bump the version, then I agree we should make it always stable

leighmcculloch: Why would the test fail? I'm suggesting that the init function should generate from a template that is always the latest stable, so that might mean that 22-rc should be initing with an sdk that is 21.

Ifropc: Oh I see what you mean. Previously test was failing because cli version is 22-rc and test checks that the template uses same major version. Should we keep that condition (test will fail on rc though) or maybe allow for -rc to use previous stable version?

PR #1741, opened 2024-11-19
Update minimal rust version to 1.82 + fix clippy 

Ifropc:
### What

After rust clippy 1.82 there have been failures. In CI we are using minimal rust version (set to 1.81) so it's not affected. This PR bumps the version and fix clippy warnings  

### Why

To be consistent with used toolchain

### Known limitations

N/A


Cargo.toml
```diff
@@ -20,7 +20,7 @@ exclude = [
 
 [workspace.package]
 version = "22.0.0-rc.1"
-rust-version = "1.81.0"
+rust-version = "1.82.0"
```
leighmcculloch, line 23: One downside of this that just occurred to me, is that the soroban-sdk only supports up to Rust 1.81, but this change makes it so you need Rust 1.82 to do a cargo install of the stellar-cli, when that version isn't actually required. That will mean devs would need both versions to install from source, which some do.

See https://github.com/stellar/rs-soroban-sdk/blob/9324422a11daa99fdf31d420bb4072dbd875de60/soroban-sdk/build.rs

Can we omit this part of the change?

```suggestion
rust-version = "1.81.0"
```

cmd/soroban-cli/src/wasm.rs
```diff
@@ -42,7 +42,7 @@ pub enum Error {
     #[error(transparent)]
     Rpc(#[from] soroban_rpc::Error),
     #[error("unexpected contract data {0:?}")]
-    UnexpectedContractToken(ContractDataEntry),
+    UnexpectedContractToken(Box<ContractDataEntry>),
```
leighmcculloch, line 45: This change requested by the linter is okay, but it's worth noting it's an optimisation, and not really required to Box it, so we could also simply silence the lint on this type since we know it's okay for the enum to have a variety of size variants.

leighmcculloch (approved):

PR #1745, opened 2024-11-19
feat(ts-bindings): Client.deploy with constructor args

### LINKED ISSUE #1562: CAP-58: Add support to deploy with/without params to typescript bindings

From the bindings perspective this means that the deploy can have required parameters same to how an invoke can have required parameters. So the ability to use the contract interface to determine which parameters are required is needed on the deploy command.

Given the deploy of a contract becomes a typed interface much like an invoke, ideally the typescript contract bindings offer a way to deploy a contract and specify the parameters / functions in the same way a contract call does.

Blocked by:
- https://github.com/stellar/js-stellar-sdk/issues/1036

chadoh:
### What

- allow omiting `--contract-id` flag to `stellar contract bindings
      typescript` command, passing only `--wasm` instead
- if no `--contract-id` is included, then the contract will not be
      fetched from a network, so no network information is logged in this
      case
- this means it's possible to create TS Bindings libraries for contracts
      that have not yet been deployed, which makes it useful to add types
      for `Client.deploy` to the generated TS Bindings
- add fixture example of `test_constructor` from test-wasms

### Why

Fixes #1562 

### Known limitations

`stellar contract bindings typescript` should probably also support `--wasm-hash`. I will do this in a follow-up PR.


cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
```diff
@@ -631,6 +627,22 @@ export interface Client {
 
 }
 export class Client extends ContractClient {
+  static async deploy<T = Client>(
+    /** Constructor/Initialization Args for the contract's `__constructor` method */
+    args: null,
```
willemneal, line 632:
```suggestion
    args: null,
```
I think for this we could not have the `args` parameter at all and then use `null` in the forwarded call.

So perhaps we should add a second example that has args?

chadoh: Good idea!

cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
```diff
@@ -84,34 +87,46 @@ impl NetworkRunnable for Cmd {
...
             let wasm: wasm::Args = wasm.into();
-            wasm.parse()?.spec
+            (wasm.parse()?.spec, None, None, None)
         } else {
+            let contract_id = self
+                .contract_id
+                .as_ref()
+                .ok_or(Error::MissingWasmOrContractId)?;
+
+            let network = self.network.get(&self.locator).ok().unwrap_or_else(|| {
```
Ifropc, line 100: I don't think we assume testnet if no network is provided anywhere else, IMO we should align this behavior across commands (so for now we shouldn't assume testnet, and if we do in the future -- we do it for all commands)

Shaptic (commented):
Generated TypeScript LGTM :call_me_hand: can't speak to the rest of it, though.

cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
chadoh: I agree with that fully, but this is not new behavior. We can address in a separate PR.

kalepail (approved):

willemneal (approved):

Issue #1739, opened 2024-11-18
Support classic accounts and trustlines in snapshots


JakeUrban:
### What version are you using?

22.0.0-rc.1

### What did you do?

Created a snapshot like so:

```
stellar snapshot create --address $myaccount --address $contractid 
```

Then used the snapshot to create an environment for my contract test using `Env::from_ledger_snapshot_file`.

Because `$myaccount` is a Stellar account, and the function tested on `$contractid` results in a SAC transfer, the test failed with the following:

```
topics:[error, Error(Storage, MissingValue)], data:"trying to get non-existing value for contract instance"
```

This appears to be because trustlines are not saved to the snapshot file or are not loaded when creating the environment.

leighmcculloch:
Requires some investigation. Snapshots are supposed to include all the trust lines of an account when an account is specified.

janewang:
@Dhanraj30 Thanks for the fix. @fnando @leighmcculloch Please take a look at this PR  https://github.com/stellar/stellar-cli/pull/1746

PR #1746, opened 2024-11-20
fix:support-classic-accounts-trustlines

Dhanraj30:
#1739
### What

support for classic accounts and trustlines in snapshots. 

### Why

Support classic accounts and trustlines in snapshots. This change is being made to ensure that legacy account and trustline information is correctly included in the ledger snapshots. Classic accounts and trustlines were previously overlooked, which caused issues

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/snapshot/create.rs
+++ b/cmd/soroban-cli/src/commands/snapshot/create.rs
@@ -20,7 +20,7 @@ use stellar_xdr::curr::{
     LedgerKeyClaimableBalance, LedgerKeyConfigSetting, LedgerKeyContractCode,
     LedgerKeyContractData, LedgerKeyData, LedgerKeyLiquidityPool, LedgerKeyOffer,
     LedgerKeyTrustLine, LedgerKeyTtl, Limited, Limits, ReadXdr, ScAddress, ScContractInstance,
-    ScVal,
+    ScVal, TrustLineAsset,
 };
 use tokio::fs::OpenOptions;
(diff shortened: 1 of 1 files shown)


janewang:
@Dhanraj30 Could you fix the coding formatting please

cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -289,7 +289,14 @@ impl Cmd {
                     }
                     let keep = match &key {
                         LedgerKey::Account(k) => current.account_ids.contains(&k.account_id),
-                        LedgerKey::Trustline(k) => current.account_ids.contains(&k.account_id),
+                        LedgerKey::Trustline(k) => {
+                            current.account_ids.contains(&k.account_id) ||
+                            current.account_ids.iter().any(|id| match &k.asset {
+                                TrustLineAsset::CreditAlphanum4(a4) => a4.issuer == *id,
+                                TrustLineAsset::CreditAlphanum12(a12) => a12.issuer == *id,
```
leighmcculloch, line 296: Pulling in every trustline for an asset unfortunately doesn't scale well with the current implementation of snapshots, which was why I left it out. If you pass in the address for a very popular asset, the file becomes huge and unusable.

For assets with much less trust lines though I can see this being very useful so maybe we should add it.

I think this use case will be better served by the following issue that hopefully we can implement in the near future:
- https://github.com/stellar/rs-soroban-sdk/issues/1448

Thoguhts @fnando ?

PR #1757, opened 2024-11-25
Remove outdated npm instruction.

fnando:
### What

This instruction doesn't make sense now that `--frontend-template` has been removed.

### Why

So the output makes sense.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
+++ b/cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
@@ -124,7 +124,7 @@ impl NetworkRunnable for Cmd {
             }
         }
         std::fs::create_dir_all(&self.output_dir)?;
-        let p: Project = self.output_dir.clone().try_into()?;
+        let project: Project = self.output_dir.clone().try_into()?;
         print.infoln(format!("Network: {}", network.network_passphrase));
         let absolute_path = self.output_dir.canonicalize()?;
         let file_name = absolute_path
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
```diff
@@ -134,18 +134,14 @@ impl NetworkRunnable for Cmd {
...
             &contract_address.to_string(),
             &network.rpc_url,
             &network.network_passphrase,
             &spec,
         )?;
         print.checkln("Generated!");
-        print.infoln(format!(
-            "Run \"npm install && npm run build\" in {:?} to build the JavaScript NPM package.",
-            self.output_dir
-        ));
```
leighmcculloch, line 148: We just added this instruction because it is relevant to the typescript bindings, which create an npm package, but the cli no longer installs and builds the package. That is now the job of the user. This command is independent of the frontend template feature that used to exist in the contract init command.

leighmcculloch (changes requested):

cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
fnando: this instruction won't work unless you follow whatever we had on the template; `npm run build` is not a standard task and can be something completely different depending on how you set up your project.

Let's say you have a hello smart contract project, and the dapp lives somewhere else, and you're using `stellar contract bindings typescript --out-dir ../hello-dapp`, then this also is misleading, because I'd run it on the smart contract dir.

fnando: oh, i just got the message. It doesn't mention the current dir, it mentions the output dir, which has everything mentioned on the task. My bad! 🤦‍♂️

PR #1758, opened 2024-11-26
feat: Amount type that allows separating numbers with `_`s

willemneal:
### What

New type which parses strings that include `_` as i64s.

### Why

When entering amounts users will want to use `_`s to better read what they are typing. Also with this approach we could add more complex parsing for example `1 XLM`.

### Known limitations

Allows for any number of `_`s as long as they don't start or end the string.We might want to be stricter. e.g. only allow them where a `,` would go e.g. `10_000_000`


fnando (approved):
lgtm

Ifropc (approved):

PR #1759, opened 2024-11-26
docs: add examples of using `tx new`, `sign`, `send`, and `op add`

willemneal:
Currently relies on unmerged PRs.
- #1758 
- #1663


Ifropc (approved):
I think we should merge this PR before it becomes stale

leighmcculloch (approved):

PR #1760, opened 2024-11-26
feat: add `tx set`

### LINKED ISSUE #1643: Make seq num an option on tx
We don't have to do this in this PR, but we should make sequence number an option, that if isn't set uses the nxt seq num lookup, otherwise uses the value passed in. A good follow up to add later.

_Originally posted by @leighmcculloch in https://github.com/stellar/stellar-cli/pull/1551#discussion_r1783846862_

willemneal:
Currently allows updating fields on Transactions V1 only.

Close #1643

- [x] stellar tx edit source-account set <SOURCE_ACCOUNT>
- [x] stellar tx edit sequence-number set <SEQUENCE_NUMBER>
- [x] stellar tx edit fee set <FEE>
- [x] stellar tx edit memo clear
- [x] stellar tx edit memo set text <MEMO_TEXT>
- [x] stellar tx edit memo set id <MEMO_ID>
- [x] stellar tx edit memo set hash <MEMO_HASH>
- [x] stellar tx edit memo set return <MEMO_RETURN>
- [ ] stellar tx edit time-bound clear
- [ ] stellar tx edit time-bound clear max
- [x] stellar tx edit time-bound set max <MAX_TIME_BOUND>
- [ ] stellar tx edit time-bound clear min
- [x] stellar tx edit time-bound set min <MIN_TIME_BOUND>
- [ ] stellar tx edit ledger-bound clear
- [ ] stellar tx edit ledger-bound min clear
- [ ] stellar tx edit ledger-bound min set <MIN_LEDGER>
- [ ] stellar tx edit ledger-bound max clear
- [ ] stellar tx edit ledger-bound max set <MAX_LEDGER>
- [ ] stellar tx edit min-seq-num clear
- [ ] stellar tx edit min-seq-num set <MIN_SEQ_NUM>
- [ ] stellar tx edit min-seq-age clear
- [ ] stellar tx edit min-seq-age set <MIN_SEQ_AGE>
- [ ] stellar tx edit min-seq-ledger-gap clear
- [ ] stellar tx edit min-seq-ledger-gap set <MIN_SEQ_LEDGER_GAP>
- [ ] stellar tx edit extra-signers clear
- [ ] stellar tx edit extra-signers add <EXTRA_SIGNER>


FULL_HELP_DOCS.md
```diff
@@ -1802,6 +1747,89 @@ Allows issuing account to configure authorization and trustline flags to an asse
...
+* `--source-account <SOURCE_ACCOUNT>` — Change the source account for the transaction
+* `--max-time-bound <MAX_TIME_BOUND>` — Set the transactions max time bound
+* `--min-time-bound <MIN_TIME_BOUND>` — Set the transactions min time bound
+* `--min-ledger <MIN_LEDGER>` — Set the minimum ledger that the transaction is valid
+* `--max-ledger <MAX_LEDGER>` — Set the max ledger that the transaction is valid. 0 or not present means to max
+* `--min-seq-num <MIN_SEQ_NUM>` — set mimimum sequence number
+* `--min-seq-age <MIN_SEQ_AGE>`
+* `--min-seq-ledger-gap <MIN_SEQ_LEDGER_GAP>` — min sequeence ledger gap
+* `--extra-signers <EXTRA_SIGNERS>` — Extra signers
+* `--no-preconditions` — Set precondition to None
```
leighmcculloch, line 1790: Playing around with this interface I think some design discussion is required for this sub-command.

When we previously discussed this (https://github.com/stellar/stellar-cli/issues/1643#issuecomment-2459724928) I had suggested we could have one sub-command per field to set, but I think this arrangement is also reasonable, but looks like there's missing functionality as it is, so we need that design to flesh out all the use cases.

For example, it looks like we can clear all the preconditions with 'no-preconditions', but it's not obvious to a user which fields are being cleared because the options are bundled in a single command the relationship between all the options appear equal / related, when they are not.

Another example, there's no way to remove a memo, or some precondition fields without clearing all preconditions.

I've started a design discussion on the issue:
- https://github.com/stellar/stellar-cli/issues/1643#issuecomment-2510447038

FULL_HELP_DOCS.md
```diff
@@ -1517,6 +1517,8 @@ Sign, Simulate, and Send transactions
 
 * `hash` — Calculate the hash of a transaction envelope
 * `new` — Create a new transaction
+* `edit` — 
+* `set` — Set various options for a transaction
```
elizabethengelman, line 1521: planning to remove this command once the transition to being nested under `edit` is complete

FULL_HELP_DOCS.md
elizabethengelman: @leighmcculloch I've pushed up some changes that I think align with the updated design discussion from #1643 - is this in line with what you had in mind? I have some ideas for refactoring things to DRY them up, but wanted to get some feedback on the UI before I cleaned up.

@fnando @Ifropc would love your feedback on this too.

FULL_HELP_DOCS.md
fnando: I'm currently using `tx edit` to do interactive editing on #1893, but we can discuss how we can make both your and mine ideas work together (I'll add this as discussion topic for today). 

It'd be nice if prs were scoped by the actual feature you're working on, in this case `tx set`. Can you please keep a separate branch with your `tx edit` version in draft mode, and only put the `tx set` changes up for review?

elizabethengelman: > I'm currently using tx edit to do interactive editing on https://github.com/stellar/stellar-cli/pull/1893, but we can discuss how we can make both of our ideas work together (I'll add this as discussion topic for today).

Sounds great!

> It'd be nice if prs were scoped by the actual feature you're working on, in this case tx set. Can you please keep a separate branch with your tx edit version in draft mode, and only put the tx set changes up for review? 

I agree—based on the discussion in #1643, it sounded like there was interest in moving `tx set` under `tx edit` rather than keeping it as a separate command. With that in mind, I was viewing this PR as implementing `tx edit set`.

This PR should definitely still be in draft—nice catch! Would it make sense to close this one and start a new PR for clarity? Happy to discuss further in the meeting later as well!

elizabethengelman:
parent issue has been closed, so closing this as well

PR #1761, opened 2024-11-27
Add Python bindings generation command

overcat:
### What

See https://github.com/stellar/stellar-cli/issues/1633#issuecomment-2500765235

### Why

If this link can be provided in the CLI, it may be easier for users to find it; otherwise, they might not know that Python bindings can be generated.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/bindings/python.rs
+++ b/cmd/soroban-cli/src/commands/contract/bindings/python.rs
@@ -1,0 +1,21 @@
+use std::fmt::Debug;
+
+use clap::Parser;
+use soroban_spec_json;
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("generate json from file: {0}")]
+    GenerateJsonFromFile(soroban_spec_json::GenerateFromFileError),
+}
+
+impl Cmd {
+    pub fn run(&self) -> Result<(), Error> {
+        println!("Python bindings are provided by external library: https://github.com/lightsail-network/stellar-contract-bindings");
+        Ok(())
+    }
+}
--- a/cmd/soroban-cli/src/commands/contract/bindings/python.rs
+++ b/cmd/soroban-cli/src/commands/contract/bindings/python.rs
@@ -1,7 +1,6 @@
 use std::fmt::Debug;
 
 use clap::Parser;
-use soroban_spec_json;
 
 #[derive(Parser, Debug, Clone)]
 #[group(skip)]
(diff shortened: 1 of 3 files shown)


cmd/soroban-cli/src/commands/contract/bindings/python.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("generate json from file: {0}")]
+    GenerateJsonFromFile(soroban_spec_json::GenerateFromFileError),
+}
+
+impl Cmd {
+    pub fn run(&self) -> Result<(), Error> {
+        println!("Python bindings are provided by external library: https://github.com/lightsail-network/stellar-contract-bindings");
+        Ok(())
```
leighmcculloch, line 19: Can you make this always return an error? The reason being is that the stellar-cli command whenever displaying help messaging, without the use of the `--help`, and without taking an action, returns an exit code to the user. This makes it really clear that the command hasn't done anything.

To see what I mean run the `stellar contract bindings` command and you'll see output with an error exit code (non-zero) which you can check with `echo $?` after the command.

In the error definition you can specify the text, and for the text I'd use this:
> python binding generation is not implemented in the stellar-cli, but is available via the tool located here: https://github.com/lightsail-network/stellar-contract-bindings

Note no title case is used on first word for error messages.

leighmcculloch (approved):
One tweak requested, see the inline comment.

I think we should merge this. Today the cli is where folks go to generate bindings, and so it makes sense as part of that story to support discoverability of a commonly used sdk.

---

We haven't started discussions about this yet, but when I saw that you implemented the Python SDK's codegen tool outside the CLI, it prompted me to start thinking about doing the same for the JS SDK. I think the pattern you've used is a good pattern to follow for the following reason. These codegen tools are generating code for using the SDKs, and so they are tightly coupled with the SDKs, and specific versions of the SDKs. We've already found it difficult to coordinate stellar-cli releases with the js-stellar-sdk, and if we moved the JS/TS binding generation from being "cli adjacent" to being "sdk adjacent" then I think it would tell a better maintainer story.

The thing I'm wrestling with is discoverability, the same issue for why this PR exists.

If we do this we wouldn't likely do it before v23. I just wanted to drop some thoughts because I think the setup we have today with bindings built into the cli may not be the setup in the near future, and so this area of code and functionality may change.

cmd/soroban-cli/src/commands/contract/bindings/python.rs
overcat: a0597b360c646fa072b94fe2541a2cfcafac9237

overcat:
Hi @leighmcculloch, I completely agree with your point of view. Another reason I didn't implement it directly in this repository is that, for me personally, developing generation scripts using a scripting language may be more efficient than using Rust. I can write and debug at the same time without having to wait for long compilation times. 😂

leighmcculloch (approved):

PR #1762, opened 2024-11-29
feat: improve asset parsing to use named keys as issuer

willemneal:
Previously the issuer had to be a fully resolved `xdr::AccountId`, now it  can be an `Address`, which then can be resolved. Thus allowing the following: `USDC:circle`

---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #1762 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #1700


cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
```diff
@@ -267,19 +268,23 @@ async fn account_merge_with_alias() {
 #[tokio::test]
 async fn set_trustline_flags() {
     let sandbox = &TestEnv::new();
-    let (test, issuer) = setup_accounts(sandbox);
-    let asset = format!("usdc:{issuer}");
-    issue_asset(sandbox, &test, &asset, 100_000, 100).await;
+    let (test, test1) = setup_accounts(sandbox);
+    let asset = "usdc:test1";
```
elizabethengelman, line 272:
```suggestion
    let asset = format!("usdc:{test1}");
```

just for consistency with line 283, and in case we ever change the name of the issuer in `setup_accounts`

elizabethengelman (approved):
🎉 🎉 🎉

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
willemneal: `test1` here is the full public key. So this will not test the new feature. But if you want we could update `setup_accounts` to allow naming the local account so that this test isn't dependent on `test1` added implicitly.

elizabethengelman: Oh! I see what you're saying. Yeah, I didn't see that `setup_accounts` adds the accounts with a name "test1".

Maybe we can even just update the `test1` var to `test1_address` or something so that it's clear its the full address vs the name. 

Definitely not a big deal, but I can add an updated suggestion.

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
```diff
@@ -267,19 +268,23 @@ async fn account_merge_with_alias() {
 #[tokio::test]
 async fn set_trustline_flags() {
     let sandbox = &TestEnv::new();
-    let (test, issuer) = setup_accounts(sandbox);
-    let asset = format!("usdc:{issuer}");
-    issue_asset(sandbox, &test, &asset, 100_000, 100).await;
+    let (test, test1) = setup_accounts(sandbox);
```
elizabethengelman, line 271:
```suggestion
    let (test, test1_address) = setup_accounts(sandbox);
```

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
```diff
@@ -267,19 +268,23 @@ async fn account_merge_with_alias() {
...
         .arg("asset")
         .arg("deploy")
         .arg("--asset")
-        .arg(&asset)
+        .arg(asset)
         .assert()
         .success();
     let id = contract_id_hash_from_asset(
-        asset.parse::<builder::Asset>().unwrap(),
+        &format!("usdc:{test1}")
```
elizabethengelman, line 283:
```suggestion
        &format!("usdc:{test1_address}")
```

cmd/crates/soroban-test/tests/it/integration/tx/operations.rs
willemneal: Sound to me!

Issue #1764, opened 2024-11-29
Support contract aliases in caller arguments


Ifropc:
### What problem does your feature solve?

When using contract aliases they can't be used in contract arguments (see example)

```bash
[ /tmp/soroban-project ] > echo $MAIN_CONTRACT 
CB4RERBQMEEXGUFPRZDIN46WTMETHR6EMTWRWXIZO6Q3527GTZI3EK7U
[ /tmp/soroban-project ] > echo $ADD_CONTRACT
CCXZXWJ4FXPEG2453SOYSU5BHCVXSAGDJX63TGNCM6KTCOFJ2KJUTQKF
[ /tmp/soroban-project ] > stellar contract alias ls
ℹ️ Aliases available for network 'Test SDF Network ; September 2015'
test_add_contract: CCXZXWJ4FXPEG2453SOYSU5BHCVXSAGDJX63TGNCM6KTCOFJ2KJUTQKF
test_main: CB4RERBQMEEXGUFPRZDIN46WTMETHR6EMTWRWXIZO6Q3527GTZI3EK7U
[ /tmp/soroban-project ] > stellar contract invoke --id $MAIN_CONTRACT --network testnet -- add_with --contract $ADD_CONTRACT --x 9 --y 10
ℹ️ Send skipped because simulation identified as read-only. Send by rerunning with `--send=yes`.
19
[ /tmp/soroban-project ] > stellar contract invoke --id test_main --network testnet -- add_with --contract $ADD_CONTRACT --x 9 --y 10 
ℹ️ Send skipped because simulation identified as read-only. Send by rerunning with `--send=yes`.
19
[ /tmp/soroban-project ] > stellar contract invoke --id test_main --network testnet -- add_with --contract test_add_contract --x 9 --y 10
❌ error: parsing argument contract: value is not parseable to Some(
    Address,
)
```

As a result, it's required to pass contract id as an env variable or just string

### What would you like to see?

CLI to be able to convert contract alias to Address internally when calling contracts 

### What alternatives are there?

Using env variables

estherbreath:
### I would like to work on this issue, please

Here are clues on how I intend to go about it:

1. Implement a feature in the CLI to automatically convert contract aliases to addresses when invoking functions.
2. Add a new internal conversion function that takes an alias and returns the corresponding full address.
3. Modify the contract invocation logic to use this conversion function for alias inputs.
4. Update the CLI's argument parsing to recognize both full addresses and aliases.
[...]

PR #1765, opened 2024-11-30
feat(contract invoke): support contract aliases when parsing address

willemneal:
fixes: #1764

Currently this disallows keys and contract aliases from overlapping and in the case that they already do  there is a warning and the contract is preferred.


cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -246,3 +246,27 @@ pub fn output_to_string(
...
+                contract @ xdr::ScAddress::Contract(_) => return Ok((contract.to_string(), None)),
+            }
+        }
+    };
+    let cmd = crate::commands::keys::address::Cmd {
+        name: addr_or_alias.to_string(),
+        hd_path: Some(0),
+        locator: config.locator.clone(),
+    };
+    Ok((account, cmd.private_key().ok()))
```
leighmcculloch: It looks like this function is really two distinct pieces of logic bundled together. Suggest breaking it up accordingly. I pushed a change to do that in 2bf2a7a96038768a986ffadc1bf6648595b8692a, but feel free to revert it if I'm missing something.

leighmcculloch (approved):

cmd/soroban-cli/src/config/sc_address.rs
```diff
@@ -0,0 +1,57 @@
+use std::str::FromStr;
+
+use crate::xdr;
+
+use super::{address, locator, ContractAddress};
+
+/// `ScAddress` can be either a resolved `xdr::ScAddress` or an alias of a `Contract` or `MuxedAccount`.
+#[derive(Clone, Debug)]
+pub enum ScAddress {
```
leighmcculloch, line 9: It's pretty confusing to me that we have ScAddress that overlaps with name of the actual ScAddress. It's also confusing that the `config::Address` wrapped a muxed and alias.

Naming here is going to be hard.

I suggest we rename both:

- `config::Address` -> `config::MuxedAccountOrAlias`
- `config::ScAddress` -> `config::ScAddressOrAlias`

leighmcculloch (approved):

cmd/soroban-cli/src/config/sc_address.rs
leighmcculloch: Or, actually:
- `config::Address` -> `config::UnresolvedMuxedAccount`
- `config::ScAddress` -> `config::UnresolvedScAddress`

Ifropc (approved):

leighmcculloch (approved):

leighmcculloch (approved):
Lgtm once tests passing

cmd/soroban-cli/src/config/sc_address.rs
```diff
@@ -0,0 +1,61 @@
...
+        let alias = match self {
+            UnresolvedScAddress::Resolved(addr) => return Ok(addr),
+            UnresolvedScAddress::Alias(alias) => alias,
+        };
+        let contract = UnresolvedContract::resolve_alias(&alias, locator, network_passphrase);
+        let muxed_account =
+            super::UnresolvedMuxedAccount::resolve_muxed_account_with_alias(&alias, locator, None);
+        match (contract, muxed_account) {
+            (Ok(contract), _) => Ok(xdr::ScAddress::Contract(xdr::Hash(contract.0))),
+            (_, Ok(muxed_account)) => Ok(xdr::ScAddress::Account(muxed_account.account_id())),
```
willemneal, line 57: Just realized that if the alias for a contract and key are the same, we choose the contract over the key. Is this the correct behavior? Should we return an error? Or do we want to ban the same names across the two namespaces when adding?

Ifropc: I don't think we should ban the same names (logically they belong to different namespaces), maybe we can add a context to this function? `resolve_key()`/`resolve_contract()`. The caller always know if it wants a contract or a key to be returned.

Ifropc: NVM I misunderstood the problem.
Maybe we should indeed error in this case and make user to specify the context directly. 
E.g. if I have aliases `main` for both keys and contract I can pass it as `main@key` or `main@contract` and we can mention this in error message
```bash
$ contract invoke my-contract -- --key main@key
# ok
$ contract invoke my-contract -- --key main@contract
# ok
$ contract invoke my-contract -- --key main
# error: duplicate aliases found for both key and contract, please specify which alias should be used: either main@key or main@contract
```

leighmcculloch: > logically they belong to different namespaces

Actually, I think we should think about them as one namespace, and the issue is they shouldn't overlap. 

It's necessary to think of them as one namespace because there is a single Soroban type that represents them both, an Address, and both can always be passed in, so therefore the alias must be unique across both.

So I think we should prevent using contract names that overlap with key names, and vice-versa.

leighmcculloch: > Or do we want to ban the same names across the two namespaces when adding?

☝🏻 We should do this.

But, given that aliases may already exist that clash, then I think we should also return an error if an alias that refers to either, cannot distinguish between them. The error can output the addresses for both, so that people who get into this situation can just copy-paste the G or C address into where they are trying to use the alias.

Ifropc: > It's necessary to think of them as one namespace because there is a single Soroban type that represents them both, an Address, and both can always be passed in, so therefore the alias must be unique across both.

Hm but don't they belong to 2 different logical entities? (address (G..) vs contract(C..))
Not sure if users think of them as the same entity 
I do agree it's easier to restrict users to not be able to have duplicate aliases

leighmcculloch: They are one logical entity in contracts. Contracts see an "Address" type, and the contents are opaque. While they're separate concepts in the CLI, I don't think we win enough by introducing a new format for people to learn like `myalias@key/contract`.

PR #1770, opened 2024-12-02
Feat/os keychain followup

elizabethengelman:
### What
This PR is based on https://github.com/stellar/stellar-cli/pull/1703 and adds some additional compatibility with the secure store like `keys rm` and `keys add`. 

- [x] `keys rm`
- [x] `keys add`
- [x] **manual test on Mac OS**
     - [x] add the key: `cargo run keys add --secure-store alice` (will be prompted for a seed phrase)
     - [x] make sure we can get the address from the new key: `cargo run keys address alice`
     - [x] make sure we can sign & send with the new key: 
     ```
         stello contract deploy --wasm target/test-wasms/hello_world.wasm  --build-only --network testnet --source alice | \
         stello tx simulate --network testnet --source alice | \
         stello tx sign --network testnet --sign-with-key alice | \
         stello tx send --network testnet
     ```
     - [x] remove the key: `stello keys rm alice`
     - [x] make sure it is no longer in the list or the keychain
- [x] **manual test on Linux**
- [x] **manual test on Windows**

### Why

I wanted to keep https://github.com/stellar/stellar-cli/pull/1703 from getting too big so it was easier to review.

### Known limitations

[TODO or N/A]


cmd/soroban-cli/src/config/locator.rs
```diff
@@ -253,7 +261,22 @@ impl Args {
...
 
-    pub fn remove_identity(&self, name: &str) -> Result<(), Error> {
+    pub fn remove_identity(&self, name: &str, global_args: &global::Args) -> Result<(), Error> {
+        let printer = Print::new(global_args.quiet);
+        let identity = self.read_identity(name)?;
+        if let Secret::SecureStore { entry_name } = identity {
+            let entry = StellarEntry::new(&entry_name)?;
+            match entry.delete_password() {
+                Ok(()) => {}
+                Err(e) if e.to_string() == keyring::Error::NoEntry.to_string() => {
```
willemneal, line 271:
```suggestion
                Err(keyring::Error::NoEntry) => {
```
Is `e` equal to the the error? Does this work?

elizabethengelman: Yeah, this line is a bit weird, thanks for reminding me! The thing I was running into is that `e` an error wrapped in custom errors twice, if that makes sense => locator::Error -> signer::keyring::Error -> keyring::Error. So, matching directly on `keyring::Error::NoEntry` wasn't working.

I just pushed up an update so we're not comparing the error strings. Lmk what you think!

willemneal (commented):
I still need to try this out locally, but correct me if I'm wrong, but it's possible to have the name used in `keys add` differ from what is used in `entry-name`?

elizabethengelman:
@willemneal 
> ... it's possible to have the name used in keys add differ from what is used in entry-name?

Good call! Yeah, at the moment, this is possible. It still ends up working because we are saving the value of `--entry-name` in the `<NAME>.toml` file, like this: `entry_name = "secure_store:org.stellar.cli-<NAME>`. And then getting that `entry_name` from the toml file when accessing the public key, or using the private key. 

Though, this may be confusing. 🤔 


Instead, we could remove the `--entry-name` flag, and when adding a new key, require that the key name is the same as the entry name in the OS's secure store.  So, this command would change to be: `cargo run keys add alice`. 

But, now I'm starting to wonder if we shouldn't allow users to add keys from the keychain at all. If we do, we'd need to make sure that users add the keys to their secure store with:
1. the key name prepended with `secure_store:org.stellar.cli-`
2. the key saved as a base64 encoded `ed25519_dalek::SigningKey`

And we could probably do that, but I'm not sure how necessary this feature is. 🤔

elizabethengelman:
Since we are now planning to store the seed phrase instead of the signing key in the OS secure storage, I think that a more useful feature would be to allow a user to add a new key with their seed phrase, which is saved in secure storage instead of in the local file. So the command could look like this:
```
cargo run keys add <NAME> --secure-store
```
which would then prompt the user to enter their seed phrase or secret key, as is the default behavior of `add`:
```
Type a secret key or 12/24 word seed phrase:
...
```

cmd/soroban-cli/src/secure_store.rs
```diff
@@ -0,0 +1,64 @@
+use sep5::SeedPhrase;
+use stellar_strkey::ed25519::PrivateKey;
+
+use crate::{
+    config::{address::KeyName, locator, secret::{self, Secret}}, print::Print, signer::keyring::{self, StellarEntry}
+};
+
+pub struct SecureStore {}
```
elizabethengelman, line 8: not sure if this needs its own mod, but wanted to pull the code out where both `add` and `generate` could use it

cmd/soroban-cli/src/config/secret.rs
```diff
@@ -42,15 +42,31 @@ pub struct Args {
...
         if let Ok(secret_key) = std::env::var("SOROBAN_SECRET_KEY") {
-            Ok(Secret::SecretKey { secret_key })
+            return Ok(Secret::SecretKey { secret_key });
+        };
+
+        println!("Type a secret key or 12/24 word seed phrase:");
+        let secret_key = read_password()?;
+
+        if self.secure_store {
+            // TODO: if secret_key is a private key, this will return error for now
```
elizabethengelman, line 61: Curious to hear what others think - but I think that we'll be able to let user's create new keys that are stored in the OS secure storage, and the values saved can be either private keys or seed phrases. That is not quite supported yet here, but I think its possible. Does that make sense, or is it too confusing?

The only downside i'm seeing right now is that when we add the ability to fetch the seed phrase for a key, we'll need to return an error if the user entered their private key directly.

cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -31,9 +43,32 @@ pub struct Cmd {
...
     pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
         let print = Print::new(global_args.quiet);
-        let secret = self.secrets.read_secret()?;
+        let secret = self.read_secret(&print)?;
         let path = self.config_locator.write_identity(&self.name, &secret)?;
         print.checkln(format!("Key saved with alias {:?} in {path:?}", self.name));
         Ok(())
     }
+
+    fn read_secret(&self, print: &Print) -> Result<Secret, Error> {
```
elizabethengelman, line 52: moved this here from `config/secret.rs` - i was seeing an interesting cyclical dependency, and moving this here helped that be a bit less confusing.

cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -31,9 +43,32 @@ pub struct Cmd {
...
+        if let Ok(secret_key) = std::env::var("SOROBAN_SECRET_KEY") {
+            return Ok(Secret::SecretKey { secret_key })
+        } else if self.secrets.secure_store {
+            let prompt = "Type a 12 or 24 word seed phrase:";
+            let secret_key = read_password(print, prompt)?;
+
+            let seed_phrase: SeedPhrase = secret_key.parse()?;
```
elizabethengelman, line 59: At the moment I only added support for a user to add a new **seed phrase** to secure storage. I'm happy to make a change so that this supports adding a private key directly as well, but wanted to get some more eyes on this to see if folks thought it was worth the effort.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -156,24 +140,6 @@ impl Cmd {
...
-        if let Ok(key) = entry.get_public_key(None) {
-            print.warnln(format!("A key for {entry_name} already exists in your operating system's secure store: {key}"));
-        } else {
-            print.infoln(format!(
-                "Saving a new key to your operating system's secure store: {entry_name}"
-            ));
-            entry.set_seed_phrase(seed_phrase)?;
-        }
-        Ok(())
-    }
```
elizabethengelman, line 176: This was all moved to a secure_store mod

cmd/soroban-cli/src/signer/secure_store.rs
```diff
@@ -0,0 +1,59 @@
+use sep5::SeedPhrase;
```
elizabethengelman, line 1: i'm not sure if this needs its own mod, or exactly where to put this, but I wanted to pull the code out where both `add` and `generate` could use it

cmd/soroban-cli/src/signer/secure_store.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    if let Secret::SecureStore { entry_name } = &secret {
+        write_to_secure_store(entry_name, seed_phrase, print)?;
+    }
+
+    return Ok(secret);
+}
+
+fn write_to_secure_store(
+    entry_name: &String,
```
willemneal, line 55:
```suggestion
    entry_name: &str,
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -260,7 +262,23 @@ impl Args {
...
+                    signer::keyring::Error::Keyring(keyring::Error::NoEntry) => {
+                        print.infoln("This key was already removed from the secure store. Removing the cli config file.");
+                    }
+                    _ => {
+                        return Err(Error::Keyring(e));
+                    }
+                },
+            }
+        }
         KeyType::Identity.remove(name, &self.config_dir()?)
```
willemneal: Not sure this will be an issue, but it seems like it could be possible for the keyring to successfully delete the key but not the Identity file? I think we can assume this isn't the case since we read the id first.

cmd/soroban-cli/src/commands/keys/add.rs
willemneal: I think it's always a mistake to use a private key directly when storing keys. It has no recovery method and almost all wallets to my knowledge use a seed phrase so if users would be importing it wouldn't be a Secret Key.

willemneal (approved):

willemneal:
🎉

cmd/soroban-cli/src/config/locator.rs
elizabethengelman: 🤔 I see what you're saying. I think you're right that since we are looking up the identity first, we know that the file path is correct, so it _should_ be able to delete cleanly. 

I wonder if its worth removing the identity file first. If that fails, then we return. And then if it succeeds, but the keychain removal fails, at least the identity won't show up in the CLI anymore. That's not ideal, but seems like a less confusing user experience.

elizabethengelman: I'm going to resolve this so we can get this merged in, but happy to keep talking/thinking about it!

PR #1771, opened 2024-12-02
feat(markdown docs): add aliases to options and subcommands

willemneal:
Move `cargo md-gen` to an ignored test. This allows moving clap-markdown to dev-dependency allowing the use of a fork and removing an unneeded dependency since the markdown generation is pre-release.

### DIFF
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -3,4 +3,4 @@
 set -ex
 
 cargo fmt --check || (cargo fmt && exit 1)
-cargo run --bin doc-gen --features clap-markdown
+cargo run --bin doc-gen
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -3,4 +3,6 @@
 set -ex
 
 cargo fmt --check || (cargo fmt && exit 1)
-cargo run --bin doc-gen
+
+echo '+cargo md-gen'
+cargo md-gen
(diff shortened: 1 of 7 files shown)


willemneal:
@leighmcculloch Is it reasonable to make an exception for `cargo-deny` it is a dev dependency under our control and only impacts generating docs. I'm not sure how long it will take my upstream changes to land.

elizabethengelman:
`clap-markdown` merged in [Willem's PR](https://github.com/ConnorGray/clap-markdown/pull/36/files)! So, we can now use the upstream dep instead of our fork. One thing to note is that in a follow up PR on clap-markdown, there was a small change to where the aliases are included in the generated markdown:
> - Print command aliases only in their dedicated section, not in the subcommands listing of the parent command.
> - Move the command aliases out of the command header (simplifying the generated anchor link text) and into a bolded section underneath.

For example, using Willem's fork, the docs for would look like this for `stellar keys public-key`:
```
## `stellar keys`

Create and manage identities including keys and addresses

**Usage:** `stellar keys <COMMAND>`

###### **Subcommands:**

* `add` — Add a new identity (keypair, ledger, OS specific secure store)
* `public-key` **[alias: `address`]** — Given an identity return its address (public key)
...
## `stellar keys public-key` **[alias: `address`]**

Given an identity return its address (public key)

**Usage:** `stellar keys public-key [OPTIONS] <NAME>`
```

But with upstream clap-markdown, it removes the [alias: ...] and adds a Command Alias section at the bottom of each command section. 
```
## `stellar keys`

Create and manage identities including keys and addresses

**Usage:** `stellar keys <COMMAND>`

###### **Subcommands:**

* `add` — Add a new identity (keypair, ledger, OS specific secure store)
* `public-key` — Given an identity return its address (public key)
...

## `stellar keys public-key`

Given an identity return its address (public key)

**Usage:** `stellar keys public-key [OPTIONS] <NAME>`

**Command Alias:** `address`
```

leighmcculloch:
👍🏻 I think it's worth using upstream even with the difference. Thanks @willemneal for contributing that upstream!

PR #1780, opened 2024-12-04
feat(ts-bindings): allow generating from wasm hash

chadoh:
### What

Allow generating TS Bindings from a wasm hash, instead of only contract ID or wasm file path.

### Why

This is useful for generating TS Bindings for wasm files that have been uploaded to the network, but for which no contracts have yet been deployed.

### Known limitations (WIP)

- [x] Await merge of #1745, then retarget this PR to `main`
- [x] Check other commands that allow `--wasm`, `--wasm-hash`, and `--contract-id` and DRY up their code.

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -290,20 +290,21 @@ Generate Rust bindings
 
 Generate a TypeScript / JavaScript package
 
-**Usage:** `stellar contract bindings typescript [OPTIONS] --output-dir <OUTPUT_DIR>`
+**Usage:** `stellar contract bindings typescript [OPTIONS] --output-dir <OUTPUT_DIR> <--wasm <WASM>|--wasm-hash <WASM_HASH>|--contract-id <CONTRACT_ID>>`
 
 ###### **Options:**
 
-* `--wasm <WASM>` — Path to wasm file on local filesystem. You must either include this OR `--contract-id`
-* `--contract-id <CONTRACT_ID>` — A contract ID/address on a network (if no network settings provided, Testnet will be assumed). You must either include this OR `--wasm`
-* `--output-dir <OUTPUT_DIR>` — Where to place generated project
-* `--overwrite` — Whether to overwrite output directory if it already exists
+* `--wasm <WASM>` — Wasm file path on local filesystem. Provide this OR `--wasm-hash` OR `--contract-id`
+* `--wasm-hash <WASM_HASH>` — Hash of Wasm blob on a network. Provide this OR `--wasm` OR `--contract-id`
+* `--contract-id <CONTRACT_ID>` — Contract ID/alias on a network. Provide this OR `--wasm-hash` OR `--wasm`
 * `--rpc-url <RPC_URL>` — RPC server endpoint
 * `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
 * `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
 * `--network <NETWORK>` — Name of network to use from config
 * `--global` — Use global config
 * `--config-dir <CONFIG_DIR>` — Location of config directory, default is "."
+* `--output-dir <OUTPUT_DIR>` — Where to place generated project
+* `--overwrite` — Whether to overwrite output directory if it already exists
 
 
 
@@ -525,13 +526,13 @@ The data outputted by this command is a stream of `SCSpecEntry` XDR values. See 
 
 Outputs no data when no data is present in the contract.
 
-**Usage:** `stellar contract info interface [OPTIONS] <--wasm <WASM>|--wasm-hash <WASM_HASH>|--id <CONTRACT_ID>>`
+**Usage:** `stellar contract info interface [OPTIONS] <--wasm <WASM>|--wasm-hash <WASM_HASH>|--contract-id <CONTRACT_ID>>`
 
 ###### **Options:**
 
-* `--wasm <WASM>` — Wasm file to extract the data from
-* `--wasm-hash <WASM_HASH>` — Wasm hash to get the data for
-* `--id <CONTRACT_ID>` — Contract id or contract alias to get the data for
+* `--wasm <WASM>` — Wasm file path on local filesystem. Provide this OR `--wasm-hash` OR `--contract-id`
+* `--wasm-hash <WASM_HASH>` — Hash of Wasm blob on a network. Provide this OR `--wasm` OR `--contract-id`
+* `--contract-id <CONTRACT_ID>` — Contract ID/alias on a network. Provide this OR `--wasm-hash` OR `--wasm`
 * `--rpc-url <RPC_URL>` — RPC server endpoint
 * `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
 * `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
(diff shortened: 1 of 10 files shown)


cmd/soroban-cli/src/commands/contract/info/shared.rs
```diff
@@ -84,18 +113,26 @@ pub async fn fetch_wasm(args: &Args) -> Result<Option<Vec<u8>>, Error> {
...
+        ))
     } else if let Some(contract_id) = &args.contract_id {
         let contract_id =
             contract_id.resolve_contract_id(&args.locator, &network.network_passphrase)?;
+        let contract_address = xdr::ScAddress::Contract(xdr::Hash(contract_id.0)).to_string();
+        print.globeln(format!("Downloading contract spec: {contract_address}"));
         let res = wasm::fetch_from_contract(&contract_id, network).await;
         if let Some(ContractIsStellarAsset) = res.as_ref().err() {
-            return Ok(None);
+            print.globeln("Contract is a Stellar asset; it has no spec");
```
Ifropc, line 131: Should it just return an error instead?

cmd/soroban-cli/src/commands/contract/info/shared.rs
```diff
@@ -54,23 +75,31 @@ pub enum Error {
...
     Rpc(#[from] soroban_rpc::Error),
     #[error(transparent)]
     Locator(#[from] locator::Error),
 }
 
-pub async fn fetch_wasm(args: &Args) -> Result<Option<Vec<u8>>, Error> {
+pub async fn fetch_wasm(
+    args: &Args,
+    print: &Print,
+) -> Result<(Option<Vec<u8>>, Option<String>, Option<Network>), Error> {
```
Ifropc, line 89: IMO the signature of this function should be
`Result<(Vec<u8>, Option<String>), Error>`: we can always get the network upstream from network args. 
For SAC I'd say we should error and it can be handled upstream (e.g. in the `contract info` we return hardcoded SAC spec)  -> this simplifies the flow and aligns with function names (fetches WASM)

cmd/soroban-cli/src/commands/contract/info/shared.rs
chadoh: We need to return `None` so that the consumer can do the right thing here. For example, the `contract info meta` command will throw with "Stellar asset contract doesn't contain meta information" in the case of SAC.

chadoh: (Also, was trying to keep existing `contract info` behavior unchanged!)

chadoh: So basically, `None` is returned here so that we can throw/return an error in the consumer.

BUT, thanks for asking about this! Now that I looked into it more, I noticed that we were throwing an error in the `bindings typescript` case, thinking we could not get a spec for the SAC. But `contract info interface` showed me how to generate a SAC spec! So `bindings typescript` no longer needs to throw an error in the `None` case here, and there's no reason to log anything here. 💪🏼

cmd/soroban-cli/src/commands/contract/info/shared.rs
chadoh: I was trying to do minimal surgery on `contract info` commands. The `Option<Vec<u8>>` is not new here. If we return an error for the case of SAC, we will need to update `contract info interface` and `contract bindings typescript` logic to inspect the error and consider it not-an-error in this specific case. That seems awkward.

As far as omitting the `Network` and requiring caller to re-derive it, I'm fine with that. But I'm not sure it's a big enough cleanup to be worth it?

FULL_HELP_DOCS.md
```diff
@@ -605,13 +606,13 @@ The data outputted by this command is a stream of `SCEnvMetaEntry` XDR values. S
...
+**Usage:** `stellar contract info env-meta [OPTIONS] <--wasm <WASM>|--wasm-hash <WASM_HASH>|--contract-id <CONTRACT_ID>>`
 
 ###### **Options:**
 
-* `--wasm <WASM>` — Wasm file to extract the data from
-* `--wasm-hash <WASM_HASH>` — Wasm hash to get the data for
-* `--id <CONTRACT_ID>` — Contract id or contract alias to get the data for
+* `--wasm <WASM>` — Wasm file path on local filesystem. Provide this OR `--wasm-hash` OR `--contract-id`
+* `--wasm-hash <WASM_HASH>` — Hash of Wasm blob on a network. Provide this OR `--wasm` OR `--contract-id`
+* `--contract-id <CONTRACT_ID>` — Contract ID/alias on a network. Provide this OR `--wasm-hash` OR `--wasm`
```
leighmcculloch, line 615: Unsure about the change from contract-id to id. That seems like an unnecessary change to make the core thing possible.

cmd/soroban-cli/src/commands/contract/info/shared.rs
```diff
@@ -84,18 +113,25 @@ pub async fn fetch_wasm(args: &Args) -> Result<Option<Vec<u8>>, Error> {
...
+        ))
     } else if let Some(contract_id) = &args.contract_id {
         let contract_id =
             contract_id.resolve_contract_id(&args.locator, &network.network_passphrase)?;
+        let contract_address = xdr::ScAddress::Contract(xdr::Hash(contract_id.0)).to_string();
+        print.globeln(format!("Downloading contract spec: {contract_address}"));
         let res = wasm::fetch_from_contract(&contract_id, network).await;
         if let Some(ContractIsStellarAsset) = res.as_ref().err() {
-            return Ok(None);
+            return Ok((None, Some(contract_address), Some(network.clone())));
```
leighmcculloch, line 131: This return value of the function, both today and before this PR is quite confusing. Before this PR, the function is confusing because it uses an `Option` to describe a union in the `Ok`. It requires a reader to dig into the code and understand it fully to understand the intent of the meaning of the return values
- `Ok(None)` – A SAC contract
- `Ok(Some(...))` – A wasm contract
- `Err(...)` – something went wrong

This PR increases the number of things a reader must understand by reading the code:
- `Ok((Some(...), None, None))` – A wasm contract loaded from file
- `Ok((Some(...), None, Some(...))` – A wasm contract loaded from network, not looked up via a contract ID
- `Ok((Some(...), Some(...), Some(...))) – A wasm contract loaded from network, looked up via a contract ID
- `Err(...)` – something went wrong

---

I don't think this needs fixing in this PR, since it seems things were already unclear, but I think we should follow up with a change where the function is renamed from `fetch_wasm` to `fetch`, and the return is changed to `Result<Fetched, ...>`:
```rust
pub fn fetch(args: &Args, print: &Print) -> Result<Fetched, Error> { ... }

pub struct Fetched {
    contract: Contract,
    source: Source,
}

pub enum Contract {
    Wasm { wasm_bytes: Vec<u8> },
    StellarAssetContract,
}

pub enum Source {
    File { path: PathBuf },
    Wasm { hash: ..., network: Network },
    Contract { contract_id: String, network: Network },
}

impl Source {
    pub fn network(&self) -> Option<&Network> {
        match self {
            FetchedFrom::File => None,
            FetchedFrom::WasmHash { ref network, .. } => Some(network),
            FetchedFrom::Contract { ref network, .. } => Some(network),
        }
    }
}
```

cmd/soroban-cli/src/commands/contract/info/shared.rs
```diff
@@ -84,18 +113,25 @@ pub async fn fetch_wasm(args: &Args) -> Result<Option<Vec<u8>>, Error> {
...
         }
-        res?
+        Ok((Some(res?), Some(contract_address), Some(network.clone())))
     } else {
-        unreachable!("One of contract location arguments must be passed");
-    };
-
-    Ok(Some(wasm))
+        return Err(Error::MalformedWasmOrWasmHashOrContractId);
```
leighmcculloch, line 135: If my suggestion above is executed I believe this error can be removed, since the calling code can confidently handle all the cases because the union will be defined specifically to handle all the cases.

cmd/soroban-cli/src/commands/contract/info/shared.rs
leighmcculloch: I think it's preferred to pass the Network back. Deriving it multiple times could lead to divergence due to a bug, and then confusing output. I left a comment about the return value of the function in another thread below.

FULL_HELP_DOCS.md
leighmcculloch: Ah, I see now `--id` still works because it's an alias, it's just showing up differently here, and clearer now 👏🏻.

leighmcculloch (approved):
👏🏻.

Could you take cleaning up the return value as a follow up?

cmd/soroban-cli/src/commands/contract/info/shared.rs
leighmcculloch: Alternatively pass the network in.

PR #1786, opened 2024-12-11
Add test for remapping paths

### LINKED ISSUE #1704: Add test case for remapping of absolute paths in panic messages
In https://github.com/stellar/stellar-cli/pull/1594 we added a fetature to remap absolute file paths in panic messages (and debuginfo) to relative paths, for reproducibility.

That issue did not land any tests - the feature is complicated to test, but we did discuss methods for testing it.

Also in that thread it was mentioned that we probably screwed up on windows specifically, because the feature as implemented does not work with spaces in paths, and windows home directories commonly contain spaces.

Let's write a test and fix the implementation on windows.

brson:
Closes #1704

### What

Adds a test case that absolute paths embedded in the wasm by rustc are remapped to relative paths, as implemented previously in https://github.com/stellar/stellar-cli/pull/1594

### Why

Previously implemented feature needs tests.

### Known limitations

This will have a minor merge conflict with https://github.com/stellar/stellar-cli/pull/1736

---

This adds the `eth_abi` example as a test fixture. It is the project I use to test the remapping feature. Not all soroban projects exhibit the problem of embedding absolute paths; this test is implemented such that it both confirms that eth_abi exhibits the problem, and stellar-cli fixes it.


leighmcculloch:
@brson Thanks for following up with this test!

janewang:
@brson Thank you for making the fix. Could you please fix the small merge conflict and we will merge this PR. Thank you!

fnando:
merged via #1832

PR #1795, opened 2024-12-13
fix(ts bindings): handle SAC spec

willemneal:
Wasm files have metadata that isn't currently used by typebindings. However, we were creating a higher level Spec (not just  `Vec<ScSpecEntry>`), from the raw bytes of the XDR.

This also adds initialization logic in `cmd/crates/soroban-spec-typescript/ts-tests` and a test to show that we
can generate typebindings from the native asset. We've cleaned up `ts-tests` logic while here, updating "soroban" references with "stellar".


chadoh (approved):
Looks great, thanks for fixing this!

PR #1807, opened 2024-12-19
keys: rename `show` to `secret`

### LINKED ISSUE #1804: keys: rename show to secret
The `stellar keys show` command outputs the secret key.

The name is ambiguous and not easily recognised.

We should rename the command to `stellar keys secret`. The command can be given the alias `show` indefinitely so that the change is backwards compatible.

leighmcculloch:
### What

Rename `keys show` to `keys secret`.

### Why

Close #1804


Ifropc (approved):
I think secret is better than private because we usually use "secret key" instead of "private key" across our docs/sdks/etc.

leighmcculloch:
Private is historically the more common term in asymmetric key management, but some libs, such as the libs in the Rust ecosystem, have used "Secret". "Secret" contrasts the key types in key pairs more clearly, and avoids problems where they are easily confusable, such as "pk" can mean private or public key.

Note that in SDKs like the js-stellar-sdk the secret terminology is used too.

PR #1808, opened 2024-12-19
keys: generate 24-word keys

### LINKED ISSUE #1803: keys: generate 24-word keys
The stellar keys generate command currently generates 12 word seed phrases.

It should generate 24 word phrases, per the recommendation from SEP-5: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0005.md#mnemonic-codes

leighmcculloch:
### What

Generate 24-word keys instead of 12-word.

### Why

Close #1803

### DIFF
--- a/cmd/soroban-cli/src/config/secret.rs
+++ b/cmd/soroban-cli/src/config/secret.rs
@@ -144,7 +144,7 @@ impl Secret {
         let seed_phrase = if let Some(seed) = seed.map(str::as_bytes) {
             sep5::SeedPhrase::from_entropy(seed)
         } else {
-            sep5::SeedPhrase::random(sep5::MnemonicType::Words12)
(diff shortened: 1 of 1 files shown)


willemneal (approved):
Thanks had wanted to do this, but at the time I think that I was working with wallet (freighter I assume) that used 12 words so I did it to make it easier to go back and forth. Also there is a method to go from 24 -> 12,  so it should still be possible to export with 12.

elizabethengelman (approved):

leighmcculloch:
I think we should default to 24 given it's safer (leaves less room for error if someone exposes one or two of their words) and wallets in the ecosystem support accepting 24. Consumer wallets might take a difference stance, but the CLI is targeting a more technical audience who are likely to prefer more secure keys?

cc @tomerweller @fnando  thoughts?

PR #1812, opened 2024-12-19
keys: valid keys when added

### LINKED ISSUE #1805: keys: add does not validate keys
The `stellar keys add <name> --seed-phrase` command accepts a seed phrase, but does not validate it.

You can pass it an 11 word phrase, or a malformed string, and it'll accept it.

For example, you can enter gibberish:

```
$ stellar keys add a4 --seed-phrase
Type a 12 word seed phrase: 
asdflkjaslkdfjklasdjflk
```
```
$ cat ~/.stellar/identity/a4.toml
seed_phrase = "asdflkjaslkdfjklasdjflk"
```

The add command when used with the seed phrase should only allow entering a valid 12 or 24 word key phrase.

leighmcculloch:
### What

Validate secret keys and seed phrases when adding.

### Why

Close #1805


cmd/soroban-cli/src/config/secret.rs
```diff
@@ -23,52 +21,34 @@ pub enum Error {
...
+    InvalidSecretOrSeedPhrase,
     #[error(transparent)]
     Signer(#[from] signer::Error),
 }
 
 #[derive(Debug, clap::Args, Clone)]
 #[group(skip)]
 pub struct Args {
-    /// Add using `secret_key`
+    /// (deprecated) Add secret (S) key
```
willemneal, line 33: Ah so we will just have a `secret` in the future? If so could we add it here? It seems weird to deprecate without an alternative.

cmd/soroban-cli/src/config/secret.rs
```diff
@@ -23,52 +21,34 @@ pub enum Error {
...
-                    .collect::<Vec<_>>()
-                    .join(" "),
-            })
         } else {
-            Err(Error::PasswordRead {})
+            println!("Type a secret key or 24 word seed phrase:");
+            let secret_key = read_password()?;
+            let secret =
+                Secret::from_str(&secret_key).map_err(|_| Error::InvalidSecretOrSeedPhrase)?;
+            Ok(secret)
```
willemneal, line 51:
```suggestion
            secret_key.parse().map_err(|_| Error::InvalidSecretOrSeedPhrase)
```

cmd/soroban-cli/src/config/secret.rs
leighmcculloch: I think we can accept both as inputs without using any options. Wdyt?

elizabethengelman: This change seems like it simplifies things!

I wonder if a bit more explanation in the doc comments & deprecations would help to let the user know that they will now be prompted to enter their key/passphrase. I was a bit confused at first how this would work, until I tried it out. 

Something like this maybe?
```suggestion
    /// (deprecated) Enter secret (S) key when prompted
```

elizabethengelman (approved):

cmd/soroban-cli/src/config/secret.rs
Ifropc: I like this change, but there should be a clear guidance of using `SOROBAN_SECRET_KEY` in help.
```
 > stellar keys generate --help
Generate a new identity with a seed phrase, currently 12 words
```

leighmcculloch: > > stellar keys generate --help
> Generate a new identity with a seed phrase, currently 12 words

- https://github.com/stellar/stellar-cli/issues/1816

Issue #1824, opened 2025-01-08
Support exporting seed phrases


elizabethengelman:
Based on this PR comment: https://github.com/stellar/stellar-cli/pull/1703#discussion_r1905619434

We should allow a user to export their seed phrase so they can use it in another wallet.

PR #1828, opened 2025-01-13
Add support exporting seed phrases

Dhanraj30:
issue: #1824 
### What

Add functionality to export seed phrases in the CLI.

### Why

To allow users to securely export their seed phrases for backup and recovery purposes.

### Known limitations

[ N/A]

### DIFF
--- a/cmd/soroban-cli/src/commands/keys/generate.rs
+++ b/cmd/soroban-cli/src/commands/keys/generate.rs
@@ -52,10 +52,23 @@ pub struct Cmd {
     /// Fund generated key pair
     #[arg(long, default_value = "false")]
     pub fund: bool,
+
+    /// Export seed phrase
+    #[arg(long)]
+    pub export_seed: bool,
 }
 
 impl Cmd {
     pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        if self.export_seed {
+            // Handle export case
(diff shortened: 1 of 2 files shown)


leighmcculloch (commented):
This PR adds a way to export (print) the seed phrase at the point in time when it is generated.

However, that only allows accessing the seed phrase at the time of generation, and it's reasonable to expect to be able to export the seed phrase at a later time.

The cli already has a comand for exporting the key, the `stellar keys show` command, but the command only outputs the key in the `S` strkey form.

We could add an option to the `show` subcommand that causes it to output in the seed phrase format.

Thoughts @Dhanraj30 @stellar/devx ?

Dhanraj30:
> This PR adds a way to export (print) the seed phrase at the point in time when it is generated.
> 
> However, that only allows accessing the seed phrase at the time of generation, and it's reasonable to expect to be able to export the seed phrase at a later time.
> 
> The cli already has a comand for exporting the key, the `stellar keys show` command, but the command only outputs the key in the `S` strkey form.
> 
> We could add an option to the `show` subcommand that causes it to output in the seed phrase format.
> 
> Thoughts @Dhanraj30 @stellar/devx ?

Thank you for the suggestion. I will check the feasibility of adding this option and proceed with the implementation and testing accordingly

PR #1839, opened 2025-01-21
Add an explanatory comment to the generated file.

fnando:
### What

Add an explanatory comment to the generated file.

### Why

So users can have a better idea of what's being generated.

### Known limitations

n/a

### DIFF
--- a/cmd/soroban-cli/src/utils/contract-template/src/lib.rs
+++ b/cmd/soroban-cli/src/utils/contract-template/src/lib.rs
@@ -4,6 +4,12 @@ use soroban_sdk::{contract, contractimpl, vec, Env, String, Vec};
 #[contract]
 pub struct Contract;
 
+// This is a sample contract. Replace this placeholder with your own contract logic.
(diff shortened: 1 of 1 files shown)


Ifropc (approved):

cmd/soroban-cli/src/utils/contract-template/src/lib.rs
```diff
@@ -4,6 +4,12 @@ use soroban_sdk::{contract, contractimpl, vec, Env, String, Vec};
 #[contract]
 pub struct Contract;
 
+// This is a sample contract. Replace this placeholder with your own contract logic.
+// A corresponding test example is available in `test.rs`.
+//
+// For comprehensive examples, visit <https://github.com/stellar/soroban-examples>.
```
Ifropc, line 10: I think we should also link stellar docs' tutorial here 
https://developers.stellar.org/docs/build/smart-contracts

fnando: Good call. Just added it.

PR #1852, opened 2025-01-28
Set local network as the default for container commands.

### LINKED ISSUE #1848: Make stellar container start and stop commands work with local by default
Currently, the `stellar container start` and `stellar container stop` commands require a network to be passed as an argument.

We should make the commands do something useful without the user needing to learn more, by defaulting the network to `local` when no network is specified.

fnando:
### What

Set local network as the default for container commands.

### Why

Close #1848

### Known limitations

N/A


Ifropc (approved):

leighmcculloch:
🥳

leighmcculloch:
Once this is released, can we merge the following?
- https://github.com/stellar/quickstart/pull/664

PR #1854, opened 2025-01-28
Add short flags for network and source account.

### LINKED ISSUE #794: Add shortcuts for arg flags
My fingees can't keep typing `--network`, `--source`, etc. I want `-n`, `-s`, etc

fnando:
### What

Add short flags for network and source account.

### Why

Close #794.

### Known limitations

N/A


Ifropc:
IIRC we decided not to add any new short flags, personally I think `-n` and `-s` make sense (as it's most commonly used options)
@janewang @leighmcculloch thoughts?

elizabethengelman (approved):
I remember that discussion about not using short flags as well, I think the reasoning was because we may need a short flag for an often used option, and we don't want to have that one unavailable. 

But I agree that `-n` and `-s` both make sense! They are needed for so many commands.

leighmcculloch (approved):

Ifropc (approved):

PR #1857, opened 2025-01-29
Update template contract to be the same as soroban-examples

elizabethengelman:
### What

This is an update to the Hello World contract that is created on `stellar contract init` to be more simliar to what is in [soroban-examples](https://github.com/stellar/soroban-examples/tree/main/hello_world).

### Why

After reviewing the Getting Started guide in the docs, I noticed that the contract that we're creating when running `stellar contract init` is different than what is in soroban-examples and perhaps it would be less confusing for users if these were the same.

### Known limitations

Related PR: https://github.com/stellar/stellar-docs/pull/1239

### DIFF
--- a/cmd/soroban-cli/src/utils/contract-template/src/lib.rs
+++ b/cmd/soroban-cli/src/utils/contract-template/src/lib.rs
@@ -2,7 +2,7 @@
 use soroban_sdk::{contract, contractimpl, vec, Env, String, Vec};
 
 #[contract]
-pub struct Contract;
+pub struct HelloContract;
 
 // This is a sample contract. Replace this placeholder with your own contract logic.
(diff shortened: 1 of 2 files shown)


fnando:
We should actually do the other way around. Changing to a generic Contract was the goal for new contracts.

elizabethengelman:
> We should actually do the other way around: update examples and docs to use Contract. Changing to a generic Contract was the goal for new contracts.

Ah, that makes sense. Thanks for the context! I have a [PR](https://github.com/stellar/stellar-docs/pull/1185) over on the docs that updates `HelloWorld` -> `Contract` that was just merged so docs should be good. And, I'll close this PR, and will plan to open one on the examples repo as well.

PR #1862, opened 2025-01-30
feat: add vscode subcommands for adding/removing txn schema from vscode

willemneal:
This PR adds support for VSCode to recognize any file ending in `.stellar_txn.json` as a stellar transaction envelope (very open to renaming this!)

- `vscode setup` - adds a json schema file in the folder of settings for VSCode. In the future we can publish it to an endpoint and then the link in the settings will be a url. It also uses jsonc-parser to parse and update the `settings.json` since it uses comments which need to preserved.

- `vscode cleanup` - is unimplemented as it isn't needed in to complete the goal and is rather a nice feature to have, but leaving the setting in place does not impact any other settings. However, it would be nice to have this as it would allow for updating the schema (which if it was a url would be automatic)

Note: Currently the schema is a fixture in the binary. However, it once the xdr cli makes the schema generation public, see: https://github.com/stellar/rs-stellar-xdr/pull/409

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2662,6 +2662,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "jsonc-parser"
+version = "0.26.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b558af6b49fd918e970471374e7a798b2c9bbcda624a210ffa3901ee5614bc8e"
+
+[[package]]
 name = "jsonrpsee-core"
 version = "0.20.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


leighmcculloch:
Can these settings be controlled by an extension? I would assume so, but it's been a while since I've done anything with vscode extensions and I don't remember if you can control arbitrary settings like that.

fnando:
Yes, that works: https://code.visualstudio.com/docs/languages/json#_mapping-a-schema-in-an-extension 

Another alternative would be adding `$schema` to the json. That means editors (including vscode) will be able to download the schema and use the completion. If you want to try, create a new json file and paste this code:

```json
{
  "$schema": "https://json.schemastore.org/coffeelint"
}
```

leighmcculloch:
TIL. So we could publish the output of `stellar xdr types schema --type ...` with each rendered JSON from the rs-stellar-xdr =O

PR #1863, opened 2025-01-30
Add features for guarding code based on the P23.

fnando:
### What

Add features for guarding code based on the P23.

### Why

The idea is to allow adding/removing code without having to keep a long running branch for either the current or next version.

This approach also allow us to easily enable features from the next protocol version when running tests, but using cargo hack.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/Cargo.toml
+++ b/cmd/soroban-cli/Cargo.toml
@@ -36,6 +36,8 @@ doctest = false
 
 [features]
 default = []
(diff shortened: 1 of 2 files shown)


elizabethengelman (commented):
This looks great! I'll update https://github.com/stellar/stellar-cli/pull/1861 to use the `version_lt_23` feature

Ifropc (approved):

PR #1869, opened 2025-02-05
Fix pre push hook

fnando:
### What

`#[allow(clippy::too_many_arguments)]` was being ignored, so let's use a module rule instead.

### Why

So pre-commit hook works as expected.

### Known limitations

N/A

### DIFF
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -1,9 +1,6 @@
 #!/bin/sh
 
-set -e
+set -ex
 
-echo '+cargo fmt --check'
 cargo fmt --check || (cargo fmt && exit 1)
-
-echo '+cargo run --bin doc-gen --features clap-markdown'
 cargo run --bin doc-gen --features clap-markdown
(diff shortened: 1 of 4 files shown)


leighmcculloch (approved):

leighmcculloch:
Do folks use the pre-push hook? I have my gitconfig configured to disable / prevent / disallow any hooks from executing, because personally I don't want commands auto-executing and I prefer to have CI alert me to issues.

willemneal:
I use it. Personally have the opposite opinion as @leighmcculloch. I only do the push, so that I get one check before wasting CI time on something that is trivial and usually it's the generated docs or `cargo fmt` which I can then amend to the last commit. And using `--no-verify` you can skip it if you know it's not needed.

leighmcculloch:
> And using `--no-verify` you can skip it

Can also skip it with:
```
git config set --global core.hooksPath no-hooks
```

leighmcculloch:
> Should we remove it entirely?

Not if folks are using it. I was just wanting to check if anyone was.

PR #1874, opened 2025-02-12
Fix cargo-deny issues

Ifropc:
### What

* Update hickory version
* Allow [multiple dependency versions](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-multiple-versions-field-optional) (set to `warn` -- default config)

### Why

https://rustsec.org/advisories/RUSTSEC-2025-0006

### Known limitations

N/A

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -131,7 +131,7 @@ registries = [
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
 [bans]
 # Lint level for when multiple versions of the same crate are detected
-multiple-versions = "deny"
+multiple-versions = "warn"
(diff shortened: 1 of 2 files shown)


deny.toml
```diff
@@ -131,7 +131,7 @@ registries = [
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
 [bans]
 # Lint level for when multiple versions of the same crate are detected
-multiple-versions = "deny"
+multiple-versions = "warn"
```
leighmcculloch, line 134: Instead of warn, can you add the duplicate dependency version to the list below? That's what we've been doing.

leighmcculloch (approved):

PR #1876, opened 2025-02-12
Fix arm cross compilation

Ifropc:
### What

Change instance of arm builder to `ubuntu-jammy-16-cores-arm64` 

### Why

Failing cross-compilation build on x64

### Known limitations

glibc will be updated to [2.35](https://packages.ubuntu.com/source/jammy/glibc) from [2.31](https://packages.ubuntu.com/source/focal/glibc) for ARM build

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -45,8 +45,8 @@ jobs:
     - run: rustup update
     - run: rustup target add ${{ matrix.sys.target }}
 
-    - if: matrix.sys.target == 'aarch64-unknown-linux-gnu'
-      run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev libdbus-1-dev
+    - if: runner.os == 'Linux'
+      run: sudo apt-get update && sudo apt-get -y install libudev-dev libdbus-1-dev
 
     - name: Setup vars
       run: |
@@ -66,11 +66,8 @@ jobs:
         echo "BUILD_WORKING_DIR=target/package/${{ matrix.crate.name }}-$VERSION" >> $GITHUB_ENV
 
     - name: Build
-      env:
-        CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: aarch64-linux-gnu-gcc
       working-directory: ${{ env.BUILD_WORKING_DIR }}
       run: |
-        sudo apt-get update && sudo apt-get -y install libdbus-1-dev
         cargo build --target-dir="$GITHUB_WORKSPACE/target" --package ${{ matrix.crate.name }} --features opt --release --target ${{ matrix.sys.target }}
 
     - name: Build provenance for attestation (release only)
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Does changing the instance to `ubuntu-22.04` make it pass? The issue might just be with us using a really old ubuntu version, that GitHub won't be supporting soon anyway.

Note that we should pick the oldest ubuntu we can so that the compiled binary works on more versions of linux.

Ifropc:
Yes, `22.04` is the oldest native ARM image available. Cross-compiling on 20.04 is possible, but challenging (the [last time I tried it](https://github.com/stellar/stellar-cli/pull/1691/commits)). We can look into cross-compiling to 20.04 in the future, but for now I think we should natively compile on `22.04-arm` for simplicity

leighmcculloch:
I mean does it work with the ubuntu-22.04 GitHub runner which is amd64.

Ifropc:
I can try, but it shouldn't work -- we don't install arm dependencies properly for ledger

.github/workflows/binaries.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         sys:
           - os: ubuntu-20.04 # Use 20.04 to get an older version of glibc for increased compat
             target: x86_64-unknown-linux-gnu
-          - os: ubuntu-20.04 # Use 20.04 to get an older version of glibc for increased compat
+          - os: ubuntu-22.04-arm # Use 22 to get an older version of glibc for increased compat
```
elizabethengelman, line 33: Since this change is removing the need for cross compilation, can we remove some of the setup that was supporting that?

```diff
    steps:
    - uses: actions/checkout@v4
    - run: rustup update
    - run: rustup target add ${{ matrix.sys.target }}
-    - if: matrix.sys.target == 'aarch64-unknown-linux-gnu'
-      run: sudo apt-get update && sudo apt-get -y install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libudev-dev libdbus-1-dev
    - name: Setup vars
      run: |
        version="$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "stellar-cli") | .version')"
        echo "VERSION=${version}" >> $GITHUB_ENV
        echo "NAME=${{ matrix.crate.name }}-${version}-${{ matrix.sys.target }}" >> $GITHUB_ENV
    - name: Package (release only)
      if: github.event_name == 'release'
      run: cargo package --no-verify --package ${{ matrix.crate.name }}
    - name: Package Extract (release only)
      if: github.event_name == 'release'
      run: |
        cd target/package
        tar xvfz ${{ matrix.crate.name }}-$VERSION.crate
        echo "BUILD_WORKING_DIR=target/package/${{ matrix.crate.name }}-$VERSION" >> $GITHUB_ENV
    - name: Build
-      env:
-        CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: aarch64-linux-gnu-gcc
      working-directory: ${{ env.BUILD_WORKING_DIR }}
      run: |
-        sudo apt-get update && sudo apt-get -y install libdbus-1-dev 
+        sudo apt-get update && sudo apt-get -y install libdbus-1-dev libudev-dev
        cargo build --target-dir="$GITHUB_WORKSPACE/target" --package ${{ matrix.crate.name }} --features opt --release --target ${{ matrix.sys.target }}
```

elizabethengelman (approved):
Looks good to me, just a comment inline about removing some of the cross compilation setup.

Ifropc:
@leighmcculloch @fnando I'll merge this PR to unblock failing builds on main, if we want to explore alternative solutions let's do it as part of another task/pr

PR #1878, opened 2025-02-17
feat: Add a stellar contract info wasm-hash command

### LINKED ISSUE #1877: Add a `stellar contract info wasm` command

It would be handy for devs to be able to retrieve the wasm hash of a contract.

```
$ stellar contract info wasm --network testnet --contract-id CDSJWGUPY2HFHYNLT3EZX7YGJ4X7YMWPQWHHGNKB6IVH6WRXAJFI7627
283f8e42a3585aa3a511c06c7d09a94d66762773cdf493cc4afc455d7b68adcf
```

There are some other ways to get the Wasm hash that a contract ID is deployed with.

#### Using RPC + XDR

The RPC and any XDR decode can be used to manually retrieve the details, however this is not convenient for a developer needing the value at the command line. For example:

```
curl -sSLX POST https://soroban-testnet.stellar.org \
  --header 'content-type: application/json' \
  --data '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getLedgerEntries",
    "params": {
      "keys": ["'"$(echo '{
        "contract_data": {
          "contract": "CDSJWGUPY2HFHYNLT3EZX7YGJ4X7YMWPQWHHGNKB6IVH6WRXAJFI7627",
          "key": "ledger_key_contract_instance",
          "durability": "persistent"
        }}' | stellar xdr encode --type LedgerKey)"'"]
    }
  }' | jq -r '.result.entries[0].xdr' | \
  stellar xdr decode --type LedgerEntryData | \
  jq -r '.contract_data.val.contract_instance.executable.wasm'
283f8e42a3585aa3a511c06c7d09a94d66762773cdf493cc4afc455d7b68adcf
```

#### Using log output

I'm pretty sure there are some commands in the cli where the wasm hash is outputted when retrieving the wasm for other purposes. I can't remember which one it is at the moment, but these commands exist for other purposes an the hash is only outputted in informational stderr logs, which is not accessible for scripting.

Dhanraj30:
### What
Closes: #1877 
This PR adds a new command wasm-hash to the stellar contract CLI, allowing developers to retrieve the Wasm hash of a deployed contract directly from the command line.

### Why
Currently, developers have to rely on more complex methods, such as using RPC calls or parsing logs, to obtain the Wasm hash of a contract. This new command simplifies the process, making it more convenient and efficient for developers to access the Wasm hash directly. It enhances the usability of the CLI by providing a straightforward command that outputs the hash in a format suitable for scripting and automation.

Get the Wasm hash of a deployed contract.

#### Usage
```bash
stellar contract wasm-hash --network <NETWORK> --contract-id <CONTRACT_ID>
```

#### Arguments
- `--network`: The network to use (e.g., `testnet`, `futurenet`, `mainnet`, or custom URL)
- `--contract-id`: The ID of the deployed contract to get the Wasm hash for

#### Example
```bash
# Get the Wasm hash for a contract on testnet
stellar contract wasm-hash --network testnet --contract-id CDSJWGUPY2HFHYNLT3EZX7YGJ4X7YMWPQWHHGNKB6IVH6WRXAJFI7627

# Output:
283f8e42a3585aa3a511c06c7d09a94d66762773cdf493cc4afc455d7b68adcf
```

#### Notes
- The contract must be deployed on the specified network
- The hash is returned in hexadecimal format
- The command requires a network connection to retrieve the contract data
```

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/info.rs
+++ b/cmd/soroban-cli/src/commands/contract/info.rs
@@ -6,6 +6,7 @@ pub mod env_meta;
 pub mod interface;
 pub mod meta;
 pub mod shared;
+pub mod wasm_hash;
 
 #[derive(Debug, clap::Subcommand)]
 pub enum Cmd {
@@ -49,6 +50,9 @@ pub enum Cmd {
     ///
     /// Outputs no data when no data is present in the contract.
     EnvMeta(env_meta::Cmd),
+
+    /// Get the wasm hash of a deployed contract
+    WasmHash(wasm_hash::Cmd),
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -59,6 +63,8 @@ pub enum Error {
     Meta(#[from] meta::Error),
     #[error(transparent)]
     EnvMeta(#[from] env_meta::Error),
+    #[error(transparent)]
+    WasmHash(#[from] wasm_hash::Error),
 }
 
 impl Cmd {
@@ -67,6 +73,7 @@ impl Cmd {
             Cmd::Interface(interface) => interface.run(global_args).await?,
             Cmd::Meta(meta) => meta.run(global_args).await?,
             Cmd::EnvMeta(env_meta) => env_meta.run(global_args).await?,
+            Cmd::WasmHash(cmd) => cmd.run(global_args).await?,
         };
         println!("{result}");
         Ok(())
--- a/cmd/soroban-cli/src/commands/contract/info/wasm_hash.rs
+++ b/cmd/soroban-cli/src/commands/contract/info/wasm_hash.rs
@@ -1,0 +1,73 @@
+use clap::{arg, command, Command, ArgMatches};
+use soroban_rpc::client::Client;
+use soroban_xdr::{ScAddress, ScVal, DecodeError};
+use stellar_xdr::XdrCodec;
+use crate::config::Config;
+use crate::error::Result;
+use clap::Parser;
+use stellar_xdr::curr as xdr;
+
+use crate::commands::{config::network, global};
+use crate::config::locator;
+use crate::rpc;
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub config_locator: locator::Args,
+
+    #[command(flatten)]
+    pub network: network::Args,
+
+    /// Contract ID to get the wasm hash for
+    #[arg(long = "contract-id")]
+    pub contract_id: stellar_strkey::Contract,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Config(#[from] locator::Error),
+    
+    #[error(transparent)]
+    Network(#[from] network::Error),
+    
+    #[error(transparent)]
+    Rpc(#[from] rpc::Error),
+}
+
+impl Cmd {
+    pub async fn run(&self, _global_args: &global::Args) -> Result<(), Error> {
+        let network = self.network.get(&self.config_locator)?;
+        let client = network.get_client()?;
+
+        // Get the contract instance ledger entry
+        let key = xdr::LedgerKey::ContractData(xdr::LedgerKeyContractData {
+            contract: self.contract_id.clone().into(),
+            key: xdr::ScVal::LedgerKeyContractInstance,
+            durability: xdr::ContractDataDurability::Persistent,
+        });
+
+        let entry = client.get_ledger_entry(&key).await?;
+        
+        // Extract the wasm hash from the contract instance
+        if let xdr::LedgerEntryData::ContractData(data) = entry.data {
+            if let xdr::ScVal::ContractInstance(instance) = data.val {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: cmd/soroban-cli/src/commands/contract/info/wasm_hash.rs [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+        Err(rpc::Error::InvalidResponse("Contract instance not found".into()).into())
+    }
+}
+
+pub fn subcommand() -> Command {
+    command!("wasm-hash")
+        .about("Retrieve the wasm hash of a contract")
+        .arg(arg!(--network <NETWORK> "The network to connect to"))
+        .arg(arg!(--contract-id <CONTRACT_ID> "The contract ID"))
+}
```
leighmcculloch, line 73: I don't think the `subcommand` fn is called or used. Could you remove it if so?

cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -91,6 +92,9 @@ pub enum Cmd {
     ///
     /// If no keys are specificed the contract itself is restored.
     Restore(restore::Cmd),
+
+    /// Get the wasm hash of a deployed contract
+    WasmHash(wasm_hash::Cmd),
```
leighmcculloch, line 97: Can you register the wasm-hash command on the `contracts info` command, rather than on the `contracts` command?

cmd/soroban-cli/src/commands/contract/info/wasm_hash.rs
```diff
@@ -0,0 +1,73 @@
...
+#[group(skip)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub config_locator: locator::Args,
+
+    #[command(flatten)]
+    pub network: network::Args,
+
+    /// Contract ID to get the wasm hash for
+    #[arg(long = "contract-id")]
```
leighmcculloch, line 24: Specifying the long name should be unnecessary because it will be given that name already based off the field name.
```suggestion
    #[arg(long)]
```

leighmcculloch (commented):
👏🏻 This looks good, thanks for contributing.

There's a few small things that need fixing up. I've left comments inline on each.

The other thing this needs is an test. Tests that use commands and interact with an RPC are located in the following directory:
> cmd/crates/soroban-test/tests/it

The way the tests work may not be intuitive because it's pretty bespoke. Let me know if that's something that you get stuck on.

Thanks again!

cmd/soroban-cli/src/commands/contract/info/wasm_hash.rs
```diff
@@ -0,0 +1,67 @@
...
+pub struct Cmd {
+    #[command(flatten)]
+    pub config_locator: locator::Args,
+
+    #[command(flatten)]
+    pub network: network::Args,
+
+    /// Contract ID to get the wasm hash for
+    #[arg(long)]
+    pub contract_id: stellar_strkey::Contract,
```
leighmcculloch, line 25: This should be a string, that is then used with alias resolution. Worth looking at other commands that accept a contract ID to see how this works.

cmd/soroban-cli/src/commands/contract/info/wasm_hash.rs
```diff
@@ -0,0 +1,67 @@
...
+        if let xdr::LedgerEntryData::ContractData(data) = entry.data {
+            if let xdr::ScVal::ContractInstance(instance) = data.val {
+                if let xdr::ContractExecutable::Wasm(hash) = instance.executable {
+                    println!("{}", hex::encode(hash.0));
+                    return Ok(());
+                }
+            }
+        }
+
+        Err(rpc::Error::InvalidResponse("Contract instance not found".into()).into())
```
leighmcculloch, line 64: This code should handle the other contract executable types, and emit the name of the executable if not a wasm-hash. See my other comment here: https://github.com/stellar/stellar-cli/pull/1878#discussion_r2315964996.

cmd/soroban-cli/src/commands/contract/info.rs
```diff
@@ -49,6 +50,9 @@ pub enum Cmd {
     ///
     /// Outputs no data when no data is present in the contract.
     EnvMeta(env_meta::Cmd),
+
+    /// Get the wasm hash of a deployed contract
+    WasmHash(wasm_hash::Cmd),
```
leighmcculloch, line 55: @fnando wdyt naming this command `executable`? Technically contracts can have an executable that is either:
- stellar-asset
- wasm-hash

In the past when we've made commands that accepted a Contract ID and errored on stellar-assets, it's been annoying that the edge case wasn't supported.

The command could output not just the hash but the type of executable, maybe in a simple json format. Or just use xdr-json of the ContractExecutable xdr type.

leighmcculloch: It also might be fair that getting a wasm-hash explicitly as the only output could be more helpful when scripting in some cases. There might be a case for both commands co-existing.

fnando:
superseded by #2576.

Issue #1827, opened 2025-01-13
stellar tx: accept xdr via an argument and not only as a file or stdin


leighmcculloch:
Related to: https://github.com/stellar/stellar-cli/issues/1822

Directly related to: https://github.com/stellar/rs-stellar-xdr/issues/406

This issue shouldn't be attempted until the above issues are resolved. The same pattern used in https://github.com/stellar/rs-stellar-xdr/issues/406 should be used in the `stellar tx` command.

Issue #1822, opened 2025-01-07
`stellar tx` takes input as files and `stdin`


janewang:
### What would you like to see?
Current `stellar tx` only takes input as `stdin`. Like `stellar xdr` it should also accept files as inputs

Slack discussion here: https://stellarfoundation.slack.com/archives/C06KTGUULUF/p1736212430852659

sagpatil:
@leighmcculloch fyi

leighmcculloch:
Recommend having a look at the stellar-xdr cli (that powers the stellar xdr command) for how it does files or stdin in its UI, to see if that same pattern would work for the stellar tx command:

https://github.com/stellar/rs-stellar-xdr/blob/28391e81229ad20a3d9ec56e1e1f382205335085/src/cli/encode.rs#L171-L183

The goal is not just to add the feature, but to maintain a UX parity between the stellar xdr command and the other commands that also work with xdr.

PR #1885, opened 2025-02-21
Implement stellar tx to accept string/file as an input

Ifropc:
### What

Makes `stellar tx` subcommands to accept stdin/string/file as an input

### Why

#1827 
#1822 

### Known limitations

* Only takes one string/file as an input
* Some copy-paste from `rs-stellar-xdr` that can be addressed later

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -1519,11 +1519,11 @@ Sign, Simulate, and Send transactions
 
 Calculate the hash of a transaction envelope
 
-**Usage:** `stellar tx hash [OPTIONS] [INPUT]`
+**Usage:** `stellar tx hash [OPTIONS] [TX_XDR]`
 
 ###### **Arguments:**
 
-* `<INPUT>` — XDR or file containing XDR to decode, or stdin if empty
+* `<TX_XDR>` — Base-64 transaction envelope XDR or file containing XDR to decode, or stdin if empty
 
 ###### **Options:**
 
@@ -1879,11 +1879,11 @@ https://developers.stellar.org/docs/learn/glossary#flags
 
 Transfers the XLM balance of an account to another account and removes the source account from the ledger
 
-**Usage:** `stellar tx operation add account-merge [OPTIONS] --source-account <SOURCE_ACCOUNT> --account <ACCOUNT> [INPUT]`
+**Usage:** `stellar tx operation add account-merge [OPTIONS] --source-account <SOURCE_ACCOUNT> --account <ACCOUNT> [TX_XDR]`
 
 ###### **Arguments:**
 
-* `<INPUT>` — XDR or file containing XDR to decode, or stdin if empty
+* `<TX_XDR>` — Base-64 transaction envelope XDR or file containing XDR to decode, or stdin if empty
 
 ###### **Options:**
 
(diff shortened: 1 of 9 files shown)


FULL_HELP_DOCS.md
```diff
@@ -1517,9 +1517,13 @@ Sign, Simulate, and Send transactions
...
 
-Calculate the hash of a transaction envelope from stdin
+Calculate the hash of a transaction envelope
 
-**Usage:** `stellar tx hash [OPTIONS]`
+**Usage:** `stellar tx hash [OPTIONS] [INPUT]`
+
+###### **Arguments:**
+
+* `<INPUT>` — XDR or file containing XDR to decode, or stdin if empty
```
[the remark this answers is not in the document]
leighmcculloch: Actually, I think we should rename "input" to something meaningful, like "XDR" or "TX-XDR".

cmd/soroban-cli/src/commands/tx/send.rs
```diff
@@ -24,8 +25,11 @@ pub enum Error {
 #[derive(Debug, clap::Parser, Clone)]
 #[group(skip)]
 /// Command to send a transaction envelope to the network
-/// e.g. `cat file.txt | soroban tx send`
+/// e.g. `soroban tx send file.txt`
```
leighmcculloch, line 28: Can we show both examples? For many of these commands piping is still the more common user flow.

cmd/soroban-cli/src/commands/tx/send.rs
```diff
@@ -24,8 +25,11 @@ pub enum Error {
...
 pub struct Cmd {
+    /// XDR or file containing XDR to decode, or stdin if empty
+    #[arg()]
+    pub input: Option<OsString>,
```
leighmcculloch, line 32:
```suggestion
    pub tx_xdr: Option<OsString>,
```

cmd/soroban-cli/src/commands/tx/xdr.rs
```diff
@@ -23,16 +20,49 @@ pub enum Error {
     TooManyOperations,
 }
 
-pub fn tx_envelope_from_stdin() -> Result<TransactionEnvelope, Error> {
-    from_stdin()
+pub fn tx_envelope_from_input(input: &Option<OsString>) -> Result<TransactionEnvelope, Error> {
+    let read: &mut dyn Read = if let Some(input) = input {
```
leighmcculloch, line 24: > &mut dyn Read

TIL, I didn't know you could have a dyn ref without the value living in a Box. Where does the file/cursor live in memory? 🤯

leighmcculloch (commented):
👏🏻 Looks good. Minor comments inline.

cmd/soroban-cli/src/commands/tx/xdr.rs
Ifropc: That's a really good question, I got to dig a bit into it and here's how I understand it:

```
let &mut file = &File::open("example.txt")
```
The file object (file descriptor, metadata, etc.) is stored on the stack, and the pointer to the stack location.
With `&mut dyn Read` instead of regular pointer we have a fat pointer

```
+-------------------+-------------------+
| Data Pointer      | VTable Pointer    |
+-------------------+-------------------+
``` 
Data pointer points to the file, while vtable pointer points to vtable, that is basically a list of pointers to all `Read` functions that is implemented by concrete type (in our case `File`)

So at compile time we will have following memory: 
```
2*usize for &File + 1 usize for &Vtable + <Number of methods in Read> * Usize
```
Now, using `read()` is a different story, because that's when we actually read the content of the file descriptor (and it's inside the implementation of `read_xdr_base64_to_end` which if I read correctly reads to multiple buffers for each separate type, and every buffer is on stack) 
The same is true for cursor (except there we know already that string is on stack because it's one of the clap's arguments) 

https://doc.rust-lang.org/stable/book/ch20-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait
https://stackoverflow.com/a/57754902

leighmcculloch (approved):

PR #1893, opened 2025-02-26
Add `tx edit`.

fnando:
### What

Add `tx edit`, so users can edit a transaction envelope using JSON in a configured editor (it reads the `STELLAR_EDITOR`, `EDITOR` and `VISUAL` env vars).

https://github.com/user-attachments/assets/54566367-c656-42a9-b57e-a2ccbabb3691

### Why

So users can have a more user friendly way of defining transactions.

### Known limitations

~There's no JSON autocompletion at the moment, but a future PR will add support for [JSON schema](https://json-schema.org)'s `$schema` in rs-stellar-xdr, which will make autocompletion work out of the box with editors like VSCode.~

Support for JSON schema has been added directly to the CLI. Unfortunately, the JSON schema will be reported as invalid because rs-stellar-xdr does not add `$schema` as a valid key on each element that allowed in the transaction envelope. This is being tracked here: https://github.com/stellar/rs-stellar-xdr/issues/417

### DIFF
--- a/cmd/soroban-cli/src/commands/tx/edit.rs
+++ b/cmd/soroban-cli/src/commands/tx/edit.rs
@@ -1,0 +1,145 @@
+use std::{
+    env, fs,
+    io::{stdin, Cursor, Write},
+    path::PathBuf,
+    process::{self, Stdio},
+};
+
+use base64::{engine::general_purpose::STANDARD as BASE64, Engine as _};
+use stellar_xdr::curr;
+
+use crate::{commands::global, print::Print};
+
+const DEFAULT_JSON: &str = r#"{
+  "tx": {
+    "tx": {
+      "source_account": "",
+      "fee": 100,
+      "seq_num": 0,
+      "cond": "none",
+      "memo": "none",
+      "operations": [],
+      "ext": "v0"
+    },
+    "signatures": []
+  }
+}
+"#;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+
+    #[error(transparent)]
+    StellarXdr(#[from] stellar_xdr::curr::Error),
+
+    #[error(transparent)]
+    SerdeJson(#[from] serde_json::Error),
+
+    #[error(transparent)]
+    Base64Decode(#[from] base64::DecodeError),
+
+    #[error("Editor returned non-zero status")]
+    EditorNonZeroStatus,
+}
+
+// Command to edit the transaction
+/// e.g. `stellar tx new manage-data --data-name hello --build-only | stellar tx edit`
+#[derive(Debug, clap::Parser, Clone, Default)]
+#[group(skip)]
+pub struct Cmd {}
+
+impl Cmd {
+    pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let print = Print::new(global_args.quiet);
+        let json: String = if atty::isnt(atty::Stream::Stdin) {
+            let mut input = String::new();
+            stdin().read_line(&mut input)?;
+            let input = input.trim();
+            xdr_to_json::<curr::TransactionEnvelope>(input)?
+        } else {
+            DEFAULT_JSON.to_string()
+        };
+
+        let path = tmp_file(&json)?;
+        open_editor(&print, &path)?;
+
+        let contents = fs::read_to_string(&path)?;
+        let xdr = json_to_xdr::<curr::TransactionEnvelope>(&contents)?;
+        fs::remove_file(&path)?;
+
+        println!("{xdr}");
+
+        Ok(())
+    }
+}
+
+fn tmp_file(contents: &str) -> Result<PathBuf, Error> {
+    let temp_dir = env::temp_dir();
+    let file_name = format!("stellar-json-xdr-{}.json", rand::random::<u64>());
+    let path = temp_dir.join(file_name);
+
+    let mut file = fs::File::create(&path)?;
+    file.write_all(contents.as_bytes())?;
+
+    Ok(path)
+}
+
+fn get_editor() -> String {
+    env::var("STELLAR_EDITOR")
+        .or_else(|_| env::var("EDITOR"))
+        .or_else(|_| env::var("VISUAL"))
+        .unwrap_or_else(|_| "vi".to_string())
+}
+
+fn open_editor(print: &Print, path: &PathBuf) -> Result<(), Error> {
+    let editor = get_editor();
+    print.infoln(format!("Using `{editor}`"));
+
+    let parts: Vec<&str> = editor.split_whitespace().collect();
+    let cmd = parts[0];
+    let args = &parts[1..];
+
+    print.infoln("Opening editor to edit the transaction envelope...".to_string());
+
+    let result = process::Command::new(cmd)
+        .stdin(Stdio::null())
+        .args(args)
+        .arg(path)
+        .spawn()?
+        .wait_with_output()?;
+
+    if result.status.success() {
+        Ok(())
+    } else {
+        Err(Error::EditorNonZeroStatus)
+    }
+}
+
+fn xdr_to_json<T>(xdr_string: &str) -> Result<String, Error>
+where
+    T: curr::ReadXdr + serde::Serialize,
+{
+    let xdr_bytes = BASE64.decode(xdr_string)?;
+    let cursor = Cursor::new(xdr_bytes);
+    let mut limit = curr::Limited::new(cursor, curr::Limits::none());
+    let value = T::read_xdr(&mut limit)?;
+    let json = serde_json::to_string_pretty(&value)?;
+
+    Ok(json)
+}
+
+fn json_to_xdr<T>(json_string: &str) -> Result<String, Error>
+where
+    T: serde::de::DeserializeOwned + curr::WriteXdr,
+{
+    let value: T = serde_json::from_str(json_string)?;
+    let mut data = Vec::new();
+    let cursor = Cursor::new(&mut data);
+    let mut limit = curr::Limited::new(cursor, curr::Limits::none());
+    value.write_xdr(&mut limit)?;
+
+    let xdr_base64 = BASE64.encode(&data);
+    Ok(xdr_base64)
+}
--- a/cmd/soroban-cli/src/commands/tx/edit.rs
+++ b/cmd/soroban-cli/src/commands/tx/edit.rs
@@ -1,9 +1,13 @@
 use std::{
-    env, fs,
+    env,
+    fs::{self, File},
     io::{stdin, Cursor, IsTerminal, Write},
     path::PathBuf,
-    process::{self, Stdio},
+    process::{self},
 };
+
+#[cfg(windows)]
+use std::process::Stdio;
 
 use base64::{engine::general_purpose::STANDARD as BASE64, Engine as _};
 use stellar_xdr::curr;
@@ -63,7 +67,9 @@ impl Cmd {
         };
 
         let path = tmp_file(&json)?;
-        open_editor(&print, &path)?;
+        let editor = get_editor();
+
+        open_editor(&print, &editor, &path)?;
 
         let contents = fs::read_to_string(&path)?;
         let xdr = json_to_xdr::<curr::TransactionEnvelope>(&contents)?;
@@ -75,8 +81,14 @@ impl Cmd {
     }
 }
 
+struct Editor {
+    cmd: String,
+    source: String,
+    args: Vec<String>,
+}
+
 fn tmp_file(contents: &str) -> Result<PathBuf, Error> {
-    let temp_dir = env::temp_dir();
+    let temp_dir = env::current_dir().unwrap_or(env::temp_dir());
     let file_name = format!("stellar-json-xdr-{}.json", rand::random::<u64>());
     let path = temp_dir.join(file_name);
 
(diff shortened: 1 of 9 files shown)


cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    #[error(transparent)]
+    Base64Decode(#[from] base64::DecodeError),
+
+    #[error("Editor returned non-zero status")]
+    EditorNonZeroStatus,
+}
+
+// Command to edit the transaction
+/// e.g. `stellar tx new manage-data --data-name hello --build-only | stellar tx edit`
```
Ifropc, line 48: I think `edit` should accept stdin/file/string as an input, the same way other `tx` commands do (using `tx_envelope_from_input`)

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,145 @@
...
+        let json: String = if stdin().is_terminal() {
+            DEFAULT_JSON.to_string()
+        } else {
+            let mut input = String::new();
+            stdin().read_line(&mut input)?;
+            let input = input.trim();
+            xdr_to_json::<curr::TransactionEnvelope>(input)?
+        };
+
+        let path = tmp_file(&json)?;
```
Ifropc, line 65: [nit] should use [tempfile](https://docs.rs/tempfile/latest/tempfile/fn.tempfile.html) to automatically cleanup files

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,145 @@
...
+// Command to edit the transaction
+/// e.g. `stellar tx new manage-data --data-name hello --build-only | stellar tx edit`
+#[derive(Debug, clap::Parser, Clone, Default)]
+#[group(skip)]
+pub struct Cmd {}
+
+impl Cmd {
+    pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let print = Print::new(global_args.quiet);
+        let json: String = if stdin().is_terminal() {
```
Ifropc, line 56: I don't think this would be needed if using `tx_envelope_from_input` function (as it should work just fine with taking stdin piped input)

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: it will, i'm just doing in parts so the pr is smaller

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: i don't want the file to be removed automatically, because i'll keep the file around in case there are errors.

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: actually, i think we need to fix other places. Right now it holds the process waiting for a stdin. Try `stellar tx sign` without anything; that's not great ux. We should give the user an error saying they need to provide _some_ input.

Ifropc: Good point: #1894

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,184 @@
...
+
+struct Editor {
+    cmd: String,
+    source: String,
+    args: Vec<String>,
+}
+
+fn tmp_file(contents: &str) -> Result<PathBuf, Error> {
+    let temp_dir = env::current_dir().unwrap_or(env::temp_dir());
+    let file_name = format!("stellar-json-xdr-{}.json", rand::random::<u64>());
```
leighmcculloch, line 92: Suggest to drop the duplicative 'json', as it's already in the file extension, and we could add the type for context in editors showing the name:
```suggestion
    let file_name = format!("stellar-{}-{}.xdr.json", type_name, rand::random::<u64>());
```

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,184 @@
...
+    file.write_all(contents.as_bytes())?;
+
+    Ok(path)
+}
+
+fn get_editor() -> Editor {
+    let (source, cmd) = env::var("STELLAR_EDITOR")
+        .map(|val| ("STELLAR_EDITOR", val))
+        .or_else(|_| env::var("EDITOR").map(|val| ("EDITOR", val)))
+        .or_else(|_| env::var("VISUAL").map(|val| ("VISUAL", val)))
```
leighmcculloch, line 105: 👍🏻

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+fn open_editor(print: &Print, editor: &Editor, path: &PathBuf) -> Result<(), Error> {
+    print.infoln(format!(
+        "Using `{source}=\"{cmd}\"`",
+        source = editor.source,
+        cmd = editor.cmd,
+    ));
+
+    print.infoln("Opening editor...".to_string());
```
leighmcculloch, line 129: I think these print lines could be merged:
```suggestion
    print.infoln(format!(
        "Opening editor with `{source}=\"{cmd}\"`",
        source = editor.source,
        cmd = editor.cmd,
    ));
```

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,184 @@
...
+
+fn xdr_to_json<T>(xdr_string: &str) -> Result<String, Error>
+where
+    T: curr::ReadXdr + serde::Serialize,
+{
+    let xdr_bytes = BASE64.decode(xdr_string)?;
+    let cursor = Cursor::new(xdr_bytes);
+    let mut limit = curr::Limited::new(cursor, curr::Limits::none());
+    let value = T::read_xdr(&mut limit)?;
+    let json = serde_json::to_string_pretty(&value)?;
```
leighmcculloch, line 167: Highly recommend using the following function instead:

```
T::from_xdr_base64(...)
```
https://docs.rs/stellar-xdr/latest/stellar_xdr/curr/trait.ReadXdr.html#method.from_xdr_base64

Why:
- It'll ensure the input is fully consumed, that the code here doesn't do and in some situations could be a security concern.
- Don't need to figure out the base64 decoder.

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,184 @@
...
+    T: serde::de::DeserializeOwned + curr::WriteXdr,
+{
+    let value: T = serde_json::from_str(json_string)?;
+    let mut data = Vec::new();
+    let cursor = Cursor::new(&mut data);
+    let mut limit = curr::Limited::new(cursor, curr::Limits::none());
+    value.write_xdr(&mut limit)?;
+
+    let xdr_base64 = BASE64.encode(&data);
+    Ok(xdr_base64)
```
leighmcculloch, line 183:
```
value.to_xdr_base64()
```
https://docs.rs/stellar-xdr/latest/stellar_xdr/curr/trait.WriteXdr.html#method.to_xdr_base64

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: what do you mean by "type"? The xdr type?

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: Awesome, I was kinda in the dark on what to do, tbh.

Do we have something similar for json->xdr, or is the current approach good enough?

cmd/soroban-cli/src/commands/tx/edit.rs
leighmcculloch: I meant the xdr type.

cmd/soroban-cli/src/commands/tx/edit.rs
leighmcculloch: I don't think so. What you're doing is 👍🏻

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: does it make sense in this case, given he xdr will always be a transaction envelope?

leighmcculloch: Oh right, good point, yeah doesn't make sense.

leighmcculloch: I wonder if a generic version of this command could also live at `stellar xdr edit` 🤔

fnando: Let's wait and see if ask for it. Porting those functions to be a generic thing would be easy enough!

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+
+fn json_to_xdr<T>(json_string: &str) -> Result<String, Error>
+where
+    T: serde::de::DeserializeOwned + curr::WriteXdr,
+{
+    let mut schema: serde_json::Value = serde_json::from_str(json_string).unwrap();
+
+    if let Some(obj) = schema.as_object_mut() {
+        obj.remove("$schema");
+    }
```
fnando, line 175: @leighmcculloch And this removes the `$schema`, before trying to convert it to xdr.

I tried a few things to use the `schema: serde_json::Value` directly when converting to xdr, but couldn't make it work. Is that possible at all?

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+where
+    T: curr::ReadXdr + serde::Serialize,
+{
+    let version = env!("CARGO_PKG_VERSION");
+    let tx = T::from_xdr_base64(xdr_string, curr::Limits::none())?;
+    let mut schema: serde_json::Value = serde_json::to_value(tx).unwrap();
+    schema["$schema"] = json!(format!(
+        "https://github.com/stellar/stellar-cli/releases/download/v{version}/stellar-xdr-{version}.json"
+    ));
+    let json = serde_json::to_string_pretty(&schema)?;
```
fnando, line 162: @leighmcculloch This adds the `$schema` pointing to a file that's now being generated by the binaries workflow. See #1901

fnando (commented):
Add support for `$schema` (marked you on the comments). Other than that, I think this is good to go.

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+    let path = temp_dir.join(file_name);
+
+    let mut file = fs::File::create(&path)?;
+    file.write_all(contents.as_bytes())?;
+
+    Ok(path)
+}
+
+fn get_editor() -> Editor {
+    let (source, cmd) = env::var("STELLAR_EDITOR")
```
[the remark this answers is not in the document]
fnando: Good call. I'll add a note on the help message. Given that this a very common pattern, I don't it's something magical that needs too much explanation. I also don't think having a flag for the editor is required, given you can override it with `EDITOR='code -w' stellar tx edit`.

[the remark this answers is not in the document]
fnando: Done via ccd0de05ef5fe8aece94e12edeb42fc52afd6775

fnando:
@leighmcculloch @Ifropc @elizabethengelman is this good to go?

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+    path::PathBuf,
+    process::{self},
+};
+
+use serde_json::json;
+use stellar_xdr::curr;
+
+use crate::{commands::global, print::Print};
+
+const DEFAULT_JSON: &str = r#"{
```
Ifropc, line 14: I'm not sure why we want this behavior of 
`stellar tx edit` (I don't think anybody would edit transactions manually)
Why doesn't it instead wait for transaction to be pasted in the terminal like other commands?

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+"#;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+
+    #[error(transparent)]
+    StellarXdr(#[from] stellar_xdr::curr::Error),
```
Ifropc, line 36: Would be nice to add a message that it's an XDR parsing error

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+    let json = serde_json::to_string_pretty(&schema)?;
+
+    Ok(json)
+}
+
+fn json_to_xdr<T>(json_string: &str) -> Result<String, Error>
+where
+    T: serde::de::DeserializeOwned + curr::WriteXdr,
+{
+    let mut schema: serde_json::Value = serde_json::from_str(json_string).unwrap();
```
Ifropc, line 171: This will make us panic if user leaves invalid JSON in the file

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    let json_string = serde_json::to_string(&schema)?;
+
+    let value: T = serde_json::from_str(json_string.as_str())?;
+    let mut data = Vec::new();
+    let cursor = Cursor::new(&mut data);
+    let mut limit = curr::Limited::new(cursor, curr::Limits::none());
+    value.write_xdr(&mut limit)?;
+
+    Ok(value.to_xdr_base64(curr::Limits::none())?)
```
Ifropc, line 185: We should have a separate error type (and message) when user edited transaction is invalid

Ifropc (changes requested):
On top of other comments:
I remember we were talking about recovering from user leaving editor with invalid transaction. 
Do we still plan to do it in this PR?

I also think panic needs to be fixed before this can be merged

```
thread 'main' panicked at cmd/soroban-cli/src/commands/tx/edit.rs:171:75:
called `Result::unwrap()` on an `Err` value: Error("trailing characters", line: 2, column: 12)
```
(On leaving editor with invalid JSON)

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+struct Editor {
+    cmd: String,
+    source: String,
+    args: Vec<String>,
+}
+
+fn tmp_file(contents: &str) -> Result<PathBuf, Error> {
+    let temp_dir = env::current_dir().unwrap_or(env::temp_dir());
```
Ifropc, line 88: This leaves tmp files in the working directory.
IMO, we should only use temp_dir

fnando: That's the idea. If the command crashes for some reason, or if you kill the process, you can still have access to the file.

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: Invalid transactions are already being handled. This is the error I got when trying to supply an invalid memo:

```
❌ error: unknown variant `invalid`, expected one of `none`, `text`, `id`, `hash`, `return` at line 1 column 81
```

or setting fee to -1:

```
❌ error: invalid value: integer `-1`, expected u32 at line 1 column 62
```

I think this is a good first step, we can iterate later on.

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: Again, I don't think keeping waiting on stdin is the best user experience. I think we should either error or give the user a default. Given `tx edit` will have JSON schema autocompletion, it seems like a good default.

@janewang wdyt? This case is where users just type `stellar tx edit` without stdin, so we're currently opening a default JSON structure that allows users to go from there.

cmd/soroban-cli/src/commands/tx/edit.rs
fnando: hardcoded this to use `https://github.com/stellar/rs-stellar-xdr/raw/main/xdr-json/curr/TransactionEnvelope.json` instead on 8a333b1bef76a9e58e380aa75178ca302d0aada3

cmd/soroban-cli/src/commands/tx/edit.rs
Ifropc: I like the Err approach! 
WDYT about modifying `tx_envelope_from_input`?
```rust
pub fn tx_envelope_from_input(input: &Option<OsString>) -> Result<TransactionEnvelope, Error> {
    let read: &mut dyn Read = if let Some(input) = input {
        let exist = Path::new(input).try_exists();
        if let Ok(true) = exist {
            &mut File::open(input)?
        } else {
        }
    } else {
        if (std().is_terminal) {
            return Err(Error::NoStdin);
        }
        &mut stdin()
    };

    let mut lim = Limited::new(SkipWhitespace::new(read), Limits::none());
    Ok(TransactionEnvelope::read_xdr_base64_to_end(&mut lim)?)
}
```
Then you can replace this code with `tx_envelope_from_input()` for having input universally handled (as a bonus, we will also be able to accept files as input. Especially useful when recovering from errors)

____________
As we discussed previously it should be low impact to change behavior of existing commands and I think it's a good idea to align functionality everywhere

fnando: this is a good suggestion! I'll do it in a separate pr, because then the intent is clear, as other places that will benefit from this, will also be on the diff.

cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -0,0 +1,172 @@
...
+    print.infoln(format!(
+        "Opening editor with `{source}=\"{cmd}\"`...",
+        source = editor.source,
+        cmd = editor.cmd,
+    ));
+
+    let mut binding = process::Command::new(editor.cmd.clone());
+    let command = binding.args(editor.args.clone()).arg(path);
+
+    #[cfg(unix)]
```
Ifropc, line 116: Could you add a comment about what's happening here and why is it necessary?

Ifropc (approved):

cmd/soroban-cli/src/commands/tx/mod.rs
```diff
@@ -14,6 +15,9 @@ pub use args::Args;
 
 #[derive(Debug, clap::Subcommand)]
 pub enum Cmd {
+    /// Edit a transaction envelope from stdin. This command respects the environment variables
```
Ifropc, line 18: Should add an example (because we only accept pipe'd transactions, and it doesn't wait for stdin):
`cat transaction.txt | stellar tx edit`

Ifropc (approved):

cmd/soroban-cli/src/commands/tx/mod.rs
fnando: I don't think the example is necessary, as the comment is clear and so is the error message (stdin is a well known pattern):

```console
$ stellar tx edit
❌ error: No stdin detected
```

PR #1894, opened 2025-02-26
Improve tx reading UX

Ifropc:
### What

Improves TX reading UX, adding an STDERR message indicating user to input transaction using stdin

### Why

Currently there are no indication of what to do, which could be confusing for the user 

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/tx/xdr.rs
+++ b/cmd/soroban-cli/src/commands/tx/xdr.rs
@@ -1,3 +1,4 @@
+use crate::print;
 use crate::xdr::{
     Limits, Operation, ReadXdr, Transaction, TransactionEnvelope, TransactionV1Envelope,
 };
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/tx/xdr.rs
```diff
@@ -29,6 +29,7 @@ pub fn tx_envelope_from_input(input: &Option<OsString>) -> Result<TransactionEnv
             &mut Cursor::new(input.clone().into_encoded_bytes())
         }
     } else {
+        eprint!("waiting for transaction input...\n");
```
fnando, line 32: Should we using the rich print instead, to it follows the same pattern?

Ifropc: Updated

fnando (approved):

PR #1896, opened 2025-02-27
Fix metadata entry not properly written

Ifropc:
### What

Fixes metadata not properly written. See previous discussion in #1694

Original implementation took artifact straight from the build directory, wrote metadata there, and then copied it to destination directory (if provided).
Naturally, if no code changes happens between rebuilds, cargo will simply do nothing, as the result WASM file is unchanged according to cargo. However, we already modified it in our code previously, and we write new metadata entries in that existing WASM file.
With this change we require output dir to be specified if we writing any meta (thus ensuring WASM is first copied and only then modified). Then, do the same steps as before.

### Why

#1694

### Known limitations

It's technically a breaking change to require output dir

### DIFF
--- a/cmd/crates/soroban-test/tests/it/build.rs
+++ b/cmd/crates/soroban-test/tests/it/build.rs
@@ -1,3 +1,5 @@
+use assert_fs::TempDir;
+use fs_extra::dir::CopyOptions;
 use predicates::prelude::predicate;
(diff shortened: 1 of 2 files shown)


leighmcculloch (commented):
@Ifropc I think we should make a similar fix to this, that'll disconnect the dependence on the out-dir. See the comment over here for some more info about what's going on and an alternative fix:
- https://github.com/stellar/stellar-cli/issues/1694#issuecomment-2709342205

leighmcculloch (approved):

Issue #1897, opened 2025-03-01
Add Java bindings generation command.


overcat:
### What problem does your feature solve?
Facilitates users in finding the tools needed to generate Java bindings.

Similar to https://github.com/stellar/stellar-cli/issues/1633 and https://github.com/stellar/stellar-cli/pull/1761

### What would you like to see?
We have already implemented it in [stellar-contract-bindings](https://github.com/lightsail-network/stellar-contract-bindings), but a command should also be added here to guide users to find it.

### What alternatives are there?
N/A

overcat:
https://github.com/stellar/stellar-cli/pull/1898

PR #1898, opened 2025-03-01
Add Java bindings generation command.

overcat:
### What
Add Java bindings generation command.

### Why

See #1897

### Known limitations

N/A


leighmcculloch (approved):
Looks good. 

One thought: I notice that the tool for doing binding gen for Java is a Python tool. Is that an odd experience for a Java dev to need to install python/pip?

overcat:
Hi @leighmcculloch, I will provide a web interface later, so users won't need to install it.

overcat:
Hi @leighmcculloch, I created a simple page: https://stellar-contract-bindings.fly.dev/

leighmcculloch:
Very nice!!!

PR #1901, opened 2025-03-05
Generate json schema.

fnando:
### What

Generate json schema for TransactionEnvelope when building binaries

### Why

So we can use this with `stellar tx edit` for autocompletion.

### Known limitations

N/A


Ifropc (approved):
LGTM 👍🏼

PR #1903, opened 2025-03-05
Update ubuntu version.

fnando:
The ubuntu-20.04 image is deprecated since Feb 1 and will be officially unsupported starting Apr 1.

### What

Update ubuntu to 22.04.

### Why

The ubuntu-20.04 image is deprecated since Feb 1 and will be officially unsupported starting Apr 1.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -28,7 +28,7 @@ jobs:
           - name: stellar-cli
             binary: stellar
         sys:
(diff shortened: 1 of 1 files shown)


Ifropc:
Json-schema is also on ubuntu-20

Also moving forward we should probably align the versions across the stack (binaries, tests, etc)

Ifropc:
This is also _technically_ a breaking change because we no longer support older glibc

Ifropc (approved):

PR #1905, opened 2025-03-06
fix incompatible frontmatter in tx cookbock files

ElliotFriend:
### What

remove the backticks from frontmatter titles in some cookbook files

### Why

docs builds are breaking

### Known limitations

the pre-push git hook is giving me some kind of error that i don't really understand


ElliotFriend:
duplicate of #1904

PR #1909, opened 2025-03-06
Feat/seq bump

elizabethengelman:
### What

Add a command to allow for bumping a transaction's sequence number.

### Why

After talking about [`tx set`](https://github.com/stellar/stellar-cli/pull/1760) in our DevX meeting yesterday, we determined that maybe we should release one feature at a time, as the community finds them useful. `tx edit seq-num bump` seems like it would be useful right off the bat, and will be good for starting to establish the foundation for this type of subcommand. 

### Known limitations

This PR will need to be merged in after https://github.com/stellar/stellar-cli/pull/1893 and may require a few updates.

The failing cargo deny advisories should be fixed in #1931


leighmcculloch (commented):
I think the interface should change slightly by avoiding the term bump. 

It looks like it bumps relatively. 

The stellar network has an operation called bump sequence that bumps an account seq absolutely not relatively. 

Because of the similar names and concepts on first read I thought it behaved differently than it does. 

Can I suggest using a name that unambiguously implies it increases the value, by using the term increment?

Separately we probably need a sub command to set the seq num, not modify it relatively. 

In regards to doing the commands piecemeal, that sgtm, but we should have a plan for all the commands so that the UX we build piecemeal is consistent and we don't use patterns for the earlier commands that don't fit the latter ones.

leighmcculloch:
We could keep this as `set` and allow doing relative modifications by using + or - at the start of the number. 

The upside is that there are less commands and we can reuse the pattern elsewhere, such as modifying times. 

The downside would be if someone wanted to build a tx for SEP-10 that has a negative seq number, they'd have to set 0 then set -1.

fnando:
I think we should simplify this even further. I'd call this `tx edit seq-number next` and avoid adding anything else until we see people are actually asking for this feature.

elizabethengelman:
I agree with keeping things simple for now to see if people use/like this feature. 

I like keeping the flexibility of allowing users to increment by more than one, as Gleb suggested above. And to me, `increment` or something similar makes the behavior of the command a bit clearer than `next`. That said, I don’t feel too strongly about it!

fnando:
I'm looking to simplify the most common usage of sequence numbers which is getting the next one. I think it's fine to have a more generic way of setting sequence numbers (you're already doing it with `increment` and `decrement`). Personally, I feel that `stellar tx new [SOMETHING] | stellar tx edit sequence-number increment` is too big of a command, making it almost impossible to remember it. 🤔

leighmcculloch:
Good call on `-1` being problematic, yeah lets not consume time on making that work.

Anyone doing something advanced with seq numbers is likely going to want to set a specific value, rather than increment it. And for the more common case of updating a tx to have a new next seq, then that probably fits in a more general command like `stellar tx update` that calls out to the network and sets it and other information ready to submit.

elizabethengelman:
Thanks for the feedback, all! Great conversation.

Summarizing what I'm seeing in the feedback -  the different use cases that we're trying to cover are (not necessarily in this pr):

1. incrementing the seq-num by 1 (for simplicity when building tx locally)
2. incrementing the seq-num by `n` (when creating multiple txns)
3. allowing the user to set the seq-num to a specific positive number
4. allowing the user to set the seq-num to a specific negative number (for SEP-10)
5. incrementing based on the account's current seq-num

I wonder if going back to `tx edit seq-num set <NUM>` could work, if we also implement `tx edit seq-num get <NUM>`. I think this would allow us to handle all of the scenarios above except the 5th. Its not the simplest solution, but it is flexible. 🤷‍♀️  And then as a separate ticket we could implement `tx update`?

I guess I'm leaning toward a `tx edit seq-num <COMMAND>` structure, since we already have it in place in this PR, and it is easily extended to include more features. What do you all think? Also happy to chat more about this in our meeting today

FULL_HELP_DOCS.md
```diff
@@ -1555,6 +1556,49 @@ Sign, Simulate, and Send transactions
 
 
 
+## `stellar tx update`
+
+Update the transaction
+
+**Usage:** `stellar tx update <COMMAND>`
```
elizabethengelman, line 1563: I decided to put this command under `stellar tx update` instead of `stellar tx edit`, since it does feel like a different set of commands than the current edit command, that opens an editor and allows the users to edit the tx json.

cmd/soroban-cli/src/commands/tx/update/sequence_number/next.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    async fn current_seq_num(&self) -> Result<i64, Error> {
+        let network = &self.config.get_network()?;
+        let client = network.rpc_client()?;
+        client
+            .verify_network_passphrase(Some(&network.network_passphrase))
+            .await?;
+
+        let muxed_account = self.config.source_account().await?;
```
Ifropc, line 65: We should get source account from transaction, not the config.
I can create a transaction sourcing from alice, and may use any non-alice acount to bump the sequence.
Currently we will use wrong account to fetch the sequence (currently used account)

FULL_HELP_DOCS.md
```diff
@@ -1585,6 +1586,49 @@ Sign, Simulate, and Send transactions
...
+
+**Usage:** `stellar tx update sequence-number next [OPTIONS] --source-account <SOURCE_ACCOUNT>`
+
+###### **Options:**
+
+* `--rpc-url <RPC_URL>` — RPC server endpoint
+* `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
+* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
+* `-n`, `--network <NETWORK>` — Name of network to use from config
+* `-s`, `--source-account <SOURCE_ACCOUNT>` — Account that where transaction originates from. Alias `source`. Can be an identity (--source alice), a public key (--source GDKW...), a muxed account (--source MDA…), a secret key (--source SC36…), or a seed phrase (--source "kite urban…"). If `--build-only` or `--sim-only` flags were NOT provided, this key will also be used to sign the final transaction. In that case, trying to sign with public key will fail
```
Ifropc, line 1625: This command shouldn't require source account, as we already know source account of transaction (all valid transactions have source account)

FULL_HELP_DOCS.md
```diff
@@ -1585,6 +1586,49 @@ Sign, Simulate, and Send transactions
...
+* `--hd-path <HD_PATH>` — If using a seed phrase, which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
```
Ifropc, line 1626: Not sure why is it here as well

cmd/soroban-cli/src/commands/tx/update/sequence_number/next.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub config: config::Args,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    XdrStdin(#[from] XdrParsingError),
+    #[error(transparent)]
+    Xdr(#[from] xdr::Error),
+    #[error("only V1 transactions are supported")]
```
Ifropc, line 22: I think it's more clear to write that v0 or fee bump transactions are not supported (as a user I know that there are regular transactions -- v1 and fee bump transactions, I assume v0 is legacy transactions)

cmd/soroban-cli/src/commands/tx/update/sequence_number/next.rs
elizabethengelman: good call! i made the update in [d3fdc92](https://github.com/stellar/stellar-cli/pull/1909/commits/d3fdc92f779c49b68c90598f9640488d1f379f68)

Ifropc (approved):
LGTM!

PR #1933, opened 2025-03-10
Update error message on missing network

Ifropc:
### What

We should encourage users using `network use` instead of env variables 

### Why

Improving error message and usability

### Known limitations

N/A


cmd/soroban-cli/src/config/network.rs
```diff
@@ -25,6 +25,8 @@ pub enum Error {
     #[error(
         r#"Access to the network is required
 `--network` or `--rpc-url` and `--network-passphrase` are required if using the network.
+Network configuration can also be set using `network use` subcommand. For example, to use
+testnet simply type `stellar network use testnet`
```
leighmcculloch, line 29:
```suggestion
testnet, run `stellar network use testnet`.
```

leighmcculloch (approved):

PR #1946, opened 2025-03-10
Update e2e test action

Ifropc:
### What

Update e2e test action to use new `action.yaml` from `system-test` repo
https://github.com/stellar/system-test/pull/119

### Why

Simplify the flow and prepare for cross-platform testing 

### Known limitations

Some options are no longer supported (e.g. toolchain version) -- this can be easily brought back in the future.
It also doesn't support changing version of quickstart yet, which is the next task in the line.


.github/workflows/e2e.yml
```diff
@@ -10,110 +10,8 @@ concurrency:
...
-
-      - name: Run system test scenarios
-        run: |
-          docker run --rm -t --name e2e_test stellar/system-test:dev \
-          --VerboseOutput $SYSTEM_TEST_VERBOSE_OUTPUT  \
-          --TestFilter "${{ matrix.scenario-filter }}" \
-          --SorobanExamplesGitHash $SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_HASH \
-          --SorobanExamplesRepoURL $SYSTEM_TEST_SOROBAN_EXAMPLES_GIT_REPO
+  systems-test:
+    uses: stellar/system-test/.github/workflows/test.yml@8b6406086ff5d6c59665d48e5ad98914b9d42443
```
fnando, line 14: this is pointing to `8b6406086ff5d6c59665d48e5ad98914b9d42443`, which is not this pr: https://github.com/stellar/system-test/pull/119

I'm assuming you want to merge https://github.com/stellar/system-test/pull/119 before this one, right? Just to keep in mind that the ref will be different. Should we point to main instead?

Ifropc: Yep, this will point to the main before the merge

Ifropc:
e2e tests are passing with this change, see last build
https://github.com/stellar/stellar-cli/actions/runs/13861805879/job/38791888870

fnando (approved):

PR #1951, opened 2025-03-12
Update README

Ifropc:
### What

Updates install guide for CLI to match the recommended way

### Known limitations

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ $ nix profile install github:stellar/stellar-cli
 
 Use GitHub Action:
 ```
-uses: stellar/stellar-cli@22.5.0
+uses: stellar/stellar-cli@v22.5.0
 ```
 
 ## Autocomplete
--- a/README.md
+++ b/README.md
@@ -46,10 +46,7 @@ $ nix profile install github:stellar/stellar-cli
 
 Use GitHub Action:
 ```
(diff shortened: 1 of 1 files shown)


fnando (approved):

fnando (changes requested):
looks like this is broken… 22.5.0 is not a valid ref. And v22.5.0 fails with:

CleanShot 2025-03-12 at 12 08 45

fnando:
seems like the version is not actually optional.

Ifropc:
It meant to be optional, but there's a bug in the action. Fixed in #1952

README.md
```diff
@@ -46,10 +46,7 @@ $ nix profile install github:stellar/stellar-cli
 
 Use GitHub Action:
 ```
-- uses: stellar/stellar-cli
-  # Optionally specify version of the cli. If not provided, latest available version will be used
-  with:
-    version: 22.4.0
+uses: stellar/stellar-cli@22.5.0
```
fnando, line 49: I believe this is still incorrect, as 22.5.0 is not a valid tag. This should be `v22.5.0` instead

```suggestion
uses: stellar/stellar-cli@v22.5.0
```

fnando (approved):

PR #1956, opened 2025-03-12
Refactoring binaries workflow

Ifropc:
### What

Refactors binaries workflow moving copy-pasted part into composite action

### Why

It grows larger with more artifacts we upload, it's a good idea to make it a bit more dry

### Known limitations

N/A


.github/workflows/binaries.yml
```diff
@@ -107,44 +85,20 @@ jobs:
...
       - name: Uncompress Artifact
         run: tar xvf ${{ env.ARTIFACT_NAME }}
-
       - name: Export JSON Schema
         run: |
           ./stellar xdr types schema --type TransactionEnvelope > ${{ env.JSON_SCHEMA_NAME }}
-
       - name: Upload Artifact
-        uses: actions/upload-artifact@v4
+        uses: ./.github/actions/artifact-upload
```
fnando, line 94: we don't really need this anymore, now that stellar/rs-stellar-xdr is publishing the json schema.

Ifropc: Should JSON schema be deleted entirely then?

fnando: yep! Somehow I missed this comment, sorry about that!

fnando (approved):

.github/actions/artifact-upload/action.yml
```diff
@@ -0,0 +1,39 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{ inputs.name }}
+        path: ${{ inputs.path }}
+
+    - name: Build provenance for attestation (release only)
+      if: github.event_name == 'release'
+      uses: actions/attest-build-provenance@v2
+      with:
+        subject-path: ${{ inputs.path }}
```
leighmcculloch, line 23: It looks like the file being attested is now the compressed file, and not the binary. Am I misunderstanding? I think we should be attesting the binary.

leighmcculloch: We can attest both if we want, but we should be at least attesting the binary.

Ifropc: Good point, didn't think about that, I'll restore attestation of binary

PR #1778, opened 2024-12-04
Experimental contract code/build linking command

leighmcculloch:
### What

Add a `contract verify` command that implements a subset of @orbitlens proposal:
- https://github.com/stellar/stellar-protocol/discussions/1573

### Try it out

```
cargo install --locked stellar-cli --git https://github.com/stellar/stellar-cli --branch contract-verify
```

### Example

The [leighmcculloch/exp-stellar-expert-verified-builds](https://github.com/leighmcculloch/exp-stellar-expert-verified-builds) repo contains a contract that has been built by the `stellar-expert/soroban-build-workflow` as well as a custom workflow that uses the attestation principles but not the stellar-expert build process.

Contract built with [`stellar-expert/soroban-build-workflow`](https://github.com/leighmcculloch/exp-stellar-expert-verified-builds/blob/713de53640b5e6b698740623db7911bfec7b918f/.github/workflows/stellar-expert-release.yml) deployed to [CBLB](https://stellar.expert/explorer/testnet/contract/CBLBYGNUMGOYZWQE7NHE3CA7SQVZXCD4PKGJRRARHVXE6DML25PIOFN4):
```console
$ stellar contract verify --id CBLBYGNUMGOYZWQE7NHE3CA7SQVZXCD4PKGJRRARHVXE6DML25PIOFN4
ℹ️ Loading wasm...
ℹ️ Wasm Hash: d751fee773a7461aef2451bb16674b8d9b67de6a9d5312c3f9a146189e011af5
ℹ️ Source Repo: github:leighmcculloch/exp-stellar-expert-verified-builds
ℹ️ Collecting GitHub attestation from https://api.github.com/repos/leighmcculloch/exp-stellar-expert-verified-builds/attestations/sha256:d751fee773a7461aef2451bb16674b8d9b67de6a9d5312c3f9a146189e011af5...
✅ Attestation found linked to GitHub Actions Workflow Run:
✅  • Repository: https://github.com/leighmcculloch/exp-stellar-expert-verified-builds
✅  • Ref:        refs/tags/v1.0.0
✅  • Path:       .github/workflows/stellar-expert-release.yml
✅  • Git Commit: 713de53640b5e6b698740623db7911bfec7b918f
✅  • Runner:     github-hosted
✅  • Run:        https://github.com/leighmcculloch/exp-stellar-expert-verified-builds/actions/runs/12158347810/attempts/1
[...]

PR #1957, opened 2025-03-13
Add command `contract info build` to display GitHub attestation (experimental)

fnando:
### What

This pr supersedes #1778 and can be merged right away. We discussed today having this as a plugin, but we can hide from the docs entirely instead. It's a great way to experiment without the burden of maintaining detached code.

CleanShot 2025-03-12 at 17 14 49

### Why

So contract builds can be verified by using github attestation.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/info.rs
+++ b/cmd/soroban-cli/src/commands/contract/info.rs
@@ -2,6 +2,7 @@ use std::fmt::Debug;
 
 use crate::commands::global;
 
+pub mod build;
 pub mod env_meta;
 pub mod interface;
 pub mod meta;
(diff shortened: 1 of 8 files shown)


cmd/soroban-cli/src/commands/contract/info/build.rs
```diff
@@ -0,0 +1,261 @@
...
+    AttestationInvalid,
+
+    #[error("Stellar asset contract doesn't contain meta information")]
+    NoSACMeta(),
+}
+
+impl Cmd {
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let print = Print::new(global_args.quiet);
+        print.warnln("\x1b[31mThe attestation verifies the build process, but please review the source code yourself.\x1b[0m".to_string());
```
leighmcculloch, line 52:
```suggestion
        print.warnln("\x1b[31mThis command displays information about the GitHub Actions run that attested to have built the wasm, and does not verify the source code. Please review the run, it's workflow, and source code.
\x1b[0m".to_string());
```

leighmcculloch (approved):
👏🏻 🎉 🥳

cmd/soroban-cli/src/commands/contract/info.rs
```diff
@@ -49,26 +50,39 @@ pub enum Cmd {
     ///
     /// Outputs no data when no data is present in the contract.
     EnvMeta(env_meta::Cmd),
+
+    /// Output the contract build information, if available.
+    ///
+    /// If the contract has a meta entry like `source_repo=github:user/repo`, this command will try
+    /// to fetch the attestation information for the WASM file.
+    #[command(hide = true)]
```
leighmcculloch, line 58: Instead of hiding this, I think it would also be fine to make it visible, and put the word "Experimental: " at the start of the comment above.

leighmcculloch: And could elaborate in the longer comment text to say that the functionality, data verified, is likely to change and is still in development.

fnando: @janewang wdyt?

PR #1959, opened 2025-03-13
Change macos runner

Ifropc:
### What

Downgrade macos runner to 13

### Why

This is the latest macos runner on Intel. We currently use ARM runner for our x86-64 builds and tests, but Intel runner should provide a native experience (i.e. we will catch if there's any potential x86-64 issue specifically on Macos, for external libraries/e2e integration, etc.) 

### Known limitations

Previously used runner is v14, but there shouldn't be any issues with downgrading

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -34,7 +34,7 @@ jobs:
             target: aarch64-unknown-linux-gnu
           - os: macos-14
             target: aarch64-apple-darwin
-          - os: macos-14
+          - os: macos-13
             target: x86_64-apple-darwin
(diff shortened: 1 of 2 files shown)


leighmcculloch (approved):
I don't see an issue with downgrading, but the PR doesn't say why this change is needed. The PR description only says why it should be fine to make the change.

Why is this change needed?

Ifropc:
It's not necessarily needed, but I noticed that we are using ARM runner for our x64 builds, so I thought it's a good idea to align that.
Also for tests it's better to have architecture to match the tested artifact

PR #1961, opened 2025-03-14
Remove default seed option from generate command

leighmcculloch:
### What
Remove default seed option from generate command.

### Why
It appears to be broken:

```
$ stellar keys generate --default-seed issuer
⚠️ Behavior of `generate` will change in the future, and it will no longer fund by default. If you want to fund please provide `--fund` flag. If you don't need to fund your keys in the future, ignore this warning. It can be suppressed with -q flag.
❌ error: invalid word in phrase
```

We could fix it, but if it's broken and no one has noticed it suggests it isn't being used anywhere. I believe it was added for testing at some point, but it also scares me a bit that there's an option on the key generation command that results in a fixed test key getting used. There's no case where that should be used in the production release.


Ifropc (approved):
I don't think we should have/maintain any commands that can be easily achieved with calling a command with some set of options

PR #1963, opened 2025-03-14
Add Asset Management cookbook and examples

leighmcculloch:
### What
Add a cookbook guide showing how to issue and manage Stellar Assets through CLI commands and contracts.

### Why
Developers need clear examples for common asset management tasks like minting, burning, freezing, and clawback.

### DIFF
--- a/cookbook/asset-management.mdx
+++ b/cookbook/asset-management.mdx
@@ -1,0 +1,146 @@
+---
+title: Asset Management
+hide_table_of_contents: false
+description: Issue a Stellar Asset, deploy it's contract, and mint, burn, freeze, and clawback.
+custom_edit_url: https://github.com/stellar/stellar-cli/edit/main/cookbook/asset-management.mdx
+---
+
+### Install
+Recommend installing from one of the supported package managers if you can, as you'll get a precompiled binary. But you can also install from source.
+On macOS:
+```
+brew install stellar-cli
+```
+On Windows:
+```
+winget install --id Stellar.StellarCLI
+```
+On anything else, or if you prefer to build from source:
+```
+cargo install --locked stellar-cli
+```
+
+### Network Config
+Network config is supported under the `stellar network` command.
+```
+stellar network use testnet
+```
+Testnet is preconfigured, and is what I'm going to use today. You can add config for your RPC for mainnet. Or your RPC for testnet if you want to use your own.
+
+### Assets
+#### Creation
+To create a Stellar Asset we need a Stellar account created, that'll be the issuer. Assets on Stellar are uniquely defined by a code and an issuer address. So they look like this:
+```
+stellar keys generate issuer
+```
+
+View the public key with:
+```
+stellar keys public-key issuer
+```
+
+Configure the issuer key to be used for signing transactions:
+```
+stellar keys use issuer
+```
+
+#### Issuer Config
+```
+stellar tx new set-options \
+  --set-revocable \
+  --set-clawback-enabled
+```
+
+If you'd like to look at the tx before sending:
+```
+stellar tx new set-options \
+  --set-revocable \
+  --set-clawback-enabled \
+  --build-only \
+  | stellar tx edit \
+  | stellar sign \
+  | stellar send
+```
+
+#### Choose Asset Code
+```
+echo ABC:$(stellar keys public-key issuer)
+```
+
+#### Contract Deployment
+```
+stellar contract asset deploy \
+  --asset ABC:$(stellar keys public-key issuer) \
+  --alias abc-asset
+```
+
+Run a few commands to get some read only data from the contract to confirm it's there.
+```
+stellar contract invoke --id abc-asset -- symbol
+stellar contract invoke --id abc-asset -- name
+stellar contract invoke --id abc-asset -- decimals
+```
+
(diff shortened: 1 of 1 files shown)


cookbook/asset-management.mdx
```diff
@@ -0,0 +1,224 @@
...
+hide_table_of_contents: false
+description: Issue a Stellar Asset, deploy it's contract, and mint, burn, freeze, and clawback.
+custom_edit_url: https://github.com/stellar/stellar-cli/edit/main/cookbook/asset-management.mdx
+---
+
+The [stellar-cli] can be used to issue and administer assets on the Stellar network.
+
+The following guide assumes that the [stellar-cli] has been configured to use testnet, or a local test network. To use testnet, run the following command.
+
+```
```
fnando: looking at the `cmd/crates/soroban-test/tests/it/integration/cookbook.rs` file, seems like all blocks must be `bash` blocks, which means a bunch of other files need updating (I saw a few using `sh`).

leighmcculloch: I've been trying to get this to pass, but I think we need to revisit the cookbook test holistically. Several issues:

- It requires that commands sit by themselves in separate code blocks, but that makes for an unpleasant reading experience when there are a group of small commands that need executing together.

- It places arbitrary demands on the output of different types of commands. Rather than simply executing the commands and testing for success or failure from the exit code, it has assertions on what type of output each command should return. Those assertions are broken by the looks of it, because they're failing on simple command usage like creating a new key.

- It requires certain commands must fail. Those assertions are in the generic run_command function, but they look like the type of assertions that would apply to a specific situation because surely the command should not always fail.

cmd/crates/soroban-test/tests/it/integration/cookbook.rs
```diff
@@ -109,27 +110,11 @@ fn run_command(
...
-            .failure()
-            .stderr(predicates::str::contains("TxSorobanInvalid"));
-    } else {
-        result.success();
-    }
+    sandbox
+        .new_assert_cmd(&cmd)
+        .args(&modified_args)
+        .assert()
+        .success();
```
leighmcculloch: This change make it so that every command executed is expected to succeed. For commands that we don't expect to succeed, they are instead marked as 'ignore' in the markdown itself. The author can decide what gets tested and what doesn't for those edge cases that are just impractical.

cookbook/asset-management.mdx
```diff
@@ -0,0 +1,303 @@
...
+stellar contract invoke --id abc -- name
+```
+
+##### Admin Setup
+
+At this point the issuer is the admin of the contract.
+Let's create another account that'll be the admin.
+
+```bash
+stellar keys generate admin
```
fnando, line 129: Is there a reason not to use `--fund` on all keys being generated instead of doing a separate `stellar keys fund` call?

cookbook/asset-management.mdx
```diff
@@ -0,0 +1,303 @@
...
+```
+
+```bash
+stellar keys fund admin
+```
+
+Using the issuer account, invoke the asset contract's `set_admin` function.
+
+```bash
+stellar keys use user
```
fnando: in terms of legibility, I think adding `--source user` makes it clearer when is just one command being executed after this command.

cmd/crates/soroban-test/tests/it/integration/util.rs
```diff
@@ -63,15 +63,6 @@ pub async fn deploy_custom_account(sandbox: &TestEnv) -> String {
...
 }
 
-pub fn setup_env_for_sandbox(sandbox: &TestEnv) {
-    env::set_var("SOROBAN_ACCOUNT", TEST_ACCOUNT);
-    env::set_var("SOROBAN_RPC_URL", sandbox.network.rpc_url.clone());
-    env::set_var("SOROBAN_NETWORK_PASSPHRASE", LOCAL_NETWORK_PASSPHRASE);
-
-    env::set_var("XDG_CONFIG_HOME", sandbox.config_dir().as_os_str());
-    env::set_var("XDG_DATA_HOME", sandbox.data_dir().as_os_str());
-}
```
leighmcculloch, line 73: These lines globber the environment and affect the state of the entire process, every test in it. That's not ideal because it means one test has a side-effect on the behaviour of the rest of the tests. It also presents us from using `stellar keys use ...` in cookbook tests.

Most of these lines aren't required anyway because we set the same parameters other ways elsewhere. The couple that are required, the config dir, can be set via the `--config-dir` option instead. The `setup_env_for_sandbox` function is only used in one place, in the `deploy_contract` function, so I've moved setting the config-dir to there.

cookbook/asset-management.mdx
leighmcculloch: I think there's value if each command does one thing and is copy-pastable to work on any network. But don't feel strongly if you think we should use --fund.

fnando (approved):

cookbook/asset-management.mdx
leighmcculloch: I think so too, but I wanted to use it for some of the tutorial, and then it felt odd to use it sometimes and not always. It also made some of the commands longer, so more things to eye on the line.

PR #1964, opened 2025-03-14
Add default json structure when calling `stellar tx edit`.

fnando:
### What

Add default json structure when calling `stellar tx edit` without piping. It also updates the help message with an example:

```
$ stellar tx edit --help
Edit a transaction envelope from stdin. This command respects the environment variables `STELLAR_EDITOR`, `EDITOR` and `VISUAL`, in that order.

Example: Start a new edit session

$ stellar tx edit

Example: Pipe an XDR transaction envelope

$ stellar tx new manage-data --data-name hello --build-only | stellar tx edit
```

### Why

It's a better ux overall. It gives users the ability to explore the transaction editing from zero.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/tx/edit.rs
+++ b/cmd/soroban-cli/src/commands/tx/edit.rs
@@ -11,6 +11,9 @@ use stellar_xdr::curr;
 
 use crate::{commands::global, print::Print};
 
+const SCHEMA_URL: &str =
+    "https://github.com/stellar/rs-stellar-xdr/raw/main/xdr-json/curr/TransactionEnvelope.json";
+
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     #[error(transparent)]
@@ -27,9 +30,6 @@ pub enum Error {
 
     #[error("Editor returned non-zero status")]
     EditorNonZeroStatus,
-
-    #[error("No stdin detected")]
-    NoStdin,
 }
 
 // Command to edit the transaction
@@ -42,7 +42,7 @@ impl Cmd {
     pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
         let print = Print::new(global_args.quiet);
         let json: String = if stdin().is_terminal() {
-            return Err(Error::NoStdin);
+            default_json()
         } else {
             let mut input = String::new();
             stdin().read_line(&mut input)?;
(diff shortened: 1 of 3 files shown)


cmd/soroban-cli/src/commands/tx/edit.rs
```diff
@@ -11,6 +11,9 @@ use stellar_xdr::curr;
 
 use crate::{commands::global, print::Print};
 
+const SCHEMA_URL: &str =
+    "https://github.com/stellar/rs-stellar-xdr/raw/main/xdr-json/curr/TransactionEnvelope.json";
```
Ifropc, line 15: Should it be pinned? It can be changed between protocol versions

fnando: Not right now. I'll update this to the tag once rs-stellar-xdr starts publishing it.

leighmcculloch:
I think this will address some of:
- https://github.com/stellar/stellar-cli/issues/1962

cmd/soroban-cli/src/commands/tx/edit.rs
leighmcculloch: We can backport the JSON to a minor version of rs-stellar-xdr if that's helpful. v23 won't be out for a bit.

fnando: Maybe we can release a minor version of rs-stellar-xdr, just to move forward and not messing with previous releases.

cmd/soroban-cli/src/commands/tx/mod.rs
```diff
@@ -17,6 +17,15 @@ pub use args::Args;
...
     /// Edit a transaction envelope from stdin. This command respects the environment variables
     /// `STELLAR_EDITOR`, `EDITOR` and `VISUAL`, in that order.
+    ///
+    /// Example: Start a new edit session
+    ///
+    /// $ stellar tx edit
+    ///
+    /// Example: Pipe an XDR transaction envelope
+    ///
+    /// $ stellar tx new manage-data --data-name hello --build-only | stellar tx edit
```
leighmcculloch, line 27: I think we should use a bump seq as an example. It results in no data left on chain. Ultimately payments would be the best example, but it requires too many parameters I think.

cmd/soroban-cli/src/commands/tx/edit.rs
leighmcculloch: I think it's fine to merge this as is. This command is flagged as experimental. And then if we can get a minor rs-stellar-xdr release out, we can open a PR and update the URL and do a quick release.

leighmcculloch (approved):

leighmcculloch:
Going to merge this as is, because I think it looks good as is, and the outstanding comments are okay to address in a follow up. I'm releasing the CLI with tx edit, and this PR, whether perfect or not, is a net-benefit.

PR #1968, opened 2025-03-17
Add env lookup by name

leighmcculloch:
### What
Add ability to lookup env config values by name with the `stellar env` command.

### Why
Allow scripts to parse specific environment variables programmatically, to retrieve them and display them places. Helpful for integrating the stellar-cli into things like a shell prompt, or use in scripting.

The format and behaviour of the command is modeled after the `go env` command that works identically.

### DIFF
--- a/cmd/soroban-cli/src/commands/env/mod.rs
+++ b/cmd/soroban-cli/src/commands/env/mod.rs
@@ -7,6 +7,10 @@ use clap::Parser;
 
 #[derive(Debug, Parser)]
 pub struct Cmd {
+    /// Env variable name to get the value of.
+    #[arg()]
+    pub name: Option<String>,
+
     #[command(flatten)]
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/env/mod.rs
```diff
@@ -7,6 +7,10 @@ use clap::Parser;
 
 #[derive(Debug, Parser)]
 pub struct Cmd {
+    /// Env variable name to get the value of.
```
fnando, line 10: I think we should add an example, just for clarity.

```suggestion
    /// Env variable name to get the value of.
    ///
    /// E.g.: $ stellar env STELLAR_ACCOUNT
```

fnando (approved):
Nice addition! One comment about adding an example, but lgtm!

PR #1977, opened 2025-03-19
additional-libs feature for keyring dependencies (secure store)

elizabethengelman:
### What
Adds an `additional-libs` feature so that users can build and use the CLI (without keyring support) without having `libdbus-1-dev` installed on their system

### Why

Allows for a smaller binary, and less system dependencies.

### Known limitations

This PR only handles the keyring dependencies, I'll put the ledger work in a separate PR: https://github.com/stellar/stellar-cli/pull/2011

~Also, I still need to do a bit more testing of this change on a linux vm.~ _tested on macos and ubuntu_


cmd/soroban-cli/Cargo.toml
```diff
@@ -40,6 +40,7 @@ version_lt_23 = []
 version_gte_23 = []
 opt = ["dep:wasm-opt"]
 emulator-tests = ["stellar-ledger/emulator-tests"]
+additional-libs = ["dep:keyring"]
```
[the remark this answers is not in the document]
elizabethengelman: I may be misunderstanding, but I thought that we did need to include this in stellar-cli, so that we could avoid including the keyring dependency (and all of it's system dependencies) when `additional-libs` is disabled.

Makefile
```diff
@@ -48,7 +48,7 @@ generate-full-help-doc:
 	cargo run --bin doc-gen --features clap-markdown
 
 test: build-test
-	cargo test --workspace --exclude soroban-test
+	cargo test --workspace --exclude soroban-test --features additional-libs
```
elizabethengelman, line 51: I wonder if we should include a test run without `additional-libs` too? I may punt on that for now, and consider how to do that when I implement the same for the ledger deps.

elizabethengelman: Maybe we can have a separate job in the `rust.yml` workflow that had additional-libs enabled 🤔

cmd/soroban-cli/Cargo.toml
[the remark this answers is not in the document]
Ifropc: yes i think so too (it wasn't in `stellar-cli` :) )

Makefile
Ifropc: i think we should just update makefile to only run additional-libs tests

cmd/soroban-cli/Cargo.toml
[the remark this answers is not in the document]
elizabethengelman: Ah, I missed that distinction! :) I'll update that now.

Makefile
elizabethengelman: > i think we should just update makefile to only run additional-libs tests

Does this change of adding `--features additional-libs` to the test command do that?

Ifropc: I think so, yeah (but it needs to be a separate command)

elizabethengelman: Are you thinking that there would be another command in the Makefile that just runs additional-libs tests?

Like, adding this:
```
additional-libs-test:
	cargo test --workspace --exclude soroban-test --features additional-libs
```
 And then would additional-lib-test be run as part of the main test command? or a separate workflow?

Ifropc: I think just adding it to test command is fine

elizabethengelman: just pushed up a change

Ifropc (approved):
LGTM!

fnando (approved):

Issue #1908, opened 2025-03-06
Add doc validation on merge


Ifropc:
### What problem does your feature solve?

We should add documentation validation on merge. Currently we might accidentally push a change that breaks stellar-docs but we can't know that

### What would you like to see?

GHA for validation.
We can clone stellar-docs and do the build on latest change (this would require change to the stellar-docs to allow to do that though)

### What alternatives are there?

N/A

Ifropc:
We should also make docs to pull docs artifact

PR #1981, opened 2025-03-20
Upload cookbook docs to artifacts

Ifropc:
### What

Uploads cookbook docs to artifacts

### Why

To decouple cookbook docs from our actual release and avoid potential breakage
docs PR: https://github.com/stellar/stellar-docs/pull/1439

### Known limitations

I manually uploaded artifact produced by the latest run to 22.6.0 release

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -120,7 +120,7 @@ jobs:
 
       - name: Compress
         run: |
-          tar czvf ${{ env.NAME }}.tar.gz cookbook
+          tar czvf ${{ env.NAME }}.tar.gz cookbook FULL_HELP_DOCS.md
 
       - name: Upload to Artifacts
         uses: ./.github/actions/artifact-upload
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -106,3 +106,24 @@ jobs:
         with:
           name: ${{ env.STELLAR_CLI_INSTALLER }}
           path: ${{ env.STELLAR_CLI_INSTALLER }}
+
+  cookbook:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup vars
+        run: |
+          version="$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "stellar-cli") | .version')"
+          echo "VERSION=${version}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


fnando (approved):
Great idea!

.github/workflows/binaries.yml
```diff
@@ -106,3 +106,24 @@ jobs:
...
+
+      - name: Setup vars
+        run: |
+          version="$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "stellar-cli") | .version')"
+          echo "VERSION=${version}" >> $GITHUB_ENV
+          echo "NAME=stellar-cli-${version}-docs-cookbook" >> $GITHUB_ENV
+
+      - name: Compress
+        run: |
+          tar czvf ${{ env.NAME }}.tar.gz cookbook FULL_HELP_DOCS.md
```
leighmcculloch, line 123: The cookbook and full help are different things. The full help docs is referred to as the user manual in the docs, so I think we should rename the job to User Manual. We could rename the .md file to USER_MANUAL too maybe for consistency.

Ifropc: Do you think we should upload FULL_HELP_DOCS.md and cookbook.tar.gz as separate artifacts?

leighmcculloch: I think they should be separate. And named by what they contain. The user manual is at least maybe interesting to others.

leighmcculloch:
> To decouple cookbook docs from our actual release and avoid potential breakage

The PR description says the aim is to decouple cookbook docs from the CLI release, however this PR seems to explicitly make the cookbooks a release artifact, therefore further coupling the cookbooks to the release. 🤔 

Is the goal to be able to regenerate and rewrite the release cookbook artifact in cases when the cookbook has an error? 🫨 

Setting up a process to rewrite release artifact's is somewhat terrifying, as the release artifacts should be built one time, from the release itself and never tampered with.

The underlying issue seems to me to be that the cookbooks in this repo are not being tested, and if they're to live in this repo instead of the stellar-docs repo, they should probably be tested here. Have we explored that option?

Ifropc:
Yes, the idea is to combine this approaches (see #1908)
The reason we want to put docs into artifacts is that if somehow we still break the docs, we can simply just replace broken artifact with a working one (though with docs validation this should never happen)

Ifropc:
We decided not to to with this approach

Issue #1913, opened 2025-03-07
New command `stellar network health`


Ifropc:
### Command description

Command that uses RPC [getHealth](https://developers.stellar.org/docs/data/rpc/api-reference/methods/getHealth) endpoint

### Command flags

Optional `--json` flag that changes output format to Json (printing  `result` object)

### Command return value

Print health status in human readable format, for example:
```
// Healthy
🟢 Healthy
Latest ledger: ...
...
```
```
// Not responding
🔴 Unhealthy
[Error details]
```

PR #1984, opened 2025-03-25
Add `network health` command

Ifropc:
### What

Adds `network health` command

```
> cargo run network health --network=local
❌ Unhealthy
❌ failed to fetch network health: Networking or low-level protocol error: HTTP error: error trying to connect: tcp connect error: Connection refused (os error 111)
> cargo run network health              
✅ Healthy
ℹ️ Latest ledger: 104038
> cargo run network health --output json-formatted
{
  "status": "healthy",
  "latestLedger": 104044,
  "oldestLedger": 256,
  "ledgerRetentionWindow": 120960
}
```

### Why

Closing #1913

### Known limitations

Depends on https://github.com/stellar/rs-stellar-rpc-client/pull/24

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,9 +64,7 @@ version = "=22.0.4"
 # Dependencies from the rs-stellar-rpc-client repo:
 [workspace.dependencies.soroban-rpc]
 package = "stellar-rpc-client"
-#version = "=22.0.0"
-git = "https://github.com/stellar/rs-stellar-rpc-client"
-branch = "get-health"
+version = "=22.1.0"
 
 [workspace.dependencies.stellar-ledger]
 version = "=22.6.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,9 @@ version = "=22.0.4"
 # Dependencies from the rs-stellar-rpc-client repo:
 [workspace.dependencies.soroban-rpc]
 package = "stellar-rpc-client"
-version = "=22.0.0"
+#version = "=22.0.0"
+git = "https://github.com/stellar/rs-stellar-rpc-client"
+branch = "get-health"
 
 [workspace.dependencies.stellar-ledger]
 version = "=22.6.0"
--- a/cmd/soroban-cli/src/commands/network/health.rs
+++ b/cmd/soroban-cli/src/commands/network/health.rs
@@ -47,7 +47,7 @@ impl Cmd {
                     } else {
                         print.warnln(format!("Status: {}", resp.status));
                     }
-                    println!("Latest ledger: {}", resp.latest_ledger);
+                    print.infoln(format!("Latest ledger: {}", resp.latest_ledger));
                 }
                 OutputFormat::Json => println!("{}", serde_json::to_string(&resp)?),
                 OutputFormat::JsonFormatted => println!("{}", serde_json::to_string_pretty(&resp)?),
--- a/cmd/soroban-cli/src/commands/network/health.rs
+++ b/cmd/soroban-cli/src/commands/network/health.rs
@@ -42,15 +42,19 @@ impl Cmd {
         match result {
             Ok(resp) => match self.output {
                 OutputFormat::Text => {
-                    print.emoji_println('🟢', "Healthy");
+                    if resp.status.to_ascii_lowercase() == "healthy" {
+                        print.checkln("Healthy");
+                    } else {
+                        print.warnln(format!("Status: {}", resp.status));
+                    }
                     println!("Latest ledger: {}", resp.latest_ledger);
                 }
                 OutputFormat::Json => println!("{}", serde_json::to_string(&resp)?),
                 OutputFormat::JsonFormatted => println!("{}", serde_json::to_string_pretty(&resp)?),
             },
             Err(err) => {
+                print.errorln("Unhealthy");
                 print.errorln(format!("failed to fetch network health: {err}"));
-                print.emoji_println('🔴', "Unhealthy");
             }
         }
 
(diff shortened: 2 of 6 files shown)


cmd/soroban-cli/src/commands/network/health.rs
```diff
@@ -0,0 +1,59 @@
...
+                    print.emoji_println('🟢', "Healthy");
+                    println!("Latest ledger: {}", resp.latest_ledger);
+                }
+                OutputFormat::Json => println!("{}", serde_json::to_string(&resp)?),
+                OutputFormat::JsonFormatted => println!("{}", serde_json::to_string_pretty(&resp)?),
+            },
+            Err(err) => {
+                print.errorln(format!("failed to fetch network health: {err}"));
+                print.emoji_println('🔴', "Unhealthy");
+            }
```
leighmcculloch, line 54: The response for the getHealth endpoint looks like this:
```json
{
  "jsonrpc": "2.0",
  "id": "0",
  "result": {
    "status": "healthy",
    "latestLedger": 94885,
    "oldestLedger": 192,
    "ledgerRetentionWindow": 120960
  }
}
```

At the moment it looks like if the status field was unhealthy or some other value that hte healthy path in his match would still be followed. Is that correct?

I think we should show the status returned instead of using the hardcoded values.

And then show the ✅ emoji, which iirc is under the `checkln` fn, only if the response is status == "healthy".

We can use the errorln for the case that the status is unhealthy too. We don't need both the ❌ and 🔴 for the unhealthy state.

cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -1,11 +1,9 @@
-use clap::Parser;
-
-use crate::rpc::{self};
-
 use super::{config::locator, global};
+use clap::Parser;
 
 pub mod add;
 pub mod default;
+mod health;
```
leighmcculloch, line 6: The commands are all pub mod so that they can be embedded into other tooling. Not too many realistically have that happen, but we made them all pub so we could if needed.
```suggestion
pub mod health;
```

cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -79,22 +82,6 @@ pub enum Error {
...
-    #[error(transparent)]
-    Rpc(#[from] rpc::Error),
-    #[error(transparent)]
-    HttpClient(#[from] reqwest::Error),
-    #[error("Failed to parse JSON from {0}, {1}")]
-    FailedToParseJSON(String, serde_json::Error),
-    #[error("Invalid URL {0}")]
-    InvalidUrl(String),
-    #[error("Inproper response {0}")]
-    InproperResponse(String),
```
leighmcculloch, line 97: Thanks for cleaning these cases up!

Cargo.toml
```diff
@@ -64,7 +64,9 @@ version = "=22.0.4"
 # Dependencies from the rs-stellar-rpc-client repo:
 [workspace.dependencies.soroban-rpc]
 package = "stellar-rpc-client"
-version = "=22.0.0"
+#version = "=22.0.0"
+git = "https://github.com/stellar/rs-stellar-rpc-client"
+branch = "get-health"
```
leighmcculloch, line 69: I've approved the PR that this branch references:
https://github.com/stellar/rs-stellar-rpc-client/pull/24

If that library has no breaking changes I think it can be released so you can reference a new version here.

cmd/soroban-cli/src/commands/network/mod.rs
```diff
@@ -52,6 +50,9 @@ pub enum Cmd {
     #[cfg(feature = "version_lt_23")]
     #[command(subcommand)]
     Container(crate::commands::container::Cmd),
+
+    /// Gets network health
```
leighmcculloch, line 54: I think we can be more specific and indicate that this is the health of the RPC we're retrieving.
```suggestion
    /// Checks the health of the configured RPC
```

cmd/soroban-cli/src/print.rs
```diff
@@ -30,6 +30,12 @@ impl Print {
         }
     }
 
+    pub fn emoji_println<T: Display + Sized, R: Display + Sized>(&self, icon: R, message: T) {
+        if !self.quiet {
+            eprintln!("{} {}", self.compute_emoji(icon), message);
+        }
+    }
+
```
leighmcculloch, line 38: Looks like this function is no longer used.

cmd/soroban-cli/src/commands/network/health.rs
```diff
@@ -0,0 +1,63 @@
...
+        let print = print::Print::new(global_args.quiet);
+        let result = self.config.get_network()?.rpc_client()?.get_health().await;
+
+        match result {
+            Ok(resp) => match self.output {
+                OutputFormat::Text => {
+                    if resp.status.to_ascii_lowercase() == "healthy" {
+                        print.checkln("Healthy");
+                    } else {
+                        print.warnln(format!("Status: {}", resp.status));
```
leighmcculloch, line 48: Nice use of warn here. 👍🏻

cmd/soroban-cli/src/commands/network/health.rs
```diff
@@ -0,0 +1,63 @@
...
+                    }
+                    println!("Latest ledger: {}", resp.latest_ledger);
```
leighmcculloch, line 50: I think this should probably be printed to the same file as the above lines. They're all reporting information from the response together.
```suggestion
                    print.infoln(format!("Latest ledger: {}", resp.latest_ledger));
```

leighmcculloch (approved):
Couple minor things, otherwise lgtm.

PR #1988, opened 2025-03-27
Adapt to wasm32v1-none target

graydon:
Part of https://github.com/stellar/rs-soroban-sdk/issues/1428 -- just moving us to rust 1.84 and wasm32v1-none for the wasm build (and fixing clippy fallout).

### DIFF
--- a/.github/workflows/bindings-ts.yml
+++ b/.github/workflows/bindings-ts.yml
@@ -40,7 +40,7 @@ jobs:
       - run: rustup update
       - run: sudo apt update && sudo apt install -y libdbus-1-dev libudev-dev
       - run: cargo build
-      - run: rustup target add wasm32-unknown-unknown
+      - run: rustup target add wasm32v1-none
       - run: make build-test-wasms
       - run: npm ci && npm run test
         working-directory: cmd/crates/soroban-spec-typescript/ts-tests
--- a/.github/workflows/bindings-ts.yml
+++ b/.github/workflows/bindings-ts.yml
@@ -8,37 +8,34 @@ on:
 jobs:
   test:
     name: test generated libraries
-    runs-on: ubuntu-22.04
-    services:
-      rpc:
-        image: stellar/quickstart:testing
-        ports:
-          - 8000:8000
-        env:
-          ENABLE_LOGS: true
-          NETWORK: local
-          ENABLE_SOROBAN_RPC: true
-        options: >-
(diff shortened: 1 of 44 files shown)


leighmcculloch (commented):
One change is needed, commented inline.

Also, we need to think about how this will work for folks with older contracts that use an older version of rustc. So I think for the contract build command we need to choose the target based on the version of Rust installed and in use.

Otherwise lgtm. We'll want to delay merging this closer to / after when the SDK is released so that the init commands that generate the contract Cargo.toml will require a late enough Rust version.

graydon:
> One change is needed, commented inline.

Pushed a fix that undoes the backticks and adds a local clippy-allow. I think this is what you wanted?

> Also, we need to think about how this will work for folks with older contracts that use an older version of rustc. So I think for the contract build command we need to choose the target based on the version of Rust installed and in use.

Not sure I understand this -- it's a rare contract that's going to not work on 1.84. Are we typically promising "never upgrading rust versions"?

> We'll want to delay merging this closer to / after when the SDK is released so that the init commands that generate the contract Cargo.toml will require a late enough Rust version.

Ok.

leighmcculloch:
I separated the rust upgrade into:
- https://github.com/stellar/stellar-cli/pull/2010

leighmcculloch:
> Are we typically promising "never upgrading rust versions"?

No, but existing contracts should be able to continue to build with the new tooling, and existing contracts explicitly refuse to build on anything newer than 1.81. Anyone who has already built contracts, and is building a new contract that they presumedly want to use a new SDK and new CLI for, when they go back and run build on their recent (not actually that old) contract.

I think we need a small tweak to branch on the version of Rust. If the version of Rust in use is >= 1.84, use `wasm32v1-none`, if < 1.84, use `wasm32-unknown-unknown`. I was wondering if we need to also branch on the SDK version in the contract, but I don't think so.

cc @stellar/devx

leighmcculloch:
A mistake that might be easy for devs to make is that they could have scripts that pull the wasm file from the old target directory, and that file could still be present on disk but not updating. Maybe we should output an ℹ️ text block? Or maybe if the CLI detects a `target/wasm32-unknown-unknown` when building for `wasm32v1-none` the CLI should delete the old directory.

graydon:
> If the version of Rust in use is >= 1.84, use `wasm32v1-none`, if < 1.84, use `wasm32-unknown-unknown`.

I think you mean "if < 1.82" right? Versions 1.82 and 1.83 are, AIUI, something we don't ever want anyone compiling with, because they _don't_ support `wasm32v1-none` but _do_ enable the objectionable target features when building for `wasm32-unknown-unknown`.

leighmcculloch:
Yes right. I wasn't sure what to do with those versions. But I guess the cli errors?

janewang:
fyi @ElliotFriend @briwylde08 We may need to update the dev docs as well. This changes the wasm path. Even our getting started guide may need to provide note for this

ElliotFriend:
Love to see this getting closer! Thanks for the hard work everyone!

fnando:
Closing this in favor of https://github.com/stellar/stellar-cli/pull/2022

PR #2012, opened 2025-04-16
Add `ledger entry get` implementation

Ifropc:
### What

### Why

### Known limitations

[TODO or N/A]


cmd/crates/soroban-test/tests/it/integration/ledger/entry.rs
```diff
@@ -0,0 +1,553 @@
...
+        parsed_key_xdr_output.entries[0].val,
+        LedgerEntryData::ContractData { .. }
+    ));
+
+    // the output should be the same regardless of key format
+    assert_eq!(parsed_key_output.entries, parsed_key_xdr_output.entries);
+}
+
+// top level test
+// todo: test --claimable-id, --pool-id,
```
elizabethengelman, line 330: similar question as with --offer -- how can i add claimable balances and pools in test chain in order to test this?

cmd/soroban-cli/src/commands/ledger/entry/fetch.rs
```diff
@@ -0,0 +1,347 @@
...
+            for x in config_setting_id {
+                let key = LedgerKey::ConfigSetting(LedgerKeyConfigSetting {
+                    config_setting_id: ConfigSettingId::try_from(*x)
+                        .map_err(|_| InvalidConfigId(*x))?,
+                });
+                ledger_keys.push(key);
+            }
+        }
+
+        if let Some(ttl) = &self.ttl {
```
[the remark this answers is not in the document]
leighmcculloch: Yup that's fine to ignore that ledger entry type. They're a special case, sort of an implementation detail that they are represented on the ledger as a separate entry. Everywhere outside of the ledger we treat the TTL as an attribute of the ledger entries themselves.

cmd/crates/soroban-test/tests/it/integration/ledger/entry.rs
janewang: @elizabethengelman you may need to create claimable balances and pools on testnet
- claimable balances: https://quest.stellar.org/learn/series/3/quest/3
- pools: https://quest.stellar.org/learn/series/3/quest/5

cmd/soroban-cli/src/commands/ledger/entry/fetch.rs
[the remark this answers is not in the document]
elizabethengelman: Sounds good, thanks for the details!

cmd/crates/soroban-test/tests/it/integration/ledger/entry.rs
elizabethengelman: Thanks for these links! I was able to create the these on the network that we're using in the gh runner 👌

elizabethengelman:
@janewang this PR is ready for review, along with https://github.com/stellar/rs-stellar-rpc-client/pull/26 which is a dependency. Once it gets merged, i think that https://github.com/stellar/stellar-cli/issues/1838 can be closed

cmd/soroban-cli/src/commands/ledger/entry/fetch/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    Contract(contract::Cmd),
+    /// Fetch the current network config by `ConfigSettingId`.
+    /// All config settings are returned if no id is provided.
+    Config(config::Cmd),
+    ///Fetch a claimable balance ledger entry by id
+    ClaimableBalance(claimable_balance::Cmd),
+    ///Fetch a liquidity pool ledger entry by id
+    LiquidityPool(liquidity_pool::Cmd),
+    /// Fetch WASM bytecode by hash
+    Wasm(wasm::Cmd),
```
leighmcculloch: The ledger entry is called "ContractCode". I think we should try and align the terminology in this command as much as possible with the network.  We can put in the help description the connection to wasm?

leighmcculloch (commented):
Taking a look at this for the first time. I have some questions, but they might be arising because I'm missing some context or discussion that happened offline or I've missed. So please let me know if I'm rehashing something that was already discussed or where I can go to catch up.

The UX/API exposed is a new abstraction over the ledger entries that exist on network, rather than being a simple 1-1 mapping of commands to network concepts. We have been trying to avoid introducing abstractions in tools because of our past experiences with abstractions in horizon diverging from the protocol over time and becoming a life of their own.

For example, data entries and trust lines are not surfaced as an "entry" to fetch, rather they are embedded under the account command. But in reality they are entries on their own, and when a developer uses the CLI, and then jumps over to querying the entries from an RPC or other form, they're working with different mental models for the same data.

Another example, that there's no command for fetching contract code entries, but instead wasm entries, but that doesn't map 1-1 with ledger entries on network.

Thoughts on rearranging these commands a bit so that they map directly to ledger entries? Or is the goal to create a new abstraction / data model here?

cmd/soroban-cli/src/commands/ledger/entry/fetch/account.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct Cmd {
+    /// Account alias or public key to lookup, default is test identity
+    pub account: String,
+
+    #[command(flatten)]
+    pub args: Args,
+
+    /// Hide the account ledger entry from the output
+    #[arg(long)]
+    pub hide_account: bool,
```
leighmcculloch: What use cases would someone use this option?

cmd/soroban-cli/src/commands/ledger/entry/fetch/account.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::{
+    commands::config::{self, locator},
+    xdr::{LedgerKey, LedgerKeyAccount, MuxedAccount},
+};
+use clap::{command, Parser};
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// Account alias or public key to lookup, default is test identity
```
leighmcculloch: Suggest rewording. It's pretty odd to talk about the test identity in the CLI UI. Does the test identity still exist, I thought we remove it anyway?
```suggestion
    /// Account alias or address to lookup
```

cmd/soroban-cli/src/commands/ledger/entry/fetch/config.rs
```diff
@@ -0,0 +1,75 @@
+use super::args::Args;
+use crate::xdr::{ConfigSettingId, LedgerKey, LedgerKeyConfigSetting};
+use clap::{command, Parser};
+use std::collections::HashMap;
+use std::fmt::Debug;
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
```
leighmcculloch: We can drop the config settings fetch I think, just because it is already available at `stellar network settings`. Either that or we keep both is fine too. Maybe in the long run we deprecate the old version.

cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs
```diff
@@ -0,0 +1,104 @@
...
+use clap::{command, Parser};
+use stellar_strkey::ed25519::PublicKey as Ed25519PublicKey;
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub args: Args,
+
+    /// Account alias or public key to lookup, default is test identity
```
leighmcculloch:
```suggestion
    /// Account alias or address to lookup
```

cmd/soroban-cli/src/commands/ledger/entry/fetch/contract_data.rs
```diff
@@ -0,0 +1,90 @@
...
+        ReadXdr, ScAddress, ScVal,
+    },
+};
+use clap::{command, Parser};
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// Contract alias or address to fetch
+    pub contract: config::UnresolvedContract,
```
leighmcculloch, line 17: Could we give this field an option? This might have changed or be changing, so defer to @fnando, but in the past all the commands accepted their parameters as named parameters and avoided positional arguments.

cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[arg(long)]
+    pub account: String,
+
+    /// Assets to get trustline info for
+    #[arg(long, required = true)]
+    pub asset: Vec<String>,
+
+    /// If identity is a seed phrase use this hd path, default is 0
```
leighmcculloch:
```suggestion
    /// If account is a seed phrase use this hd path, default is 0
```

cmd/soroban-cli/src/commands/ledger/entry/fetch/offer.rs
```diff
@@ -0,0 +1,74 @@
...
+};
+use clap::{command, Parser};
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub args: Args,
+
+    /// Account alias or public key to lookup, default is test identity
```
leighmcculloch:
```suggestion
    /// Account alias or address to lookup
```

cmd/soroban-cli/src/commands/ledger/entry/fetch/account.rs
elizabethengelman: good catch - this was an old artifact of when `stellar ledger entry fetch account` was also returning trustlines, etc. This should be removed!

cmd/soroban-cli/src/commands/ledger/entry/fetch/account.rs
elizabethengelman: Good catch - we had removed that. I think this was a copy paste error. I'll also clean up `stellar keys secret`'s docs which included this mention of a test identity.

cmd/soroban-cli/src/commands/ledger/entry/fetch/contract_data.rs
elizabethengelman: ah gotcha, i think that makes sense. we seem to have a few commands that have positional args instead of named params, but maybe updating those can be future work. 

but in the mean time, when adding new commands we can make sure to use named args. 

the only downside i see to using only named params is that it makes commands a bit more verbose, in that the user needs to put the name of the arg. but perhaps that isn't a huge downside, and is worth it for the clarity.

cmd/soroban-cli/src/commands/ledger/entry/fetch/config.rs
elizabethengelman: I agree that we may not need config fetch-able in both places. The one thing I like about the `stellar ledger entry fetch config` in this PR is that it allows for fetching just one config item by the config setting id, which helps with output readability. but, that feature could also probably be added `stellar network settings` if we find it useful.

cmd/soroban-cli/src/commands/ledger/entry/fetch/contract_data.rs
elizabethengelman: updated in 181f0533d6757524f8ea980ea71f82da7190eb1d

PR #2014, opened 2025-04-25
Add subcommand lookup for missing commands.

fnando:
### What

Add support for subcommand plugins lookup. If you have a binary like `stellar-contract-bindings-typescript`, that means you can call it by using `stellar contract bindings typescript`.

```console
$ which stellar-hello
/Users/fnando/Projects/stellar/stellar-cli/target/debug/stellar-hello

$ stellar hello
running `stellar hello` with args:

$ stellar hello 1
running `stellar hello` with args: 1

$ which stellar-contract-hello
/Users/fnando/Projects/stellar/stellar-cli/target/debug/stellar-contract-hello

$ stellar contract hello
running `stellar contract hello` with args:

$ stellar contract hello foo bar --nice 1
running `stellar contract hello` with args: foo bar --nice 1

$ stellar env
STELLAR_ACCOUNT=me      # use
STELLAR_NETWORK=testnet # use
```

### Why

This allows publishers to extend the CLI by bringing their own subcommands.

### Known limitations

For now, we had to remove command suggestion, but clap already does command suggestion.

```console
$ stellar contrac
error: unrecognized subcommand 'contrac'

  tip: some similar subcommands exist: 'container', 'contract'

Usage: stellar [OPTIONS] <COMMAND>

```

### DIFF
--- a/cmd/crates/soroban-test/tests/it/plugin.rs
+++ b/cmd/crates/soroban-test/tests/it/plugin.rs
@@ -64,7 +64,7 @@ fn has_no_path_failure() {
         .unwrap_or_else(|_| assert_cmd::Command::new("stellar"))
         .arg("hello")
         .assert()
-        .stderr(predicates::str::contains("error: no such command: `hello`"));
+        .stderr(predicates::str::contains("unrecognized subcommand 'hello'"));
 }
 
 fn target_bin() -> PathBuf {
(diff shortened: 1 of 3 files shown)


leighmcculloch (approved):
👍🏻 I also think it's fine to leave command suggestion off of plugins if it's too cumbersome / complex to add. 

Are there any commands we should disallow adding plugins to?

fnando:
> Are there any commands we should disallow adding plugins to?

@leighmcculloch Good question… I don't think so, because we even want to allow things like adding additional signing mechanisms.

PR #2017, opened 2025-04-30
Remove 'Waiting for transaction input' message when reading from stdin

leighmcculloch:
### What
Remove the 'Waiting for transaction input' message when reading from stdin.

### Why
The message creates noise, especially when chaining many tx commands together, it ends up outputting several times. Due to the way that commands run concurrently when piping, the messages come out immediately all at once too with no context on what is waiting for what.


cmd/soroban-cli/src/commands/tx/xdr.rs
```diff
@@ -30,7 +30,6 @@ pub fn tx_envelope_from_input(input: &Option<OsString>) -> Result<TransactionEnv
             &mut Cursor::new(input.clone().into_encoded_bytes())
         }
     } else {
-        print::Print::new(false).infoln("Waiting for transaction input...");
```
elizabethengelman, line 33: I think the reason we added this print line was because if a user forgot to pipe a transaction into the command, it would just hang without any prompting. Which was confusing. But I can also see how this would be annoying if it printed it over and over! 

Maybe we can get this merged in, and then add a new ticket to rework this a bit to make the "Waiting for transaction input..." only print once. 🤔

leighmcculloch: > it would just hang without any prompting.

Even though this is not a great experience, it is a very common experience across many composable command line tools.

The existing output is optimising for making the failing case easier to understand, at the cost of making the happy path noisy and confusing.

For example:

```
$ stellar contract invoke --id asset --build-only -- \
    mint \
    --to CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABSC4 \
    --amount 123 \
  | stellar tx simulate \
  | ...
  | stellar tx simulate \
  | stellar tx sign --sign-with-key issuer \
  | stellar tx send
ℹ️ Waiting for transaction input...
ℹ️ Waiting for transaction input...
ℹ️ Waiting for transaction input...
ℹ️ Waiting for transaction input...
```

fnando: I said this before, but I think commands that expect stdin and don't receive one should fail. `tx edit` used to do that, but now it's opening a default json structure instead.

https://github.com/stellar/stellar-cli/blob/main/cmd/soroban-cli/src/commands/tx/edit.rs#L44

leighmcculloch: I think that sounds fine too. 👍🏻

leighmcculloch: I've copied the pattern you're using in 6fa9c37.

leighmcculloch:
```
$ echo -n 'AAAAAgAAAABC6aNIWicaOZvbWjNa...' | stellar tx hash
0940b37a69679463ee60f0bda525826f41a0859a244b71eb423b14f0376bddd7
```
```
$ stellar tx hash
❌ error: no transaction provided
```

fnando (approved):

PR #2019, opened 2025-05-01
Add `stellar plugin search`.

fnando:
### What

Add `stellar plugin ls` and `stellar plugin search`.

CleanShot 2025-05-01 at 15 09 26

The docs have an accompanying pull request: https://github.com/stellar/stellar-docs/pull/1525

### Why

Hopefully, this will help with discoverability of plugins.

### Known limitations

N/A


cmd/soroban-cli/src/commands/plugin/ls.rs
```diff
@@ -0,0 +1,22 @@
...
+#[derive(Debug, clap::Parser, Clone)]
+#[group(skip)]
+pub struct Cmd;
+
+impl Cmd {
+    pub fn run(&self) -> Result<(), Error> {
+        let plugins = default::list().unwrap_or_default();
+
+        if plugins.is_empty() {
+            println!("No Plugins installed. E.g. stellar-hello");
```
leighmcculloch, line 15: The example doesn't feel like it has enough context. Suggest we add:
```suggestion
            println!("No Plugins installed. Plugins are commands available on the path that start with 'stellar-'. E.g. stellar-hello");
```

cmd/soroban-cli/src/commands/plugin/search.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Debug, clap::Parser, Clone)]
+#[group(skip)]
+pub struct Cmd;
+
+impl Cmd {
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let print = Print::new(global_args.quiet);
+        let url =
+            "https://api.github.com/search/repositories?q=topic:stellar-cli-plugin&per_page=100";
```
leighmcculloch, line 29: I think we should alter the search query:
- sort by stars with `&sort=stars&order=desc`
- exclude forks with `fork:false`
- exclude archived repos with `archived:false`

cmd/soroban-cli/src/commands/plugin/search.rs
```diff
@@ -0,0 +1,82 @@
...
+        if search.total_count == 0 {
+            print.searchln(format!("No plugins found."));
+            return Ok(());
+        }
+
+        let wording = if search.total_count == 1 {
+            "plugin"
+        } else {
+            "plugins"
+        };
```
leighmcculloch, line 43: 😍

leighmcculloch (approved):
One request otherwise lgtm. 👏🏻

cmd/soroban-cli/src/commands/plugin/search.rs
fnando: Great! Didn't think about that!

cmd/soroban-cli/src/commands/plugin/ls.rs
fnando: Good call!

Changed it to give this output instead:

```console
$ stellar --list
No plugins installed.

Plugins are commands available on the path
that start with 'stellar-'. E.g. stellar-hello

https://developers.stellar.org/docs/tools/cli/plugins
```

PR #2027, opened 2025-05-13
update npm dependency stellar-sdk for TS bindings

chadoh:
### What

Update `@stellar/stellar-sdk` npm dependencies to allow anything in the 13.x major version

### Why

So generated TS libraries have the latest & greatest.

When the bindings were new, they needed specific features in specific patch-level updates of js-stellar-sdk. That's no longer true. Anything in the 13.x range will work fine.

### DIFF
--- a/cmd/crates/soroban-spec-typescript/fixtures/test_constructor/src/index.ts
+++ b/cmd/crates/soroban-spec-typescript/fixtures/test_constructor/src/index.ts
@@ -33,9 +33,6 @@ if (typeof window !== 'undefined') {
 
 
 
-export const Errors = {
-
-}
 
 export interface Client {
   /**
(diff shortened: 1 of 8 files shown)


fnando (approved):

cmd/crates/soroban-spec-typescript/fixtures/test_constructor/src/index.ts
```diff
@@ -64,7 +61,7 @@ export class Client extends ContractClient {
   static async deploy<T = Client>(
         /** Constructor/Initialization Args for the contract's `__constructor` method */
         {counter}: {counter: u32},
-    /** Options for initializing a Client as well as for calling a method, with extras specific to deploying. */
+    /** Options for initalizing a Client as well as for calling a method, with extras specific to deploying. */
```
fnando, line 64:
```suggestion
    /** Options for initializing a Client as well as for calling a method, with extras specific to deploying. */
```

PR #2028, opened 2025-05-13
Remove contract alias length limit

chadoh:
### What

Remove 30-char length limit on aliases.

### Why

Does this serve a purpose? Will there be actual problems caused by people (or tooling) creating humongously long aliases?

I ran into this limit while incorporating [OpenZeppelin's `fungible-token-interface-example`](https://github.com/OpenZeppelin/stellar-contracts/blob/main/examples/fungible-token-interface/Cargo.toml) into https://github.com/AhaLabs/scaffold-stellar-frontend. The alias is not for human use; it's for tooling. It is generated automatically.

But even if people want to create ridiculously long aliases, why should the CLI limit them?

### Known limitations

None???

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/deploy/utils.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/utils.rs
@@ -2,14 +2,12 @@ use regex::Regex;
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error(
-        "alias must be 1-30 chars long, and have only letters, numbers, underscores and dashes"
-    )]
+    #[error("alias must have only letters, numbers, underscores and dashes")]
     InvalidAliasFormat { alias: String },
 }
 
 pub fn alias_validator(alias: &str) -> Result<String, Error> {
-    let regex = Regex::new(r"^[a-zA-Z0-9_-]{1,30}$").unwrap();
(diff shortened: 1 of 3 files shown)


cmd/soroban-cli/src/commands/contract/deploy/utils.rs
```diff
@@ -2,14 +2,12 @@ use regex::Regex;
...
-    #[error(
-        "alias must be 1-30 chars long, and have only letters, numbers, underscores and dashes"
-    )]
+    #[error("alias must have only letters, numbers, underscores and dashes")]
     InvalidAliasFormat { alias: String },
 }
 
 pub fn alias_validator(alias: &str) -> Result<String, Error> {
-    let regex = Regex::new(r"^[a-zA-Z0-9_-]{1,30}$").unwrap();
+    let regex = Regex::new(r"^[a-zA-Z0-9_-]{1,}$").unwrap();
```
Shaptic, line 10: I think this
```suggestion
    let regex = Regex::new(r"^[a-zA-Z0-9_-]+$").unwrap();
```
is more idiomatic regex

leighmcculloch (approved):

leighmcculloch:
Looks like a test needs updating.

PR #2033, opened 2025-05-21
Spike rpc subcommand

elizabethengelman:
### What

This is a spike to see what the rpc commands may look like if nested under an `rpc` subcommand.

### Why

### Known limitations

[TODO or N/A]


elizabethengelman:
Not going to implement this subcommand

PR #2040, opened 2025-06-05
Feat/only fee

elizabethengelman:
### What
Based on #2035 

Adds another command for seeing the breakdown of the transaction's fee. 

`stellar tx fetch fee <HASH>`

```
stellar tx fetch fee --hash dfc58e0b3a96ac27d8c0e823d4f38c3500de18b4a8b18c20e01bdf4c7010f37e --network local
     Running `target/debug/soroban tx fetch fee --hash dfc58e0b3a96ac27d8c0e823d4f38c3500de18b4a8b18c20e01bdf4c7010f37e --network local`
┌──────────────────────────────────────────────────────────────────────────────┐
│ tx.fee: 59680                                                                │
├──────────────────────────────────────────────────────────────────────────────┤
│ tx.v1.sorobanData.resourceFee: 59580                    │ inclusion fee: 100 │
├──────────────────────────────────────────────────────────────────────────────┤
│ fixed resource fee: 59501 │ refundable resource fee: 79 │ inclusion fee: 100 │
└──────────────────────────────────────────────────────────────────────────────┘
```

### Why

To help with understanding and clarity around fees. 

### Known limitations

I'd love some feedback/ideas about the specifics of how we display the fee table.

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -2394,7 +2394,7 @@ Fetch a transaction from the network by hash If no subcommand is passed in, the 
 
 * `result` — Fetch the transaction result
 * `meta` — Fetch the transaction meta
-* `fee` — 
+* `fee` — Fetch the transaction fee information
 
 ###### **Options:**
 
@@ -2475,6 +2475,8 @@ Fetch the transaction meta
 
 
 ## `stellar tx fetch fee`
+
+Fetch the transaction fee information
 
 **Usage:** `stellar tx fetch fee [OPTIONS] --hash <HASH>`
 
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -2385,19 +2385,21 @@ Fetch a transaction from the network by hash
 
 ###### **Subcommands:**
 
-* `result` — 
-* `meta` — 
-* `envelope` — 
+* `result` — Fetch the transaction result
+* `meta` — Fetch the transaction meta
+* `envelope` — Fetch the transaction envelope
 
 
 
 ## `stellar tx fetch result`
 
+Fetch the transaction result
+
 **Usage:** `stellar tx fetch result [OPTIONS] <HASH>`
 
 ###### **Arguments:**
 
-* `<HASH>`
+* `<HASH>` — Transaction hash to fetch
 
 ###### **Options:**
 
@@ -2422,11 +2424,13 @@ Fetch a transaction from the network by hash
 
 ## `stellar tx fetch meta`
 
+Fetch the transaction meta
+
 **Usage:** `stellar tx fetch meta [OPTIONS] <HASH>`
 
 ###### **Arguments:**
 
-* `<HASH>`
+* `<HASH>` — Transaction hash to fetch
 
 ###### **Options:**
 
@@ -2451,11 +2455,13 @@ Fetch a transaction from the network by hash
 
 ## `stellar tx fetch envelope`
 
+Fetch the transaction envelope
+
 **Usage:** `stellar tx fetch envelope [OPTIONS] <HASH>`
 
 ###### **Arguments:**
 
-* `<HASH>`
+* `<HASH>` — Transaction hash to fetch
 
 ###### **Options:**
 
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -2494,8 +2494,6 @@ Fetch the transaction meta
     JSON output of the ledger entry with parsed XDRs (one line, not formatted)
   - `json-formatted`:
     Formatted (multiline) JSON output of the ledger entry with parsed XDRs
-  - `xdr`:
-    Original RPC output (containing XDRs)
   - `table`:
     Formatted in a table comparing fee types
 
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -1584,7 +1584,7 @@ Sign, Simulate, and Send transactions
 * `send` — Send a transaction envelope to the network
 * `sign` — Sign a transaction envelope appending the signature to the envelope
 * `simulate` — Simulate a transaction envelope from stdin
-* `fetch` — Fetch a transaction from the network by hash
+* `fetch` — Fetch a transaction from the network by hash If no subcommand is passed in, the transaction envelope will be returned
 
 
 
@@ -2379,15 +2379,35 @@ Simulate a transaction envelope from stdin
 
 ## `stellar tx fetch`
 
-Fetch a transaction from the network by hash
+Fetch a transaction from the network by hash If no subcommand is passed in, the transaction envelope will be returned
 
-**Usage:** `stellar tx fetch <COMMAND>`
+**Usage:** `stellar tx fetch [OPTIONS]
+       fetch <COMMAND>`
 
 ###### **Subcommands:**
 
 * `result` — Fetch the transaction result
 * `meta` — Fetch the transaction meta
-* `envelope` — Fetch the transaction envelope
+
+###### **Options:**
+
+* `--hash <HASH>` — Hash of transaction to fetch
+* `--rpc-url <RPC_URL>` — RPC server endpoint
+* `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
+* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
+* `-n`, `--network <NETWORK>` — Name of network to use from config
+* `--output <OUTPUT>` — Format of the output
+
+  Default value: `json`
+
+  Possible values:
+  - `json`:
+    JSON output of the ledger entry with parsed XDRs (one line, not formatted)
+  - `json-formatted`:
+    Formatted (multiline) JSON output of the ledger entry with parsed XDRs
+  - `xdr`:
+    Original RPC output (containing XDRs)
+
 
 
 
@@ -2395,14 +2415,11 @@ Fetch a transaction from the network by hash
 
 Fetch the transaction result
 
-**Usage:** `stellar tx fetch result [OPTIONS] <HASH>`
-
-###### **Arguments:**
-
-* `<HASH>` — Transaction hash to fetch
+**Usage:** `stellar tx fetch result [OPTIONS] --hash <HASH>`
 
 ###### **Options:**
 
+* `--hash <HASH>` — Transaction hash to fetch
 * `--rpc-url <RPC_URL>` — RPC server endpoint
 * `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
 * `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
@@ -2426,45 +2443,11 @@ Fetch the transaction result
 
 Fetch the transaction meta
 
-**Usage:** `stellar tx fetch meta [OPTIONS] <HASH>`
-
-###### **Arguments:**
-
-* `<HASH>` — Transaction hash to fetch
+**Usage:** `stellar tx fetch meta [OPTIONS] --hash <HASH>`
 
 ###### **Options:**
 
-* `--rpc-url <RPC_URL>` — RPC server endpoint
-* `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
-* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
-* `-n`, `--network <NETWORK>` — Name of network to use from config
-* `--output <OUTPUT>` — Format of the output
-
-  Default value: `json`
-
-  Possible values:
-  - `json`:
-    JSON output of the ledger entry with parsed XDRs (one line, not formatted)
-  - `json-formatted`:
-    Formatted (multiline) JSON output of the ledger entry with parsed XDRs
-  - `xdr`:
-    Original RPC output (containing XDRs)
-
-
-
-
-## `stellar tx fetch envelope`
-
-Fetch the transaction envelope
-
-**Usage:** `stellar tx fetch envelope [OPTIONS] <HASH>`
-
-###### **Arguments:**
-
-* `<HASH>` — Transaction hash to fetch
-
-###### **Options:**
-
+* `--hash <HASH>` — Transaction hash to fetch
 * `--rpc-url <RPC_URL>` — RPC server endpoint
 * `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
 * `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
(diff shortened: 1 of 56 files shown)


cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    #[command(flatten)]
+    pub network: network::Args,
+
+    /// Output format for fee command
+    #[arg(long, default_value = "json")]
+    pub output: FeeOutputFormat,
+}
+
+impl FeeArgs {
```
elizabethengelman: This could probably use some DRYing up. I'm happy to look into doing that tomorrow!

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -0,0 +1,231 @@
...
+        FeeTable {
+            fee,
+            resource_fee,
+            inclusion_fee: fee - resource_fee,
+            non_refundable_resource_fee,
+            refundable_resource_fee,
+        }
+    }
+
+    fn print(&self) {
```
elizabethengelman, line 283: This produces a table like this:
```
┌───────────────────────────────────────────────────────────────────────────────────┐
│ tx.fee: 1235854                                                                   │
├───────────────────────────────────────────────────────────────────────────────────┤
│ tx.v1.sorobanData.resourceFee: 1235754                       │ inclusion fee: 100 │
├───────────────────────────────────────────────────────────────────────────────────┤
│ fixed resource fee: 57152 │ refundable resource fee: 1178602 │ inclusion fee: 100 │
└───────────────────────────────────────────────────────────────────────────────────┘
```

- Should I add the exact language from https://developers.stellar.org/docs/learn/fundamentals/fees-resource-limits-metering#resource-fee for each of these cells?
- Anything else we should include?

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -0,0 +1,224 @@
...
+                                message: "SorobanTransactionMetaExt::V0 not supported".to_string(),
+                            })
+                        }
+                        SorobanTransactionMetaExt::V1(v1) => (
+                            v1.total_non_refundable_resource_fee_charged,
+                            v1.total_refundable_resource_fee_charged,
+                        ),
+                    }
+                } else {
+                    return Err(Error::NotSupported {
```
leighmcculloch: All the cases that are not supported, I think you could assume zero is the value for each of those fee values. The tx fetch fee command should be able to show the break down for classic and soroban txs. For classic txs it is a very simple table that gets drawn.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
leighmcculloch: If it's possible to reference the exact locations with the field names, that would be super informative. But if its impractical then using the simple names is good I think.

With the field names, they may change for different txs, with new versions of meta, etc, so if we were to use field names they'd need to update with that I guess, which would be more work than would deliver value here probably.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -0,0 +1,224 @@
...
+    NotFound { tx_hash: Hash, network: String },
+    #[error(transparent)]
+    Rpc(#[from] rpc::Error),
+}
+
+impl Cmd {
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let resp = self.args.fetch_transaction(global_args).await?;
+        let tx_result = resp.clone().result.unwrap();
+        let tx_meta = resp.clone().result_meta.unwrap();
```
leighmcculloch: Instead of cloning the respond it might be better to clone the result and result_meta, but I doubt the impact of one way or the other will be super important.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
leighmcculloch: This looks good as is imo.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
elizabethengelman: Do you mean that it would be good to include `TransactionMeta::V3.soroban_meta.ex.total_non_refundable_resource_fee_charged` and `TransactionMeta::V3.soroban_meta.ex.total_refundable_resource_fee_charged` for example?

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
elizabethengelman: Good call!

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
elizabethengelman: That's super helpful 🙏 

For classic tx, would the fee just be comprised of the inclusion fee? I am seeing the transaction result returned has `fee_charged` as 100.

elizabethengelman: That's super helpful. Thank you!

FULL_HELP_DOCS.md
```diff
@@ -2394,6 +2394,7 @@ Fetch a transaction from the network by hash If no subcommand is passed in, the
 
 * `result` — Fetch the transaction result
 * `meta` — Fetch the transaction meta
+* `fee` —
```
leighmcculloch: 👀

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -0,0 +1,215 @@
...
+
+#[derive(Parser, Debug, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub(crate) args: FeeArgs,
+}
+
+#[derive(Debug, Clone, clap::Args)]
+pub struct FeeArgs {
```
leighmcculloch: What's the benefit of setting the args up this way where they are defined separately then flattened? This question doesn't need anything to change, just asking for my own learning.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
leighmcculloch: Yeah that but seems too long and wouldn't work well in the table, so I don't think we should.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -0,0 +1,215 @@
...
+        FeeTable {
+            fee,
+            resource_fee,
+            inclusion_fee: fee - resource_fee,
+            non_refundable_resource_fee,
+            refundable_resource_fee,
+        }
+    }
+
+    fn print(&self) {
```
leighmcculloch, line 283: > ```
> ┌───────────────────────────────────────────────────────────────────────────────────┐
> │ tx.fee: 1235854                                                                   │
> ├───────────────────────────────────────────────────────────────────────────────────┤
> │ tx.v1.sorobanData.resourceFee: 1235754                       │ inclusion fee: 100 │
> ├───────────────────────────────────────────────────────────────────────────────────┤
> │ fixed resource fee: 57152 │ refundable resource fee: 1178602 │ inclusion fee: 100 │
> └───────────────────────────────────────────────────────────────────────────────────┘
> ```

Something else: The table makes it looks like the bottom row should add up to the second and first rows, but the second and first rows are a max fee that'll be paid, while the last row is the actual fee that was paid. So presumably for transactions that do see a refund amount, there should be a gap at the bottom to represent some amount refunded. 

Or am I missing something?

elizabethengelman: Ah, good call. I think I had previously been incorrectly using the max fee, instead of fee_charged for the first and second rows. That has been fixed, and this is what the output looks like now:


```
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ tx.fee: 206593                                                                                           │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ tx.v1.sorobanData.resourceFee: 206493                                               │ inclusion fee: 100 │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ non-refundable resource fee: 45841                │ refundable resource fee: 160652 │ inclusion fee: 100 │
│                                                   │                                 │                    │
│ calculated based on tx.v1.sorobanData.resources.* │                                 │                    │
│                                                   │                                 │                    │
│ instructions                                      │ rent                            │                    │
│ read                                              │ events                          │                    │
│ write                                             │ return value                    │                    │
│ bandwidth (size of tx)                            │                                 │                    │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

does that look correct?

also, i'm on the fence about including the resource field names that the fee is calculated from - what do you think about including that in the table?

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
elizabethengelman: Good question! I think I had a refactor in mind to re-use the `fetch_transaction` fn from args::Args, but forgot to finish it. 

I'll get something pushed up to clean this up shortly.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        let mut table = Table::new();
+
+        table.set_format(table_format);
+
+        table.add_row(Row::new(vec![Cell::new(&format!("tx.fee: {}", self.fee))
+            .style_spec("b")
+            .with_hspan(3)]));
+
+        if self.resource_fee == 0 {
```
elizabethengelman: I added this piece to simplify the fee table for classic txns and fee bump txs. @leighmcculloch @jane Does this look ok, or do we want to add some additional info? 

```
┌─────────────────┐
│ tx.fee: 100     │
└─────────────────┘
```

janewang: Hi @elizabethengelman, 

For classic, that looks good. 

Also I think we should show:
- Max fee: Max fee specified in the transaction itself – the maximum XLM amount the source account willing to pay. Each transaction sets a fee that is paid by the source account. 
- Transaction fee (what you have): Transaction fee is what is actually charged during the transaction execution, which can be lower than the Max fee specified. (Transaction fee  = Resource fee + Inclusion fee)

janewang:
Btw, this work really elucidates the fees for soroban transactions which is hugely beneficial for devs! Thank you @elizabethengelman

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
leighmcculloch: What I was getting at is the same thing @janewang mentioned below (https://github.com/stellar/stellar-cli/pull/2040#discussion_r2150991651) in another thread, that the current table shows a mix of max vs actual fee charged, but it isn't displaying where that difference is. So I think there should be a box on the bottom row, or somewhere, that separates out the amount that wasn't charged.

elizabethengelman: Oh, gotcha! That makes sense!

elizabethengelman:
Just pushed up a change to the table, to include max fees. For a soroban tx:
```
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Transaction Fee Charged: 185119                                                                          │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Resource Fee Charged: 185019                                                        │ Inclusion Fee: 100 │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Non-refundable Resource Fee: 59343                │ Refundable Resource Fee: 125676 │ Inclusion Fee: 100 │
│                                                   │                                 │                    │
│ calculated based on tx.v1.sorobanData.resources.* │                                 │                    │
│                                                   │                                 │                    │
│ instructions                                      │ rent                            │                    │
│ read                                              │ events                          │                    │
│ write                                             │ return value                    │                    │
│ bandwidth (size of tx)                            │                                 │                    │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Max Fee Set: 248869                                                                                      │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Max Resource Fee: 248769                                                            │ Inclusion Fee: 100 │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────┘

```

And for a fee bump tx:
```
┌───────────────────────────────────┐
│ Transaction Fee Charged: 200      │
├───────────────────────────────────┤
│ Max Fee Set: 400                  │
└───────────────────────────────────┘
```

For the fee bump txns, should the inclusion fee be included in the table? Or are we okay with just the fee charged and max fee?

leighmcculloch:
The table presentation was very effective when all of the rows visually communicated that fees came together to make up the larger amounts (https://github.com/stellar/stellar-cli/pull/2040#issuecomment-2977967422). But in the new table (https://github.com/stellar/stellar-cli/pull/2040#issuecomment-2981813867), that property is lost a bit because the max fee is in the same table and its cell is the same size.

Is there a way we can represent the refund maybe as a component? The refund is a first class concept. The way that it shows up in events for example in protocol 23 is that there are two events, an initial fee event charging the max fee, and then a refund fee that pays back an amount.

For example, something like:
```
┌────────────────────────────────────────────────────────────────────────────────────┐
│ Fee Proposed: 185119                                                               │
├────────────────────────────────────────────────────────────────────────────────────┤
│ Inclusion Fee: 100 │ Resource Fee: 185019                                          │
├────────────────────────────────────────────────────────────────────────────────────┤
│ Inclusion Fee: 100 │ Non-Refundable: 59343 │ Refundable: 125676                    │
│                    │                       │                                       │
│                    │ cpu instructions      │ return value                          │
│                    │ storage read/write    │ storage rent                          │
│                    │ tx size               │ events                                │
└────────────────────────────────────────────────────────────────────────────────────┘
                           👆 Proposed Fee  👇 Final Fee
┌─────────────────────────────────────────────────────────────────┐┌─────────────────┐
│ Inclusion Fee: 100 │ Non-refundable: 59343 │ Refundable: 105676 ││ Refunded: 20000 │
├─────────────────────────────────────────────────────────────────││─────────────────┤
│ Inclusion Fee: 100 │ Resource Fee: 165019                       ││ Refunded: 20000 │
├─────────────────────────────────────────────────────────────────││─────────────────┤
│ Fee Charged: 165119                                             ││ Refunded: 20000 │
└─────────────────────────────────────────────────────────────────┘└─────────────────┘
```

elizabethengelman:
@leighmcculloch @janewang updated. I can also play with bolding and making text different colors as well. 
Screenshot 2025-06-18 at 12 30 52 PM

janewang:
@elizabethengelman https://github.com/stellar/stellar-cli/pull/2035 is merged. Could we resolve conflicts and merge this one too?

janewang (approved):

elizabethengelman:
@janewang conflicts have been resolved! Besides waiting for CI to finish, i think we'd also need an approval to get this merged in.

PR #2044, opened 2025-06-10
Add support for Protocol 23.

fnando:
### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/cmd/soroban-cli/src/log/event.rs
+++ b/cmd/soroban-cli/src/log/event.rs
@@ -24,11 +24,20 @@ pub fn contract(events: &[DiagnosticEvent], print: &Print) {
                         type_: ContractEventType::Contract,
                         ..
                     },
+                in_successful_contract_call,
                 ..
             } => {
                 let topics = serde_json::to_string(&topics).unwrap();
                 let data = serde_json::to_string(&data).unwrap();
-                print.eventln(format!("{contract_id} - Event: {topics} = {data}"));
+                let status = if in_successful_contract_call {
+                    "Success"
+                } else {
+                    "Failure"
+                };
+
+                print.eventln(format!(
+                    "{contract_id} - {status} - Event: {topics} = {data}"
+                ));
             }
 
             DiagnosticEvent {
@@ -39,11 +48,18 @@ pub fn contract(events: &[DiagnosticEvent], print: &Print) {
                         type_: ContractEventType::Diagnostic,
                         ..
                     },
+                in_successful_contract_call,
                 ..
             } => {
                 if topics[0] == xdr::ScVal::Symbol(str_to_sc_symbol("log")) {
+                    let status = if in_successful_contract_call {
+                        "Success"
+                    } else {
+                        "Failure"
+                    };
+
                     let data = serde_json::to_string(&data).unwrap();
-                    print.logln(format!("{contract_id} - Log: {data}"));
+                    print.logln(format!("{contract_id} - {status} - Log: {data}"));
                 }
             }
 


Cargo.toml
```diff
@@ -40,35 +40,40 @@ path = "./cmd/crates/soroban-spec-typescript"
...
 [workspace.dependencies.soroban-rpc]
 package = "stellar-rpc-client"
-version = "=22.1.0"
-
-[workspace.dependencies.stellar-ledger]
-version = "=22.8.2"
-path = "cmd/crates/stellar-ledger"
+version = "23.0.0-rc.4"
+git = "https://github.com/stellar/rs-stellar-rpc-client.git"
+branch = "main"
```
leighmcculloch: This is unreleased? We need to release it, otherwise the publish of the cli will fail because it references a git repo.

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -105,8 +105,11 @@ pub fn sign_soroban_authorizations(
             // See if we have a signer for this authorizationEntry
             // If not, then we Error
             let needle = match address {
+                ScAddress::MuxedAccount(_) => todo!("muxed accounts are not supported yet"),
+                ScAddress::ClaimableBalance(_) => todo!("claimable balance not supported yet"),
+                ScAddress::LiquidityPool(_) => todo!("liquidity pool not supported yet"),
```
leighmcculloch: I'd remove the word "yet", I don't think they'll ever be supported.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -121,6 +121,7 @@ impl Spec {
             | ScType::I256
             | ScType::String
             | ScType::Bool => String::new(),
+            ScType::MuxedAddress => String::from("Can be a muxed account (M13..)"),
```
leighmcculloch: Can also be a G. 

"Can be an account (G13...) or a muxed account (M13..)"

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1041,7 +1044,12 @@ fn sc_address_to_json(v: &ScAddress) -> Value {
         ScAddress::Account(AccountId(PublicKey::PublicKeyTypeEd25519(Uint256(k)))) => {
             Value::String(stellar_strkey::ed25519::PublicKey(*k).to_string())
         }
-        ScAddress::Contract(Hash(h)) => Value::String(stellar_strkey::Contract(*h).to_string()),
+        ScAddress::Contract(ContractId(h)) => {
+            Value::String(stellar_strkey::Contract(h.clone().into()).to_string())
+        }
+        ScAddress::MuxedAccount(_) => todo!("MuxedAddress is not supported yet"),
```
leighmcculloch, line 1052: We should be able to support it using the strkey lib like above.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1089,6 +1097,7 @@ impl Spec {
             ScType::U256 => Some("u256".to_string()),
             ScType::I256 => Some("i256".to_string()),
             ScType::String => Some("String".to_string()),
+            ScType::MuxedAddress => todo!("MuxedAddress is not supported yet"),
```
leighmcculloch: "MuxedAddress"

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1279,6 +1289,7 @@ impl Spec {
             ScType::Udt(ScSpecTypeUdt { name }) => {
                 self.example_udts(name.to_utf8_string_lossy().as_ref())
             }
+            ScType::MuxedAddress => todo!("MuxedAddress is not supported yet"),
```
leighmcculloch: Let's use the same example as Address, a G address. While M is supported, G will still be the common case.

cmd/crates/soroban-spec-typescript/src/types.rs
```diff
@@ -207,7 +207,7 @@ impl From<&ScSpecTypeDef> for Type {
             ScSpecTypeDef::Symbol => Type::Symbol,
             ScSpecTypeDef::Error => Type::Error { message: None },
             ScSpecTypeDef::Bytes => Type::Bytes,
-            ScSpecTypeDef::String => Type::String,
+            ScSpecTypeDef::String | ScSpecTypeDef::MuxedAddress => Type::String,
```
leighmcculloch: Should be its own type I think.

cmd/soroban-cli/src/log/event.rs
leighmcculloch, line 1: It's not clear to me what is changing in this file. Are we changing the format of the logs, or this is just a refactor?

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
```diff
@@ -309,6 +322,7 @@ mod test {
     use super::*;
 
     #[test]
+    #[ignore]
```
leighmcculloch, line 325: What's behind the need to ignore these tests?

cmd/soroban-cli/src/commands/contract/extend.rs
```diff
@@ -186,34 +188,46 @@ impl NetworkRunnable for Cmd {
...
-            return Err(Error::LedgerEntryNotFound);
-        }
+                operations[0].changes.clone()
+            }
+            TransactionMeta::V3(TransactionMetaV3 { operations, .. }) => {
+                // Simply check if there is exactly one entry here. We only support extending a single
+                // entry via this command (which we should fix separately, but).
+                if operations.is_empty() {
+                    return Err(Error::LedgerEntryNotFound);
+                }
```
leighmcculloch, line 214: It comments about checking for a single entry, but looks like it is checking for no operations. What's happening here?

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -282,6 +283,9 @@ fn resolve_address(addr_or_alias: &str, config: &config::Args) -> Result<String,
             match addr {
                 xdr::ScAddress::Account(account) => account.to_string(),
                 contract @ xdr::ScAddress::Contract(_) => contract.to_string(),
+                stellar_xdr::curr::ScAddress::MuxedAccount(_)
```
leighmcculloch: We should support muxed accounts, we need to be able to parse them. The other types though do not need to be supported, and so should return an error instead of a todo or a panic.

Cargo.toml
fnando: I thought I had released the latest changes. Here you go: https://github.com/stellar/rs-stellar-rpc-client/pull/44

cmd/crates/soroban-spec-tools/src/lib.rs
fnando: Would this be correct? `String::from("Can be a muxed account (M13..) or a public key (G13..)")`?

leighmcculloch: Yup. I would lead with the G case because it's still the common case.

cmd/soroban-cli/src/commands/tx/fetch/fee.rs
fnando: The tests are using a subset of xdr json, so it's failing with P23. @elizabethengelman will take a look at this tomorrow.

cmd/soroban-cli/src/commands/contract/extend.rs
fnando: No idea. I just kept the previous comment as it is. 😬

cmd/soroban-cli/src/log/event.rs
fnando: There was some problems with types due to P23, but I wasn't able to fix them, so refactoring helped.

cmd/crates/soroban-spec-typescript/src/types.rs
fnando: Any idea how this translates to muxed address in `@stellar/js-xdr`? Would it be a string in the end?

edit: I can see Type::Address becomes this, but can't find the actual implementation anywhere: `types::Type::Address => "addressToScVal(i)".to_owned(),`

edit: Ok, looks like in the end we'd need to call something like `MuxedAccount.fromString(addr, seqnum)`, which doesn't have an abstraction right now. I guess we can use `addressToScVal(i)` too, given it's the most common path, then we can support M-accounts.

leighmcculloch: I'd use `addressToScVal` here too, and open an issue on the js-stellar-sdk. At least G addresses passed to muxed address fields will work fine, which is the common case.

This is a good example of something that'd benefit from the typescript codegen living in close proximity to the js-stellar-sdk. cc @quietbits @jeesunikim @Shaptic

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -121,6 +121,9 @@ impl Spec {
             | ScType::I256
             | ScType::String
             | ScType::Bool => String::new(),
+            ScType::MuxedAddress => {
+                String::from("Can be public key (G13..) or a muxed account (M13..)")
+            }
             ScType::Address => String::from(
                 "Can be public key (G13..), a contract hash (6c45307) or an identity (alice), ",
```
leighmcculloch: Oh wow, this is wrong from a long time ago:
```suggestion
                "Can be public key (G13..), a contract ID (C13...) or an identity (alice), ",
```

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -121,6 +121,9 @@ impl Spec {
             | ScType::I256
             | ScType::String
             | ScType::Bool => String::new(),
+            ScType::MuxedAddress => {
+                String::from("Can be public key (G13..) or a muxed account (M13..)")
```
leighmcculloch: Sorry, actually it can be even more things:
```suggestion
                String::from("Can be public key (G13..), contract ID (C13...), or a muxed account (M13..), or an identity")
```

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -133,6 +136,7 @@ impl Spec {
                     | ScSpecEntry::UdtUnionV0(ScSpecUdtUnionV0 { doc, .. })
                     | ScSpecEntry::UdtEnumV0(ScSpecUdtEnumV0 { doc, .. })
                     | ScSpecEntry::UdtErrorEnumV0(ScSpecUdtErrorEnumV0 { doc, .. }) => doc,
+                    ScSpecEntry::EventV0(_) => todo!("EventV0 not supported yet"),
```
leighmcculloch: We should be able to extract the `doc` field out, as per the other lines.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1147,6 +1158,7 @@ impl Spec {
                     ScSpecEntry::UdtUnionV0(union) => self.arg_value_union(union, depth),
                     ScSpecEntry::UdtEnumV0(enum_) => Some(arg_value_enum(enum_)),
                     ScSpecEntry::FunctionV0(_) | ScSpecEntry::UdtErrorEnumV0(_) => None,
+                    ScSpecEntry::EventV0(_) => todo!("EventV0 is not supported yet"),
```
leighmcculloch: Should be `None` like the line above, because an event type cannot be used as a user defined type.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1317,6 +1332,7 @@ impl Spec {
                 enum_.cases.iter().next().map(|c| c.value.to_string())
             }
             Some(ScSpecEntry::FunctionV0(_) | ScSpecEntry::UdtErrorEnumV0(_)) | None => None,
+            Some(ScSpecEntry::EventV0(_)) => todo!("EventV0 is not supported yet"),
```
leighmcculloch: This shuld return `None` like the line above because an event can't be used as a function parameter.

cmd/crates/soroban-spec-typescript/src/types.rs
```diff
@@ -250,6 +252,7 @@ impl From<&ScSpecEntry> for Entry {
                 name: e.name.to_utf8_string_lossy(),
                 cases: e.cases.iter().map(Into::into).collect(),
             },
+            ScSpecEntry::EventV0(_) => todo!("EventV0 is not implemented yet"),
```
leighmcculloch, line 255: Leaving the todo in here will make the CLI panic if it encounters a spec that contains an event. Ideally we fill this in, but if we can't yet, I think we need to do something else, such as eprintln! a warning that the spec generated is incomplete, or silently hide the ommission. Otherwise it won't be possible to call anything that uses this code with newer contracts.

If we're releasing only an rc, then this is probably fine temporarily.

cmd/soroban-cli/src/log/event.rs
```diff
@@ -1,90 +1,62 @@
...
+                        contract_id: Some(contract_id),
+                        body: ContractEventBody::V0(ContractEventV0 { topics, data, .. }),
+                        type_: ContractEventType::Contract,
+                        ..
+                    },
+                ..
+            } => {
+                let topics = serde_json::to_string(&topics).unwrap();
+                let data = serde_json::to_string(&data).unwrap();
+                print.eventln(format!("{contract_id} - Event: {topics} = {data}"));
```
leighmcculloch: It's pretty important that we emit whether the event was successful or not, if we're emitting the event from the diagtostic event stream. That might be a bug we have in the CLI today. Events are still included in the diagnostic events even if the invocation that published them fails and aborts. There's a field on the `DiagnosticEvent` to indicate they were part of a successful call.

leighmcculloch (approved):

leighmcculloch (approved):

cmd/crates/soroban-spec-typescript/src/types.rs
fnando: I'm going to keep it like this for this initial release, and I'll work on that tomorrow. :)

PR #2045, opened 2025-06-12
Feat: --optimize flag for contract build

BlaineHeffron:
### What

Adds an --optimize flag to `contract build`. This uses the same logic from `contract optimize`; the optimized wasm is output with the usual name. 

Example output: 

```
ℹ️  Build Summary (Optimized):
   Wasm File: target/wasm32v1-none/release/stellar_example.wasm
   Wasm Hash: 98f44b624d792833e498b4ca33cdf2a8cebc0b84a0e402d35d17634a45969f20
   Wasm Size: 537 bytes
   Exported Functions: 2 found
     • _
     • hello
   Optimized. 574 → 537 bytes
✅ Build Complete
```

### Why

Saves an extra CLI call to `optimize`. 

### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -321,46 +380,74 @@ impl Cmd {
...
-            .parse_all(&wasm_bytes)
-            .filter_map(Result::ok)
-            .filter_map(|payload| {
-                if let wasmparser::Payload::ExportSection(exports) = payload {
-                    Some(exports)
-                } else {
-                    None
+        print.blankln(format!("Wasm Size: {} bytes", wasm_bytes.len()));
+
+        if show_exports {
```
willemneal: When is this false?

cmd/soroban-cli/src/wasm.rs
```diff
@@ -138,3 +138,23 @@ pub async fn fetch_from_contract(
     }
     Err(UnexpectedContractToken(Box::new(data_entry)))
 }
+
+#[cfg(feature = "opt")]
+pub fn optimize_wasm(input: &Path, output: &Path) -> Result<(), wasm_opt::OptimizationError> {
```
willemneal: Could move this into the `Args` struct and add the error cases there.
```
wasm::optimize_wasm(&self.wasm.wasm, &wasm_out).map_err(Error::OptimizationError)?;
```
Becomes
```
self.wasm.optimize(&wasm_out)?
```

cmd/soroban-cli/src/commands/contract/build.rs
BlaineHeffron: Oops, this was leftover code from when I was producing both wasms and I didn't want to print out the exports both times. Ill remove it.

cmd/soroban-cli/src/wasm.rs
BlaineHeffron: done

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -46,6 +46,25 @@ pub struct Cmd {
...
+    ///
+    /// If provided, wasm files can be found in the cargo target directory, and
+    /// the specified directory.
+    ///
+    /// If ommitted, wasm files are written only to the cargo target directory.
+    #[arg(long)]
+    pub out_dir: Option<std::path::PathBuf>,
+    /// Add key-value to contract meta (adds the meta to the `contractmetav0` custom section)
+    #[arg(long, num_args=1, value_parser=parse_meta_arg, action=clap::ArgAction::Append, help_heading = "Metadata")]
+    pub meta: Vec<(String, String)>,
```
leighmcculloch: This PR seems to both add new functionality and refactor existing functionality. Can we separate the two, stay hyperfocused on the new functionality being added and surgically add the new functionality? The changes here to reorganise the options into flattened additional structures within the scope seems unnecessary and unclear to me its a net benefit.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -199,27 +215,72 @@ impl Cmd {
...
+                use crate::wasm::Args as WasmArgs;
+                let orig_size = fs::metadata(&final_path).map(|m| m.len()).unwrap_or(0);
+                let tmp_optimized = final_path.with_extension("optimized.wasm");
+                let wasm_args = WasmArgs {
+                    wasm: final_path.clone(),
+                };
+                wasm_args.optimize(&tmp_optimized)?;
+                let opt_size = fs::metadata(&tmp_optimized).map(|m| m.len()).unwrap_or(0);
+                // Overwrite original file with optimized artifact:
+                fs::rename(&tmp_optimized, &final_path).map_err(Error::WritingWasmFile)?;
```
leighmcculloch: I'm not sure this rename is safe. On at least one platform Rust will use hard-linked files between the final .wasm file and intermediary files found in the target directory that it maintains for incremental builds. Simply writing into an existing .wasm file can mess with the incremental file system, resulting in future builds having surprising behaviour. Note that this might not be obvious on every contract build because it depends on whether the incremental intermediary gets used as is in the final build step or not. This is why the `handle_contract_metadata_args` function entirely removes a file (removes the link) and recreates it from scratch to ensure that the new write does not use the existing link.

This should become a non-issue if the optimize writes to a new file, which I think could be preferred because that's the behaviour of the `optimize` command.

leighmcculloch (commented):
Great idea to expose optimize functionality on the build command.

This change as is results in different behaviour between the build and optimize commands. The optimize command doesn't rewrite the file, instead creates a new file. While the build command with the optimize option rewrites the existing file. Can you make it so that they operate the same, and build results in a .optimized.wasm file? The build summary can then output the build info as is, then output an additional couple lines about the optimized file and its size.

cmd/soroban-cli/src/commands/contract/build.rs
BlaineHeffron: This was done in order to appease a clippy warning regarding not having more than 3 boolean args. I'll change it back to how it was and ignore the warning.

cmd/soroban-cli/src/commands/contract/build.rs
BlaineHeffron: Sounds good, I'll have it produce the same output as the optimize command.

cmd/soroban-cli/src/wasm.rs
```diff
@@ -92,6 +95,29 @@ impl Args {
...
+        // Formerly Soroban supported only the MVP feature set, but Rust 1.70 as
+        // well as Clang generate code with sign-ext + mutable-globals enabled,
+        // so Soroban has taken a change to support them also.
+        options.converge = true;
+        options.mvp_features_only();
+        options.enable_feature(Feature::MutableGlobals);
+        options.enable_feature(Feature::SignExt);
+        options
+            .run(&self.wasm, output)
+            .map_err(Error::OptimizationError)
```
leighmcculloch: Small tweak here so that the commentary is read to only apply to the lines that implement what the comment is talking about.

```suggestion
        let mut options = OptimizationOptions::new_optimize_for_size_aggressively();
        options.converge = true;

        // Explicitly set to MVP + sign-ext + mutable-globals, which happens to
        // also be the default featureset, but just to be extra clear we set it
        // explicitly.
        //
        // Formerly Soroban supported only the MVP feature set, but Rust 1.70 as
        // well as Clang generate code with sign-ext + mutable-globals enabled,
        // so Soroban has taken a change to support them also.
        options.mvp_features_only();
        options.enable_feature(Feature::MutableGlobals);
        options.enable_feature(Feature::SignExt);

        options
            .run(&self.wasm, output)
            .map_err(Error::OptimizationError)
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -199,27 +205,63 @@ impl Cmd {
...
-                    fs::copy(target_file_path, &out_file_path).map_err(Error::CopyingWasmFile)?;
-                    out_file_path
-                } else {
-                    target_file_path
-                };
+    fn build_package(
+        &self,
+        print: &Print,
+        package: &Package,
+        target_dir: &Utf8PathBuf,
```
leighmcculloch, line 221: We don't need introduce the direct use of camino here I think yeah?
```suggestion
        target_dir: &Path,
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -199,27 +205,63 @@ impl Cmd {
...
+            target_file_path.clone()
+        };
 
-                Self::print_build_summary(&print, &final_path)?;
+        #[cfg(feature = "additional-libs")]
+        let mut optimized_path = None;
+        #[cfg(not(feature = "additional-libs"))]
+        let optimized_path = None;
+
+        if self.optimize {
```
leighmcculloch: Instead of creating multiple variables, one which is mutable, you can have the value built in the block returned into the value.
```suggestion
        let optimized_path = if self.optimize {
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -199,27 +205,63 @@ impl Cmd {
...
 
-                let final_path = if let Some(out_dir) = &self.out_dir {
-                    fs::create_dir_all(out_dir).map_err(Error::CreatingOutDir)?;
-                    let out_file_path = Path::new(out_dir).join(&file);
-                    fs::copy(target_file_path, &out_file_path).map_err(Error::CopyingWasmFile)?;
-                    out_file_path
-                } else {
-                    target_file_path
-                };
+    fn build_package(
```
leighmcculloch: This function that logic is being moved into is named `build_package`, but the function doesn't appear to build the package. The package is built further up with the call to rustc. Maybe:
```suggestion
    fn post_build_steps(
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -321,46 +363,94 @@ impl Cmd {
...
+    fn print_build_summary(
+        print: &Print,
+        target_file_path: &PathBuf,
+        optimized_path: Option<&PathBuf>,
+    ) -> Result<(), Error> {
+        if optimized_path.is_some() {
+            print.infoln("Build Summary (Optimized):");
+        } else {
+            print.infoln("Build Summary:");
+        }
```
leighmcculloch: I think we can keep the output consistent for the top half.
```suggestion
        print.infoln("Build Summary:");
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -321,46 +363,94 @@ impl Cmd {
...
-        let parser = wasmparser::Parser::new(0);
-        let export_names: Vec<&str> = parser
-            .parse_all(&wasm_bytes)
-            .filter_map(Result::ok)
-            .filter_map(|payload| {
-                if let wasmparser::Payload::ExportSection(exports) = payload {
-                    Some(exports)
-                } else {
-                    None
+        if show_size {
```
leighmcculloch: The logic for printing the output has become rather complex with the need for labels, forks on show size and optional behaviour. I think this code would be simpler if the existing print logic remained unchanged, and 2-3 additional println calls were added for when an optimized path is present. Sometimes DRY is harder to read and maintain.

leighmcculloch (commented):
Nice. Small asks inline.

Suggestion for some small modifications to the output:
```
ℹ️  Build Summary:
   Wasm File: target/wasm32v1-none/release/stellar_example.wasm (574 bytes)
   Wasm Hash: c30e7244eb261b2204f446d15c46b8efd364e9befb5c52921454774dc9edeed3
   Exported Functions: 2 found
     • _
     • hello
   Optimized: reduced by 37 bytes
   Wasm File: target/wasm32v1-none/release/stellar_example.optimized.wasm (537 bytes)
   Wasm Hash: 98f44b624d792833e498b4ca33cdf2a8cebc0b84a0e402d35d17634a45969f20
✅ Build Complete
```

willemneal:
@leighmcculloch in stellar scaffold we place the resulting binary in `target/stellar`. What is your opinion of in the case where the is `--out-dir` and `--optimize` it overwrites the file put in this dir? This shouldn't cause the issues mentioned above with touching the binary generated by `rustc`.

cmd/soroban-cli/src/commands/contract/build.rs
BlaineHeffron: metadata.target_directory has the type `Utf8PathBuf`:
```
    /// Build directory
    pub target_directory: Utf8PathBuf,
```

cmd/soroban-cli/src/commands/contract/build.rs
BlaineHeffron: done

cmd/soroban-cli/src/commands/contract/build.rs
BlaineHeffron: I altered it so that the original print fn is the same (minus the Build Complete line) and added an optimized print which calls the original print fn and adds the lines as you requested. Then the calling fn adds the Build Complete line.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -199,27 +205,62 @@ impl Cmd {
...
+                wasm_args.optimize(&optimized_file_path)?;
+                Self::print_build_summary_with_optimization(
+                    print,
+                    &final_path,
+                    &optimized_file_path,
+                )?;
             }
+            #[cfg(not(feature = "additional-libs"))]
+            {
+                print.warnln("Optimization not available. Install with \"additional-libs\" feature to use --optimize.");
```
leighmcculloch: One last thing, this should be a hard error imo. The command was asked to optimize and the feature isn't available, so someone should have to rerun the command without --optimize.

leighmcculloch (approved):
👏🏻

leighmcculloch:
@fnando wdyt about this command outputting the optimized file as a separate file? I suggested it to @BlaineHeffron (sorry!) because that would make this command more consistent with the `optimize` command.

But, with optimize being added here, we maybe want to deprecate the `optimize` command in favour of this. I can't think of a lot of reason to use the separate `optimize` command if you can just add `--optimize` to the build command. And consistency doesn't really matter if they both don't live on.

And, when I think about most compilers, never do compilers output an unoptimized and optimized version. They output whatever the version is that was requested and only that version.

So my earlier ask was probably misguided. We could flip it back after merging this, so that @BlaineHeffron can get this merged.

cmd/soroban-cli/src/commands/contract/build.rs
BlaineHeffron: Got it

fnando:
Closing in favor of #2241

PR #2046, opened 2025-06-12
Feat/ledger latest

elizabethengelman:
### What

Add a `stellar ledger latest` command to get the latest ledger for the given network. Defaults to testnet.

It also defaults to the `text` output, which looks like this:
```
stellar ledger latest

Sequence: 1468935
Protocol Version: 22
ID: 41a18d68e814201db155ee754c1f4523f899dc31a234cb1d6435030d97ddb94f
```

`--output json` and `--output json-formatted` are also available.
### Why

To allow for users to easily look up that latest ledger from the CLI.

### Known limitations

N/A


cmd/soroban-cli/src/commands/ledger/latest.rs
```diff
@@ -0,0 +1,54 @@
...
+    Text,
+    /// JSON result of the RPC request
+    Json,
+    /// Formatted (multiline) JSON output of the RPC request
+    JsonFormatted,
+}
+
+#[derive(Debug, clap::Parser)]
+pub struct Cmd {
+    pub seq: Option<i64>,
```
leighmcculloch: This option looks unused, is that right?

leighmcculloch (commented):
Looks good. One question inline.

cmd/soroban-cli/src/commands/ledger/latest.rs
elizabethengelman: good catch!

elizabethengelman: just pushed a fix!

FULL_HELP_DOCS.md
```diff
@@ -2704,3 +2705,40 @@ List installed plugins
...
 
+## `stellar ledger`
+
+Fetch ledger information
+
+**Usage:** `stellar ledger <COMMAND>`
+
+###### **Subcommands:**
+
+* `latest` —
```
leighmcculloch: 👀 Looks like docs are missing for this subcommand.

cmd/soroban-cli/src/commands/ledger/latest.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let network = self.network.get(&global_args.locator)?;
+        let client = network.rpc_client()?;
+        let ledger = client.get_latest_ledger().await?;
+
+        match self.output {
+            OutputFormat::Text => {
+                println!("Sequence: {}", ledger.sequence);
+                println!("Protocol Version: {}", ledger.protocol_version);
+                println!("ID: {}", ledger.id);
```
leighmcculloch: We should refer tot his field as hash. I realise that the stellar-rpc does refer to it as an `id` on one endpoint. But that's a mistake I think we should correct. The value is the ledger hash and we refer it to that in a lot of places, but not on the latest ledger endpoint for some reason.
```suggestion
                println!("Hash: {}", ledger.id);
```

cmd/soroban-cli/src/commands/ledger/latest.rs
```diff
@@ -0,0 +1,52 @@
...
+    JsonFormatted,
+}
+
+#[derive(Debug, clap::Parser)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub network: network::Args,
+
+    /// Format of the output
+    #[arg(long, default_value = "text")]
```
leighmcculloch: Because the type has default implemented on it, you should be able to reference the type for its default. I think the syntax is:
```suggestion
    #[arg(long, default_value_t)]
```

FULL_HELP_DOCS.md
elizabethengelman: good call, just pushed up a fix!

cmd/crates/soroban-test/tests/it/integration/ledger.rs
```diff
@@ -0,0 +1,16 @@
...
+    sandbox
+        .new_assert_cmd("ledger")
+        .arg("latest")
+        .arg("--network")
+        .arg("testnet")
+        .assert()
+        .success()
+        .stdout(predicates::str::contains("Sequence:"))
+        .stdout(predicates::str::contains("Protocol Version:"))
+        .stdout(predicates::str::contains("ID:"));
```
leighmcculloch:
```suggestion
        .stdout(predicates::str::contains("Hash:"));
```

cmd/soroban-cli/src/commands/ledger/mod.rs
```diff
@@ -0,0 +1,24 @@
+use crate::commands::global;
+use clap::Subcommand;
+mod latest;
+
+#[derive(Debug, Subcommand)]
+pub enum Cmd {
+    /// Get the latest ledger sequence formation from the network
```
leighmcculloch:
```suggestion
    /// Get the latest ledger sequence and information from the network
```

leighmcculloch (approved):
Small tweaks suggested, otherwise looks great 👍🏻

PR #2055, opened 2025-07-02
default to testnet  when no network specified

sagpatil:
### Why
Solves
https://github.com/stellar/stellar-cli/issues/2049

### DIFF
--- a/cmd/soroban-cli/src/config/network.rs
+++ b/cmd/soroban-cli/src/config/network.rs
@@ -105,12 +105,6 @@ impl Args {
             self.network_passphrase.clone(),
         ) {
             (None, None, None) => {
-                // Try to read default network from config first
-                if let Ok(config) = crate::config::Config::new() {
-                    if let Some(default_network) = config.defaults.network {
-                        return Ok(locator.read_network(&default_network)?);
-                    }
-                }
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/config/network.rs
```diff
@@ -104,7 +104,16 @@ impl Args {
...
             self.network_passphrase.clone(),
         ) {
-            (None, None, None) => Err(Error::Network),
+            (None, None, None) => {
+                // Try to read default network from config first
+                if let Ok(config) = crate::config::Config::new() {
+                    if let Some(default_network) = config.defaults.network {
+                        return Ok(locator.read_network(&default_network)?);
+                    }
+                }
```
elizabethengelman: I think that this piece of code may be redundant / never be executed. We get the defaults from the config in the [main cli fn](https://github.com/stellar/stellar-cli/blob/34b7fbd59f90a03d84f606850dd7e08b10f36bd4/cmd/soroban-cli/src/cli.rs#L43) that is run whenever the cli is executed, and set them as env vars. So by the time we reach this fn we should already have the network env var (STELLAR_NETWORK) set and won't make it into the `(None, None, None)` match arm. 

I think just line 115 is all we need for this!

PR #2065, opened 2025-07-09
Fix/refactor signing paths

elizabethengelman:
### What

Closes https://github.com/stellar/stellar-cli/issues/2008

### Why
We previously had 2 paths to signing tx, one was used in the `tx sign` (sign_with), and the other path was used in the rest of the tx signing (using signer.sign directly). Making sure that all signing goes through one path seems less likely to introduce bugs when working on the signing flow. 

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/config/mod.rs
+++ b/cmd/soroban-cli/src/config/mod.rs
@@ -8,7 +8,7 @@ use std::{
 use crate::{
     print::Print,
     signer,
-    xdr::{self, SequenceNumber, Transaction, TransactionEnvelope},
+    xdr::{self, SequenceNumber, Transaction, TransactionEnvelope, TransactionV1Envelope, VecM},
     Pwd,
 };
 use network::Network;
@@ -40,6 +40,8 @@ pub enum Error {
     Rpc(#[from] soroban_rpc::Error),
     #[error(transparent)]
     Signer(#[from] signer::Error),
+    #[error(transparent)]
+    SignWith(#[from] sign_with::Error),
     #[error(transparent)]
     StellarStrkey(#[from] stellar_strkey::DecodeError),
     #[error(transparent)]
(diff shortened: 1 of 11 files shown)


cmd/soroban-cli/src/config/sign_with.rs
```diff
@@ -66,6 +67,29 @@ impl Args {
...
+            .await
+    }
+
+    async fn sign_env(
+        &self,
+        tx: &TransactionEnvelope,
+        locator: &locator::Args,
+        network: &Network,
+        quiet: bool,
+        source_account: Option<UnresolvedMuxedAccount>,
```
leighmcculloch: What does the `_with_default_signer` mean? Does it mean that the signing will occur with the source account? It looks to me like that is what is happening.

I think it might be easier to follow to remove the first function, and add a comment to `sign_env` (maybe renamed to `sign_tx`), that indicates if a `source_account` is provided, it will sign with it if no other signer is specified. That is what `sign_env` does anyway.

leighmcculloch (approved):
👏🏻 Thanks for making this change so surgically. I had expected it to be more disruptive.

One concern inline, but it is a small aspect.

cmd/soroban-cli/src/config/sign_with.rs
elizabethengelman: Good question! Like you said, this allows for defaulting to signing with the source account if the user doesn't pass a signer. This was the behavior for all of the signing flows except for `stellar tx sign`, so I wanted to make sure to keep that behavior as-is.

But you're right, we can use just one fn since the source_account field is optional. I'll get a fix pushed up for this shortly!

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -90,12 +90,12 @@ impl Args {
...
         Ok(self
             .sign_with
-            .sign_tx_env_with_default_signer(
+            .sign_tx_env(
                 &tx_env,
                 &self.locator,
                 &self.network.get(&self.locator)?,
                 false,
-                self.source_account.clone(),
+                Some(self.source_account.clone()),
```
leighmcculloch: Can the function take an `Option<&_>` so that the clone is unnecessary?

elizabethengelman: good call - just pushed an update

PR #2076, opened 2025-07-17
Use 8 cores instead.

fnando:
### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/.github/workflows/rpc-tests.yml
+++ b/.github/workflows/rpc-tests.yml
@@ -6,7 +6,9 @@ on:
   pull_request:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref_protected == 'true' && github.sha || github.ref }}
+  group:
+    ${{ github.workflow }}-${{ github.ref_protected == 'true' && github.sha ||
+    github.ref }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/rpc-tests.yml
```diff
@@ -33,20 +36,26 @@ jobs:
...
         if: runner.os == 'Linux'
         run: |
           gnome-keyring-daemon
       - run: cargo build --features additional-libs
       - run: rustup target add wasm32v1-none
       - run: make build-test-wasms
-      - run: cargo test --features it --package soroban-test --test it -- integration --test-threads=1
+      - run:
+          cargo test --features it --package soroban-test --test it --
+          integration --test-threads=1
```
Shaptic, line 61: Wouldn’t you want to use 8 threads here?

fnando: honestly, no idea why is using only 1 thread and I just assumed it was something that I didn't have the context.

@leighmcculloch do you know?

Shaptic: definitely not a merge blocker but just stood out to me given the context

.github/workflows/rpc-tests.yml
```diff
@@ -33,20 +36,26 @@ jobs:
       - uses: actions/checkout@v4
       - uses: stellar/actions/rust-cache@main
       - run: rustup update
-      - run: sudo apt update && sudo apt install -y libudev-dev libdbus-1-dev gnome-keyring
+      - run:
+          sudo apt update && sudo apt install -y libudev-dev libdbus-1-dev
+          gnome-keyring
```
Shaptic, line 41: Will this work with a line break?? We always add `\`s but it'd be great to know we don't need to lol

Shaptic (approved):

.github/workflows/rpc-tests.yml
fnando: yamllint formatter. It's better not to fight it. YAML's is single lined by default unless you use `|`.

.github/workflows/rpc-tests.yml
leighmcculloch: Typically you would only add `--test-threads=1` to `cargo test` if you wanted the tests to run sequentially, instead of in parallel. Parallel is the default. It might be that there was some case where the tests didn't work properly when executing concurrently. I don't know.

PR #2082, opened 2025-07-18
Do not spawn a fire-and-forget thread for upgrade check.

fnando:
### What

Improve how we do upgrade checks.

### Why

To improve reliability.

### Known limitations

N/A


leighmcculloch (commented):
The fire and forget was intentional, to not slow down cli commands that are supposed to be fast. The existing design is based on what we saw in some other commands that do checks in a non-instrusive way.

Basically it's okay to kill the upgrade check before it's finished if the user is doing something that's finishing fast. The upgrade check shouldn't intrusively slow down what a user is doing.

Issue #1721, opened 2024-11-11
stellar contract build --meta results in one new section per entry


leighmcculloch:
> the cli is writing a whole new section per entry, but it should be joining the new entries together and writing them in one new entry, otherwise the wasm is unnecessarily larger due to managing one section per meta.

_Originally posted by @leighmcculloch in https://github.com/stellar/stellar-cli/issues/1707#issuecomment-2463682157_
            

cc @elizabethengelman

PR #2095, opened 2025-07-24
Fixes duplicate contractmetav0 custom sections when users add meta on build

elizabethengelman:
### What

Closes https://github.com/stellar/stellar-cli/issues/1721

Updates the logic to add new contract meta data to the contract wasm on build so that there is only one `contractmetav0` custom section.

Here's a diff of the before (on the left) vs after with this branch (on the right): https://www.diffchecker.com/fshP64p8/

These are the commands that were run to get the outputs for the diff:
```bash
// before (left)
stellar contract build --manifest-path ./cmd/crates/soroban-test/tests/fixtures/workspace-with-default-members/Cargo.toml --meta key1=value1 --meta key2=value2

wasm-tools parse cmd/crates/soroban-test/tests/fixtures/workspace-with-default-members/target/wasm32v1-none/release/add.wasm -t


// with this branch (right)
cargo run contract build --manifest-path ./cmd/crates/soroban-test/tests/fixtures/workspace-with-default-members/Cargo.toml --meta key1=value1 --meta key2=value2

wasm-tools parse cmd/crates/soroban-test/tests/fixtures/workspace-with-default-members/target/wasm32v1-none/release/add.wasm -t
```

Output from uploading and deploying the resulting wasm:
```
❯ cargo run contract upload --wasm cmd/crates/soroban-test/tests/fixtures/workspace-with-default-members/target/wasm32v1-none/release/add.wasm --source ee --network testnet
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/soroban contract upload --wasm cmd/crates/soroban-test/tests/fixtures/workspace-with-default-members/target/wasm32v1-none/release/add.wasm --source ee --network testnet`
⚠️ A new release of stellar-cli is available: 22.8.2 -> 23.0.0
ℹ️ Simulating install transaction…
ℹ️ Signing transaction: 866da95a8df01f6c63756367c383764c4a38d556268cbd7b78574aaa896849a8
🌎 Submitting install transaction…
3f7bebccdd3f43bb03551844e95b0cb3c6a53b2393bb8c2fdd268090447e5dff


❯  cargo run contract deploy --wasm-hash 3f7bebccdd3f43bb03551844e95b0cb3c6a53b2393bb8c2fdd268090447e5dff --source ee --network testnet
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/soroban contract deploy --wasm-hash 3f7bebccdd3f43bb03551844e95b0cb3c6a53b2393bb8c2fdd268090447e5dff --source ee --network testnet`
ℹ️ Using wasm hash 3f7bebccdd3f43bb03551844e95b0cb3c6a53b2393bb8c2fdd268090447e5dff
⚠️ A new release of stellar-cli is available: 22.8.2 -> 23.0.0
ℹ️ Transaction hash is c93ec680d376f64ea14c65fe94c92f780777b40c97857ef34daf9c3dbcbd998f
🔗 https://stellar.expert/explorer/testnet/tx/c93ec680d376f64ea14c65fe94c92f780777b40c97857ef34daf9c3dbcbd998f
ℹ️ Signing transaction: c93ec680d376f64ea14c65fe94c92f780777b40c97857ef34daf9c3dbcbd998f
🔗 https://stellar.expert/explorer/testnet/contract/CDNDIUQLAVWBRNXEYEWPKLPNLN74I5TVC4DCROJMROX5FCCJ5GUH4WTP
✅ Deployed!
CDNDIUQLAVWBRNXEYEWPKLPNLN74I5TVC4DCROJMROX5FCCJ5GUH4WTP
```

### Why

This will make it so the wasm isn't larger than it needs to be with duplicate `contractmetav0` sections. This hopefully will also make it easier for the data team to fetch contract data.

### Known limitations

This does reencode the whole wasm file, instead of just appending a new section. Given the tooling, this was the only way I could figure out how to remove the default `contractmetav0` and include just the new, updated section with the user's meta.

### DIFF
--- a/.github/workflows/bindings-ts.yml
+++ b/.github/workflows/bindings-ts.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: stellar/quickstart@main
         with:
           tag: testing
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: "20.x"
       - uses: actions/checkout@v5
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,7 +10,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@v9
+      - uses: actions/stale@v10
         with:
           debug-only: false
           days-before-stale: 90
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -1706,6 +1706,7 @@ Create a new transaction
 ###### **Subcommands:**
 
 * `account-merge` — Transfer XLM balance to another account and remove source account
+* `begin-sponsoring-future-reserves` — Begin sponsoring future reserves for another account
 * `bump-sequence` — Bump sequence number to invalidate older transactions
 * `change-trust` — Create, update, or delete a trustline
 * `claim-claimable-balance` — Claim a claimable balance by its balance ID
@@ -1714,13 +1715,16 @@ Create a new transaction
 * `create-account` — Create and fund a new account
 * `create-claimable-balance` — Create a claimable balance that can be claimed by specified accounts
 * `create-passive-sell-offer` — Create a passive sell offer on the Stellar DEX
+* `end-sponsoring-future-reserves` — End sponsoring future reserves
 * `liquidity-pool-deposit` — Deposit assets into a liquidity pool
+* `liquidity-pool-withdraw` — Withdraw assets from a liquidity pool
 * `manage-buy-offer` — Create, update, or delete a buy offer
 * `manage-data` — Set, modify, or delete account data entries
 * `manage-sell-offer` — Create, update, or delete a sell offer
 * `path-payment-strict-send` — Send a payment with a different asset using path finding, specifying the send amount
 * `path-payment-strict-receive` — Send a payment with a different asset using path finding, specifying the receive amount
 * `payment` — Send asset to destination account
+* `revoke-sponsorship` — Revoke sponsorship of a ledger entry or signer
 * `set-options` — Set account options like flags, signers, and home domain
 * `set-trustline-flags` — Configure authorization and trustline flags for an asset
 
@@ -1752,6 +1756,35 @@ Transfer XLM balance to another account and remove source account
 * `--sign-with-lab` — Sign with https://lab.stellar.org
 * `--sign-with-ledger` — Sign with a ledger wallet
 * `--account <ACCOUNT>` — Muxed Account to merge with, e.g. `GBX...`, 'MBX...'
+
+
+
+## `stellar tx new begin-sponsoring-future-reserves`
+
+Begin sponsoring future reserves for another account
+
+**Usage:** `stellar tx new begin-sponsoring-future-reserves [OPTIONS] --source-account <SOURCE_ACCOUNT> --sponsored-id <SPONSORED_ID>`
+
+###### **Options:**
+
+* `--fee <FEE>` — fee amount for transaction, in stroops. 1 stroop = 0.0000001 xlm
+
+  Default value: `100`
+* `--cost` — Output the cost execution to stderr
+* `--instructions <INSTRUCTIONS>` — Number of instructions to simulate
+* `--build-only` — Build the transaction and only write the base64 xdr to stdout
+* `--rpc-url <RPC_URL>` — RPC server endpoint
+* `--rpc-header <RPC_HEADERS>` — RPC Header(s) to include in requests to the RPC provider
+* `--network-passphrase <NETWORK_PASSPHRASE>` — Network passphrase to sign the transaction sent to the rpc server
+* `-n`, `--network <NETWORK>` — Name of network to use from config
+* `-s`, `--source-account <SOURCE_ACCOUNT>` [alias: `source`] — Account that where transaction originates from. Alias `source`. Can be an identity (--source alice), a public key (--source GDKW...), a muxed account (--source MDA…), a secret key (--source SC36…), or a seed phrase (--source "kite urban…"). If `--build-only` or `--sim-only` flags were NOT provided, this key will also be used to sign the final transaction. In that case, trying to sign with public key will fail
+* `--global` — ⚠️ Deprecated: global config is always on
+* `--config-dir <CONFIG_DIR>` — Location of config directory. By default, it uses `$XDG_CONFIG_HOME/stellar` if set, falling back to `~/.config/stellar` otherwise. Contains configuration files, aliases, and other persistent settings
+* `--sign-with-key <SIGN_WITH_KEY>` — Sign with a local key or key saved in OS secure storage. Can be an identity (--sign-with-key alice), a secret key (--sign-with-key SC36…), or a seed phrase (--sign-with-key "kite urban…"). If using seed phrase, `--hd-path` defaults to the `0` path
+* `--hd-path <HD_PATH>` — If using a seed phrase to sign, sets which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
+* `--sign-with-lab` — Sign with https://lab.stellar.org
+* `--sign-with-ledger` — Sign with a ledger wallet
+* `--sponsored-id <SPONSORED_ID>` — Account that will be sponsored
 
 
 
@@ -1901,7 +1934,7 @@ Clawback a claimable balance by its balance ID
 * `--hd-path <HD_PATH>` — If using a seed phrase to sign, sets which hierarchical deterministic path to use, e.g. `m/44'/148'/{hd_path}`. Example: `--hd-path 1`. Default: `0`
 * `--sign-with-lab` — Sign with https://lab.stellar.org
 * `--sign-with-ledger` — Sign with a ledger wallet
-* `--balance-id <BALANCE_ID>` — Balance ID of the claimable balance to clawback. Accepts multiple formats: - API format with type prefix (72 chars): 000000006f2179b31311fa8064760b48942c8e166702ba0b8fbe7358c4fd570421840461 - Direct hash format (64 chars): 6f2179b31311fa8064760b48942c8e166702ba0b8fbe7358c4fd570421840461 - StrKey format (base32): BAAMLBZI42AD52HKGIZOU7WFVZM6BPEJCLPL44QU2AT6TY3P57I5QDNYIA
+* `--balance-id <BALANCE_ID>` — Balance ID of the claimable balance to clawback. Accepts multiple formats: - API format with type prefix (72 chars): 000000006f2179b31311fa8064760b48942c8e166702ba0b8fbe7358c4fd570421840461 - Direct hash format (64 chars): 6f2179b31311fa8064760b48942c8e166702ba0b8fbe7358c4fd570421840461 - Address format (base32): BAAMLBZI42AD52HKGIZOU7WFVZM6BPEJCLPL44QU2AT6TY3P57I5QDNYIA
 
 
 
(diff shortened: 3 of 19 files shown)


cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -295,8 +299,54 @@ impl Cmd {
...
+        let wasm_bytes = fs::read(target_file_path).map_err(Error::ReadingWasmFile)?;
+        let existing_meta: Vec<ScMetaEntry> = Spec::new(&wasm_bytes).unwrap().meta;
+
+        let mut module = Module::new();
+
+        for payload in WasmParser::new(0).parse_all(&wasm_bytes) {
+            match payload.unwrap() {
+                Payload::CustomSection(section) => {
+                    if section.name() == META_CUSTOM_SECTION_NAME {
+                        let updated_meta = self.append_new_meta(&existing_meta).unwrap();
```
leighmcculloch: This code assumes that the file contains only a single `contractmetav0` custom section. While that happens to be the case today during build, it's somewhat brittle to behave unexpected if that changes about how LLVM (what rustc uses to compile and construct the wasm file) builds custom sections in the future. If that changed, this logic would rewrite the existing meta multiple times. This logic would be more resilient I think if it ensure only a single meta section was writtne.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -295,8 +299,54 @@ impl Cmd {
...
+        let updated_wasm_bytes = module.finish();
 
+        // Deleting .wasm file effectively unlinking it from /release/deps/.wasm preventing from overwrite
+        // See https://github.com/stellar/stellar-cli/issues/1694#issuecomment-2709342205
+        fs::remove_file(target_file_path).map_err(Error::DeletingArtifact)?;
+        fs::write(target_file_path, updated_wasm_bytes).map_err(Error::WritingWasmFile)
+    }
+
+    fn append_new_meta(&self, existing_meta: &[ScMetaEntry]) -> Result<Vec<u8>, Error> {
+        let mut updated_meta = existing_meta.to_owned();
```
leighmcculloch: This line makes a copy of the vec in memory, which is unnecessary because the only use of the vec being passed in is to append. So you can accept a `&mut Vec<ScMetaEntry>` and modify it directly to avoid the copy.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -295,8 +299,54 @@ impl Cmd {
             return Ok(());
         }
 
-        let mut wasm_bytes = fs::read(target_file_path).map_err(Error::ReadingWasmFile)?;
+        // get existing wasm bytes
+        let wasm_bytes = fs::read(target_file_path).map_err(Error::ReadingWasmFile)?;
+        let existing_meta: Vec<ScMetaEntry> = Spec::new(&wasm_bytes).unwrap().meta;
```
leighmcculloch: Why do we use the `Spec` class to read the meta out of the `wasm_bytes`? I would expect that we wouldn't need to parse the existing meta. The new meta can be encoded to XDR on its own and appended to the existing meta's bytes without decoding the existing meta.

leighmcculloch (commented):
This change isn't quite what I meant in #1721, but it's better!

The issue was reporting that:
> the cli is writing a whole new section per entry, but it should be joining the new entries together and writing them in one new entry

The issue was that when appending N new meta after build, the file would contain N new custom sections, instead of 1 new custom section. The intent was to still append a new custom section without modifying any of the existing custom sections, but to add all the meta entries into 1 new custom section.

The change in this PR is a bit different because it is fully decoding the file and rewriting the entire file. I had avoided rewriting the file in the past out of a fear we could introduce a problem with the contract by fiddling with the existing sections. But we were likely going to end up needing to rewrite files at some point anyway someday, and this change looks good and straight forward.

Rewriting has the outcome that the resulting file should be a tiny bit smaller than if we kept with appending a 2nd section.

I like that we've shifting from wasm-gen to wasm-encoder, because the former has not had updates for years and the latter is maintained by the Bytecode Alliance.

Few issues inline.

cmd/soroban-cli/src/commands/contract/build.rs
leighmcculloch: For how to address this, I suggest iterating over the entire file, and collecting the contents of all `contractmetav0` into a `Vec<u8>` and _not_ adding them into the new `module`. All other sections, including all other custom sections, can be added via the raw section logic that exists in the PR further down. Then, serialize the new meta appending it into the collected contents of existing meta sections, and add the `contractmetav0` section with the collected + new meta into the `module`. Following that process should result in only a single meta section no matter how many existed prior to this code being called.

cmd/soroban-cli/src/commands/contract/build.rs
elizabethengelman: Good catch - I had meant to go back and fix that!

elizabethengelman: update in 03e3b73ae4e43be4edfdbee8301b90559e472d28

cmd/soroban-cli/src/commands/contract/build.rs
elizabethengelman: I replaced `append_new_meta` with `encode_new_meta` so think that this should no longer be an issue

cmd/soroban-cli/src/commands/contract/build.rs
elizabethengelman: Good call with this! I just pushed up a change that collects all of the existing metas, and then appends them with the new ones, and finally adds them to the module.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -295,8 +302,58 @@ impl Cmd {
...
+                        // collect all existing meta data into one collection to merge with new meta, and then add to the module at the end
+                        existing_meta.extend_from_slice(section.data());
+                    } else {
+                        let custom = CustomSection {
+                            name: section.name().into(),
+                            data: section.data().into(),
+                        };
+                        module.section(&custom);
+                    }
+                }
```
leighmcculloch: Can this section be simplified as follows? It looks to me like the else serves the same purpose as the raw section copy below it and that it probably doesn't need both.
```suggestion
                Payload::CustomSection(section) if section.name() == META_CUSTOM_SECTION_NAME => {
                    // collect all existing meta data into one collection to merge with new meta, and then add to the module at the end
                    existing_meta.extend_from_slice(section.data());
                }
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -295,8 +302,58 @@ impl Cmd {
...
+                other => {
+                    // Reconstruct raw section bytes and add them to the new module
+                    if let Some((id, range)) = other.as_section() {
```
leighmcculloch, line 319: The docs for `.as_section()` say that:
> Not all payloads refer to entire sections, such as the `Version` and `CodeSectionEntry` variants. These variants will return `None` from this function.

Does that mean when rewriting the wasm file in this matter, where we are only copying over entire sections, that we're discarding some information, or that information is being replaced by the `wasm-encoder`? This might be fine, I'm just wondering if we're going to get bitten by that.

elizabethengelman: Good question! From what I understand, it looks like  `CodeSectionEntry` sections are kind of "subsections", and are nested under a top level section as a stream. Since we are included the full content of a section into a `RawSection` that information should be included. 

And then I think that for the `Version` section that is something that is automatically generated by encoder on `finish()`.

This info is mostly from chatgpt, so i'm taking it with a grain of salt, and will keep doing some more research to make sure its correct.

elizabethengelman: I was able to confirm that the Version is added to a new module automatically by wasm_encoder, however it seems to be added in the `new` fn: https://github.com/bytecodealliance/wasm-tools/blob/c2fc211e3259292ea91afdabacd93b1e1c7470fb/crates/wasm-encoder/src/core.rs#L137

elizabethengelman: To verify the `CodeSectionEntry` piece, from the doc comment:
    /// An entry of the code section, a function, was parsed from a WebAssembly
    /// module.
    ///
    /// This entry indicates that a function was successfully received from the
    /// code section, and the payload here is the window into the original input
    /// where the function resides. Note that the function itself has not been
    /// parsed, it's only been outlined. You'll need to process the
    /// `FunctionBody` provided to test whether it parses and/or is valid.

I also printed out the "other" payloads, and this seems to match my chatgpt-aided understanding
```
===========> other: CodeSectionStart { count: 3, range: 135..390, size: 254 }
===========> other: CodeSectionEntry("...")
===========> other: CodeSectionEntry("...")
===========> other: CodeSectionEntry("...")
===========> other: CustomSection(CustomSectionReader { name: "contractspecv0", data_offset: 407, data: "...", range: 392..467 })
===========> other: CustomSection(CustomSectionReader { name: "contractenvmetav0", data_offset: 487, data: "...", range: 469..499 })
```

leighmcculloch: @graydon do you have thoughts about how safe you think it is for the stellar-cli to reconstruct teh entire wasm file using [`wasm-encoder`]?

Today the stellar-cli only appends new custom sections, and so doesn't rewrite or touch anything already existing in a file. This would change that behaviour to rewriting the entire file and rebuilding it using the wasm-encoder crate.

For this specific change we could avoid rewriting the whole wasm file by still just appending the meta into a single new contractmetav0 custom section as I described in https://github.com/stellar/stellar-cli/pull/2095#pullrequestreview-3053867954, but we have other open issues that'd require rewriting the file anyway so if we avoid doing it here we probably aren't avoiding doing it in the future.

[`wasm-encoder`]: https://crates.io/crates/wasm-encoder

leighmcculloch: Based on manual testing of some verify basic wasms built with and without this change, nothing jumps out at me as being a big concern today. My concern is really if over time issues arise.

@elizabethengelman Have you done any testing with contracts that have been built and are in use on mainnet? E.g. Blend, Soroswap, Axelar, so on, we could try building them with and without this logic, then inspect the binary to ensure nothing significant is changing other than the custom sections.

elizabethengelman: @leighmcculloch I have not, but I'm happy to do so!

elizabethengelman: @leighmcculloch I did some testing with Blend v2 contracts, and it looks like the wasm generated without additional meta added at build, is the same between building with stellar-cli v 23 and building with this branch. Here’s a diff on the pool_factory contract to show that: https://www.diffchecker.com/u3D6LcUT/

When comparing adding meta on build, the diff seems to be just the `contractmetav0` section(s), which is expected. Here’s the diff, on the left is what is generated with stellar-cli v23, and the right is what is generated with this branch: https://www.diffchecker.com/ufMyl13F/

I’ll also note that there was more of a diff between building with stellar-cli v23 and what is currently on mainnet. Not sure if this is anything to worry about, but I wanted to point it out.

I also tested Blend's backstop contract and found similar results.

elizabethengelman: The Soroswap [contracts that I found ](https://github.com/soroswap/core/tree/main/contracts) seemed to be using an old soroban-sdk version, so i was seeing some errors when trying to build with an up-to-date cli version.
```
error: unexpected `cfg` condition value: `testutils`
  --> /Users/ebethme/Desktop/projects/stellar/soroswap-core/contracts/factory-interface/src/lib.rs:12:1
   |
12 | #[contractclient(name = "SoroswapFactoryClient")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

leighmcculloch: The CLI just calls cargo, so this would suggest maybe the version of Rust in use is not compatible.

According to the CLI the rust version used to build the Soroswap Factory contract on mainnet is 1.75.0:
```
$ stellar contract info meta --id CA4HEQTL2WPEUYKYKCDOHCDNIV4QHNJ7EL4J4NQ6VADP7SYHVRYZ7AW2
ℹ️ Network: Public Global Stellar Network ; September 2015
🌎 Downloading contract spec: CA4HEQTL2WPEUYKYKCDOHCDNIV4QHNJ7EL4J4NQ6VADP7SYHVRYZ7AW2
Contract meta:
 • rsver: 1.75.0 (Rust version)
 • rssdkver: 20.2.0#6e198b79a51c48ccc8f22b02dcc4046d8cb7a887 (Soroban SDK version and its commit hash)
```

Could try using `rustup` to install and use 1.75.0 to see?

elizabethengelman: I tried downgrading to rust version 1.75.0 but was seeing other errors when trying to build the stellar cli because it doesn't support a rust version that old, nor do a lot of its dependencies.
```
error: failed to parse lock file at: /Users/ebethme/Desktop/projects/stellar/stellar-cli/Cargo.lock

Caused by:
  lock file version `4` was found, but this version of Cargo does not understand this lock file, perhaps Cargo needs to be updated?

❯ cargo build
error: package `soroban-test v23.0.1 (/Users/ebethme/Desktop/projects/stellar/stellar-cli/cmd/crates/soroban-test)` cannot be built because it requires rustc 1.89.0 or newer, while the currently active rustc version is 1.75.0
```

I suppose I could try to run the underlying cargo rustc command, or downgrade the stellar cli version... but I'm not sure how to then compare the wasm it generates with wasm with the new meta logic. 🤔

graydon: @leighmcculloch I don't feel especially strongly. I mean, I guess I am worried/skeptical but not about any _specific_ failure mode. I would say:

1. wasm itself (really any object file format) will have a bunch of not-obvious-from-the-APIs invariants and relationships that have to be upheld for the resulting file to be correct as per the spec.
2. if you happen to generate something incorrect, it might or might not be caught by tools like the validator/parser. being caught and rejected is the good case! it's also possible we'll just wind up with some incorrect wasms on chain forever that "happen to work" for a while, until in some other utterly inscrutable context (eg. after a bugfix to wasmparser) they don't.
3. the same is also true of the soroban env host/guest interface. a bunch of rules and assumptions, some baked in to different levels of the tooling and SDK, probably a few of them might break if you do this, and probably not all such breakage will be immediately caught, might linger on chain.
4. the same is _also_ true of the internal invariants and assumptions of the rust language. you might literally break some part of rust's semantics by doing this. and that's even less likely to be caught by anyone.
5. the same is _also_ true of the internal invariants and assumptions of a given contract (eg. accounting/authorization/storage/etc. logic), that nothing other than their own unit tests will notice you breaking. it's fully possible to miscompile (or mis-edit) user code and make it parse, validate, run, and just _do something the user didn't want_ when it runs.

given the number of layers at which something can go wrong, I would recommend against perturbing much of the stack if you can avoid it. but if you can't avoid it, I mean, there's nothing _inherently_ wrong with doing this, in the sense that "everything will definitely break". everything might be ok!

the soroban env in any case has no idea which PL it's running much less which toolchain produced or post-processed it, nor the accounting semantics of the guest code or whatever.

but there is a ton of stuff that _might_ break, so I'd aim to change as little as possible. everything is delicate and only as good as its tests.

leighmcculloch: I share the same concerns. @elizabethengelman @fnando I think we should to revert some of this change, so that it addresses just the original issue #1721, by returning to only appending a custom section, but just one instead of multiple.

leighmcculloch: If we end up needing to do this again for another feature where we absolutely need to rewrite the file, we can use this PR as a template for how to do it, because I think for how to do it this PR did a good job of that.

elizabethengelman: thanks for that synopsis @graydon - that is super helpful info. 

I agree that reverting some of this change, and just appending to the custom section sounds like the way to go. I'll plan to either clean this PR up, or just close it and open a new one shortly

PR #2097, opened 2025-07-27
Add automatically fund a key on testnet

### LINKED ISSUE #2048: Automatically fund a key on testnet when one is not set
The stellar-cli could automatically fund a key on testnet when one is not set or provided for commands that interact with testnet.

That key could be named something general like 'default'.

The flow would mean that devs getting started would have less to do upfront before a `stellar contract deploy` would just work.

Importantly, this would be a randomly generated key, not some zero value, or default value, or fixed test key.

cc @stellar/devx

hakymulla:
### What

closes #2048 

automatically creates and funds a key on testnet when none is provided for testnet.
```
// before 
stellar contract deploy --wasm ../soroban-hello-world/target/wasm32v1-none/release/hello_world.wasm
error: the following required arguments were not provided:
  --source-account <SOURCE_ACCOUNT>

Usage: stellar contract deploy --source-account <SOURCE_ACCOUNT> <--wasm <WASM>|--wasm-hash <WASM_HASH>> [-- <CONTRACT_CONSTRUCTOR_ARGS>...]

// After
stellar contract deploy --wasm ../soroban-hello-world/target/wasm32v1-none/release/hello_world.wasm
ℹ️  Skipping install because wasm already installed
ℹ️  Using wasm hash c803e2c6e0dd242b72a4fcbaa916a0541e7bdfde55740b95f64429d611fb2c23
ℹ️  Simulating deploy transaction…
ℹ️  Transaction hash is a222c5018bbf6908010f67cab97be992da54594edc1a85f8df11162cb634fbf2
🔗 https://stellar.expert/explorer/testnet/tx/a222c5018bbf6908010f67cab97be992da54594edc1a85f8df11162cb634fbf2
ℹ️  Signing transaction: a222c5018bbf6908010f67cab97be992da54594edc1a85f8df11162cb634fbf2
🔗 https://stellar.expert/explorer/testnet/contract/CA2HYU6OLNI2B5JULUWTJW53TBW55QWQU4JX2UEERFFDQHJ7BNZHH5UT
✅ Deployed!
CA2HYU6OLNI2B5JULUWTJW53TBW55QWQU4JX2UEERFFDQHJ7BNZHH5UT
```
### Why

It Randomly generates a key and funds it meaning that devs getting started would have less setup to do.

### Known limitations

Still have to supply the --wasm flag

### DIFF
--- a/cmd/soroban-cli/src/config/mod.rs
+++ b/cmd/soroban-cli/src/config/mod.rs
@@ -85,30 +85,27 @@ impl Args {
         match &self.source_account {
             Some(UnresolvedMuxedAccount::AliasOrSecret(alias)) if alias == "default" => {
                 let default_name = KeyName("default".to_string());
-                if self.locator.read_identity(&default_name).is_err() {
-                    println!("Generating new default account");
-                    let network = self.network.get(&self.locator)?;
-                    let should_fund = network.network_passphrase == network::passphrase::TESTNET;
-                    let generate_cmd = generate::Cmd {
-                        name: default_name.clone(),
-                        // #[cfg(feature = "version_lt_23")]
-                        // no_fund: !should_fund,
-                        seed: None, // Random seed for security
-                        as_secret: false,
-                        secure_store: false,
-                        config_locator: self.locator.clone(),
-                        hd_path: None,
-                        network: self.network.clone(),
-                        fund: should_fund,
-                        overwrite: false, // Prevent overwriting
-                    };
-                    let _ = generate_cmd
-                        .run(&global::Args {
-                            quiet: true,
-                            ..Default::default()
-                        })
-                        .await;
-                }
+                let network = self.network.get(&self.locator)?;
+                let should_fund = network.network_passphrase == network::passphrase::TESTNET;
+                let generate_cmd = generate::Cmd {
+                    name: default_name.clone(),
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/config/mod.rs
```diff
@@ -53,14 +57,20 @@ pub struct Args {
     #[command(flatten)]
     pub network: network::Args,
 
-    #[arg(long, short = 's', visible_alias = "source", env = "STELLAR_ACCOUNT")]
+    #[arg(
+        long,
+        short = 's',
+        visible_alias = "source",
+        env = "STELLAR_ACCOUNT",
+        default_value = "default"
```
leighmcculloch, line 65: What if instead of introducing a magic value, we leveraged the env var by setting the env var at a high level. The idea I suggested in the issue was less about having a magic 'default', but more about identifying when a default key hasn't been set via the `use` command, and there are no generated keys, and creating a key and then setting it as if the `use` command had been called. Outputting messages saying it is doing that.

hakymulla: ooh, so to be clear If no STELLAR_ACCOUNT is set and no default key exists, the source_account method should create and set a default key?

leighmcculloch: That's what I meant in the original issue I think.

PR #2100, opened 2025-07-29
Add network config settings command

leighmcculloch:
### What
  Add command to fetch network configuration settings in JSON format.

  ### Why
  A convenient way to retrieve the current network configuration for a network without needing to install, setup, and boot stellar-core.

### DIFF
--- a/cmd/soroban-cli/src/commands/network/config_settings.rs
+++ b/cmd/soroban-cli/src/commands/network/config_settings.rs
@@ -1,0 +1,86 @@
+use crate::commands::global;
+use crate::config::network;
+use crate::rpc::FullLedgerEntries;
+use crate::{config, print};
+use clap::command;
+use stellar_xdr::curr::{
+    ConfigSettingId, ConfigUpgradeSet, LedgerEntryData, LedgerKey, LedgerKeyConfigSetting,
+};
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Config(#[from] config::Error),
+    #[error(transparent)]
+    Network(#[from] network::Error),
+    #[error(transparent)]
+    Serde(#[from] serde_json::Error),
+}
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, clap::ValueEnum, Default)]
+pub enum OutputFormat {
+    /// JSON result of the RPC request
+    #[default]
+    Json,
+    /// Formatted (multiline) JSON output of the RPC request
+    JsonFormatted,
+}
+
+#[derive(Debug, clap::Parser, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    #[command(flatten)]
+    pub config: config::ArgsLocatorAndNetwork,
+    /// Format of the output
+    #[arg(long, default_value = "json")]
+    pub output: OutputFormat,
+}
+
+impl Cmd {
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let print = print::Print::new(global_args.quiet);
+        let result = self
+            .config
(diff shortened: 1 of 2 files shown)


cmd/soroban-cli/src/commands/network/config_settings.rs
```diff
@@ -0,0 +1,86 @@
...
+                let entries = entries
+                    .into_iter()
+                    .filter_map(|e| match e.val {
+                        LedgerEntryData::ConfigSetting(setting) => Some(setting),
+                        _ => None,
+                    })
+                    .collect::<Vec<_>>();
+                let set = ConfigUpgradeSet {
+                    updated_entry: entries.try_into().unwrap(),
+                };
```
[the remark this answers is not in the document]
leighmcculloch: The conversion is very unlikely to fail. The conversion is required for the VecM type that validates and unfortaantely even for unlimited size VecM's, doesn't have a .into() conversion.

cmd/soroban-cli/src/commands/network/config_settings.rs
```diff
@@ -0,0 +1,86 @@
...
+        // CLI doesn't know about. Warn, because the output of this command might provide an
+        // incomplete view of the network's config settings.
+        let network_version = rpc.get_version_info().await?.protocol_version;
+        let self_version = Version::parse(stellar_xdr::VERSION.pkg)?.major;
+        if self_version < network_version.into() {
+            print.warnln(format!("Network protocol version is {network_version} but the stellar-cli supports {self_version}. The config fetched may not represent the complete config settings for the network. Upgrade the stellar-cli."));
+        }
+
+        // Collect the ledger entries for all the config settings.
+        let keys = ConfigSettingId::variants().map(|id| {
```
Shaptic: TIL these are ledger keys (I figured they were but had never seen an example); nice!

leighmcculloch: Yeah it's pretty convenient the config is so accessible just like any other ledger entry. Glad it wasn't some unique construction stored somewhere else.

FULL_HELP_DOCS.md
```diff
@@ -1212,6 +1212,7 @@ Configure connection to networks
 * `add` — Add a new network
 * `rm` — Remove a network
 * `ls` — List networks
+* `config-settings` — Fetch the config settings for the network
```
fnando: can this be named `config` or `settings` instead?

leighmcculloch: The reason I used the term `config-settings` is because that's what each of the config items are called. Each is defined as a "ConfigSettingId", and then a ConfigSettingEntry. I can reduce it to `config`, but that might be more ambiguous and folks might think it's doing something with local network config.

Maybe I shouldn't be bundling this under the network command.

I'm thinking I should move this under the ledger subcommand, since this is pulling information from the ledger.

leighmcculloch: - https://github.com/stellar/stellar-cli/issues/1838

leighmcculloch: I guess this should live near the `network info` command.

leighmcculloch: - https://github.com/stellar/stellar-cli/issues/1917

leighmcculloch: Wdyt, rename to `config` is okay? `fetch-config`?

leighmcculloch: I've renamed it to `settings`, since the term settings seems to have the least overlap with other terms already used in the CLI for other things.

leighmcculloch:
This is ready for review again. I've changed the command to be simply 'settings'. And the command now has an `--internal` option and won't output internal configs unless that option is passed.

PR #2111, opened 2025-08-05
Avoid detecting contracts as read-only when build-only flag is set.

### LINKED ISSUE #2109: `--build-only` not respected without `--send=yes` on `stellar contract invoke`

v23.0.0

### What did you do?

```
stellar contract invoke --build-only --id CCW67TSZV3SSS2HXMBQ5JFGCKJNXKZM7UQUWUZPUTHXSTZLEO7SJMI75 -- balance --id CCW67TSZV3SSS2HXMBQ5JFGCKJNXKZM7UQUWUZPUTHXSTZLEO7SJMI75
ℹ️  Simulation identified as read-only. Send by rerunning with `--send=yes`.
"23650000"
```

The XDR of the transaction.

### What did you see instead?

The simulated result.

If I add `--send`, it works:

```
stellar contract invoke --build-only --send=yes --id CCW67TSZV3SSS2HXMBQ5JFGCKJNXKZM7UQUWUZPUTHXSTZLEO7SJMI75 -- balance --id CCW67TSZV3SSS2HXMBQ5JFGCKJNXKZM7UQUWUZPUTHXSTZLEO7SJMI75
AAAAAgAAAADW+EG7Ro4psijThkm7KsZTPwDj2i+tlfltrJ8FAE4/9wAAAGQDeXnXAAAAAQAAAAAAAAAAAAAAAQAAAAAAAAAYAAAAAAAAAAGt785ZruUpaPdgYdSUwlJbdWWfpClqZfSZ7ynlZHfklgAAAAdiYWxhbmNlAAAAAAEAAAASAAAAAa3vzlmu5Slo92Bh1JTCUlt1ZZ+kKWpl9JnvKeVkd+SWAAAAAAAAAAAAAAAA
```

fnando:
When build_only is specified, skip the should_send_tx check that could prevent transaction assembly and building.

### What

Close #2109.

```console
# before
$ stellar contract invoke --id hello --build-only -- hello --to 'fnando'
ℹ️ Simulation identified as read-only. Send by rerunning with `--send=yes`.
["Hello","fnando"]

# after
$ stellar contract invoke --id hello --build-only -- hello --to 'fnando'
AAAAAgAAAADYcrin+iEjGmimHUdHO1ayN/n4bkbl6AqSvsit9+nL8QAAAGQAC0v9AAAACAAAAAAAAAAAAAAAAQAAAAAAAAAYAAAAAAAAAAE6MO1DdzCGq9XRsczB3lcR8GiBJQcHysbuKULvAtrpAAAAAAVoZWxsbwAAAAAAAAEAAAAOAAAABmZuYW5kbwAAAAAAAAAAAAAAAAAA

$ pbpaste | stellar xdr decode --type TransactionEnvelope
{"tx":{"tx":{"source_account":"GDMHFOFH7IQSGGTIUYOUORZ3K2ZDP6PYNZDOL2AKSK7MRLPX5HF7DEA4","fee":100,"seq_num":"3179774742626312","cond":"none","memo":"none","operations":[{"source_account":null,"body":{"invoke_host_function":{"host_function":{"invoke_contract":{"contract_address":"CA5DB3KDO4YINK6V2GY4ZQO6K4I7A2EBEUDQPSWG5YUUF3YC3LUQAB2P","function_name":"hello","args":[{"string":"fnando"}]}},"auth":[]}}}],"ext":"v0"},"signatures":[]}}
```

### Why

Because build-only implies getting the xdr back, so it makes no sense detecting invocations as read-only in this case.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/invoke.rs
+++ b/cmd/soroban-cli/src/commands/contract/invoke.rs
@@ -247,7 +247,12 @@ impl NetworkRunnable for Cmd {
         let assembled = self
             .simulate(&host_function_params, &default_account_entry(), &client)
             .await?;
-        let should_send = self.should_send_tx(&assembled.sim_res)?;
+
+        let should_send = if self.fee.build_only {
+            ShouldSend::Yes
+        } else {
+            self.should_send_tx(&assembled.sim_res)?
+        };
 
         let account_details = if should_send == ShouldSend::Yes {
             client


cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -247,7 +247,12 @@ impl NetworkRunnable for Cmd {
         let assembled = self
             .simulate(&host_function_params, &default_account_entry(), &client)
             .await?;
-        let should_send = self.should_send_tx(&assembled.sim_res)?;
```
elizabethengelman, line 250: This doesn't have to be done in this PR, but I wonder if we should consider renaming `should_send_tx`. Like this issue shows, this fn doesn't necessarily determine if the tx is sent to the network or not. 🤔

elizabethengelman (approved):
LGTM! 🎉 

One comment inline, that doesn't have to be addressed in this PR, just an idea.

cmd/soroban-cli/src/commands/contract/invoke.rs
fnando: would `skip_readonly_check` be better?

edit: i'm going to dock the renaming for now, because it envolves changing the enum and all references in a bunch of places. 🫠

elizabethengelman: yeah, that sounds good. `skip_readonly_check` makes sense, but def not an urgent change.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -247,7 +247,12 @@ impl NetworkRunnable for Cmd {
...
+
+        let should_send = if self.fee.build_only {
```
leighmcculloch, line 251: If build only is specified, I would expect no simulation to be taking place. Somethiing still doesn't seem right about the flow in this command.

PR #2112, opened 2025-08-06
Only simulate invocation when absolutely necessary.

fnando:
### What

Build only doesn't require simulation, so let's do it only when required.

### Why

To avoid an unnecessary simulation.

### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -269,6 +271,7 @@ impl NetworkRunnable for Cmd {
                 );
             }
 
+            let assembled = simulate_if_needed().await?;
```
leighmcculloch, line 274: If should send is not yes, and not build only, won't it simulate twice? It looks like it calls that function twice.

PR #2113, opened 2025-08-06
Swap version checking, so it works if value is cached.

### LINKED ISSUE #2081: stellar-cli upgrade check
two issues I ran into:
- when I was on `22.8.2`, it didn't ask me to upgrade to the latest `23`
- after upgrading it to the latest `23`, I get the following error: `• A new release of stellar-cli is available: 22.1.0 -> 22.8.2` which is incorrect

Image

fnando:
### What

Swaps the version checking. This fixes the case where latest version is cached and the current version is greater than the cached value, showing the upgrade message to users that don't need it.

### Why

Close #2081.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/upgrade_check.rs
+++ b/cmd/soroban-cli/src/upgrade_check.rs
@@ -87,7 +87,7 @@ pub async fn upgrade_check(quiet: bool) {
     let current_version = Version::parse(current_version).unwrap();
     let latest_version = get_latest_version(&current_version, &stats);
 
-    if *latest_version > current_version {
+    if current_version < *latest_version {
         let printer = Print::new(quiet);
         printer.warnln(format!(
             "A new release of stellar-cli is available: {current_version} -> {latest_version}"


elizabethengelman (approved):
Approving this because it looks good to me as a fix for the first issue!
 >when I was on 22.8.2, it didn't ask me to upgrade to the latest 23

I am not sure I fully understand how this will address the second issue, because this makes it seem like the executable still thought it was at version 22.1.0. 🤔 Which would still be an issue, even not considering the potential crates.io caching, right?

> after upgrading it to the latest 23, I get the following error: • A new release of stellar-cli is available: 22.1.0 -> 22.8.2 which is incorrect

fnando:
The asking to upgrade part is tricky, because it depends on execution time. Even if cache was disabled completely, commands could still not show it, so I'm not going to bother with it.

This fix only handles the case where the latest installed cli wouldn't ever fail if cache was stale (e.g. v23 wouldn't fail if latest cached was v22). I think there's something fishy on the reported CLI version, because it returned 22.1.0, whether user error or something like brew not having finished installed the latest version.

PR #2116, opened 2025-08-06
Ensure simulation only runs once.

fnando:


cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -244,16 +244,14 @@ impl NetworkRunnable for Cmd {
...
         let (function, spec, host_function_params, signers) = params;
 
-        let simulate_if_needed = || async {
-            self.simulate(&host_function_params, &default_account_entry(), &client)
-                .await
-        };
-
-        let should_send = if self.fee.build_only {
-            ShouldSend::Yes
+        let (should_send, cached_simulation) = if self.fee.build_only {
```
elizabethengelman, line 249: I wonder if we should add a comment explaining that build only txns won't actually be sent, and will be checked again later on. Whenever I look at this file, it always takes me a bit of time to realize that though we're using `ShouldSend::Yes` here, we won't actually be sending the tx. 

Maybe something like this:

```suggestion
        // `self.fee.build_only` will be checked again below and the fn will return a TxnResult::Txn if the user passed the --build-only flag
        let (should_send, cached_simulation) = if self.fee.build_only {
```

elizabethengelman (approved):
This looks good to me - one comment inline.

PR #2117, opened 2025-08-08
Fix clippy warnings on rust 1.89.

fnando:
### What

Fix all clippy errors on rust 1.89.

### Why

Good hygiene is good for you!

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/assembled.rs
+++ b/cmd/soroban-cli/src/assembled.rs
@@ -14,7 +14,7 @@ pub(crate) const DEFAULT_TRANSACTION_FEES: u32 = 100;
 pub async fn simulate_and_assemble_transaction(
     client: &soroban_rpc::Client,
     tx: &Transaction,
-) -> Result<Assembled, Box<Error>> {
+) -> Result<Assembled, Error> {
     let sim_res = client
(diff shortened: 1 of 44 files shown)


cmd/soroban-cli/src/assembled.rs
```diff
@@ -487,8 +496,10 @@ mod tests {
         );
 
         match result {
-            Err(Error::UnexpectedSimulateTransactionResultSize { length }) => {
-                assert_eq!(0, length);
+            Err(e) if matches!(*e, Error::UnexpectedSimulateTransactionResultSize { .. }) => {
+                if let Error::UnexpectedSimulateTransactionResultSize { length } = *e {
```
leighmcculloch: I've found boxing errors to be very unfun experience when pattern matching them like this. I think it would also be fine to change the clippy run to ignore this lint so that boxing isn't required. Unless it is causing a OOM, I don't think it's a big deal.

leighmcculloch (approved):
Looks good, but I think we should disable the lint forcing us to box the errors. Huge pain, and in reality afaik we are not seeing any material problems with the unboxed errors today.

Issue #2088, opened 2025-07-22
Support DEX operations


janewang:
Add support to the following operations

- [x] manage_buy_offer
- [x] manage_sell_offer
- [x] create_passive_sell_offer 
- [x] path_payment_strict_send
- [x] path_payment_strict_receive

Relevant thread: https://stellarfoundation.slack.com/archives/C06KTGUULUF/p1753190736370519

PR #2124, opened 2025-08-11
Add `stellar tx new create-passive-sell-offer`.

fnando:
### What

Add CREATE_PASSIVE_SELL_OFFER operation.

```console
$ stellar tx new create-passive-sell-offer --selling native --buying USDC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5 --amount 100000000 --price 1:1 --build-only | stellar xdr decode --type Transac
tionEnvelope --output json-formatted
{
  "tx": {
    "tx": {
      "source_account": "GDMHFOFH7IQSGGTIUYOUORZ3K2ZDP6PYNZDOL2AKSK7MRLPX5HF7DEA4",
      "fee": 100,
      "seq_num": "3179774742626316",
      "cond": "none",
      "memo": "none",
      "operations": [
        {
          "source_account": null,
          "body": {
            "create_passive_sell_offer": {
              "selling": "native",
              "buying": {
                "credit_alphanum4": {
                  "asset_code": "USDC",
                  "issuer": "GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5"
                }
              },
              "amount": "100000000",
              "price": {
                "n": 1,
                "d": 1
              }
            }
          }
        }
      ],
      "ext": "v0"
    },
    "signatures": []
  }
}
```

### Why

#2088 

### Known limitations

N/A

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -2952,6 +2952,7 @@ Fetch ledger information
 ###### **Subcommands:**
 
 * `latest` — Get the latest ledger sequence and information from the network
+* `fetch` — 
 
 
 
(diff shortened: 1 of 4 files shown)


cmd/soroban-cli/src/commands/tx/new/create_passive_sell_offer.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    #[clap(flatten)]
+    pub op: Args,
+}
+
+#[derive(Debug, clap::Args, Clone)]
+pub struct Args {
+    /// Asset to sell
+    #[arg(long)]
+    pub selling: builder::Asset,
```
elizabethengelman, line 19: Nit pick: not sure how useful it would be, but I wonder if we could add a bit more description about how to input the assets, like perhaps giving possible values: "i.e. `native`, `BTC:GAAA...COINBASE`, etc".

Definitely doesn't have to be part of this PR because I think that we'd need to make this change to several commands, but just a thought.

elizabethengelman (approved):
LGTM! 🎉 

One inline comment that is an idea for future updates.

cmd/soroban-cli/src/commands/tx/new/create_passive_sell_offer.rs
fnando: good idea! I'll update asset descriptions everywhere once all these commands are merged.

leighmcculloch: Also aliases can be used for hte issuer part of the name, so good to demo that too, it's easy to forget that's possible.

PR #2132, opened 2025-08-12
Conditionally set rustflags based on the target that should be used.

fnando:
### What

Use the correct env var (`TARGET_wasm32-unknown-unknown_RUSTFLAGS` or `TARGET_wasm32v1-none_RUSTFLAGS`) based on the target that will be used.

### Why

So the behavior matches the actual target.

### Known limitations

N/A


elizabethengelman (approved):

leighmcculloch (commented):
This PR, because it touched the `build` command was supposed to be blocked on a review by someone from the contract-committers group, we must have had a change to what blocks PRs, because I see the group on the review list but the PR still merged.

In any case the change looks good 👍🏻 .

Good catch @elizabethengelman @fnando.

PR #2146, opened 2025-08-22
Run tx simulation in restore command

ElliotFriend:
### What

Adding transaction simulation to the `stellar contract restore ...` command. I basically copied the way it's done in the `extend` command.

### Why

Currently, contract restoration can't (easily) be done in the CLI. They fail with `TxSorobanInvalid`.

### Known limitations

N/A

### DIFF
--- a/cookbook/restore-contract-instance.mdx
+++ b/cookbook/restore-contract-instance.mdx
@@ -7,7 +7,7 @@ custom_edit_url: https://github.com/stellar/stellar-cli/edit/main/cookbook/resto
 
 If your contract instance has been archived, it can easily be restored using the Stellar CLI.
 
-```bash cookbooktest.fail cookbooktest.stderr=TxSorobanInvalid
+```bash
 stellar contract restore \
     --source S... \
     --network testnet \
--- a/cookbook/restore-contract-instance.mdx
+++ b/cookbook/restore-contract-instance.mdx
@@ -7,7 +7,7 @@ custom_edit_url: https://github.com/stellar/stellar-cli/edit/main/cookbook/resto
 
 If your contract instance has been archived, it can easily be restored using the Stellar CLI.
 
-```bash
+```bash cookbooktest.fail cookbooktest.stderr=TxSorobanInvalid
 stellar contract restore \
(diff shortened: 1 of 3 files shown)


ElliotFriend:
Maybe I got a little eager and fired this off too early hehe

Testing it to try and restore a contract executable gives me a new error:

```sh
❯ stellar contract restore --network testnet --source guestbookhost --wasm-hash b02bf4e7e3e14022d08ba6ef3d94c5f6e7755434a668951a3edd1ec4eb49bbe7
ℹ️  Signing transaction: 5e188cac1d770da8627fd5318d7c9d5f0476fdac2011ec6bcfbd46fd2b42ea85
❌ error: Ledger entry not found
```

At least, we're signing the right transaction hash, though. So, progress? lol

dmkozh:
> Testing it to try and restore a contract executable gives me a new error:

FWIW this error happens due to incorrect parsing of meta. If we're able to fix CI, then we could merge this as is and fix meta as a followup. Or maybe fix meta now, but we'll need to also add handling for the restored entry state. 
Also, `Ledger entry not found` error is misleading, I think we should open an issue for this to be cleaned up everywhere.

cookbook/restore-contract-instance.mdx
```diff
@@ -7,7 +7,7 @@ custom_edit_url: https://github.com/stellar/stellar-cli/edit/main/cookbook/resto
 
 If your contract instance has been archived, it can easily be restored using the Stellar CLI.
 
-```bash
+```bash cookbooktest.fail cookbooktest.stderr=TxSorobanInvalid
```
dmkozh, line 10: Why did you roll this back, is it still not working?

ElliotFriend: Correct. @sagpatil suggested if the error on the cookbooks test remains, we can revert it so the test failures can be addressed in another PR. (if i'm understanding correctly)

dmkozh: Ok. It's really concerning to have a test failure documented in the cookbook, I hope we can fix this soon.

PR #2149, opened 2025-08-25
Fix cookbook integration tests

elizabethengelman:
### What

Fixes to the deploy-contract.mdx to allow for an alias flag.

### Why

This was previously failing in integration tests.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/restore.rs
+++ b/cmd/soroban-cli/src/commands/contract/restore.rs
@@ -225,12 +225,11 @@ impl NetworkRunnable for Cmd {
         tracing::debug!("Changes:\nlen:{}\n{changes:#?}", changes.len());
 
         if changes.is_empty() {
+            print.infoln("No changes detected, transaction was a no-op.");
             let entry = client.get_full_ledger_entries(&entry_keys).await?;
             let extension = entry.entries[0].live_until_ledger_seq;
 
-            if entry.latest_ledger < i64::from(extension) {
-                return Ok(TxnResult::Res(extension));
-            }
+            return Ok(TxnResult::Res(extension));
         }
 
         Ok(TxnResult::Res(


cmd/soroban-cli/src/commands/contract/extend.rs
```diff
@@ -222,9 +222,12 @@ impl NetworkRunnable for Cmd {
             let entry = client.get_full_ledger_entries(&keys).await?;
             let extension = entry.entries[0].live_until_ledger_seq;
 
-            if entry.latest_ledger + i64::from(extend_to) < i64::from(extension) {
+            let new_ext = entry.latest_ledger + i64::from(extend_to);
+            if new_ext < i64::from(extension) {
                 return Ok(TxnResult::Res(extension));
             }
+
```
elizabethengelman: In the cookbook integration tests, I was seeing situation where we getting here - where `changes` is empty, but the fn does not return. This was then panicking below when trying to get the 0 or 1 index of an empty collection: `match (&changes[0], &changes[1])`.

I'm still not entirely sure why we weren't seeing this issue before - I wonder if it has something to do with the ledger entries being extended in other test scenarios, and therefore we don't end up in the ` if new_ext < i64::from(extension) {` branch? 🤔

elizabethengelman: A similar change will presumably need to be made in `restore.rs` as well

cmd/soroban-cli/src/commands/contract/extend.rs
```diff
@@ -222,9 +222,12 @@ impl NetworkRunnable for Cmd {
             let entry = client.get_full_ledger_entries(&keys).await?;
             let extension = entry.entries[0].live_until_ledger_seq;
 
-            if entry.latest_ledger + i64::from(extend_to) < i64::from(extension) {
+            let new_ext = entry.latest_ledger + i64::from(extend_to);
```
dmkozh: I don't understand what this block tries to do. If changes are empty, then extension didn't happen. If extension didn't happen, then chances are we shouldn't have sent this transaction at all. There is only an edge case where someone extends the entry just before you, so that the transaction becomes no-op only at apply time. Even then, I think it's better to just say that transaction was a no-op and entry had sufficient TTL. But I think we shouldn't send the transaction when it's going to be no-op, similarly to what we need to do for the restore command.

elizabethengelman: Thanks for taking a look @dmkozh 

There is a comment a few lines above:
>         // The transaction from core will succeed regardless of whether it actually found & extended
>        // the entry, so we have to inspect the result meta to tell if it worked or not.

That makes me think that the original intent of this block is to allow for a return value from the tx, even if the tx was a no-op. But this is confusing, and I'm not sure why the `if entry.latest_ledger + i64::from(extend_to) < i64::from(extension)` was there to begin with.

Instead, does the following make sense? If there are no changes, we print something to the command line, saying as much, and then return the existing TTL ledger from the fn?

```suggestion
          
            if changes.is_empty() {
                println!("No changes, this transaction was a noop")
                let entry = client.get_full_ledger_entries(&keys).await?;
                let extension = entry.entries[0].live_until_ledger_seq;
                
                return Ok(TxnResult::Res(extension)); 

           
```

dmkozh: Yeah, this makes much more sense to me. As I've mentioned in the comment, we still need to look at the simulation results to avoid sending no-ops in the first place, but this block still makes sense for the edge cases. We can update the simulation parsing logic as a followup.

elizabethengelman: > we still need to look at the simulation results to avoid sending no-ops in the first place

Makes sense 👍

PR #2161, opened 2025-08-28
Fix CI failing job

sagpatil:
Adding a no op PR to check CI system
----

[fix: ignore RUSTSEC-2025-0052 advisory for async-std in test dependencies](https://github.com/stellar/stellar-cli/pull/2161/commits/98a4c4c17a3d54c693b7639da2b9adccde19bd6c)

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -34,6 +34,7 @@ yanked = "warn"
 ignore = [
     "RUSTSEC-2024-0388", # derivative is no longer maintained, but that has no known material impact on the this repo
     "RUSTSEC-2024-0436", # paste is no longer maintained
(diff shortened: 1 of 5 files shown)


deny.toml
```diff
@@ -34,6 +34,7 @@ yanked = "warn"
 ignore = [
     "RUSTSEC-2024-0388", # derivative is no longer maintained, but that has no known material impact on the this repo
     "RUSTSEC-2024-0436", # paste is no longer maintained
+    "RUSTSEC-2025-0052", # async-std has been discontinued - used only in test dependencies
```
leighmcculloch, line 37: +1 I think we can ignore since the use is only in test dependencies, good call.

leighmcculloch (approved):

PR #2173, opened 2025-08-29
Uptake stellar-asset-spec crate from refactoring in rs-soroban-sdk PR#1536

anupsdf:
### What
StellarAssetSpec got refactored in https://github.com/stellar/rs-soroban-sdk/pull/1536
This PR is to adapt to the new crate stellar-asset-spec v23.0.0-rc.3

Plus pick up the dependencies from other crates released within rs-soroban-sdk v23.0.0-rc.3.

### DIFF
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -1090,22 +1090,18 @@ fn sc_address_from_json(s: &str) -> Result<ScVal, Error> {
 fn sc_muxed_address_from_json(s: &str) -> Result<ScVal, Error> {
     stellar_strkey::Strkey::from_string(s)
         .map_err(|_| Error::InvalidValue(Some(ScType::MuxedAddress)))
-        .map(|parsed| match parsed {
-            stellar_strkey::Strkey::PublicKeyEd25519(p) => Some(ScVal::Address(
-                ScAddress::Account(AccountId(PublicKey::PublicKeyTypeEd25519(Uint256(p.0)))),
-            )),
-            stellar_strkey::Strkey::Contract(c) => {
-                Some(ScVal::Address(ScAddress::Contract(ContractId(Hash(c.0)))))
+        .and_then(|parsed| match parsed {
+            stellar_strkey::Strkey::PublicKeyEd25519(_) | stellar_strkey::Strkey::Contract(_) => {
+                sc_address_from_json(s)
             }
-            stellar_strkey::Strkey::MuxedAccountEd25519(m) => Some(ScVal::Address(
+            stellar_strkey::Strkey::MuxedAccountEd25519(m) => Ok(ScVal::Address(
                 ScAddress::MuxedAccount(MuxedEd25519Account {
                     ed25519: Uint256(m.ed25519),
                     id: m.id,
                 }),
             )),
-            _ => None,
-        })?
-        .ok_or(Error::InvalidValue(Some(ScType::MuxedAddress)))
+            _ => Err(Error::InvalidValue(Some(ScType::MuxedAddress))),
+        })
 }
 
 fn to_lower_hex(bytes: &[u8]) -> String {
@@ -1552,6 +1548,24 @@ mod tests {
                 assert!(matches!(addr, ScVal::Address(ScAddress::MuxedAccount(_))));
             }
             Err(e) => panic!("Unexpected error parsing MuxedAddress: {e}"),
+        }
+
+        // Test that regular Ed25519 addresses also work with MuxedAddress parser
+        match sc_muxed_address_from_json("GA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQHES5")
+        {
+            Ok(addr) => {
+                assert!(matches!(addr, ScVal::Address(ScAddress::Account(_))));
+            }
+            Err(e) => panic!("Unexpected error parsing Ed25519 address as MuxedAddress: {e}"),
+        }
+
+        // Test that contract addresses also work with MuxedAddress parser
+        match sc_muxed_address_from_json("CA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQGAXE")
+        {
+            Ok(addr) => {
+                assert!(matches!(addr, ScVal::Address(ScAddress::Contract(_))));
+            }
+            Err(e) => panic!("Unexpected error parsing Contract address as MuxedAddress: {e}"),
         }
     }
 
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -1552,7 +1552,7 @@ mod tests {
                     0x31, 0x64, 0xe3, 0x91, 0xba, 0x47, 0x8f, 0x88, 0x54, 0x05, 0xa3, 0x84, 0xc5,
                     0x03, 0x9b, 0x5e, 0xd3, 0x78, 0x63,
                 ]);
-                let expected_id = 1754924385537090737u64;
+                let expected_id = 1_754_924_385_537_090_737_u64;
 
                 assert_eq!(ed25519, expected_ed25519);
                 assert_eq!(id, expected_id);
@@ -1575,7 +1575,7 @@ mod tests {
                     0x31, 0x64, 0xe3, 0x91, 0xba, 0x47, 0x8f, 0x88, 0x54, 0x05, 0xa3, 0x84, 0xc5,
                     0x03, 0x9b, 0x5e, 0xd3, 0x78, 0x63,
                 ]);
-                let expected_id = 1754924385537090737u64;
+                let expected_id = 1_754_924_385_537_090_737_u64;
 
                 assert_eq!(ed25519, expected_ed25519);
                 assert_eq!(id, expected_id);
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -1553,24 +1553,6 @@ mod tests {
             }
             Err(e) => panic!("Unexpected error parsing MuxedAddress: {e}"),
         }
-
-        // Test that regular Ed25519 addresses also work with MuxedAddress parser
-        match sc_muxed_address_from_json("GA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQHES5")
-        {
-            Ok(addr) => {
-                assert!(matches!(addr, ScVal::Address(ScAddress::Account(_))));
-            }
-            Err(e) => panic!("Unexpected error parsing Ed25519 address as MuxedAddress: {e}"),
-        }
-
-        // Test that contract addresses also work with MuxedAddress parser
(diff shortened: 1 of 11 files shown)


cookbook/payments-and-assets.mdx
```diff
@@ -50,13 +50,13 @@ stellar keys use alice
 ```
 
 ```bash
-stellar contract invoke --id <ASSET_CONTRACT_ID> -- transfer --to bob --from alice --amount 100
+stellar contract invoke --id <ASSET_CONTRACT_ID> -- transfer --to $(stellar keys address bob) --from alice --amount 100
```
leighmcculloch: Why didn't the alias work? The alias bob should work and it shouldn't need the additional process?

cmd/crates/soroban-test/tests/fixtures/test-wasms/hello_world/src/lib.rs
```diff
@@ -61,10 +72,12 @@ impl Contract {
...
     /// Logs a string with `hello ` in front.
     pub fn log(env: Env, str: Symbol) {
-        env.events().publish(
-            (Symbol::new(&env, "hello"), Symbol::new(&env, "")),
-            str.clone(),
-        );
+        // For the event, we'll create a simple hello message
+        // Since string concatenation is complex in no_std, we'll use a fixed message format
+        HelloEvent {
+            message: String::from_str(&env, "hello message logged"),
```
anupsdf: TODO: Need to fix this.

cookbook/payments-and-assets.mdx
anupsdf: Something with the MuxedAddress change tripped this test and couldn't parse the "to" field. We need to look into this.

anupsdf: Let me revert so we can see the error in GH actions.

sagpatil: > sagar@sagar-MGXX0Y41VY-mbp CLI-2 % stellar contract invoke --id CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC -- transfer --from alice --to bob --amount 100
> :information_source:  Signing transaction: 10ca4d5db81579c6e16372323a063a7eca6c6ce77566ea652050582e5ed00210
> :date: CDLZFC3SYJYDZT7K67VZ75HPJVIEUVNIXF47ZG2FB2RMQQVU2HHGCYSC - Success - Event: [{"symbol":"transfer"},{"address":"GBDUM7GOCQPOPSMLWNB5AXDXCZDKOWRJZTY33TLXBTNRH3GPN2ZSARVG"},{"address":"GD5UA45RYP7YWNT3JCA47P6N5OWX2WVRH4GUGTKMVKJYJ62KHODIJJBE"},{"string":"native"}] = {"i128":"100"}


works locally fails in ci



Runn

> ing command 9: stellar contract invoke --id <ASSET_CONTRACT_ID> -- transfer --from alice --to bob --amount 100 ({})
> Executing command: contract invoke --id CDMLFMKMMD7MWZP3FKUBZPVHTUEDLSX4BYGYKH4GCESXYHS3IHQ4EIG4 -- transfer --from alice --to bob --amount 100
> :x: error: parsing argument to: expected value at line 1 column 1
> thread 'integration::cookbook::tests::test_all_mdx_files' panicked at /home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5:
> Unexpected failure.
> code=1

leighmcculloch: That's not the only failure. There's a whole heap of very strange failures.

leighmcculloch: Oh the field is a muxed address, we might have a bug in the cli's muxed address arg parsing logic as part of invoke. We'll need to fix that, not replace it with the address string.

anupsdf: Yeah, let's see if this commit fixes parsing of MuxedAddress, https://github.com/stellar/stellar-cli/pull/2173/commits/2e080fcd794bbc73a89f6885faa6981d11a6e6aa

anupsdf: Plus this to see if it does fix parsing MuxedAddress, https://github.com/stellar/stellar-cli/pull/2173/commits/1bd67f6b4517a6195870e689810d18c189a46912

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1517,6 +1542,62 @@ mod tests {
...
+            "MA5XIGA5C7QTPTWXQHY6MCJRMTRZDOSHR6EFIBNDQTCQHG262N4GGGC2XZXD7G7EWH7U6",
+        ) {
+            Ok(addr) => {
+                assert!(matches!(addr, ScVal::Address(ScAddress::MuxedAccount(_))));
+            }
+            Err(e) => panic!("Unexpected error parsing MuxedAddress: {e}"),
+        }
+
+        // Test that regular Ed25519 addresses also work with MuxedAddress parser
+        match sc_muxed_address_from_json("GA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQHES5")
```
anupsdf: We don't need this G or the C address check below for this MuxedAddress test.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1517,6 +1542,62 @@ mod tests {
...
+            Ok(addr) => {
+                assert!(matches!(addr, ScVal::Address(ScAddress::MuxedAccount(_))));
+            }
+            Err(e) => {
+                panic!("Unexpected error parsing MuxedAddress with from_string_primitive: {e}")
+            }
+        }
+
+        // Test that regular addresses work with MuxedAddress type too
+        let ed25519_addr = "GA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQHES5";
```
anupsdf: Not needed. Will clean this up.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1083,6 +1087,27 @@ fn sc_address_from_json(s: &str) -> Result<ScVal, Error> {
         .ok_or(Error::InvalidValue(Some(ScType::Address)))
 }
 
+fn sc_muxed_address_from_json(s: &str) -> Result<ScVal, Error> {
+    stellar_strkey::Strkey::from_string(s)
+        .map_err(|_| Error::InvalidValue(Some(ScType::MuxedAddress)))
+        .map(|parsed| match parsed {
+            stellar_strkey::Strkey::PublicKeyEd25519(p) => Some(ScVal::Address(
```
dmkozh: Maybe we should defer to sc_address_from_json for non-muxed-account keys? This will reduce duplication a bit.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1517,6 +1542,33 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_sc_muxed_address_from_json() {
+        // Test muxed address parsing
+        match sc_muxed_address_from_json(
+            "MA5XIGA5C7QTPTWXQHY6MCJRMTRZDOSHR6EFIBNDQTCQHG262N4GGGC2XZXD7G7EWH7U6",
+        ) {
+            Ok(addr) => {
+                assert!(matches!(addr, ScVal::Address(ScAddress::MuxedAccount(_))));
```
dmkozh: We should assert on the full expected XDR, I'm pretty sure strkey tests should have samples

dmkozh: Wouldn't assert_eq work?

cmd/crates/soroban-spec-tools/src/lib.rs
anupsdf: Good point. Fixed to remove duplication.

cmd/crates/soroban-spec-tools/src/lib.rs
anupsdf: Done for G and C addresses.

anupsdf: As per Claude,

```
assert_eq! wouldn't be appropriate here because:

assert_eq! compares for equality between two values
assert!(matches!(...)) checks if a value matches a specific pattern
The current usage is correct because we're checking if the returned ScVal::Address contains the right variant (MuxedAccount, Account, or Contract) without needing to compare the exact inner values.

If you wanted to use assert_eq!, you would need to construct the exact expected values, but that would make the tests more brittle and harder to maintain. The current assert!(matches!(...)) pattern is the idiomatic way to test enum variants in Rust when you only care about the variant type, not the exact inner values.
```

dmkozh: What I mean is that we should use `assert_eq` with the precise payload, i.e. `MuxedEd25519Account { ed25519: <public key>, id: <id> },`. Otherwise the test would still pass even if we remove the strkey call and replace it by arbitrary value.

dmkozh: Don't trust Claude here (or I guess in general) :) The test is brittle as it is now. We should make sure that the address parsing actually happens and that the fields are routed appropriately.

anupsdf: Great point! Now the tests have the full payload MuxedEd25519Account { ed25519: <public key>, id: <id> } similar to how we have in https://github.com/stellar/rs-stellar-strkey/blob/main/tests/tests.rs repo. Does it look okay?

anupsdf: Yeah, it led me the wrong path. Now the test are using assert_eq! for the ed25519 and id parts of the M address.

anupsdf: Oh wait. I should derive the M address from the edd25519 and id to make the test better.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -1083,6 +1087,23 @@ fn sc_address_from_json(s: &str) -> Result<ScVal, Error> {
         .ok_or(Error::InvalidValue(Some(ScType::Address)))
 }
 
+fn sc_muxed_address_from_json(s: &str) -> Result<ScVal, Error> {
```
leighmcculloch: `sc_address_from_json` already handles muxed accounts, so I think that function can just be used as is without adding this new function. I'm not sure why we allow passing an M address to all addresses today. There might be a reason for that, so I wouldn't try to address that anomaly in this PR, but we open an issue to follow up with that.

leighmcculloch (commented):
One change requested inline, otherwise this lgtm.

cmd/crates/soroban-spec-tools/src/lib.rs
anupsdf: ah that's right. Ok, will remove this new fn sc_muxed_address_from_json

anupsdf: Removed the fn.

cmd/crates/soroban-spec-tools/src/lib.rs
anupsdf: Fixed the test. However, removed the new fn as per Leigh's suggestion. I didn't realize that the existing sc_address_from_json fn handles muxed accounts. I left this test though since it contructs a M addr from its components while there is an existing test which starts directly from M addr.

sagpatil:
Thanks a lot @anupsdf

PR #2185, opened 2025-09-03
Only run macos/windows tests in certain cases.

fnando:
### What

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -1608,6 +1608,8 @@ Sign, Simulate, and Send transactions
 * `sign` — Sign a transaction envelope appending the signature to the envelope
 * `simulate` — Simulate a transaction envelope from stdin
 * `fetch` — Fetch a transaction from the network by hash If no subcommand is passed in, the transaction envelope will be returned
+* `decode` — Decode a transaction envelope from XDR to JSON
+* `encode` — Encode a transaction envelope from JSON to XDR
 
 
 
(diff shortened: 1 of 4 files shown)


.github/workflows/binaries.yml
```diff
@@ -32,49 +32,61 @@ jobs:
...
+    if: github.event_name == 'release' || (github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')))
+    strategy:
+      fail-fast: false
+      matrix:
+        crate:
+          - name: stellar-cli
+            binary: stellar
+        sys:
           - os: macos-14
             target: aarch64-apple-darwin
```
leighmcculloch, line 54: Heads up if you wanted to keep the build binary workflow inside the binaries.yml, another way would be to include a new parameter here in the matrix for the macos and windos builds, something like "build-on-prs: false" and then on the linux ones have "build-on-prs: true", and then filter within the existing workflow with that.

leighmcculloch (approved):
I hit the wrong button, meant to approve on the last review message.

.github/workflows/binaries.yml
fnando: I thought about that, but I think breaking into separate actions makes it more readable in this case (the file was too complex).

If you don't have an issue with that, I'll keep it as it is.

leighmcculloch: No issue with going either route. 🚀

PR #2186, opened 2025-09-04
Filter out the constructor from invoke command function list

### LINKED ISSUE #2179: Stellar contract invoke command incorrectly displays __constructor as invokable function
The stellar contract invoke command currently displays `__constructor` as an available function to invoke, but this is misleading because:

- The `__constructor` function is automatically called during contract deployment
- It should not be listed as an available function for the invoke command
- Users may attempt to invoke it manually, which would fail or be confusing

This creates confusion for users who might think they can manually invoke the constructor after deployment, when in fact it's only called once during the deploy process. The invoke command should filter out constructor functions to avoid this confusion.

### Steps to reproduce
1. Deploy a contract with a constructor
2. Run `stellar contract invoke --id <contract_id> --list` or similar
3. Observe that `__constructor` appears in the list of invokable functions

### Expected behavior
The `__constructor` function should not appear in the list of functions available for invocation.

### Example

```console
$ stellar contract invoke --id testc
Usage: stellar contract invoke --id testc -- [COMMAND]

Commands:
  __constructor
  name
  symbol
  decimals
  total_supply
  balance_of
  transfer
  transfer_from
  approve
  allowance
  mint
  burn
  help           Print this message or the help of the given subcommand(s)

Options:
  -h, --help
          Print help
```

sagpatil:
### What

Fixes #2179

### Why

Filter out the constructor from invoke command function list so __constructor is not seen as invokable function


janewang:
@sagpatil Could you convert this PR from a draft to a regular PR?

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -261,6 +261,25 @@ impl NetworkRunnable for Cmd {
...
+            let txn = Box::new(build_create_contract_tx(
+                wasm_hash,
+                1, // dummy sequence number for build-only
+                self.fee.fee,
+                source_account,
+                contract_id_preimage,
+                constructor_params.as_ref(),
+            )?);
+            print.checkln("Transaction built!");
+            return Ok(TxnResult::Txn(txn));
```
elizabethengelman: 🤔 I'm not sure if we'd necessarily want a dummy seq number in here. I'm not sure frequently users do this, but I think that sometimes people use `--build-only` to get the tx's xdr, and then pipe that into a `stellar tx simulate/sign/send/etc`. If we remove the ability to get a real seq number here, then users would need to do an additional pipe to update the seq number before signing and sending. i.e. `

elizabethengelman (commented):
I left a comment inline in `deploy/wasm.rs` - all looks good, though I'm not sure about using a dummy seq number here

elizabethengelman:
I took a quick look at the failing specs, and it kind of looks like a docker issue - I wonder if kicking the tests off again would help

PR #2187, opened 2025-09-04
 Optimize test build performance for soroban-test

sagpatil:
This commit introduces a development profile override in `cmd/crates/soroban-test/Cargo.toml` to significantly speed up local test execution.

By setting `debug = 0` for all dependency packages (`*`) in the `dev` profile, we prevent the generation of debug information for third-party crates during `cargo test` runs. This reduces the amount of data the linker has to process, leading to much faster link times, especially on macOS where linking can be a bottleneck.

This change only affects dependencies, preserving full debug capabilities for the `soroban-test` crate itself, while improving the developer feedback loop.

### DIFF
--- a/cmd/crates/soroban-test/Cargo.toml
+++ b/cmd/crates/soroban-test/Cargo.toml
@@ -60,5 +60,11 @@ emulator-tests = ["stellar-ledger/emulator-tests"]
 version_lt_23 = []
 version_gte_23 = []
 
+[profile.dev]
+# Speeds up compilation in development.
+opt-level = 0
+
 [profile.dev.package."*"]
+# Speeds up compilation of dependencies in development.
+opt-level = 2
 debug = 0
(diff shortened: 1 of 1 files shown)


leighmcculloch:
What is the time saving with this change? It's surprising to me that it would be a significant win. The builds and test are already cached using sccache, and so even if the first time a crate is built is slow, it'll typically be fast for future builds.

sagpatil:
> What is the time saving with this change? It's surprising to me that it would be a significant win. The builds and test are already cached using sccache, and so even if the first time a crate is built is slow, it'll typically be fast for future builds.


I am doing this as an experiment, will run a test with and without this chnage and timing the build

sagpatil:
Ran a local test and compared it to main branch and found that the test build performance is no different than main branch. Closing this PR.

PR #2197, opened 2025-09-08
Add `stellar tx clawback-claimable-balance` and `stellar tx op add clawback-claimable-balance`.

fnando:
### What

Add clawback-claimable-balance commands.

### Why

So we can clawback claimable balances using the CLI.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/tx/new/clawback_claimable_balance.rs
+++ b/cmd/soroban-cli/src/commands/tx/new/clawback_claimable_balance.rs
@@ -53,34 +53,12 @@ fn parse_balance_id(balance_id: &str) -> Result<Vec<u8>, tx::args::Error> {
     // 3. Direct hash format (64 hex chars): 6f2179b3...
 
     if balance_id.starts_with('B') && balance_id.len() > 50 {
-        // StrKey format - decode manually since stellar-strkey doesn't have claimable balance type yet
-        match decode_strkey_to_hex(balance_id) {
-            Ok(hex_string) => {
-                // The StrKey decodes to the full API format (with 00000000 prefix)
-                // So we need to strip the prefix like we do for API format
-                let cleaned_balance_id =
-                    if hex_string.len() == 72 && hex_string.starts_with("00000000") {
-                        &hex_string[8..]
-                    } else {
-                        &hex_string
-                    };
-
-                let balance_id_bytes =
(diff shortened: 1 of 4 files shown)


cmd/soroban-cli/src/commands/tx/new/clawback_claimable_balance.rs
```diff
@@ -0,0 +1,263 @@
...
+    #[clap(flatten)]
+    pub op: Args,
+}
+
+#[derive(Debug, clap::Args, Clone)]
+pub struct Args {
+    /// Balance ID of the claimable balance to clawback. Accepts multiple formats:
+    /// - API format with type prefix (72 chars): 000000006f2179b31311fa8064760b48942c8e166702ba0b8fbe7358c4fd570421840461
+    /// - Direct hash format (64 chars): 6f2179b31311fa8064760b48942c8e166702ba0b8fbe7358c4fd570421840461
+    /// - StrKey format (base32): BAAMLBZI42AD52HKGIZOU7WFVZM6BPEJCLPL44QU2AT6TY3P57I5QDNYIA
```
leighmcculloch, line 20: Love that this supports multiple formats.

cmd/soroban-cli/src/commands/tx/new/clawback_claimable_balance.rs
```diff
@@ -0,0 +1,263 @@
...
+    let payload_end = bytes.len() - 2; // Exclude 2-byte CRC
+    let mut payload = &bytes[1..payload_end];
+
+    // If payload starts with 00 (claimable balance type indicator), skip it to get the 32-byte hash
+    if payload.len() == 33 && payload[0] == 0x00 {
+        payload = &payload[1..]; // Skip the leading 00 to get the actual 32-byte hash
+    }
+
+    // Return with claimable balance type prefix (00000000)
+    Ok(format!("00000000{}", hex::encode(payload)))
```
leighmcculloch: The stellar_strkey package supports balance address, so you should be able to use it instead of decoding manually here. Why does the CLI decode it manually?

fnando: Ah, the version we were using didn't have ClaimableBalance… upgrading to the latest made it work!

PR #2199, opened 2025-09-10
Fix/restore

elizabethengelman:
### What

Update the parsing done in `stellar contract restore` cmd to allow for more LedgerEntryChange patterns.

### Why

We previously had a select few patterns that we were matching from the restore tx's changes, to get the updated TTL to return to the user on the CLI. However, we were missing some valid matches i.e. 
```
[Restored(LedgerEntry { 
	last_modified_ledger_seq: 3895, 
	data: Ttl(TtlEntry { 
		key_hash: Hash(83a1270ef2b7bbe3e14e5c4b4538d071ad0a436ab4ef22a5f16a5dda26604b53),
		live_until_ledger_seq: 4694 }), 
	ext: V0 }),

 Restored(LedgerEntry { 
	last_modified_ledger_seq: 247, 
	data: ContractData(ContractDataEntry { 
			ext: V0, 
			contract: Contract(ContractId(Hash(bcb0fd9dfa6eadbf03049c1469716ac103ce5a37e917d9a4efde75b3cae0b5a1))), 
			key: Symbol(ScSymbol(StringM(COUNTER))), 
			durability: Persistent,
			val: U32(1)
		 }), 
	ext: V0 })]
```

### Known limitations

I did a refactor that would allow for any LedgerEntryChange types, as long as the data field is a ttl entry (`LedgerEntryData::Ttl`), and then take the largest `live_until_ledger_seq` value of the change set. I would love another set of eyes on this for confirmation that this is ok. Since this is just impacting parsing and displaying the tx's result it should be fairly safe, as it won't effect the tx or restore logic at all.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/restore.rs
+++ b/cmd/soroban-cli/src/commands/contract/restore.rs
@@ -239,51 +239,22 @@ impl NetworkRunnable for Cmd {
 }
 
 fn parse_changes(changes: &[LedgerEntryChange]) -> Option<u32> {
-    match changes.len() {
-        // Handle case with 2 changes (original expected format)
-        2 => match (&changes[0], &changes[1]) {
-            (
-                LedgerEntryChange::State(_),
-                LedgerEntryChange::Restored(LedgerEntry {
-                    data:
-                        LedgerEntryData::Ttl(TtlEntry {
-                            live_until_ledger_seq,
-                            ..
-                        }),
-                    ..
-                })
-                | LedgerEntryChange::Updated(LedgerEntry {
-                    data:
-                        LedgerEntryData::Ttl(TtlEntry {
-                            live_until_ledger_seq,
-                            ..
-                        }),
-                    ..
-                })
-                | LedgerEntryChange::Created(LedgerEntry {
-                    data:
-                        LedgerEntryData::Ttl(TtlEntry {
-                            live_until_ledger_seq,
-                            ..
-                        }),
-                    ..
-                }),
-            ) => Some(*live_until_ledger_seq),
-                (  LedgerEntryChange::Restored(LedgerEntry {
-                    data:
-                        LedgerEntryData::Ttl(TtlEntry {
-                            live_until_ledger_seq,
-                            ..
-                        }),
-                    ..
-                }), LedgerEntryChange::Restored(_)) => Some(*live_until_ledger_seq),
-            _ => {
-                None
-            },
-        },
-        // Handle case with 1 change (single "Restored" type change)
-        1 => match &changes[0] {
-            LedgerEntryChange::Restored(LedgerEntry {
+    if changes.len() == 3 {
+        return None;
+    }
+
+    changes
+        .iter()
+        .filter_map(|change| match change {
+            LedgerEntryChange::State(LedgerEntry {
+                data:
+                    LedgerEntryData::Ttl(TtlEntry {
+                        live_until_ledger_seq,
+                        ..
+                    }),
+                ..
+            })
+            | LedgerEntryChange::Restored(LedgerEntry {
                 data:
                     LedgerEntryData::Ttl(TtlEntry {
                         live_until_ledger_seq,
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/contract/restore.rs
```diff
@@ -239,41 +239,22 @@ impl NetworkRunnable for Cmd {
...
-                }),
-            ) => Some(*live_until_ledger_seq),
-            _ => None,
-        },
-        // Handle case with 1 change (single "Restored" type change)
-        1 => match &changes[0] {
-            LedgerEntryChange::Restored(LedgerEntry {
+    if changes.len() == 3 {
+        return None;
+    }
```
elizabethengelman: We have a unit test below that specifically for 3 changes, and expects that this fn returns None. Adding this here is a little random, but I wasn't sure where else would be appropriate.

dmkozh: This test doesn't make much sense to me. If you always restore a single entry, then you should expect either 0 or 2 meta entries to be present. 0 case should be a very rare edge case for when someone restores an entry just before you. If an entry is already restored as of simulation, then we shouldn't submit the transaction to the network at all (have we done this change btw?).

So I'd say that for the edge case of 0 entries CLI should descriptive a descriptive non-error message, like 'entry is already live, someone must have just restored it simultaneously with you'. In case of 2 entries you can parse the restored entry TTL as you do now. Anything else should raise a descriptive error, like 'internal CLI error - unexpected meta format encountered'. Hopefully if users ever encounter the latter error, they would reach out to us. The same error should be returned for unexpected changes format when we have 2 elements.

cmd/soroban-cli/src/commands/contract/restore.rs
```diff
@@ -239,41 +239,22 @@ impl NetworkRunnable for Cmd {
...
+
+    changes
+        .iter()
+        .filter_map(|change| match change {
+            LedgerEntryChange::State(LedgerEntry {
```
dmkozh: I don't think we ever return State for restored entries as of p23 (this made sense while CLI v23 was used before mainnet upgrade, but it's redundant now).

cmd/soroban-cli/src/commands/contract/restore.rs
elizabethengelman: I was a little confused by that test as well, and was going to remove it, but wanted to get another set of eyes on it. 👍 

> If an entry is already restored as of simulation, then we shouldn't submit the transaction to the network at all (have we done this change btw?).

We haven't done this change yet, I just opened an issue to make sure that we get this prioritized https://github.com/stellar/stellar-cli/issues/2201

> ...for the edge case of 0 entries CLI should descriptive a descriptive non-error message

We currently have this as the message when an entry is still live. It could use some refining like you say, but does this fit the general idea of what you were thinking?
```
❯ stellar contract restore --id hello-world --source ee --network local --key 
ℹ️ Signing transaction: 0fa1d83165e56505f4b3e584b2766ff342a0573ad34db67b9dfc79600a12f64c
New ttl ledger: 58926
```

dmkozh: `New ttl ledger` is misleading, because nothing happened. Otherwise this seems fine for the case when the entry is live as of simulation.

The edge case I'm thinking about is a bit different: it is possible that during the simulation the entry is archived, but when tx is submitted, someone else restores the entry just before your transaction has been applied. In this case the transaction will contain no meta. This is rather unlikely, but we still probably don't want to return a confusing error message for that case.

elizabethengelman: Also, from what you're saying above, it seems like we should not expect to see only 1 change either. Is that right?

We recently added the ability to parse just one change in https://github.com/stellar/stellar-cli/pull/2168. I'm wondering if that is still needed or if we should remove that support too. Though, with this refactor of `parse_changes` we're also able to handle just one change too.

CC @sagpatil

cmd/soroban-cli/src/commands/contract/restore.rs
elizabethengelman: Sounds good, I can remove that

cmd/soroban-cli/src/commands/contract/restore.rs
dmkozh: For the restore operation I'm not sure how we could end up with any other number of changes besides 2 or 0 (very rarely). But I may be missing something. I see that the change you've linked also contains update/create meta, which shouldn't happen for restore op either. So maybe the function has different usage contexts?

cmd/soroban-cli/src/commands/contract/restore.rs
dmkozh: I see that's still there.

elizabethengelman: it should be removed now

cmd/soroban-cli/src/commands/contract/restore.rs
elizabethengelman: > maybe the function has different usage contexts

Hm, that is an interesting point. I don't see that we're using `parse_changes` anywhere else in the CLI code. 

I am thinking that I may merge this in as-is, since this is just for parsing the response, it shouldn't effect the tx itself. Does that sound ok @dmkozh @sagpatil ?

dmkozh: The only thing some sort of assertion would be useful for is to discover a behavior change on protocol boundaries. But that seems really unlikely to happen, so the current implementation seems fine.

Issue #2129, opened 2025-08-12
Implement missing classic operations


fnando:
- [x] ~allow-trust~ (deprecated) 
- [x] create-claimable-balance
- [x] claim-claimable-balance
- [x] begin-sponsoring-future-reserves
- [x] end-sponsoring-future-reserves
- [x] revoke-sponsorship
- [x] clawback
- [x] clawback-claimable-balance
- [x] liquidity-pool-deposit
- [x] liquidity-pool-withdraw
- [x] ~invoke-host-function~[^1] (use `stellar contract invoke` instead)
- [x] ~extend-footprint-ttl~[^1] (use `stellar contract extend` instead)
- [x] ~restore-footprint~[^1] (use `stellar contract restore` instead)

https://developers.stellar.org/docs/learn/fundamentals/transactions/list-of-operations

[^1]: hard to add right now because it'd require a lot of changes to support `TransactionExt::V1(SorobanTransactionData)`

PR #2209, opened 2025-09-15
Add `stellar tx new liquidity-pool-deposit` and `stellar tx op add liquidity-pool-deposit`.

fnando:
### What

Add two commands:

- `stellar tx new liquidity-pool-deposit`
- `stellar tx op add liquidity-pool-deposit`

### Why

#2129

### Known limitations

N/A


elizabethengelman (approved):
Looks good! I wonder if its worth adding a test too

PR #2210, opened 2025-09-15
Hack fix to the address resolution in the snapshot create command

leighmcculloch:
### What

Hack fix to the address resolution in the snapshot create command.

### Why

To unblock @kalepail affected by https://github.com/stellar/stellar-cli/issues/2207.

### Known limitations

This change probably doesn't address the full issue with address resolution. I see multiple functions that all seem to have a mistaken assumption that ScAdderss only contains contract addresses, when it can also contain an account address. This PR is just open teporarily to unblock @kalepail and isn't intended for merging. But it may help inform the final fix for the issue too.

### Usage

To use this change instead of the release stellar-cli, install from source using the following command:

```
cargo install --locked --git https://github.com/stellar/stellar-cli --branch snapshot-address-resolution stellar-cli
```


fnando:
@leighmcculloch do we want to merge this in? Does it need anything else?

Issue #2198, opened 2025-09-09
Populate fields in snapshots created with stellar snapshot create


leighmcculloch:
The stellar snapshot create command should populate the ttl fields based on the config settings of the network at that ledger.

Today those fields are 0:
```
  "timestamp": 0,
  "min_persistent_entry_ttl": 0,
  "min_temp_entry_ttl": 0,
  "max_entry_ttl": 0,
```

THe missing ttl fields especially cause problems when using the snapshot in contract tests because a max_entry_ttl of 0 means that storage cannot be extended as it normally would be able to.

PR #2219, opened 2025-09-22
Fetch snapshot values dynamically rather than hardcoding them.

fnando:
### What

#2198

### Why

So we have a better snapshot representation.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/snapshot/create.rs
+++ b/cmd/soroban-cli/src/commands/snapshot/create.rs
@@ -165,6 +165,47 @@ impl Cmd {
         print.infoln(format!("Network Passphrase: {network_passphrase}"));
         print.infoln(format!("Network id: {}", hex::encode(network_id)));
 
+        // Get ledger close time and base reserve from RPC
+        let mut ledger_close_time = 0u64;
+        let mut base_reserve = 1u32; // Default value
+        if let Ok(network) = self.network.get(&global_args.locator) {
+            if let Ok(client) = network.rpc_client() {
+                match client
+                    .get_ledgers(
+                        crate::rpc::LedgerStart::Ledger(ledger),
+                        Some(1),
+                        Some("json".to_string()),
+                    )
+                    .await
+                {
+                    Ok(result) if !result.ledgers.is_empty() => {
+                        let ledger_data = &result.ledgers[0];
+
+                        if let Ok(parsed_time) = ledger_data.ledger_close_time.parse::<u64>() {
+                            ledger_close_time = parsed_time;
+                            print.infoln(format!("Ledger Close Time: {ledger_close_time}"));
+                        } else {
+                            print.warnln(format!(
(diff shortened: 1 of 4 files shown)


leighmcculloch:
The values should be retrievable out of the archive itself, instead of from an rpc. So that we don't need two data sources and so that this command continues to function for ledgers older than 7 days.

cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -287,6 +312,18 @@ impl Cmd {
                     if seen.contains(&key) {
                         continue;
                     }
+
+                    // Extract TTL settings from StateArchival config entries before filtering
+                    let Some(val) = val else { continue };
```
leighmcculloch: This line needs to come after the `seen.insert`, otherwise a `DeadEntry` for a key (see a few lines up where the bucket entrys are pulled out) will be lost. We have to "see" the dead entry so that the next time we see a live entry for a key we ignore it. Keys can appear multiple times and we only look at the first time we see a key.

So I think how I'd structure the logic here, is "keep" the config setting and treat it the same as other records, as a seen key, then handle pulling the values out of it in the `match &val.data` block the same as we do other values.

cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -287,6 +312,18 @@ impl Cmd {
...
+                    if let LedgerEntryData::ConfigSetting(ConfigSettingEntry::StateArchival(
+                        state_archival,
+                    )) = &val.data
```
leighmcculloch: There's a bug with this logic that old configs could be override new configs, because the config could be in multiple buckets but we should only take the first time we see it. If we implement the logic as per https://github.com/stellar/stellar-cli/pull/2219#discussion_r2374160865, then this issue should be resolved.

leighmcculloch (commented):
Couple issues that need addressing.

cmd/soroban-cli/src/commands/snapshot/create.rs
fnando: Done! Thanks for the suggestion.

cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -284,29 +309,47 @@ impl Cmd {
...
+                            false
+                        }
+
                         LedgerEntryData::ContractData(e) => {
                             // If a contract instance references contract
                             // executable stored in another ledger entry, add
                             // that ledger entry to the filter so that Wasm for
                             // any filtered contract is collected too in the
                             // second pass.
-                            if keep && e.key == ScVal::LedgerKeyContractInstance {
```
fnando, line 309: `keep` was always `true` at this point (there's a guard to continue the loop if it's false on line 326).

leighmcculloch (approved):

PR #2225, opened 2025-09-26
Add WASM fetch by hash support

### LINKED ISSUE #2223: stellar contract fetch should support --wasm-hash
The `stellar contract fetch` command should support `--wasm-hash`.

Today it only supports fetching a .wasm with a contract id, but it should also be possible to fetch based on the wasm-hash.

leighmcculloch:
### What
  Support fetch contract wasm using a hash instead of contract ID. Add new --wasm-hash argument to contract fetch command.

  ### Why
  Enable fetching WASM files directly by their content hash for faster deployment and verification. Support hash-based retrieval alongside existing ID-based approach.

Close #2223

### DIFF
--- a/cmd/crates/soroban-test/tests/it/integration/contract/fetch.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/contract/fetch.rs
@@ -1,0 +1,54 @@
+use crate::integration::util::{deploy_contract, DeployOptions, HELLO_WORLD};
+
+use soroban_test::TestEnv;
+
+#[tokio::test]
+async fn tx_fetch_with_hash() {
+    let sandbox = &TestEnv::new();
+    let test_account_alias = "test";
+    let wasm_bytes = HELLO_WORLD.bytes();
+    let wasm_hash = HELLO_WORLD.hash().unwrap();
+    let _contract_id = deploy_contract(
+        sandbox,
+        HELLO_WORLD,
+        DeployOptions {
+            deployer: Some(test_account_alias.to_string()),
+            ..Default::default()
+        },
+    )
+    .await;
+
+    sandbox
+        .new_assert_cmd("contract")
+        .arg("fetch")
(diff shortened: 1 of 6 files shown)


cmd/crates/soroban-test/tests/it/integration/contract/fetch.rs
```diff
@@ -0,0 +1,54 @@
...
+        .arg("fetch")
+        .arg("--wasm-hash")
+        .arg(wasm_hash.to_string())
+        .assert()
+        .success()
+        .stdout(predicates::ord::eq(wasm_bytes));
+}
+
+#[tokio::test]
+async fn tx_fetch_with_id() {
```
leighmcculloch, line 32: There was no tests for the `stellar contract fetch` command so this test is testing the existing functionality.

cmd/soroban-cli/src/commands/contract/fetch.rs
```diff
@@ -22,7 +22,10 @@ use crate::{
 pub struct Cmd {
     /// Contract ID to fetch
     #[arg(long = "id", env = "STELLAR_CONTRACT_ID")]
-    pub contract_id: config::UnresolvedContract,
+    pub contract_id: Option<config::UnresolvedContract>,
+    /// Wasm to fetch
+    #[arg(long = "wasm-hash", conflicts_with = "contract_id")]
```
fnando, line 27: contract_id has an env var to it. Should we add one for wasm-hash as well? `STELLAR_WASM_HASH` maybe?

leighmcculloch: I don't think we use a env var for the wasm hash on other commands, so I don't think so. Having an env var on the first option is also odd, but I don't think we need to remove it to create consistency without it.

fnando (approved):

PR #2230, opened 2025-09-30
Fix/soroban auth signing

elizabethengelman:
### What

Closes https://github.com/stellar/stellar-cli/issues/2039

### Why

Previously we were trying to get a signer key's private key directly to sign an invoke tx and when signing auth entries, but this doesn't work for secure store or ledger device signers since they don't expose their private key directly. 

This PR updates the invoke tx signing, and signing auth entries to use a signer to sign, rather than a signing key directly so we don't need explicit access to the private key. 

#### local source key, local auth key

 ```
stellar contract invoke --source local-key-1 --id hello-world --network local  \
-- auth  --addr  local-key-2 --world hi
```

- worked before this change ✅ 
- works after this change ✅ 

#### local source key, secure auth key

 ```
stellar contract invoke --source local-key-1 --id hello-world --network local  \
-- auth  --addr secure-key-1 --world hi
```

- worked before this change ❌ 
    - `❌ error: Missing signing key for account GB765REIZB4KQEN5TYFEPN3FQQMGIZHVXE5SKDD3REULDSTJWTR5E7GL`
- works after this change ✅ 

#### secure source key, local auth key

 ```
stellar contract invoke --source secure-key-1 --id hello-world --network local  \
-- auth  --addr local-key-1 --world hi
```

- worked before this change
    - `❌ error: Secure Store does not reveal secret key`
- works after this change ✅

#### secure source key, secure auth key

 ```
stellar contract invoke --source secure-key-1 --id hello-world --network local  \
-- auth  --addr secure-key-2 --world hi
```

- worked before this change
    - `❌ error: Secure Store does not reveal secret key`
- works after this change ✅ 

### Known limitations

This PR does not address ledger devices yet, that will be a follow up PR to avoid this one getting too large.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/arg_parsing.rs
+++ b/cmd/soroban-cli/src/commands/contract/arg_parsing.rs
@@ -528,19 +528,23 @@ fn get_available_functions(spec: &Spec) -> String {
 fn is_primitive_type(type_def: &ScSpecTypeDef) -> bool {
     matches!(
         type_def,
-        ScSpecTypeDef::U64
-            | ScSpecTypeDef::I64
+        ScSpecTypeDef::U32
+            | ScSpecTypeDef::U64
             | ScSpecTypeDef::U128
-            | ScSpecTypeDef::I128
-            | ScSpecTypeDef::U32
-            | ScSpecTypeDef::I32
             | ScSpecTypeDef::U256
+            | ScSpecTypeDef::I32
+            | ScSpecTypeDef::I64
+            | ScSpecTypeDef::I128
             | ScSpecTypeDef::I256
             | ScSpecTypeDef::Bool
             | ScSpecTypeDef::Symbol
             | ScSpecTypeDef::String
+            | ScSpecTypeDef::Bytes
+            | ScSpecTypeDef::BytesN(_)
             | ScSpecTypeDef::Address
             | ScSpecTypeDef::MuxedAddress
+            | ScSpecTypeDef::Timepoint
+            | ScSpecTypeDef::Duration
             | ScSpecTypeDef::Void
     )
 }
@@ -636,7 +640,7 @@ fn parse_argument_with_validation(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use stellar_xdr::curr::{ScSpecTypeDef, ScSpecTypeOption, ScSpecTypeVec};
+    use stellar_xdr::curr::{ScSpecTypeBytesN, ScSpecTypeDef, ScSpecTypeOption, ScSpecTypeVec};
 
     #[test]
     fn test_get_type_name_primitives() {
@@ -670,6 +674,41 @@ mod tests {
             element_type: Box::new(ScSpecTypeDef::String),
         }));
         assert_eq!(get_type_name(&vec_type), "vector of string");
+    }
+
+    #[test]
+    fn test_is_primitive_type_all_primitives() {
+        assert!(is_primitive_type(&ScSpecTypeDef::U32));
+        assert!(is_primitive_type(&ScSpecTypeDef::I32));
+        assert!(is_primitive_type(&ScSpecTypeDef::U64));
+        assert!(is_primitive_type(&ScSpecTypeDef::I64));
+        assert!(is_primitive_type(&ScSpecTypeDef::U128));
+        assert!(is_primitive_type(&ScSpecTypeDef::I128));
+        assert!(is_primitive_type(&ScSpecTypeDef::U256));
+        assert!(is_primitive_type(&ScSpecTypeDef::I256));
+
+        assert!(is_primitive_type(&ScSpecTypeDef::Bool));
+        assert!(is_primitive_type(&ScSpecTypeDef::Symbol));
+        assert!(is_primitive_type(&ScSpecTypeDef::String));
+        assert!(is_primitive_type(&ScSpecTypeDef::Void));
+        assert!(is_primitive_type(&ScSpecTypeDef::Bytes));
+        assert!(is_primitive_type(&ScSpecTypeDef::BytesN(
+            ScSpecTypeBytesN { n: 32 }
+        )));
+        assert!(is_primitive_type(&ScSpecTypeDef::BytesN(
+            ScSpecTypeBytesN { n: 64 }
+        )));
+
+        assert!(is_primitive_type(&ScSpecTypeDef::Address));
+        assert!(is_primitive_type(&ScSpecTypeDef::MuxedAddress));
+        assert!(is_primitive_type(&ScSpecTypeDef::Timepoint));
+        assert!(is_primitive_type(&ScSpecTypeDef::Duration));
+
+        assert!(!is_primitive_type(&ScSpecTypeDef::Vec(Box::new(
+            ScSpecTypeVec {
+                element_type: Box::new(ScSpecTypeDef::U32),
+            }
+        ))));
     }
 
     #[test]
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -269,6 +269,17 @@ impl Spec {
                     ScType::U128 | ScType::I128 | ScType::U256 | ScType::I256 => {
                         Ok(Value::String(s.to_owned()))
                     }
+                    ScType::Timepoint | ScType::Duration => {
+                        // timepoint and duration both expect a JSON object with the value
+                        // being the u64 number as a string, and key being the type name
+                        let key = match t {
+                            ScType::Timepoint => "timepoint",
+                            ScType::Duration => "duration",
+                            _ => unreachable!(),
+                        };
+
+                        Ok(json!({ key: s }))
+                    }
                     _ => Ok(val),
                 },
             )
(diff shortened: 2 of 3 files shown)


cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -102,60 +105,60 @@ pub fn build_host_function_parameters(
...
-                    tracing::debug!(
-                        "file {arg_path:?}, has contents:\n{file_contents}\nAnd type {:#?}\n{}",
-                        i.type_,
-                        file_contents.len()
-                    );
-                    spec.from_string(&file_contents, &i.type_)
-                        .map_err(|error| Error::CannotParseArg { arg: name, error })
+    let mut parsed_args = Vec::with_capacity(func.inputs.len());
+    let mut signers = Vec::<Signer>::new();
+    for i in func.inputs.iter() {
```
elizabethengelman, line 186: A big chunk of this change was using a for loop instead of map so that I could more easily await `resolve_signer`.

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
elizabethengelman: `build_host_function_parameters` was changed to an async fn so that we could await getting the signer from a `Secret`.

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -51,9 +52,11 @@ pub enum Error {
     HelpMessage(String),
     #[error("Unsupported ScAddress {0}")]
     UnsupportedScAddress(String),
+    #[error(transparent)]
+    Signer(#[from] signer::Error),
 }
 
-pub type HostFunctionParameters = (String, Spec, InvokeContractArgs, Vec<SigningKey>);
+pub type HostFunctionParameters = (String, Spec, InvokeContractArgs, Vec<Signer>);
```
elizabethengelman, line 82: by returning a `Signer` instead of a `ed25519_dalek::SigningKey` we're able to sign the auth entries without getting the private get directly

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -102,60 +105,60 @@ pub fn build_host_function_parameters(
 
     let func = spec.find_function(function)?;
     // create parsed_args in same order as the inputs to func
-    let mut signers: Vec<SigningKey> = vec![];
```
elizabethengelman, line 180: a large part of this change was changing the map to a for loop so i could more easily await `resolve_signer`

cmd/soroban-cli/src/config/mod.rs
```diff
@@ -104,20 +111,21 @@ impl Args {
     pub async fn sign_soroban_authorizations(
         &self,
         tx: &Transaction,
-        signers: &[ed25519_dalek::SigningKey],
+        signers: &[Signer],
     ) -> Result<Option<Transaction>, Error> {
         let network = self.get_network()?;
-        let source_key = self.key_pair()?;
+        let source_signer = self.source_signer().await?;
```
elizabethengelman, line 117: `key_pair()` tries to get the private key directly, so this is where we were failing when using a secure store key as the source account, so I switched this to also be a `Signer` instead of a `SigningKey`.

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -61,10 +63,10 @@ fn requires_auth(txn: &Transaction) -> Option<xdr::Operation> {
 
 // Use the given source_key and signers, to sign all SorobanAuthorizationEntry's in the given
 // transaction. If unable to sign, return an error.
-pub fn sign_soroban_authorizations(
+pub async fn sign_soroban_authorizations(
     raw: &Transaction,
-    source_key: &ed25519_dalek::SigningKey,
-    signers: &[ed25519_dalek::SigningKey],
+    source_signer: &Signer,
+    signers: &[Signer],
```
elizabethengelman, line 69: I had thought about appending the `source_signer` to the `signers` vec, but we are handling the `source_signer` slightly differently below, so I left it as is, but happy to rethink that.

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -83,76 +85,78 @@ pub fn sign_soroban_authorizations(
 
     let network_id = Hash(Sha256::digest(network_passphrase.as_bytes()).into());
 
-    let verification_key = source_key.verifying_key();
-    let source_address = verification_key.as_bytes();
-
```
elizabethengelman, line 88: similar to above, this was changed from a map to a for loop

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -264,6 +261,37 @@ impl Signer {
             _ => Err(Error::UnsupportedTransactionEnvelopeType),
         }
     }
+
+    // when we implement this for ledger we'll need it to be awaited
+    #[allow(clippy::unused_async)]
+    pub async fn get_public_key(&self) -> Result<[u8; 32], Error> {
+        match &self.kind {
+            SignerKind::Local(local_key) => Ok(*local_key.key.verifying_key().as_bytes()),
+            SignerKind::Ledger(_ledger) => todo!("ledger key"),
```
elizabethengelman: I am planning to implement this for ledger devices in a follow up PR

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -319,4 +355,10 @@ impl SecureStoreEntry {
         let signature = Signature(signed_tx_hash.clone().try_into()?);
         Ok(DecoratedSignature { hint, signature })
     }
+
+    pub fn sign_payload(&self, payload: [u8; 32]) -> Result<Ed25519Signature, Error> {
+        let signed_bytes = secure_store::sign_tx_data(&self.name, self.hd_path, &payload)?;
+        let sig = Ed25519Signature::from_bytes(signed_bytes.as_slice().try_into()?);
+        Ok(sig)
+    }
```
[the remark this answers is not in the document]
elizabethengelman: Error does have a TryFromSliceError variant

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -264,6 +261,37 @@ impl Signer {
...
+                local_key.sign_payload(p)
+            }
+            SignerKind::Ledger(_ledger) => todo!("ledger"),
+            SignerKind::Lab => Err(Error::ReturningSignatureFromLab),
+            SignerKind::SecureStore(secure_store_entry) => {
+                let p = <[u8; 32]>::try_from(payload)?;
+                secure_store_entry.sign_payload(p)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
elizabethengelman: Error does have a TryFromSliceError variant

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -83,76 +85,78 @@ pub fn sign_soroban_authorizations(
...
+        let mut signer: Option<&Signer> = None;
+        for s in signers {
+            if needle == &s.get_public_key().await? {
+                signer = Some(s);
+            }
+        }
+
+        if needle == &source_signer.get_public_key().await? {
+            signer = Some(source_signer);
+        }
```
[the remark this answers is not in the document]
elizabethengelman: i removed the await from this fn - for now `get_public_key` can be synchronous. if/when we implement this for a ledger device, we will need to make this async

cmd/soroban-cli/src/signer/mod.rs
elizabethengelman: implementing this for ledger devices does not seem to be a priority at the moment - there is an [issue](https://github.com/stellar/stellar-cli/issues/1872) to track this work if we decide to pick it up again in the future

cmd/soroban-cli/src/signer/mod.rs
mootz12: IMO keeping them separate was a good call

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
```diff
@@ -215,24 +219,40 @@ fn parse_single_argument(
         ) {
             let trimmed_s = s.trim_matches('"');
             let addr = resolve_address(trimmed_s, config)?;
-            if let Some(signer) = resolve_signer(trimmed_s, config) {
+            if let Some(signer) = resolve_signer(trimmed_s, config).await {
```
mootz12, line 222: we should do this outside of argument parsing so it only occurs once.

In testing, I was getting prompted multiple times for access to my key before the CLI attempted to simulate, and I assume this is the culprit

mootz12 (changes requested):
I'm not all that familiar with keychain for storing keys, but I assume this is expected? Bombadil is my secure store key on mac.

```
stellar-cli % devcli keys secret bombadil                                                                             

❌ error: Secure Store does not reveal secret key
```

cmd/soroban-cli/src/commands/contract/arg_parsing.rs
elizabethengelman: This is a good catch. I think I'd like to handle this in a separate PR though, because I think that being prompted multiple times by keychain for access to the key could be related to something else instead/in addition to this.

I was previously seeing it on key generation, and fetching the public key... and now that isn't consistently happening. So i think this is worthy of a separate investigation.

elizabethengelman: Just created a new issue for this: https://github.com/stellar/stellar-cli/issues/2291, feel free to add any notes!

PR #2241, opened 2025-10-07
Deprecate `stellar contract optimize` in favor of `stellar contract build --optimize`.

fnando:
### What

```console
$ stellar contract build --optimize
ℹ️ CARGO_BUILD_RUSTFLAGS=--remap-path-prefix=/Users/fnando/.cargo/registry/src= cargo rustc --manifest-path=contracts/foo/Cargo.toml --crate-type=cdylib --target=wasm32v1-none --release
    Finished `release` profile [optimized] target(s) in 0.04s
ℹ️ Build Summary:
   Wasm File: target/wasm32v1-none/release/foo.wasm (537 bytes (original size was 574 bytes))
   Wasm Hash: 9276bc278792b46e89b299bb55beffd6ca2fa2d17d723abe1ec615e6d5752485
   Wasm Size: 537 bytes optimized (original size was 574 bytes)
   Exported Functions: 2 found
     • _
     • hello
✅ Build Complete

$ stellar contract build
ℹ️ CARGO_BUILD_RUSTFLAGS=--remap-path-prefix=/Users/fnando/.cargo/registry/src= cargo rustc --manifest-path=contracts/foo/Cargo.toml --crate-type=cdylib --target=wasm32v1-none --release
ℹ️ Build Summary:
   Wasm File: target/wasm32v1-none/release/foo.wasm (574 bytes)
   Wasm Hash: b39694f92f640414ccf28326ecf1d3b1b48981167eb00fc00739f71dac344013
   Wasm Size: 574 bytes
   Exported Functions: 2 found
     • _
     • hello
✅ Build Complete
```

### Why

Supersedes #2045.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -265,8 +265,8 @@ impl Cmd {
                     optimize::optimize(true, vec![final_path.clone()], Some(path.clone()))?;
                     optimized_wasm_bytes = fs::read(&path).map_err(Error::ReadingWasmFile)?;
 
-                    fs::copy(&path, &final_path).map_err(Error::CopyingWasmFile)?;
-                    fs::remove_file(&path).map_err(Error::DeletingArtifact)?;
+                    fs::remove_file(&final_path).map_err(Error::DeletingArtifact)?;
+                    fs::rename(&path, &final_path).map_err(Error::CopyingWasmFile)?;
                 }
 
                 Self::print_build_summary(&print, &final_path, wasm_bytes, optimized_wasm_bytes);
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -17,7 +17,11 @@ use std::{
 };
 use stellar_xdr::curr::{Limited, Limits, ScMetaEntry, ScMetaV0, StringM, WriteXdr};
 
-use crate::{commands::global, print::Print};
+use crate::{
+    commands::{contract::optimize, global},
+    print::Print,
+    wasm,
+};
 
 /// Build a contract from source
 ///
@@ -32,6 +36,7 @@ use crate::{commands::global, print::Print};
 /// To view the commands that will be executed, without executing them, use the
 /// --print-commands-only option.
 #[derive(Parser, Debug, Clone)]
+#[allow(clippy::struct_excessive_bools)]
 pub struct Cmd {
     /// Path to Cargo.toml
     #[arg(long)]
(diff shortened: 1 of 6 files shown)


cmd/soroban-cli/src/commands/contract/mod.rs
```diff
@@ -81,7 +81,7 @@ pub enum Cmd {
     ///     stellar contract invoke ... -- --help
     Invoke(invoke::Cmd),
 
-    /// Optimize a WASM file
+    /// (Deprecated in favor of `build --optimize` subcommand) Optimize a WASM file
```
leighmcculloch: Suggest a deprecated-use statement that is simpler language.
```suggestion
    /// (Deprecated, use `build --optimize`) Optimize a WASM file
```

Shaptic:
Is there any harm to this being the default behavior and dropped the need for a flag entirely? As in, why wouldn't we always just include an optimized WASM as part of the build process?

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -219,7 +256,17 @@ impl Cmd {
                     target_file_path
                 };
 
-                Self::print_build_summary(&print, &final_path)?;
+                let optimized_file_path = if self.optimize {
+                    let mut path = final_path.clone();
+                    path.set_extension("optimized.wasm");
```
leighmcculloch: I thought we already implemented this feature, but maybe we just talked about it.

When we had discussed this feature previously we discussed continuing to use the .optimized.wasm extension for consistency with the optimize command. But if the plan is to deprecate the optimize command, then I don't think we need to worry too much about consistency and instead should design this command for the most native 'rust-like' experience.

When using rust to optimize the final target file is still always the same just the same .wasm file, so I think we should consider doing the same here.

The important thing if we do that is we need to remove the target file before writin the optimized version, so that on some operating systems, like mac, we don't accidetnally corrupt the intermediary files. (On mac and maybe some other platforms the final .wasm file may be a hardlink to an intermediary file.)

leighmcculloch (approved):
Looks good, but suggestion inline about the file name that gets written, I think we can omit the .optimized if we are deprecating.

leighmcculloch:
> Is there any harm to this being the default behavior and dropped the need for a flag entirely? As in, why wouldn't we always just include an optimized WASM as part of the build process?

I think we need to be really confident about wasm-opt and what it does to files if it is the default behaviour on. I haven't spent enough time to confirm that, so maybe lets open an issue about that and explore that separately to this issue.

Shaptic:
> I think we need to be really confident about wasm-opt and what it does to files if it is the default behaviour on

To be clear, I think we should output both, not just the one optimized variant always.

cmd/soroban-cli/src/commands/contract/build.rs
fnando: Done!

~I'm generating the optimized file separately, then copying it over the original path. That should work, right?~ Ended up removing the origional file and renaming the optimized file.

PR #2251, opened 2025-10-09
Feat: Replace Standalone with Local in TS Boilerplate

zachfedor:
### What

Generated TS bindings create an object with keys for each network. This will replace the legacy `"standalone"` key with the new `"local"` key. Network passphrases are not changed, just the key to reference them.

### Why

This will simplify configuration needed for Scaffold Stellar projects. A single environment variable can be set to match one of the keys on this `network` object and spread out defaults to instantiate the contract's RPC client with the RPC URL and network passphrase, rather than having each individual piece defined as an environment variable.

This shouldn't be a breaking change as we'll still allow these individual environment variables to overwrite the defaults for a given a network.

### Known limitations

[N/A]


leighmcculloch (commented):
Using the 'local' term to match quickstart is preferred, good call.

I would think this would be a breaking change because the generated `networks` object has a value for its `standalone` field today, but that will be `undefined` after this is merged.

For example, looking at the tests:
```ts
import {
  Client,
  networks,
  contract as ContractClient,
} from "test-custom-types";

//...

const contract = new Client({
  ...networks.standalone,
  ...networks.local,
  rpcUrl,
  allowHttp: true,
  publicKey: root.keypair.publicKey(),
  ...signer,
});
```

I note in the description you say this is not a breaking change. Why is that?

Is there a way to make this non-breaking, such as by having `networks` contain both `standalone` and `local` map to the same value?

leighmcculloch:
Reverted the formatting changes in the ts test files so the diff is just the `networks.standalone` → `networks.local` swap.

Also pushed a change that makes this non-breaking by emitting both keys when the passphrase is the Standalone Network passphrase. The primary key is `local`; `standalone` is kept as an alias with a JSDoc `@deprecated` so consumers regenerating bindings get an IDE warning but their code keeps compiling.

Generated output for the local case:

```ts
export const networks = {
  local: {
    networkPassphrase: "Standalone Network ; February 2017",
    contractId: "...",
  },
  /** @deprecated Use `local` instead. */
  standalone: {
    networkPassphrase: "Standalone Network ; February 2017",
    contractId: "...",
  },
} as const
```

Other networks (testnet, futurenet, unknown) are unchanged.

@fnando or @Ryang-21, could one of you take a look?

Ryang-21 (approved):

PR #2263, opened 2025-10-16
Inject CLI version into contract meta during build

leighmcculloch:
### What
  Inject the CLI version into contract meta during contract build.

  ### Why
  CLI version is helpful information when understanding how a contract was built, especially as we move the optimization feature of the cli closer into regular builds and encourage its use. To reproduce builds it will be necessary to know what wasm-opt build and configuration was used, and at the moment the easiest way to identify that is with the stellar-cli's version.

 Close https://github.com/stellar/stellar-cli/issues/2226


fnando (approved):

fnando:
I imagine something like `src=stellar-cli` or `src=some-typescript-cli` and so on.

leighmcculloch:
Maybe I should add an option to not add it, for any sort of low level use cases where someone needs to build using the cli without injecting any meta.

PR #2272, opened 2025-11-03
Group CLI options into sections in help docs

leighmcculloch:
### What
Move the doc-gen binary code into its own crate, separate from the soroban-cli crate, and have it be dependent on the unmerged pull request to clap-markdown that adds support for grouping cli options:
- https://github.com/ConnorGray/clap-markdown/pull/48

### Why
The grouped options experience is a massive improvement. It makes the help documentation easier to scan and understand. It's the way the options are organised on the command line and the help markdown documentation should follow suite. While the pull request adds the feature upstream, the pull request hasn't been accepted yet.

The clap-markdown dependency isn't needed for the stellar-cli other than doc-gen, so it is in any case good to get the dependency outside of the soroban-cli crate, so that it isn't downloaded by cargo everytime someone installs the stellar-cli. Moving doc-gen into its own crate is trivial because the soroban-cli is designed to be embedded, and achieves that. Additionally because the doc-gen crate does not need to be published it can reference the unreleased pull request without issue.

### DIFF
--- a/cmd/doc-gen/README.md
+++ b/cmd/doc-gen/README.md
@@ -4,6 +4,18 @@ A generator for the help documentation using clap-markdown.
 
 ## Usage
 
+Run the following command from the workspace root.
+
 ```
 cargo run --package doc-gen
+```
+
+The command will update the file `FULL_HELP_DOCS.md` located in the root of the workspace.
+
+Typically this command is not run directly, but in the root `Makefile` the `docs` target will run this command in conjunction with formatting the markdown.
+
+So typically run this command via make:
+
+```
+make docs
 ```


.github/workflows/dependency-check.yml
leighmcculloch, line 1: I've removed hte dependency check job because it's been superseded by the cargo deny.toml file in the repo that also does a git-rev check, but is more feature rich and allows us to specify git-rev's that are exception cases.

cmd/doc-gen/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+readme = "README.md"
+version.workspace = true
+edition = "2021"
+rust-version.workspace = true
+publish = false
+
+[dependencies]
+soroban-cli = { workspace = true }
+# clap-markdown from https://github.com/ConnorGray/clap-markdown/pull/48
+clap-markdown = { version = "0.1.5", git = "https://github.com/ConnorGray/clap-markdown", rev = "42956b342cef3325d9060fc43995d595e7c8aa66" }
```
marcelosalloum, line 17: Do you plan to remove this comment before merging the PR?

cmd/doc-gen/README.md
```diff
@@ -0,0 +1,9 @@
+# Stellar CLI Documentation Generator
+
+A generator for the help documentation using clap-markdown.
+
+## Usage
+
+```
+cargo run --package doc-gen
+```
```
marcelosalloum, line 11: It'd be good to have an example here. Something as simple as what we see [here](https://github.com/ConnorGray/clap-markdown/blob/558062836bb8e3fb8334b24d56ffa3cc8997b8fc/tests/test_markdown.rs#L103-L154) would work.

deny.toml
```diff
@@ -258,7 +258,10 @@ unknown-git = "deny"
 # if not specified. If it is specified but empty, no registries are allowed.
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
 # List of URLs for allowed Git repositories
-allow-git = []
+allow-git = [
+  # Only used by the unpublished doc-gen crate, temporarily until PR is merged: https://github.com/ConnorGray/clap-markdown/pull/48
+  "https://github.com/ConnorGray/clap-markdown?rev=42956b342cef3325d9060fc43995d595e7c8aa66",
+]
```
marcelosalloum, line 264: Just flagging this for visibility

FULL_HELP_DOCS.md
marcelosalloum, line 1: 🤚 Could these generated docs be updated to display the crate version? Thinking about AI agents, they often return commands from older tool versions, so having them (and users) know which version their knowledge relates to is valuable.

marcelosalloum (approved):
LGTM!

cmd/doc-gen/Cargo.toml
leighmcculloch: No. The plan is to merge this PR, get CLI using that fork/pr of clap-markdown, and we can remove that when clap-markdown accepts the change. If clap-markdown doesn't ever accept the change we might move the pr into a fork and publish it on crates.io.

FULL_HELP_DOCS.md
leighmcculloch: That'd be a contribution to clap-markdown, or we fork and add it. Good idea though if you're up for contributing that upstream.

cmd/doc-gen/README.md
leighmcculloch: 1d11a3e93aea7dc6f5d074367dbdc05ca2725502

FULL_HELP_DOCS.md
marcelosalloum: Cool, opened a PR here: https://github.com/ConnorGray/clap-markdown/pull/52

PR #2279, opened 2025-11-06
Add better emoji support for unknown terminals.

fnando:
### What

Running VHS in a docker container prints the wrong emoji spacing. `$TERM_PROGRAM` is not set at all when running with ttyd (the thing vhs uses), so we can default to `unknown`. Worse case scenario, people will see an additional space (but I believe all real terminals set that).

Before fix:

https://github.com/user-attachments/assets/1ecca099-4f35-4172-9796-31335fddb902

After fix:

https://github.com/user-attachments/assets/65dc2e47-1146-4308-a152-2ac84f57b1c5

### Why

So we can have better emoji support.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/print.rs
+++ b/cmd/soroban-cli/src/print.rs
@@ -7,8 +7,6 @@ use crate::{
     config::network::Network, utils::explorer_url_for_transaction, utils::transaction_hash,
 };
 
-const TERMS: &[&str] = &["Apple_Terminal", "vscode", "unknown"];
-
 #[derive(Clone)]
 pub struct Print {
     pub quiet: bool,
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/print.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
     config::network::Network, utils::explorer_url_for_transaction, utils::transaction_hash,
 };
 
-const TERMS: &[&str] = &["Apple_Terminal", "vscode"];
+const TERMS: &[&str] = &["Apple_Terminal", "vscode", "unknown"];
```
leighmcculloch: This variable doesn't have a lot of context where it is, we have to look at the should_add_additional_space fn to know what it is. So we should probably move it into the functions. Constants can be defined inside functions too.

PR #2283, opened 2025-11-11
fix(ts-bindings): client transaction option type

chadoh:
### What

Rather than hard-coding a subset of AssembledTransactionOptions, this imports the type and uses it.

### Why

Avoids needing to do goofy things like this in [the frontend](https://github.com/theahaco/scaffold-stellar-frontend/pull/123/files#diff-b73cfec3be134b268d337f3e5e20880fb584de84a5ef2a2087e81ff606596c79R27)

```ts
const tx = await game.guess(
  { a_number: BigInt(theGuess), guesser: address },
  // @ts-expect-error js-stellar-sdk has bad typings; publicKey is, in fact, allowed
  { publicKey: address },
);
```

Also ensures that these generated methods will always be up-to-date with `AssembledTransactionOptions` as it evolves.

### Known limitations

TypeScript typeahead in this situation is garbage! You just see `AssembledTransactionOptions` instead of actually-useful expanded types. This was the reason we originally hard-coded the options, but I think the trade-off is worth it. Also, maybe a future version of TypeScript won't be so stupid?

Captura de pantalla 2025-11-11 a la(s) 11 12 57

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -112,7 +112,7 @@ jobs:
       - name: Uncompress Artifact
         run: tar xvf ${{ env.ARTIFACT_NAME }}
       - shell: powershell
-        run: winget install --id JRSoftware.InnoSetup --scope machine --silent --accept-package-agreements --accept-source-agreements
+        run: winget install --id JRSoftware.InnoSetup --scope machine --silent --accept-package-agreements --accept-source-agreements --force
       - shell: powershell
         run: |
           $innoPath = "C:\Program Files (x86)\Inno Setup 6"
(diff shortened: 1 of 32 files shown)


fnando (approved):
Is there any documentation that needs to be updated with this on stellar-docs? If the answer is yes, can we prepare those pull requests and leave them in draft mode until we make a CLI release?

chadoh:
@fnando the biggest effect this will have on docs and consuming projects like Scaffold Stellar is that we can remove a `@ts-expect-error` directive from examples. I searched the docs and `ts-expect-error` does not appear. We are happy to update Scaffold Stellar appropriately once this is released.

PR #2288, opened 2025-11-14
Error on Rust 1.91.0 due to wasm compilation bug

leighmcculloch:
### What
  Add validation to disallow Rust 1.91.0 and display an error message directing users to upgrade to 1.91.1+ with a reference to the official Rust blog post.

  ### Why
  Rust 1.91.0 contains a known bug that breaks wasm compilation in some situations, and it may be possible for it to affect contracts. 

More details at https://blog.rust-lang.org/2025/11/10/Rust-1.91.1/

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -145,6 +145,9 @@ pub enum Error {
     #[error("use rust 1.81 or 1.84+ to build contracts (got {0})")]
     RustVersion(String),
 
+    #[error("rust 1.91.0 contains a bug building wasm, use 1.91.1+ instead (https://blog.rust-lang.org/2025/11/10/Rust-1.91.1/)")]
+    RustVersion1910Disallowed,
+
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -145,6 +145,9 @@ pub enum Error {
     #[error("use rust 1.81 or 1.84+ to build contracts (got {0})")]
     RustVersion(String),
 
+    #[error("rust 1.91.0 contains a bug building wasm, use 1.91.1+ instead (https://blog.rust-lang.org/2025/11/10/Rust-1.91.1/)")]
+    RustVersion1910Disallowed,
```
anupsdf, line 149: Is the CI failure because its expecting a string to be passed in to RustVersion1910Disallowed like the one above?

leighmcculloch: I don't think the CI docs validation failure has anything to do with this change.

PR #2292, opened 2025-11-17
Update rpc client to get_events by passing topics as [TopicFilter]

elizabethengelman:
### What

closes https://github.com/stellar/stellar-cli/issues/2021

related rpc-client pr: https://github.com/stellar/rs-stellar-rpc-client/pull/57

Updates the topic filter structure that is being passed to the rpc server `getEvents` method, to align with it's expected API. 

The GetEventsRequest, from the rpc server, is expecting the topics to be an array of TopicFilters, and a TopicFilter is an array of SegmentFilters. So it is expecting an array of arrays.
```
type TopicFilter []SegmentFilter
type SegmentFilter struct {
	Wildcard *string    `json:"-"`
	ScVal    *xdr.ScVal `json:"-"`
}
type EventFilter struct {
	EventType   EventTypeSet  `json:"type,omitempty"`
	ContractIDs []string      `json:"contractIds,omitempty"`
	Topics      []TopicFilter `json:"topics,omitempty"`
}
```

```
cargo run events --start-ledger 757126 --type contract --topic 'AAAADwAAAAh0cmFuc2Zlcg=='  --topic 'AAAADwAAAAh0cmFuc2Zlcg==,**" --network testnet
```

### Why
We are currently just passing in a segment filter, without nesting it in an array of arrays, and are seeing an error when doing so.

```
❌ error: ErrorObject { code: InvalidParams, message: "invalid parameters", data: Some(RawValue("json: cannot unmarshal string into Go struct field EventFilter.filters.topics of type protocol.TopicFilter")) }
```

### Known limitations

n/a

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,8 +74,7 @@ version = "23.0.1"
 # Dependencies from the rs-stellar-rpc-client repo:
 [workspace.dependencies.soroban-rpc]
 package = "stellar-rpc-client"
-git = "https://github.com/stellar/rs-stellar-rpc-client"
-branch = "main"
+version = "23.2.1"
 
 # Dependencies from elsewhere shared by crates:
 [workspace.dependencies]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5377,7 +5377,9 @@ dependencies = [
 
 [[package]]
 name = "stellar-rpc-client"
-version = "23.2.0"
+version = "23.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "061da047abb73d3de3b93cef460757ca1e99850e68bdf677e1102d575b8345a7"
 dependencies = [
  "clap",
  "hex",


cmd/soroban-cli/src/commands/events.rs
```diff
@@ -185,6 +168,40 @@ impl Cmd {
         Ok(())
     }
 
+    fn parse_topics(&self) -> Result<rpc::TopicFilters, Error> {
+        let mut topic_filters: rpc::TopicFilters = Vec::new();
+        for topic in &self.topic_filters {
```
mootz12, line 182: we should probably error if a user passes more than 5 topic filters. Looks like RPC only accepts 5.

ref: https://developers.stellar.org/docs/data/apis/rpc/api-reference/methods/getEvents

Cargo.toml
```diff
@@ -74,7 +74,8 @@ version = "23.0.1"
 # Dependencies from the rs-stellar-rpc-client repo:
 [workspace.dependencies.soroban-rpc]
 package = "stellar-rpc-client"
-version = "23.2.0"
+git = "https://github.com/stellar/rs-stellar-rpc-client"
+branch = "fix/events-for-base64-topics"
```
mootz12: - [x] use published version (leaving this here so we don't forget)

mootz12 (changes requested):
LG2M. Noticed one check that was missing.

cmd/soroban-cli/src/commands/events.rs
```diff
@@ -55,7 +59,7 @@ pub struct Cmd {
     /// into a single filter (i.e. combination of type, IDs, and topics).
     #[arg(
         long = "topic",
-        num_args = 1..=5,
+        num_args = 1.., // allowing 1+ arguments here, and doing additional validation in parse_topics
```
elizabethengelman, line 64: We need to include `num_args` here in order to allow users to pass topics like this 
`--topic topic1 topic 2 topic3`, etc. But with the existing validation, if the uses passes in an extra topic, the error message isn't very useful:

```
$ stellar events --topic topic1 topic2 topic3 topic4 topic5 topic6

error: unexpected argument 'topic6' found
```

with this change, we're relying on the validation in parse_topics, which gives us a nicer error message

`❌ error: max of 5 topic filters allowed per request, received 6`

cmd/soroban-cli/src/commands/events.rs
elizabethengelman: updated in [94112d9](https://github.com/stellar/stellar-cli/pull/2292/commits/94112d92cd146f2160ead5e845b6ce23dc4fd527)

PR #2304, opened 2025-11-24
Add `stellar snapshot merge` to merge several snapshots into one file.

### LINKED ISSUE #2228: Allow to load or combine multiple snapshot files

When you use a snapshot for testing, you sometimes have to consider another contract or address. When that happens, you need to recreate a snapshot and make sure to consider all the addresses you need.

One issue is that if you manually edited the snapshot for some reason (e.g. adjusting an address or some balances), then you need to remember to do that again.

I would like to be able to load multiple snapshots in the env. It could be at once when calling `Env::from_ledger_snapshot_file` or we could have an append function or else.

Regenerate a file every time and make sure you don't forget manual adjustments if any.

fnando:
### What

Implement snapshot merge, so users can easily create one aggregation file with several snapshots. The last inputted file has precedence over previous files:

- "Last wins" conflict resolution for duplicate ledger keys
- Uses last snapshot's metadata

https://github.com/user-attachments/assets/7cf987f6-9066-434b-aaf1-175eeb021d3e

### Why

Close #2228

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-spec-typescript/fixtures/test_constructor/src/index.ts
+++ b/cmd/crates/soroban-spec-typescript/fixtures/test_constructor/src/index.ts
@@ -17,7 +17,6 @@ import type {
   i128,
   u256,
   i256,
-  AssembledTransactionOptions,
   Option,
   Typepoint,
   Duration,
@@ -40,7 +39,7 @@ export interface Client {
    * Construct and simulate a counter transaction. Returns an `AssembledTransaction` object which will have a `result` field containing the result of the simulation. If this transaction changes contract state, you will need to call `signAndSend()` on the returned object.
    * Counter value
    */
-  counter: (options?: AssembledTransactionOptions<u32>) => Promise<AssembledTransaction<u32>>
+  counter: (options?: MethodOptions) => Promise<AssembledTransaction<u32>>
 
 }
 export class Client extends ContractClient {
--- a/cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
+++ b/cmd/crates/soroban-spec-typescript/fixtures/test_custom_types/src/index.ts
@@ -17,7 +17,6 @@ import type {
   i128,
   u256,
   i256,
-  AssembledTransactionOptions,
   Option,
   Typepoint,
   Duration,
@@ -56,6 +55,8 @@ export type TupleStruct = readonly [Test,  SimpleEnum];
 
 export type ComplexEnum = {tag: "Struct", values: readonly [Test]} | {tag: "Tuple", values: readonly [TupleStruct]} | {tag: "Enum", values: readonly [SimpleEnum]} | {tag: "Asset", values: readonly [string, i128]} | {tag: "Void", values: void};
 
+export type RecursiveEnum = {tag: "List", values: readonly [Array<RecursiveEnum>]} | {tag: "Void", values: void};
+
 export const Errors = {
   /**
    * Please provide an odd number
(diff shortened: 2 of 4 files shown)


cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -75,7 +82,7 @@ pub struct Cmd {
     wasm_hashes: Vec<Hash>,
 
     /// Format of the out file.
-    #[arg(long)]
+    #[arg(long, default_value_t = Output::Json)]
```
mootz12: good catch

cmd/soroban-cli/src/commands/snapshot/create.rs
```diff
@@ -39,6 +39,14 @@ pub enum Output {
     Json,
 }
 
+impl std::fmt::Display for Output {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Output::Json => write!(f, "json"),
+        }
+    }
+}
```
mootz12: Not sure I follow this addition. Is it just used for displaying help text to CLI?

cmd/soroban-cli/src/commands/snapshot/merge.rs
```diff
@@ -0,0 +1,123 @@
...
+    // Use a HashMap to track entries by key, with last-wins semantics
+    let mut merged_entries: HashMap<LedgerKey, (Box<stellar_xdr::curr::LedgerEntry>, Option<u32>)> =
+        HashMap::new();
+
+    // Iterate through snapshots in order, so later entries override earlier ones
+    for snapshot in &snapshots {
+        for (key, (entry, ttl)) in &snapshot.ledger_entries {
+            merged_entries.insert((**key).clone(), (entry.clone(), *ttl));
+        }
+    }
```
mootz12, line 33: [picky] we could probably reduce the amount of cloning we do by not borrowing here.

e.g.

```
    // Iterate through snapshots in order, so later entries override earlier ones
    for snapshot in snapshots {
        for (key, entry) in snapshot.ledger_entries {
            merged_entries.insert(*key, entry);
        }
    }
```

We would need to define local variables for the metadata before building merged entries to prevent the borrow checker from being mad about `last_snapshot`.

Def not necessary, but thought I'd throw it out there.

cmd/soroban-cli/src/commands/snapshot/create.rs
fnando: Yeah, this is required by the `default_value_t` above.

cmd/soroban-cli/src/commands/snapshot/merge.rs
fnando: okay, i reworked that part. :)

cmd/soroban-cli/src/commands/snapshot/create.rs
leighmcculloch: I recommend doing this another way, which results in less need to repeat implementations manually:

Use `value_enum` along with `default_value_t` without passing a value:
```rust
    #[arg(short, long, value_enum, default_value_t)]
    format: Format,
```

On the enum, derive `Default` and specify the default with `#[default]` and derive `ValueEnum`:
```
#[derive(Default, Clone, ValueEnum)]
enum Format {
    #[default]
    Json,
    Xdr,
}
```

fnando: I don't get it. This is exactly what's already there. The difference is that to be able to set the default value, I had to implement the display trait.

https://github.com/stellar/stellar-cli/blob/main/cmd/soroban-cli/src/commands/snapshot/create.rs#L36-L40

leighmcculloch: You need the `value_enum` attribute on the arg itself.

leighmcculloch: And specify the default using `#[default]` on the type itself instead of in the arg.

fnando: Ah!

```diff
-    #[arg(long, default_value_t = Output::Json)]
+    #[arg(long, value_enum, default_value_t)]
```

PR #2310, opened 2025-12-02
Print all env vars prefixed with `STELLAR`.

fnando:
### What

```console
$ stellar env
STELLAR_ACCOUNT=default          # use
STELLAR_CONFIG_HOME=/tmp/stellar # env
STELLAR_NETWORK=testnet          # use

$ stellar env STELLAR_CONFIG_HOME
/tmp/stellar
```

### Why

So we can see all env vars, not just `STELLAR_ACCOUNT` and `STELLAR_NETWORK`.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
@@ -327,7 +327,7 @@ impl NetworkRunnable for Cmd {
             data::write(get_txn_resp.clone().try_into()?, &network.rpc_uri()?)?;
         }
 
-        if let Some(url) = utils::explorer_url_for_contract(&network, &contract_id) {
+        if let Some(url) = utils::lab_url_for_contract(&network, &contract_id) {
             print.linkln(url);
         }
 
--- a/cmd/soroban-cli/src/utils.rs
+++ b/cmd/soroban-cli/src/utils.rs
@@ -41,19 +41,24 @@ static EXPLORERS: phf::Map<&'static str, &'static str> = phf_map! {
     "Public Global Stellar Network ; September 2015" => "https://stellar.expert/explorer/public",
 };
 
+static LAB_CONTRACT_URLS: phf::Map<&'static str, &'static str> = phf_map! {
+    "Test SDF Network ; September 2015" => "https://lab.stellar.org/r/testnet/contract/{contract_id}",
+    "Public Global Stellar Network ; September 2015" => "https://lab.stellar.org/r/mainnet/contract/{contract_id}",
+};
+
 pub fn explorer_url_for_transaction(network: &Network, tx_hash: &str) -> Option<String> {
     EXPLORERS
         .get(&network.network_passphrase)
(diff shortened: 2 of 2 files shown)


cmd/soroban-cli/src/commands/env/mod.rs
```diff
@@ -29,14 +29,15 @@ impl Cmd {
...
         let mut vars: Vec<EnvVar> = Vec::new();
 
-        if let Some(v) = EnvVar::get("STELLAR_NETWORK") {
-            vars.push(v);
-        }
-
-        if let Some(v) = EnvVar::get("STELLAR_ACCOUNT") {
-            vars.push(v);
-        }
+        std::env::vars().for_each(|(key, _)| {
```
leighmcculloch: I think we should only print out env vars tha are actually used by the cli. Otherwise this makes it look like other envs not used are. 

Would it be reasonable to get a list of envs from clap? Or not realistic?

fnando: Fair point. I think it's just easier to enumerate those env vars (we already have a list that's used to support SOROBAN/STELLAR prefixes).

fnando: Turns out, the list of env vars was outdated. I updated in both places (here, and the compatibility layer with `SOROBAN_`).

cmd/soroban-cli/src/commands/env/mod.rs
```diff
@@ -28,15 +28,31 @@ impl Cmd {
...
         let print = Print::new(global_args.quiet);
         let mut vars: Vec<EnvVar> = Vec::new();
-
-        if let Some(v) = EnvVar::get("STELLAR_NETWORK") {
-            vars.push(v);
-        }
-
-        if let Some(v) = EnvVar::get("STELLAR_ACCOUNT") {
-            vars.push(v);
+        let supported = vec![
```
leighmcculloch:
```suggestion
        // Warning: This list needs to stay in sync with the list at cmd/soroban-cli/src/cli.rs.
        let supported = vec![
```

cmd/soroban-cli/src/cli.rs
```diff
@@ -20,17 +20,21 @@ pub async fn main() {
     // Map SOROBAN_ env vars to STELLAR_ env vars for backwards compatibility
     // with the soroban-cli prior to when the stellar-cli was released.
     let vars = &[
```
leighmcculloch:
```suggestion
    // Warning: This list needs to stay in sync with the list at cmd/soroban-cli/src/commands/env/mod.rs.
    let vars = &[
```

leighmcculloch (approved):
Would be easier to maintain this list if it was in one place, but I assume it's not for a reason. So I suggest adding the below comments so when we edit it we know we need to modify the other place too.

cmd/soroban-cli/src/commands/env/mod.rs
fnando: I centralized the list and they're now shared between both files.

PR #2313, opened 2025-12-03
Fix/secure store prompting

### LINKED ISSUE #2291: Investigate multiple prompting for secure storage key
Related to this comment https://github.com/stellar/stellar-cli/pull/2230#discussion_r2525171488

We have seen the keychain / secure store prompting the user for their password multiple times for one command on `stellar keys generate`, `stellar keys public-key` and `stellar contract invoke`. Though recently I've noticed that that is inconsistent. 

With older keys, I am still seeing the multiple prompting when fetching the public-key. But with newly created keys, that is not happening - I am not being prompted for permission to access the key at all.

elizabethengelman:
### What

closes #2291 

Caches seedphrase on the secure store entry, to prevent multiple prompts from the OS for permission to use the key from keychain. 

This required plumbing the caching the key that is retrieved in config::locator layer so that the cached seedphrase could be reused for each step of the command. 

### Why
We were seeing multiple prompts each time a secure store key was used. For example, for the command, we were being prompted 6 times from the keychain to approve the command's actions. 

```
cargo run contract invoke --source my-secure-store-key --network local --id hello-world -- inc
```

### Known limitations

On my mac, each time I access my keychain it prompts for permission twice - once for access to the keychain, and once for access to the specific key. I am not sure if it is possible to get this down to just one prompt.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,7 +38,7 @@ jobs:
         build-and-test-windows,
         publish-dry-run,
       ]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     steps:
       - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
         run: exit 1


deny.toml
```diff
@@ -35,6 +35,8 @@ ignore = [
     "RUSTSEC-2024-0388", # derivative is no longer maintained, but that has no known material impact on the this repo
     "RUSTSEC-2024-0436", # paste is no longer maintained
     "RUSTSEC-2025-0052", # async-std has been discontinued - used only in test dependencies
+    "RUSTSEC-2025-0134", # rustls-pemfile is archived; used indirectly via bollard which should be updated in the future
```
elizabethengelman: I am working on trying to upgrade bollard in another PR, which will hopefully allow for removing this ID from the list of ignores.

cmd/soroban-cli/src/signer/mod.rs
```diff
@@ -326,31 +332,3 @@ impl Lab {
         Err(Error::ReturningSignatureFromLab)
     }
 }
-
-pub struct SecureStoreEntry {
-    pub name: String,
-    pub hd_path: Option<usize>,
-}
-
-impl SecureStoreEntry {
```
elizabethengelman, line 335: I moved this over to secure_store_entry.rs for 2 reasons:
1. so we could merge this `SecureStoreEntry` functionality with what was in `secure_store.rs` - this just removed a layer that we didnt really need, and allows SecureStoreEntry access the keyring entry directly 
2. we want to feature flag some of the imports required for the keyring crate so that users can build the binary without the additional-libs feature and it seemed better to do that in a separate file

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -117,6 +121,10 @@ pub struct Args {
     /// Contains configuration files, aliases, and other persistent settings.
     #[arg(long, global = true, help_heading = HEADING_GLOBAL)]
     pub config_dir: Option<PathBuf>,
+
+    #[clap(skip)]
+    // This saves us from reading the same key from the file system more than once for one cmd
+    pub cached_keys: OnceLock<Arc<Mutex<CachedKeys>>>,
```
elizabethengelman, line 126: By caching the keys that are read from the fs here, we can save the secure story entry's seed phrase on the key for the entirity of the command

cmd/soroban-cli/src/config/secret.rs
```diff
@@ -55,13 +62,22 @@ pub struct Args {
...
+    SeedPhrase {
+        seed_phrase: String,
+    },
     Ledger,
-    SecureStore { entry_name: String },
+    SecureStore {
+        entry_name: String,
+        #[serde(skip)]
+        #[serde(default)]
+        cached_entry: Arc<OnceLock<SecureStoreEntry>>,
```
elizabethengelman, line 78: Here is where we are caching the specific SecureStoreEntry to reuse, so we don't get prompted multiple times per command

cmd/soroban-cli/src/signer/keyring.rs
```diff
@@ -17,30 +20,40 @@ pub enum Error {
...
 pub struct StellarEntry {
+    inner: Arc<StellarEntryInner>,
+}
+
+#[derive(Debug)]
+pub struct StellarEntryInner {
     name: String,
     #[cfg(feature = "additional-libs")]
     keyring: Entry,
+    cached_seed: Mutex<Option<SeedPhrase>>,
```
elizabethengelman, line 36: And here is where we are caching the seedphrase - this was helpful because most of the fns on StellarEntry use the `use_key` fn, which calls `self.get_seed_phrase()` - thus creating another set of prompts from the keychain. If we cache the seed here, we can just reuse it in `use_key`.

cmd/soroban-cli/src/signer/secure_store.rs
```diff
@@ -1,95 +0,0 @@
-use sep5::SeedPhrase;
```
elizabethengelman, line 1: this functionality was moved into secure_story_entry

cmd/soroban-cli/src/signer/secure_store_entry.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    #[error("Secure Store keys are not allowed: additional-libs feature must be enabled")]
+    FeatureNotEnabled,
+}
+
+#[derive(Debug, Clone)]
+pub struct SecureStoreEntry {
+    pub hd_path: Option<usize>,
+    #[cfg(feature = "additional-libs")]
+    pub entry: Arc<StellarEntry>,
```
elizabethengelman, line 44: Since this is wrapped in an Arc we are able to clone the same `StellarEntry` to reuse in the whole command

cmd/soroban-cli/src/signer/keyring.rs
```diff
@@ -50,14 +63,18 @@ impl StellarEntry {
...
         Ok(())
     }
 
     pub fn delete_seed_phrase(&self, print: &Print) -> Result<(), Error> {
-        match self.keyring.delete_credential() {
-            Ok(()) => Ok(()),
+        match self.inner.keyring.delete_credential() {
+            Ok(()) => {
+                // clear the cached seed
+                self.inner.cached_seed.lock().unwrap().take();
```
elizabethengelman: This sets the cached_seed back to none once the key is removed.

cmd/soroban-cli/src/signer/keyring.rs
```diff
@@ -50,14 +66,18 @@ impl StellarEntry {
...
     }
 
     pub fn delete_seed_phrase(&self, print: &Print) -> Result<(), Error> {
-        match self.keyring.delete_credential() {
-            Ok(()) => Ok(()),
+        match self.inner.keyring.delete_credential() {
+            Ok(()) => {
+                // clear the cached seed
+                self.inner.cached_seed.lock().map_err(|_| Error::MutexPoison)?.take();
+                Ok(())
```
elizabethengelman, line 83: This sets the cached_seed back to none once the key is removed.

elizabethengelman: I wonder if it is also worth doing this to clear out the cached_seed on drop as well

mootz12: I don't think its necessary, but probably smart to do for completeness.

cmd/soroban-cli/src/signer/keyring.rs
```diff
@@ -69,7 +93,15 @@ impl StellarEntry {
     }
 
     fn get_seed_phrase(&self) -> Result<SeedPhrase, Error> {
-        Ok(self.keyring.get_password()?.parse()?)
+        let mut guard = self.inner.cached_seed.lock().unwrap();
```
mootz12, line 96: check for MutexPoison?

mootz12 (changes requested):
Looks good overall. Few minor comments.

Couple overall thoughts:
1. We might want to consider just storing the pubkey for all accessed keys (on CachedKey or something)? This would be a pretty agnostic way to reduce key interactions.
2. If 1 is easy to do, we could consider just retaining the `StellarEntry` cache layer and not storing seed phrase in memory at all. Thus, we would be down to 3 interactions (entry access, pubkey derivation, and signing data).

cc @leighmcculloch @fnando

cmd/soroban-cli/src/signer/secure_store_entry.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    pub fn create_and_save(
+        _entry_name: &str,
+        _seed_phrase: &SeedPhrase,
+        _print: &Print,
+    ) -> Result<String, Error> {
+        Err(Error::FeatureNotEnabled)
+    }
+
+    pub fn sign_tx_data(_data: &[u8]) -> Result<Vec<u8>, Error> {
```
fnando, line 120:
```suggestion
    pub fn sign_tx_data(&self, _data: &[u8]) -> Result<Vec<u8>, Error> {
```

Issue #2282, opened 2025-11-11
Add a leeway arg for the fee


kalepail:
I'm getting a ton of timeouts with the CLI due to the fee not being sufficiently higher than the base so the tx just keeps getting punted to the next ledger but never being included (thanks $KALE 🙄)

```
% stellar contract invoke --id CAK6Z6KFMB3V2ENEJ7THVKXUYQ5EG7EL2TM5UQ2FLDXI37FS6DRIMIZH --source rich -- add_game --id CDB6IODG5BNNVILLJXBXYZVR7NP4HDO2NL7WALWIXGIDMA6VY4V75CEX
ℹ️  Signing transaction: d7787e6b6423cdad1944763384cab2a905b44add08ad248258a3c4d4cab011b3
❌ error: transaction submission timeout
% stellar contract invoke --id CAK6Z6KFMB3V2ENEJ7THVKXUYQ5EG7EL2TM5UQ2FLDXI37FS6DRIMIZH --source rich -- add_game --id CDB6IODG5BNNVILLJXBXYZVR7NP4HDO2NL7WALWIXGIDMA6VY4V75CEX
ℹ️  Signing transaction: d7787e6b6423cdad1944763384cab2a905b44add08ad248258a3c4d4cab011b3
❌ error: transaction submission timeout
```

I can hardcode a `--fee` but that's too brute force. I'd prefer to be able to make an additive fee to the simulated fee vs just overwriting it.

Another fix you should consider is using the fee endpoint of the RPC to be smarter about what fee to use. Use p50 or even higher to get past the base fee Launchtube is using.

tupui:
I don't know about that in terms of API. I think having it fail and tell me exactly why and what fees would be needed at least and how to adjust that would be clearer than having some blanket p90. I would always prefer to set a max absolute value.

PR #2321, opened 2025-12-10
Issues/2282 fee args

mootz12:
### What

Rename the `--fee` argument to `--inclusion-fee`, and deprecate the fee argument.

Introduces a new command `stellar fees` to replace `stellar feestats`. This has two subcommands:

1. stats, a copy/paste of feestats
```bash
$ stellar fees stats
```
2. use, adds the ability to store default inclusion fee settings based on raw stroops of feestats results

-> w/ stroops
```bash
$ stellar fees use --amount 1000                                           
ℹ️  The default inclusion fee is set to `1000`
```
-> w/ fee stats
```bash
$ stellar fees use --fee-metric p99 
ℹ️  The default inclusion fee is set to `200`
```
-> to remove
```bash
$ stellar fees unset
ℹ️  The default inclusion fee has been cleared
```

This change also includes some minor reorganization of arguments now that `fee` and `inclusion-fee` has been removed from `fee::Args`. `fee::Args` was renamed to `resources` to better represent the purpose of these args (edit transaction resources), and all fee / build / send related args were moved under `Options` instead of `Options (RPC)`.

### Why

This solves #2282 by allowing users to specify default inclusion fee args easily directly from the CLI. Also, it clears up ambiguity around the meaning of `--fee` and other arguments that modify `soroban_data`.

### Known limitations

There is a potentially weird edge case caused by how `--inclusion-fee` is given precendence over `--fee`. The behavior is detailed in a test here: https://github.com/stellar/stellar-cli/blob/issues/2282-fee-args/cmd/crates/soroban-test/tests/it/integration/fee_args.rs#L82-L94

TL;DR - is a user has some default config for `--inclusion-fee` but provides a `--fee` arg directly, the arg will be ignored in favor of the `--inclusion-fee`. This is mainly caused by us not knowing if clap sourced the `--inclusion-fee` value from an env variable or directly via an argument.

### DIFF
--- a/cmd/soroban-cli/src/resources.rs
+++ b/cmd/soroban-cli/src/resources.rs
@@ -20,7 +20,7 @@ pub struct Args {
     #[arg(long, help_heading = HEADING_RPC)]
     pub instruction_leeway: Option<u64>,
     /// Output the cost execution to stderr
-    #[arg(long = "cost", help_heading = HEADING_RPC)]
+    #[arg(long, help_heading = HEADING_RPC)]
     pub cost: bool,
 }
 


cmd/soroban-cli/src/soroban_data.rs
```diff
@@ -7,34 +7,31 @@ use crate::assembled::Assembled;
...
-    #[arg(long, default_value = "100", env = "STELLAR_FEE", help_heading = HEADING_RPC)]
-    pub fee: u32,
-    /// Output the cost execution to stderr
-    #[arg(long = "cost", help_heading = HEADING_RPC)]
-    pub cost: bool,
-    /// Number of instructions to simulate
+    /// Set the fee for smart contract resource consumption, in stroops. 1 stroop = 0.0000001 xlm. Overrides the simulated resource fee
+    #[arg(long, env = "STELLAR_RESOURCE_FEE", help_heading = HEADING_RPC)]
+    pub resource_fee: Option<u64>,
+    /// ⚠️ Deprecated, use `--instruction_leeway` to increase instructions. Number of instructions to allocate for the transaction
```
fnando:
```suggestion
    /// ⚠️ Deprecated, use `--instruction-leeway` to increase instructions. Number of instructions to allocate for the transaction
```

cmd/soroban-cli/src/resources.rs
```diff
@@ -7,34 +7,31 @@ use crate::assembled::Assembled;
...
 pub struct Args {
-    /// fee amount for transaction, in stroops. 1 stroop = 0.0000001 xlm
-    #[arg(long, default_value = "100", env = "STELLAR_FEE", help_heading = HEADING_RPC)]
-    pub fee: u32,
-    /// Output the cost execution to stderr
-    #[arg(long = "cost", help_heading = HEADING_RPC)]
-    pub cost: bool,
-    /// Number of instructions to simulate
+    /// Set the fee for smart contract resource consumption, in stroops. 1 stroop = 0.0000001 xlm. Overrides the simulated resource fee
+    #[arg(long, env = "STELLAR_RESOURCE_FEE", help_heading = HEADING_RPC)]
```
fnando, line 14: Should this and further args be using the global options scope, as per the pr's description?

>  send related args were moved under Options instead of Options (RPC)

cmd/soroban-cli/src/commands/fees/default.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl Cmd {
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let printer = Print::new(global_args.quiet);
+
+        if std::env::var("STELLAR_INCLUSION_FEE").is_ok()
+            && std::env::var("STELLAR_INCLUSION_FEE_SOURCE").is_err()
+        {
+            printer.warnln("Environment variable STELLAR_INCLUSION_FEE is set, which will override this default inclusion fee.");
```
fnando, line 74: 1. we need to add `INCLUSION_FEE` to cmd/soroban-cli/src/env_vars.rs
2. there used to be `STELLAR_FEE` as well. Is that still being used or did it get replaced by the inclusion fee env var? If it got replaced, we need to warn users in the deprecation process as well.

cmd/soroban-cli/src/soroban_data.rs
fnando, line 1: should we rename this file to something like `resource_args` instead (or even `fee_args`)? I feel soroban_data is not clear enough (also it has the name soroban, which we're trying to avoid)

cmd/soroban-cli/src/commands/keys/default.rs
```diff
@@ -21,8 +21,14 @@ pub struct Cmd {
 impl Cmd {
     pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
         let printer = Print::new(global_args.quiet);
-        let _ = self.config_locator.read_identity(&self.name)?;
 
+        if std::env::var("STELLAR_ACCOUNT").is_ok()
```
fnando, line 23: I don't think this is totally correct, right? We're not showing anything when users override the argument with `--source`/`--source-account`. I think we should remove these warnings from all commands and think of something else that can handle all cases (I personally feel this can get annoying, so maybe this should be a trace log instead?)

I think we should get users used to `stellar env`, which shows exactly the source for arguments (use vs env), and assume `--option` will override both.

cmd/soroban-cli/src/cli.rs
```diff
@@ -102,6 +102,9 @@ fn set_env_from_config() {
     if let Ok(config) = Config::new() {
         set_env_value_from_config("STELLAR_ACCOUNT", config.defaults.identity);
         set_env_value_from_config("STELLAR_NETWORK", config.defaults.network);
+        if let Some(inclusion_fee) = config.defaults.inclusion_fee {
+            set_env_value_from_config("STELLAR_INCLUSION_FEE", Some(inclusion_fee.to_string()));
+        }
```
fnando: I think we can change the `set_env_value_from_config` function to something like this instead:

```rust
fn set_env_value_from_config<T: std::fmt::Display>(name: &str, value: Option<T>) {
    let Some(value) = value else {
        return;
    };

    std::env::remove_var(format!("{name}_SOURCE"));

    if std::env::var(name).is_err() {
        std::env::set_var(name, value.to_string());
        std::env::set_var(format!("{name}_SOURCE"), "use");
    }
}
```

cmd/soroban-cli/src/soroban_data.rs
mootz12: I'm good with `resource_args`. I originally chose to match the XDR type name in the transaction these args directly impact. But, even if we choose to add the ability to mutate footprints in the future, `resource_args` makes sense.

cmd/soroban-cli/src/commands/keys/default.rs
mootz12: There is a "silent failure" case I was hoping to avoid here where users run a `stellar keys use` or similar command.

Let's imagine user has env with this:
```
export STELLAR_ACCOUNT=samwise
```

Some time passes and:
```
$ stellar keys use frodo
ℹ️  The default source account is set to `frodo`
$ stellar contract invoke ...
```
This will submit the transaction with `samwise`, even though the CLI made it clear the default is frodo.

A warning here during `use` feels appropriate that the env will override this default config.

cmd/soroban-cli/src/resources.rs
mootz12: I kept `soroban_data` / `resource_fee` under the RPC umbrella for now mainly because `instruction-leeway` is used as an RPC parameter to edit the simulation results.

I think a more thorough effort to reorganize help headings is probably warranted, so I left these unchanged for now.

The PR focused on ensuring send / build-only / sign / inclusion fee options were in the same place.

cmd/soroban-cli/src/commands/fees/default.rs
mootz12: 1. good catch
2. it still updates the deprecated `fee` value, so it will be used unless inclusion fee is specified.

Would a warning for usage also be warranted? We could do any `fee` usage, or limit to just env vars.

e.g. in `config.rs` `get_inclusion_fee` we could do:
```rust
    pub fn get_inclusion_fee(&self) -> Result<u32, Error> {
        if self.fee.is_some() {
            print_ln!("⚠️ Warning, `--fee [env: STELLAR_FEE]` is deprecated, use `--inclusion-fee [env: STELLAR_INCLUSION_FEE]` instead.
        }
        Ok(self.inclusion_fee.or(self.fee).unwrap_or(100))
    }
```

fnando: > Would a warning for usage also be warranted?

That's a good idea!

cmd/soroban-cli/src/commands/keys/default.rs
fnando: should we actually update the precedence for values? Seems like a good idea to me. Priority would be `switch options (if set) -> Defaults -> env vars`.

cmd/soroban-cli/src/resources.rs
```diff
@@ -6,34 +6,31 @@ use crate::assembled::Assembled;
...
     pub instructions: Option<u32>,
-    /// Allow this many extra instructions when budgeting resources during transaction simulation
+    /// Allow this many extra instructions when budgeting resources with transaction simulation
     #[arg(long, help_heading = HEADING_RPC)]
     pub instruction_leeway: Option<u64>,
-    /// Build the transaction and only write the base64 xdr to stdout
-    #[arg(long, help_heading = HEADING_RPC)]
-    pub build_only: bool,
+    /// Output the cost execution to stderr
+    #[arg(long = "cost", help_heading = HEADING_RPC)]
```
fnando: We can remove the string here, right?

```suggestion
    #[arg(long, help_heading = HEADING_RPC)]
```

fnando:
The relationship between `--inclusion-fee` and `--resource-fee` could be better documented; we should create a cookbook file that shows examples on when you'd use one or the other. This can be addressed in a separate pr.

cmd/soroban-cli/src/commands/keys/default.rs
fnando: btw, this could also be addressed in a separate pr.

fnando (approved):

cmd/soroban-cli/src/commands/keys/default.rs
mootz12: added issue to track, as it feels out of scope for this PR: https://github.com/stellar/stellar-cli/issues/2343

PR #2328, opened 2025-12-11
Update deny.toml to ignore rustls-pemfile error for now

elizabethengelman:
### What
Add RUSTSEC-2025-0134 to list of advisories to ignore. This advisory is regarding the rustls-pemfile crate, which has been archived. We are using rustls-pemfile indirectly via bollard.

### Why

So that cargo deny check will no longer fail.

### Known limitations

I am working on trying to upgrade bollard in another PR, which will hopefully allow for removing this ID from the list of ignores.


leighmcculloch:
The warning might go away if we upgade the crates upstream of the dependency.

elizabethengelman:
I'll close this PR, and hope we'll be able to fix this with upgrading bollard 🤞

PR #2332, opened 2025-12-16
Add `stellar keys use --clear` to clear the default identity.

fnando:
### What

```console
$ stellar keys use default
ℹ️ The default source account is set to `default`

$ stellar env
STELLAR_ACCOUNT=default # use
STELLAR_NETWORK=local   # use

$ stellar keys use --clear
ℹ️ The default source account has been cleared

$ stellar env
STELLAR_NETWORK=local # use
```

### Why

So users have a way to clear the default identity.

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-test/tests/it/config.rs
+++ b/cmd/crates/soroban-test/tests/it/config.rs
@@ -200,7 +200,7 @@ fn use_env() {
     sandbox
         .new_assert_cmd("keys")
         .env(
-            "SOROBAN_SECRET_KEY",
+            "STELLAR_SECRET_KEY",
             "SDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCQYFD",
         )
         .arg("add")
@@ -225,7 +225,7 @@ fn set_default_identity() {
     sandbox
         .new_assert_cmd("keys")
         .env(
-            "SOROBAN_SECRET_KEY",
+            "STELLAR_SECRET_KEY",
             "SC4ZPYELVR7S7EE7KZDZN3ETFTNQHHLTUL34NUAAWZG5OK2RGJ4V2U3Z",
         )
         .arg("add")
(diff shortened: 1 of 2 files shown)


leighmcculloch:
> ```
> stellar keys use --clear
> ```

The option on this keys use command doesn't feel natural or intuitive to me.

If we look at other commands for inspiration, k8s uses a separate command to clear the selected:
```
kubectl config use-context <context-name>
kubectl config unset current-context
```

What about something like:

```
stellar keys clear-use
```

And then the same with the network:

```
stellar network clear-use
```

cmd/soroban-cli/src/commands/keys/default.rs
```diff
@@ -4,13 +4,23 @@ use crate::{commands::global, config::locator, print::Print};
...
 #[derive(Debug, clap::Parser, Clone)]
 #[group(skip)]
 pub struct Cmd {
     /// Set the default network name.
-    pub name: String,
+    pub name: Option<String>,
+
+    /// Clear the default source account.
+    #[arg(long)]
+    pub clear: bool,
```
leighmcculloch, line 23: Should we omit the clear: bool, and just make it if name is None, then clear it?

cmd/soroban-cli/src/commands/keys/default.rs
```diff
@@ -19,14 +29,26 @@ pub struct Cmd {
...
-        let _ = self.config_locator.read_identity(&self.name)?;
 
-        self.config_locator.write_default_identity(&self.name)?;
+        if self.clear && self.name.is_some() {
+            return Err(Error::NameWithClear);
+        }
+
+        if !self.clear && self.name.is_none() {
+            return Err(Error::NameRequired);
+        }
```
leighmcculloch, line 39: If we omitted the bool, then we wouldn't need these condition checks.

leighmcculloch:
I don't feel strongly about it. An upside is that the --clear is discoverable inside the help of the command.

cmd/soroban-cli/src/commands/keys/default.rs
mootz12: Having clap manage this might make it easier:
```
clap::ArgGroup::new("Key")
    .required(true)
    .args(& ["name", "clear"]),
```

PR #2339, opened 2025-12-22
Update SOROBAN env vars to be STELLAR instead.

fnando:
### What

Update env var from SOROBAN to STELLAR.

### Why

We've been pushing STELLAR naming lately, so it makes sense.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/version.rs
+++ b/cmd/soroban-cli/src/commands/version.rs
@@ -39,8 +39,14 @@ pub fn git() -> &'static str {
 
 pub fn long() -> String {
     let xdr = stellar_xdr::VERSION;
+    let git_rev = if git().is_empty() {
+        String::new()
+    } else {
+        format!(" ({})", git())
+    };
+
     [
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/cli.rs
```diff
@@ -20,7 +20,11 @@ pub async fn main() {
     // Map SOROBAN_ env vars to STELLAR_ env vars for backwards compatibility
     // with the soroban-cli prior to when the stellar-cli was released.
     //
-    let vars = env_vars::unprefixed();
+    let mut vars = env_vars::unprefixed();
+
+    // Manually add SECRET_KEY so it doesn't leak on `stellar env`.
+    vars.push("SECRET_KEY");
```
mootz12, line 26: Is this needed? `stellar env` only prints env vars in the supported list.

fnando: It is. This is where the value is resolved from either SOROBAN_NAME or STELLAR_NAME

PR #2353, opened 2026-01-08
Implement the build side of spec shaking

### LINKED ISSUE #2030: `stellar contract build` should strip contract specs, removing types that are not referenced by functions or events
We have an issue in the soroban-sdk that any exported `contracttypes` defined in the SDK, or any library for that matter, that don't get used in the importing contract still end up in the contract spec.

The problem results in some awkward behaviour where we avoid exporting contracttypes in the SDK, even though doing so would be convenient and enable using contracttypes more frequently there.

Because of how proc-macros run in isolation and cannot coordinate in wasm builds in Rust (the ctor crate cannot be used like we use it in non-wasm builds in tests to coordinate), there's nothing the soroban-sdk can do to identify whether a contracttype that sometimes needs to be exported hasn't been used and can be excluded.

The stellar-cli contract build command though does have an opportunity to help here. It can inspect the contract specs and identify unused types that are using up space for now reason. It could either do this widely for any type that is unused, or do it only for types exported from the soroban sdk, because types can specify what library they came from.

This will require some care because it will require rewriting the wasm file after build, which is something the stellar-cli doesn't currently do. Currently the stellar-cli does modify the wasm, but only in an append fashion.

leighmcculloch:
### What

Filter unused type and event definitions from the `contractspecv0` WASM section during `stellar contract build` using markers embedded by the SDK.

**Changes:**
- Add `replace_custom_section()` in new `soroban_spec_tools::wasm` module for WASM custom section replacement
- Integrate spec shaking into the build process using `soroban_spec::shaking` from the SDK
- Gate filtering on the `rssdkfeat=experimental_spec_shaking_v2` contract meta entry
- Set `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2` env var during builds to signal SDK support
- Add comprehensive integration tests and test fixture

### Why

Contract WASM files can contain unused type and event definitions in the `contractspecv0` custom section that inflate binary size unnecessarily.

**The problem:** Any exported `contracttypes` or `contractevents` defined in the SDK or any library that don't get used in the importing contract still end up in the contract spec. This results in awkward behaviour where we avoid exporting contracttypes in the SDK, even though doing so would be convenient and enable using contracttypes more frequently there.

**Why the SDK can't fix this alone:** Because of how proc-macros run in isolation and cannot coordinate in WASM builds in Rust (the ctor crate cannot be used like we use it in non-wasm builds in tests), there's nothing the soroban-sdk can do to identify whether a contracttype that sometimes needs to be exported hasn't been used and can be excluded.

**Solution:** The SDK now embeds markers in the WASM data section for each type/event that is actually used. These markers survive dead code elimination only if the corresponding code is reachable. The CLI extracts these markers and filters the spec accordingly.

**Examples of unused entries that get filtered:**
- Error enums from imported libraries that are never used by any function
- Types that are defined but never referenced in reachable code
- Events that are defined but never published
- Transitive type dependencies that no longer apply after refactoring

Close #2030
Close https://github.com/stellar/rs-soroban-sdk/issues/1569
Close https://github.com/stellar/rs-soroban-sdk/issues/1330

### How

The filtering uses markers embedded by the SDK (via `soroban_spec::shaking`):

**Marker format (14 bytes):**
- Bytes 0-5: `SpEcV1` magic prefix
- Bytes 6-13: first 64 bits of SHA256 hash of the spec entry XDR bytes

**Filtering process:**
1. After cargo build completes, parse the built WASM
2. Check contract meta for `rssdkfeat=experimental_spec_shaking_v2` — if absent, skip filtering (SDK does not support spec shaking)
3. Scan the WASM data section for `SpEcV1` markers via `soroban_spec::shaking::find_all()`
4. Filter spec entries via `soroban_spec::shaking::filter()`:
   - Functions: always kept (they define the contract's API)
   - Types/events: kept only if a matching marker exists in the data section
5. Replace the `contractspecv0` custom section with the filtered version

**Why this works:** The SDK's proc-macros generate code that references a static marker when a type/event is used. If that code path survives dead code elimination (DCE), the marker remains in the data section. If the type/event is never used, DCE removes the code and the marker disappears.

**Build integration:** The CLI sets `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2=1` during builds to inform the SDK that the build system supports marker-based filtering.

**WASM manipulation:** The new `soroban_spec_tools::wasm::replace_custom_section()` function rebuilds the WASM, copying all sections verbatim except the target custom section, which is replaced with new content at the end. If multiple custom sections share the same name, all are consolidated into one.

### SDK Support Required

This feature requires SDK support for embedding markers. See: https://github.com/stellar/rs-soroban-sdk/pull/1571

### Try It Out

Install the modified `stellar-cli` that strips unused specs:
```
cargo install --locked --git https://github.com/stellar/stellar-cli --branch spec-clean stellar-cli
```

Import the modified `soroban-sdk` that marks used specs:
```
[dependencies]
soroban-sdk = { git = "https://github.com/stellar/rs-soroban-sdk", branch = "spec-markers", features = ["experimental_spec_shaking_v2"] }

[dev-dependencies]
soroban-sdk = { git = "https://github.com/stellar/rs-soroban-sdk", branch = "spec-markers", features = ["testutils", "experimental_spec_shaking_v2"] }
```

Build using the stellar-cli:
```
stellar contract build
```

Inspect the contract interface and it should only show types actually used at the boundary of the contract (inputs, outputs, and events):
```
stellar contract info interface --wasm target/wasm32v1-none/release/contract.wasm
```

### DIFF
--- a/.github/workflows/test-with-openzeppelin-contracts.yml
+++ b/.github/workflows/test-with-openzeppelin-contracts.yml
@@ -81,7 +81,6 @@ jobs:
       - name: Build ${{ matrix.dir }}
         run: stellar contract build --manifest-path ${{ matrix.dir }}/Cargo.toml
       - name: Find wasm
-        if: ${{ startsWith(matrix.dir, 'examples/') }}
         id: find-wasm
         run: |
           wasm=$(find . -name '*.wasm' -path '*/wasm32v1-none/release/*' -not -path '*/deps/*' | head -1)


cmd/crates/soroban-spec-tools/src/contract.rs
```diff
@@ -296,3 +377,169 @@ fn format_name(lib: &StringM<80>, name: &StringM<60>) -> String {
...
+        // This test checks that filtering a contract with events but no UDT references
+        // keeps the events (as per design) but would filter any unused UDTs
+
+        // Skip if the file doesn't exist (it's in a different repo)
+        let wasm_path = "/Users/leighmcculloch/Code/rs-soroban-sdk/tests/empty/out/test_empty.wasm";
+        if !std::path::Path::new(wasm_path).exists() {
+            return;
+        }
+
+        let wasm_bytes = std::fs::read(wasm_path).unwrap();
```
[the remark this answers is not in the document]
leighmcculloch: Fixed - removed this test entirely since it relied on a local path. The filtering logic is comprehensively covered by unit tests in `filter.rs`.

cmd/crates/soroban-spec-tools/src/contract.rs
```diff
@@ -296,3 +377,169 @@ fn format_name(lib: &StringM<80>, name: &StringM<60>) -> String {
...
+    #[test]
+    fn test_filter_on_custom_types_contract() {
+        // Test filtering on the custom_types contract wasm
+        let wasm_path = "/Users/leighmcculloch/Code/stellar-cli-spec-clean/target/wasm32v1-none/release/test_custom_types.wasm";
+        if !std::path::Path::new(wasm_path).exists() {
+            eprintln!("Skipping test: wasm file not found at {}", wasm_path);
+            return;
+        }
+
+        let wasm_bytes = std::fs::read(wasm_path).unwrap();
```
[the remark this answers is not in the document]
leighmcculloch: Fixed - removed this test entirely since it relied on a local path. The filtering logic is comprehensively covered by unit tests in `filter.rs`.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -256,6 +262,7 @@ impl Cmd {
                     .join(&file);
 
                 self.inject_meta(&target_file_path)?;
+                Self::filter_spec(&target_file_path)?;
```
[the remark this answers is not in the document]
leighmcculloch: The filtering logic is covered by 10 unit tests in `filter.rs` that test: removing unused types, keeping directly/transitively referenced types, preserving functions, handling error enums in Result types, circular references, and Vec/Map with UDTs. The build integration is straightforward - it calls `filter_spec()` after `inject_meta()`. Adding a full integration test would require setting up a fixture contract with unused types, which could be done as a follow-up if needed.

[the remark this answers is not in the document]
mootz12: > Adding a full integration test would require setting up a fixture contract with unused types, which could be done as a follow-up if needed.

This seems like a good idea to me.

cmd/crates/soroban-spec-tools/src/filter.rs
```diff
@@ -0,0 +1,787 @@
...
+};
+
+/// Magic bytes that identify a spec marker: `SpEc`
+pub const SPEC_MARKER_MAGIC: [u8; 4] = [b'S', b'p', b'E', b'c'];
+
+/// Length of the hash portion (truncated SHA256 - first 8 bytes / 64 bits).
+pub const SPEC_MARKER_HASH_LEN: usize = 8;
+
+/// Length of the marker: 4-byte prefix + 8-byte truncated SHA256 hash.
+pub const SPEC_MARKER_LEN: usize = 4 + SPEC_MARKER_HASH_LEN;
```
mootz12: (picky) Thoughts on moving these consts to the SDK so both the generators and cleaners can share? Maybe `soroban-spec`?

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -415,14 +426,104 @@ impl Cmd {
...
+
+        // Parse the spec from the wasm
+        let spec = Spec::new(&wasm_bytes)?;
+
+        // Check if the contract meta indicates spec shaking v2 is enabled.
+        // The SDK embeds a contractmeta entry with key "rssdkfeat" and value
+        // "experimental_spec_shaking_v2" when the feature is active.
+        let has_spec_shaking = spec.meta.iter().any(|entry| {
+            let ScMetaEntry::ScMetaV0(ScMetaV0 { key, val }) = entry;
+            key.to_utf8_string_lossy() == "rssdkfeat"
```
mootz12: [picky] should this come from the SDK (probably `soroban_spec::shaking`)? It might be useful if any tools want to write spec filters themselves.

If we intend to remove this meta entry on the p27 release, then it's OK.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -415,14 +426,104 @@ impl Cmd {
...
+
+        // Append new meta to existing meta
+        let new_meta = self.encoded_new_meta()?;
+        existing_meta.extend(new_meta);
+
+        let meta_section = CustomSection {
+            name: META_CUSTOM_SECTION_NAME.into(),
+            data: existing_meta.into(),
+        };
+        module.section(&meta_section);
```
mootz12: [picky] is it worth it to try and remove some code duplication with `soroban-spec-tools::wasm::replace_custom_section`?

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -273,6 +279,10 @@ impl Cmd {
                 cmd.env("CARGO_BUILD_RUSTFLAGS", rustflags);
             }
 
+            // Set env var to inform the SDK that this CLI supports spec
+            // optimization using markers.
+            cmd.env("SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2", "1");
```
mootz12, line 284: [picky] ditto on V2 comment from `rs-soroban-sdk` PR

mootz12 (commented):
Looks great! Left a couple of pedantic comments inline.

Leaving review as comment for now - will revise after doing some manual testing.

leighmcculloch:
Opened a couple related issues as a result of this work:
- https://github.com/stellar/stellar-cli/issues/2423
- https://github.com/stellar/stellar-cli/issues/2425

I've opened PRs for both, if those PRs merged before this one, this change will benefit from them.

cmd/soroban-cli/src/commands/contract/build.rs
leighmcculloch: 👍🏻 I've moved it there and added a constant.

cmd/soroban-cli/src/commands/contract/build.rs
leighmcculloch: I've reverted the changes in this area because they are unnecessary.

graydon (approved):

PR #2355, opened 2026-01-12
Apply inclusion fee and resource fee to simulated tx correctly

mootz12:
### What

We were overriding inclusion fee and not applying resource fee inputs to the simulated transaction

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -35,7 +35,7 @@ jobs:
             target: aarch64-unknown-linux-gnu
     runs-on: ${{ matrix.sys.os }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: ./.github/actions/build-binary
         with:
           target: ${{ matrix.sys.target }}
@@ -59,7 +59,7 @@ jobs:
             target: x86_64-apple-darwin
     runs-on: ${{ matrix.sys.os }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: ./.github/actions/build-binary
         with:
           target: ${{ matrix.sys.target }}
@@ -82,7 +82,7 @@ jobs:
             ext: .exe
     runs-on: ${{ matrix.sys.os }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: ./.github/actions/build-binary
         with:
           target: ${{ matrix.sys.target }}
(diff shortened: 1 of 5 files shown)


cmd/soroban-cli/src/assembled.rs
```diff
@@ -219,16 +217,26 @@ fn assemble(
         });
     }
 
+    let mut transaction_data = simulation.transaction_data()?;
     let min_resource_fee = if let Some(rf) = resource_fee {
-        tracing::trace!(
-            "setting resource fee to {rf} from {}",
+        if let Ok(rf_i64) = i64::try_from(rf) {
```
leighmcculloch: To me the bug seems like `resource_fee` is a u64, not that we're missing a conversion from `u64` to `i64` and need a fallback value if a user enters a value too large.

Does the issue stem from the `Args` specifying fees on the command line as `u64` instead of `i64`?
https://github.com/stellar/stellar-cli/blob/068052d607e8524f5baa56b2d31666babda5a7f5/cmd/soroban-cli/src/resources.rs#L12-L15

We should collect it as the correct type so that if a user enters an invalid value, the command line hard errors right there, forcing the developer to discover the issue rather than hide the mistake.

mootz12: No, this was just overlooked during the first PR adding this.

The real issue was twofold:
1. we didn't have the `transaction_data.resource_fee` getting set when a user supplied this arg
2. the tx.fee was being set based on the `min_tx_fee`, not the `raw.fee`

This being said, I can make this change with the additional check that it is positive to improve the error handling, as it is likely the right way to do it.

cmd/soroban-cli/src/assembled.rs
```diff
@@ -219,16 +217,26 @@ fn assemble(
...
-            simulation.min_resource_fee
-        );
-        rf
-    } else {
-        simulation.min_resource_fee
+    let mut transaction_data = simulation.transaction_data()?;
+    let min_resource_fee = match resource_fee {
+        Some(rf) => {
+            tracing::trace!(
+                "setting resource fee to {rf} from {}",
```
leighmcculloch: Suggest slightly different output. The to-from statement wasn't as clear to me what was going on.
```suggestion
                "overriding resource fee to {rf} (simulation suggested {})",
```

cmd/soroban-cli/src/assembled.rs
leighmcculloch: Yeah sorry I was pretty confused when I first posted this because I didn't realise that the type of the fee is different in different places 😕. Your original change was probably fine. Sorry!

leighmcculloch (approved):

PR #2357, opened 2026-01-14
Sign windows installer.

fnando:
### What

Sign Windows Installer when making a new release.

CleanShot 2026-01-14 at 17 22 00@2x

### Why

So we can get rid of the nasty warning on Windows. See https://github.com/stellar/ops/issues/4170 for reference.

### Known limitations

Given the limited number of signatures we can issue (1000), we're only signing actual releases.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -136,8 +136,7 @@ jobs:
 
       - name: Setup Software Trust Manager
         if:
-          github.event_name == 'release' || github.ref == 'refs/heads/main' ||
-          startsWith(github.ref, 'refs/heads/release/') || startsWith(github.head_ref, 'release/')
+          github.event_name == 'release' || startsWith(github.ref, 'refs/heads/release/') || startsWith(github.head_ref, 'release/')
         id: stm-setup
         uses: digicert/code-signing-software-trust-action@v1.0.1
         with:


.github/workflows/binaries.yml
```diff
@@ -104,25 +104,47 @@ jobs:
...
+      - name: Setup Software Trust Manager
+        if:
+          github.event_name == 'release' || github.ref == 'refs/heads/main' ||
+          startsWith(github.ref, 'refs/heads/release/') || startsWith(github.head_ref, 'release/')
+        id: stm-setup
+        uses: digicert/code-signing-software-trust-action@v1.0.1
+        with:
+          simple-signing-mode: true
+          keypair-alias: key_1412258126
+          input: ${{ env.STELLAR_CLI_INSTALLER }}
```
mootz12, line 145: it looks like these jobs will run on more than just a published release (e.g. the initial PR for a release and any commits it gets)

Does this impact the 1000 signing limit?

fnando: we may consume more than 1 signature per release, and that should be fine. We usually don't do that many extraneous commits on a release pr.

mootz12: Ah - got confused with the `github.ref == 'refs/heads/main'`. Maybe we remove this condition on all the filters for this workflow?

If `on: push` gets added to this workflow it could consume a lot of signatures.

fnando: yeah, let's be conservative and not sign `main` merges.

mootz12 (approved):

PR #2364, opened 2026-01-21
Remove network runnable abstraction.

fnando:
### What

Remove network runnable abstraction in favor of a more direct approach.

### Why

So it's easier to follow.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
@@ -205,7 +205,7 @@ impl Cmd {
                     ignore_checks: self.ignore_checks,
                     build_only: is_build,
                 }
-                .execute(config, quiet, !no_cache)
+                .execute(config, quiet, no_cache)
                 .await?
                 .into_result()
                 .expect("the value (hash) is expected because it should always be available since build-only is a shared parameter")
--- a/cmd/soroban-cli/src/commands/contract/invoke.rs
+++ b/cmd/soroban-cli/src/commands/contract/invoke.rs
@@ -248,6 +248,7 @@ impl Cmd {
         .await?)
     }
 
+    #[allow(clippy::too_many_lines)]
     pub async fn execute(
         &self,
         config: &config::Args,
@@ -272,11 +273,21 @@ impl Cmd {
 
         let client = network.rpc_client()?;
 
+        let global_args = global::Args {
+            locator: config.locator.clone(),
+            filter_logs: Vec::default(),
+            quiet,
+            verbose: false,
+            very_verbose: false,
+            list: false,
+            no_cache,
+        };
+
         let spec_entries = get_remote_contract_spec(
             &contract_id.0,
             &config.locator,
             &config.network,
-            None,
+            Some(&global_args),
             Some(config),
         )
         .await


cmd/soroban-cli/src/commands/container/start.rs
```diff
@@ -236,20 +236,22 @@ impl Runner {
...
-        } else {
-            String::new()
+        if self.network == Network::Local {
+            if let Some(limits) = self.args.limits.as_ref() {
+                return format!("--limits {limits}");
+            }
         }
+
+        String::new()
     }
```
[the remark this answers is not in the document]
fnando: needed to get rid of clippy warnings.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -283,7 +276,7 @@ impl NetworkRunnable for Cmd {
             &contract_id.0,
             &config.locator,
             &config.network,
-            global_args,
+            None,
```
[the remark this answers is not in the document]
fnando: Fixed in 1f9e228f57cd3a436cfc361e30441d2a5d237604

cmd/soroban-cli/src/commands/contract/deploy/asset.rs
```diff
@@ -98,7 +97,7 @@ pub struct Cmd {
 impl Cmd {
     pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
         let res = self
-            .run_against_rpc_server(Some(global_args), None)
+            .execute(&self.config, global_args.quiet, global_args.no_cache)
```
leighmcculloch, line 100: Do you think we should just parse GlobalArgs to all these fns so as not to have to add a new parameter each time? Or does GlobalArgs introduce too much of a burden in tests?

cmd/soroban-cli/src/commands/contract/deploy/wasm.rs
```diff
@@ -213,7 +205,7 @@ impl NetworkRunnable for Cmd {
                     ignore_checks: self.ignore_checks,
                     build_only: is_build,
                 }
-                .run_against_rpc_server(global_args, Some(config))
+                .execute(config, quiet, !no_cache)
```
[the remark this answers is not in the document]
fnando: Fixed in 1f9e228f57cd3a436cfc361e30441d2a5d237604

cmd/soroban-cli/src/commands/contract/read.rs
```diff
@@ -88,10 +87,27 @@ pub enum Error {
 
 impl Cmd {
     pub async fn run(&self) -> Result<(), Error> {
-        let entries = self.run_against_rpc_server(None, None).await?;
+        let entries = self
+            .execute(&config::Args {
```
leighmcculloch, line 91: Is the config args being manually built because we don't or can't use config::Args in the parameter definition for this command?

cmd/soroban-cli/src/commands/contract/deploy/asset.rs
fnando: Good question! I opted for explicit parameters over passing the full GlobalArgs for a few reasons:

- the `execute()` signature makes it explicit what the command actually uses
- tests only need to provide the specific values they care about (e.g., `cmd.execute(&config, false, false)`) rather than constructing a full GlobalArgs struct with many irrelevant fields
- commands shouldn't need to know about global CLI concerns they don't use (like verbose, filter_logs, etc.)

For now, most commands only need these two boolean flags from GlobalArgs, so it seemed like a good middle ground.

cmd/soroban-cli/src/commands/contract/read.rs
fnando: The read command's `run()` method doesn't have access to a full `config::Args`.

- the command stores `config: config::ArgsLocatorAndNetwork` (which only has locator and network fields)
- `execute()` needs a full `config::Args` (which includes source_account, sign_with, fee, inclusion_fee)

Since the read command doesn't actually use those extra fields, we construct a minimal config::Args with default values for the unused fields. This is a pattern present in  other commands that similarly don't need the full config (like events, fetch, etc.).

An alternative would be to change these commands to use config::Args everywhere, but that would require adding unused fields to their structs.

PR #2367, opened 2026-01-27
Error without overflow checks

mootz12:
### What

Update checks for Cargo during `stellar contract build` to throw errors instead of warnings. 

### Why

Prevent the CLI from being used to build contracts without `overflow-checks = true`

### Known limitations

None

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,12 +17,13 @@
         stellardev = {
           name = "stellar";
           src = ./.;
-          nativeBuildInputs = pkgs.lib.optionals (pkgs.stdenv.isDarwin) [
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -120,6 +120,9 @@ predicates = "3.1.2"
 testcontainers = { version = "0.26.0" }
 httpmock = "0.7.0"
 
+[profile.release]
+overflow-checks = true
```
mootz12, line 124: this is needed for cookbook tests to pass, as stellar CLI does not pass `profile.test-wasms` in, and I felt this was better than any additional profile arg to the cookbook

PR #2377, opened 2026-01-29
  Resolve contract error names from WASM spec during invocation

BlaineHeffron:
### What

  When a contract invocation fails, resolve the numeric error code (e.g. `Error(Contract, #1)`) to its human-readable name and documentation string from the contract's WASM spec, and display it directly next to the error code in the CLI output.

  Before:
  ❌ error: transaction simulation failed: HostError: Error(Contract, #1)

  Event log (newest first):
     0: [Diagnostic Event] ...

  After:
  ❌ error:  transaction simulation failed: HostError: Error(Contract, #1)
  NotFound: The requested resource was not found.

  Event log (newest first):
     0: [Diagnostic Event] ...

  ### Why

  Contract authors define error enums with meaningful names and documentation in their Soroban contracts (e.g. `NotFound = 1` with doc `"The requested resource was not found."`), but the CLI only showed the raw numeric code `#1`. This made debugging contract failures unnecessarily difficult — users had to manually cross-reference error codes against the contract source. The WASM spec already contains this information; the CLI just wasn't using it.

  This change handles all three error paths in contract invocation (first simulation, second simulation, and transaction submission), parses error codes from both the Display format (`Error(Contract, #N)`) and Debug format (`Contract(N)`), and searches all error enums in the spec rather than only the one named "Error" to support contracts with multiple error enums from dependencies.

  ### Known limitations

  - Error code extraction relies on string pattern matching against the RPC error message format (`Contract, #N` and `Contract(N)`). If the upstream error format changes, the parsing would need to be updated.
  - The resolved error name is inserted before the event log by splitting at the first blank line (`\n\n`) in the error string. If the RPC response format changes to not include a blank line separator, the resolved name would fall back to appearing at the end of the message.
  - Non-contract Soroban errors (Budget, Auth, etc.) are intentionally not resolved — only `ScError::Contract(u32)` codes are matched.


cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -452,6 +455,110 @@ enum ShouldSend {
...
+/// - `Error(Contract, #N)` from the Soroban host display format (simulation errors)
+/// - `Contract(N)` from Rust Debug format of `ScError::Contract(u32)` (submission errors)
+///
+/// The Display format uses the prefix `Contract, #` to distinguish contract errors
+/// from other Soroban error types (Budget, Auth, etc.) which also use `#N`.
+///
+/// The Debug format is used by `TransactionSubmissionFailed` errors which
+/// pretty-print (`{:#?}`) the `TransactionResult`, where the number may
+/// appear on a separate line with surrounding whitespace.
+fn extract_contract_error_code(msg: &str) -> Option<u32> {
```
leighmcculloch: We should be able to extract the error programatically, without string parsing, from the transaction result.

cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -222,6 +222,29 @@ impl Spec {
         Err(Error::MissingErrorCase(value))
     }
 
+    /// Search all error enums in the spec for a case matching the given value.
+    ///
+    /// Unlike `find_error_type`, which only looks at the error enum named
+    /// "Error", this method searches across all error enums in the contract
+    /// spec. This handles contracts that include multiple error enums from
+    /// dependencies.
+    pub fn find_error_type_any(
```
leighmcculloch: For functions that have in their signature a Result return value with a specific error type, I think this logic should probably limit its search to that specific error type.

It's always possible for a contract to panic with some other error inside that function. But if a contract function says in its spec those are the errors it will return, we should probably limit displaying those errors.

leighmcculloch (commented):
👏🏻 Very cool idea.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -374,7 +376,8 @@ impl Cmd {
 
         let res = client
             .send_transaction_polling(&config.sign(*txn, quiet).await?)
-            .await?;
+            .await
+            .map_err(|e| enhance_error(Error::Rpc(e), &spec))?;
```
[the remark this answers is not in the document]
BlaineHeffron: Added unit tests to more thoroughly cover this scenario although this is difficult to test directly in a testing environment.

cmd/soroban-cli/src/commands/contract/invoke.rs
BlaineHeffron: I changed it so now it will check the results from the simulated transaction response, then it will check the events if no error code was returned in the response. In my testing it generally seems to be returned in the events. The events need to be scanned in reverse order for the correct code since the outermost function call will be last in the event log array.

cmd/crates/soroban-spec-tools/src/lib.rs
BlaineHeffron: Now it only resolves codes that are declared in the calling function's spec, falling back to original behavior if it can't match the code there.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -452,6 +483,170 @@ enum ShouldSend {
...
+
+/// Attempt to resolve a contract error code from the simulation response and
+/// enhance the error message with the error name and documentation from the spec.
+fn enhance_simulation_error(
+    error_msg: &str,
+    sim_res: &SimulateTransactionResponse,
+    spec: &soroban_spec_tools::Spec,
+    function: &str,
+) -> Error {
+    let events = sim_res.events().ok();
```
mootz12, line 542: Sadly, diagnostic events are generally not available from mainnet RPC providers. We should attempt to find a workable solution from just the simulation result and the spec.

I think it's OK if we report contract errors as-is, e.g. if we get error 5, we attempt to match the spec to error 5. This is the interface provided by Soroban. I do still agree with @leighmcculloch that if the function defines a specific return error type, we should only match to error codes within the given error type.

We should just be able to do something along the lines of:

* if sim_res is error and we have an error message from the RPC
* check if `Error(Contract, #N)` exists before the first newline
    * Pretty sure this is always `HostError: Error(Contract, #N)`, but no reason to do string validation on more than we need to.
* extract `N` and run `find_error_for_function`

mootz12: We might want to consider improving the simulation error result from RPC. Maybe adding a new field that returns the error as a base-64 XDR string?

mootz12 (commented):
Thanks for the contribution - this is great! Will be a nice UX improvement once this is merged.

Comment on using diagnostic events which are not always available! Let me know if you have any questions.

cmd/soroban-cli/src/commands/contract/invoke.rs
BlaineHeffron: Good catch, thanks for pointing that out. I wasn’t aware diagnostic events are often unavailable from mainnet RPC providers.

I can parse `Error(Contract, #N)` from the simulation error string as suggested, and that works as a general fallback. The limitation is cross-contract traps: diagnostic events are currently the only reliable way we have to distinguish a non-try cross-contract panic (inner contract code) from an error defined by the invoked contract’s own enum when codes overlap.

So on mainnet providers that omit diagnostic events, that specific case can’t be disambiguated reliably. In that scenario, mapping `#N` to the invoked function’s error type is best-effort and can mislabel overlapping codes. Unfortunately I'm not aware of another means of doing so.  Like  you mention, perhaps an upstream change to the RPC server is the best option.

cmd/soroban-cli/src/commands/contract/invoke.rs
```diff
@@ -452,6 +483,183 @@ enum ShouldSend {
...
+        format!("{msg}\n{detail}")
+    }
+}
+
+fn extract_contract_error_from_error_msg(msg: &str) -> Option<u32> {
+    let first_line = msg.lines().next().unwrap_or(msg);
+    let marker = "Error(Contract, #";
+    let start = first_line.find(marker)? + marker.len();
+    let end = first_line[start..].find(')')?;
+    first_line[start..start + end].parse::<u32>().ok()
```
willemneal: `regex` is already a dep, which would make this more readable.

cmd/soroban-cli/src/commands/contract/invoke.rs
BlaineHeffron: We use diagnostic events for error extraction (`ScError::Contract(code)`) when present. String parsing is a fallback when mainnet RPC doesn't surface diagnostic events. mootz12's suggested base64-XDR field on the simulation error would give us a structured source on this path and let us drop the string parse.

PR #2383, opened 2026-02-11
Passthrough `--locked` on contract build

willemneal:
### What

Add the option to pass "--locked" to `cargo rustc`. This ensures that the deps found in the lockfile won't be updated.

### Why

Compiling contracts should be as deterministic as possible. While this won't ensure that the final wasm binary is the same it will ensure that the deps used remain the same.

### Known limitations

[TODO or N/A]


graydon (dismissed):

leighmcculloch (approved):
Looks good to me. I think when I last left comments there was a race condition from when I started looking at the PR to submitting my comment, and the changes, because originally this PR I thought made all builds locked.

It looks like the PR exposes and passes through the `--locked` option, which is standard on cargo build and so good to surface here 👍🏻.

Thanks!

_Tip: Opening issues first helps everyone better understand the problem being solved._

PR #2385, opened 2026-02-11
Fix BytesN parsing when using valid hex values.

### LINKED ISSUE #2245: contract invoke BytesN parsing fails to parse valid hex values
> why `"0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"` succeed while `"4142434400000000000000000000000000000000000000000000000000000000"` fails? Both are properly formatted 64-chars hex strings representing 32 bytes. 

 _Originally posted by @jsmaxi in [#2239](https://github.com/stellar/stellar-cli/issues/2239#issuecomment-3380828974)_

fnando:
### What

Properly parse BytesN when using valid hex values.

### Why

Fix #2245.

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -882,9 +882,6 @@ pub fn from_json_primitives(v: &Value, t: &ScType) -> Result<ScVal, Error> {
         (ScType::Address | ScType::MuxedAddress, Value::String(s)) => sc_address_from_json(s)?,
 
         // Bytes parsing
-        (ScType::BytesN(_), Value::Number(_)) => {
-            return Err(Error::InvalidValue(Some(t.clone())));
-        }
         (ScType::BytesN(bytes), Value::String(s)) => ScVal::Bytes(ScBytes({
             if bytes.n == 32 {
                 // Bytes might be a strkey, try parsing it as one. Contract devs should use the new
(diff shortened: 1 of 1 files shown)


cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -288,6 +288,9 @@ impl Spec {
                     ScType::U128 | ScType::I128 | ScType::U256 | ScType::I256 => {
                         Ok(Value::String(s.to_owned()))
                     }
+                    ScType::Bytes | ScType::BytesN(_) if matches!(val, Value::Number(_)) => {
+                        Ok(Value::String(s.to_owned()))
+                    }
```
leighmcculloch, line 293: Adding this means that the existing `(ScType::BytesN(_), Value::Number(n))` and `(ScType::Bytes, Value::Number(n))` arms in `from_json_primitives` (lines 86-88, 104-106) are unreachable for inputs that flowing through `from_string` -> `from_json` -> `from_json_primitives`. This new code converts numbers to strings earlier. They're still reachable if someone calls `from_json` directly with a pre-built `Value::Number`.

leighmcculloch (approved):
Looks good, one comment inline because this shadows a code path that becomes dead and wondering if there's an error relating to that.

cmd/crates/soroban-spec-tools/src/lib.rs
fnando: Good call. I've replaced both `Value::Number` arms with explicit `InvalidValue` errors instead of the `format!("{n}")` conversion. Hex data in `Bytes`/`BytesN` should always be passed as quoted strings (e.g., `["4554"]` not `[4554]`), which avoids the `f64` precision loss entirely. Added tests covering both the direct `from_json_primitives` path and the nested `Vec<Bytes>` path.

Does that make sense?

leighmcculloch: Won't quotes often be lost to shell expansion?

leighmcculloch: Oh you're meaning specifically in JSON, yes that makes sense.

PR #2400, opened 2026-02-18
Add signer plugins

mootz12:
### What

Add signer plugins that can be invoked during the transaction submission process to add arbitrary logic while signing authorization entries or transaction envelopes.

### Why

Writing custom code to support signing for non-standard account types is not well supported today. A few methods have emerged:

1. Write plugins that can be used in nested commands apply signatures ([stellar-cli-sign-auth-ed25519](https://github.com/stellar/soroban-examples/tree/main/multisig_1_of_n_account/stellar-cli-sign-auth-ed25519))
2. Write plugins that build, sign, and submit the transaction on their own ([smart-account-sign](https://github.com/brozorec/smart-account-sign/tree/main))

Having an integration point within the transaction submission process would simplify the process of writing, sharing, and using plugins made for signing. Further, these could be supported by future plugins as well, like protocol specific ones that help interact with protocols, e.g. [stellar-private-payments](https://github.com/tomerweller/stellar-private-payments/tree/feat/cli-plugin/cli).

Closes: https://github.com/stellar/stellar-cli/issues/2026

### How

_NOTE: CLI API under construction. The current version is just a proof of concept that writing the plugin is reasonable. Please provide feedback!_

Two new arguments are added to `sign_with.rs` to allow users to specify singing plugins for a given Address/Alias.

* `--sign-with-plugin` - Takes in a `plugin_name=Address` that ensures each time Address is required to sign, the given plugin with `plugin_name` is invoked, such that the plugin is an executable at `stellar-signer-{plugin_name}`. Multiple relationships can be created.
* `--plugin-arg` - Takes in a string of `plugin_name:arg_name=value`, such that each time `plugin_name` is called, in the arg json string provided, `{ arg_name: value }` is included.

Plugins need to support two different signing methods, each with a different JSON-string provided as an argument.

#### Sign Authorization Entries
* Input:
```json
{
  "mode": "sign_auth",
  "network_passphrase": "Test...",
  "address": "ScAddress as a Base64 XDR string",
  "nonce": 0,
  "signature_expiration_ledger": 100,
  "root_invocation": "SorobanAuthorizedInvocation as a Base64 XDR string",
  "args": {// based on --plugin-arg fields}
}
```
* Output: Base64 XDR string of the ScVal representing the credential signature for the auth entry

#### Sign Transaction
* Input:
```json
{
  "mode": "sign_tx",
  "network_passphrase": "Test...",
  "tx_env_xdr": "TransactionEnvelope as a Base64 XDR string",
  "tx_hash": "hex encoded tx_hash",
  "args": {// based on --plugin-arg fields}
}
```
* Output: JSON array of base64 XDR strings`representing DecoratedSignatures to add to the transaction envelope

### Try It Out

An example signer plugin was created for Stellar multisig accounts (`cmd/crates/stellar-multisig-plugin`). The same plugin was also created for comparison purposes that works with the existing CLI, using nested commands (`cmd/crates/stellar-old-multisig-plugin`).

**Setup CLI**

* Checkout this branch!

* Install Stellar CLI:

```bash
make
make install
```

* Install plugins:

```bash
make use-signing-plugin
make use-old-signing-plugin
```

*Setup test environment*

* Create two funded stellar accounts on a standalone or test network. 1 should be a multisig where the master key cannot submit a transaction (`alice` in this example). 1 should be a regular stellar account (`bob` in this example).

* Deploy the `hello-world` contract:
```
stellar contract --source bob upload --wasm ./target/wasm32v1-none/test-wasms/test_hello_world.wasm
``` 

**Custom Signing for Source Account**

* Using the singing plugin:
```bash
stellar contract invoke \ 
  --source alice 
  --sign-with-plugin multisig=alice \
  --plugin-arg multisig:signers=S..,S.. \ 
  --id C... \
  -- auth --addr alice --world Test1
```

* Using the nested commands:
```bash
stellar contract invoke --source alice --id C.. --build-only -- auth --addr alice --world TestOld \
  | stellar tx simulate \
  | stellar sign-auth-multisig \
      --signers S..,S.. \
      --signature-expiration-ledger 102345 \
      --network-passphrase 'Standalone Network ; February 2017' \
      --sign-tx \
  | stellar tx send
```

**Custom Signing for Authorization Entries**

* Using the singing plugin:
```bash
stellar contract invoke \ 
  --source samwise 
  --sign-with-plugin multisig=alice \
  --plugin-arg multisig:signers=S..,S.. \ 
  --id C... \
  -- auth --addr alice --world Test1
```

* Using the nested commands:
```bash
stellar contract invoke --source samwise --id C.. --build-only -- auth --addr alice --world TestOld \
  | stellar tx simulate \
  | stellar sign-auth-multisig \
      --signers S..,S.. \
      --signature-expiration-ledger 102345 \
      --network-passphrase 'Standalone Network ; February 2017' \     
  | stellar tx simulate \
  | stellar tx sign --sign-with-key samwise \
  | stellar tx send
```

### TODO

- [ ] Explore allowing "key/contract" config to store signing plugin data and arguments. That is, you would setup signing plugins directly on the key/contract itself, `stellar keys use-plugin alice --name multisig -- --signers S..,S..`, preventing the need for the weird plugin arg CLI API to be used. This is likely to touch a lot of code, so was omitted until we decide on a direction

- [ ] Support different arguments based if more than one address uses the same plugin. Currently, the same arguments are passed to all invocations of a singing plugin.

- [ ] General code polishing / testing

- [ ] Remove plugin examples from CLI repo


leighmcculloch:
> > Another thing to consider is: how should a plugin that requires asynchronous / detached signing work
> 
> The example doesn't show this, but the CLI will wait for the singing plugin process to terminate, and it does attach stderr so the plugin process can send prompts / URLs / etc to the user.

This flow will work for short lived asynchronous signing, but not asynchronous signing.

But, as long as the signing plugins can also be used with `stellar tx sign`, that should make it possible to use them asynchronously.

PR #2403, opened 2026-02-19
feat: allow stellar key secrets to be input via pipe

mootz12:
### What

Allow secret keys to be input via a pipe.

Example:
```
echo "S..." | stellar keys add my-new-key
```

### Why

Better support for CI environments to safely create keys with the CLI.

Closes: https://github.com/stellar/stellar-cli/issues/2388

### Known limitations

None

### DIFF
--- a/cmd/crates/soroban-test/tests/it/config.rs
+++ b/cmd/crates/soroban-test/tests/it/config.rs
@@ -1,4 +1,4 @@
-use crate::util::{add_key, add_test_id, SecretKind, DEFAULT_SEED_PHRASE};
+use crate::util::{add_key, add_test_id, SecretKind, GENERATED_SEED_PHRASE};
 use predicates::prelude::predicate;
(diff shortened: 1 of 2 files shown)


cmd/crates/soroban-test/tests/it/config.rs
```diff
@@ -218,6 +218,29 @@ fn use_env() {
         .stdout("SDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCQYFD\n");
 }
 
+#[test]
+fn add_key_from_stdin() {
```
leighmcculloch, line 222: Will this support an S key and a 12/24 word recovery phrase? Or only an S key? Could we add a test for the latter?

mootz12: yep, added tests for both.

PR #2404, opened 2026-02-20
Fix `snapshot create` unconditionally adding entries when match result was unused.

### LINKED ISSUE #2397: `snapshot create` match result appears unused

In `cmd/soroban-cli/src/commands/snapshot/create.rs`, the second `match` on `val.data` (around line 383) returns `true` or `false` from each arm, which looks like it was intended to filter whether an entry should be saved into the snapshot. However the result of the `match` is discarded (semicolon on the closing brace), and the `snapshot.ledger_entries.push(...)` and `count_saved += 1` that follow execute unconditionally for every entry:

https://github.com/stellar/stellar-cli/blob/a1c94559a7fb05b58b4697d82e918be44eab4df5/cmd/soroban-cli/src/commands/snapshot/create.rs#L383-L414

```rust
match &val.data {
    LedgerEntryData::ConfigSetting(ConfigSettingEntry::StateArchival(
        state_archival,
    )) => {
        // ...
        false
    }
    LedgerEntryData::ContractData(e) => {
        // ...
        keep
    }
    _ => false,
};  // <-- result discarded
snapshot
    .ledger_entries
    .push((Box::new(key), (Box::new(val), Some(u32::MAX))));
count_saved += 1;
```

Does this mean all matched entries (including `ConfigSetting` entries that return `false`) are being unconditionally included in the final snapshot? I think so.

The `match` result should be used to conditionally save the entry, if that was the original intent.

If all entries are intentionally being saved, the `false` return values in the match arms are dead code and should be cleaned up to avoid confusion.

fnando:
### What

Fix `snapshot create` unconditionally adding entries when match result was unused.

```console
$ grep -c config_setting /tmp/before.json
17

$ grep -c config_setting /tmp/after.json
0
```

### Why

A previous refactor broke filtering. Close #2397.

### Known limitations

N/A

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -274,7 +274,6 @@ Generate code client bindings for a contract
 
 ###### **Subcommands:**
 
-- `json` — Generate Json Bindings
 - `rust` — Generate Rust bindings
 - `typescript` — Generate a TypeScript / JavaScript package
 - `python` — Generate Python bindings
@@ -282,16 +281,6 @@ Generate code client bindings for a contract
 - `flutter` — Generate Flutter bindings
 - `swift` — Generate Swift bindings
 - `php` — Generate PHP bindings
-
-## `stellar contract bindings json`
-
-Generate Json Bindings
-
-**Usage:** `stellar contract bindings json --wasm <WASM>`
-
-###### **Options:**
-
-- `--wasm <WASM>` — Path to wasm binary
 
 ## `stellar contract bindings rust`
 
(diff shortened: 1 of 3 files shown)


FULL_HELP_DOCS.md
```diff
@@ -274,7 +274,6 @@ Generate code client bindings for a contract
 
 ###### **Subcommands:**
 
-- `json` — Generate Json Bindings
```
fnando, line 277: Reminiscent of #2396.

PR #2406, opened 2026-02-20
Return error when fetching ledger entry for trustline using native asset.

### LINKED ISSUE #2394: `stellar ledger entry fetch trustline --asset xlm` shouldn't be supported

The `stellar ledger entry fetch trustline --asset` flag accepts `"XLM"` (mapped to `TrustLineAsset::Native`), but trustlines do not hold the native asset — the native balance is held on the account entry itself. Passing `--asset xlm` silently returns an empty result instead of telling the user the input is invalid:

```
$ stellar ledger entry fetch trustline --asset xlm --account GB6RQWLXTCVLJZX6LACXVD4WFZHX4YGIPPBSH2YQQMZIMEKA4BM7QLOU
{"entries":[],"latestLedger":61271020}
```

https://github.com/stellar/stellar-cli/blob/a1c94559a7fb05b58b4697d82e918be44eab4df5/cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs#L60-L61

Return an error when `"XLM"` or `"native"` is passed as an `--asset` value, since trustlines cannot hold the native asset. For example:

```
error: native asset (XLM) does not have a trustline, use `stellar ledger entry fetch account` instead
```

Leave the behavior as-is, silently returning an empty result.

fnando:
### What

Return error when fetching ledger entry for trustline using native asset.

### Why

So it behaves like it should with native asset. Close #2394.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs
+++ b/cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs
@@ -38,6 +38,8 @@ pub enum Error {
     InvalidAsset(String),
     #[error("provided data name is invalid: {0}")]
     InvalidDataName(String),
+    #[error("native assets do not have trustlines")]
+    NativeAsset,
     #[error(transparent)]
     Locator(#[from] locator::Error),
     #[error(transparent)]
@@ -61,7 +63,9 @@ impl Cmd {
                 if parts.next().is_some() {
                     Err(Error::InvalidAsset(asset.clone()))?;
                 }
-                let source_bytes = Ed25519PublicKey::from_string(issuer).unwrap().0;
+                let source_bytes = Ed25519PublicKey::from_string(issuer)
+                    .map_err(|_| Error::InvalidAsset(asset.clone()))?
+                    .0;
                 let issuer = AccountId(PublicKey::PublicKeyTypeEd25519(Uint256(source_bytes)));
 
                 let asset_code: AssetCode = code
@@ -75,6 +79,8 @@ impl Cmd {
                         TrustLineAsset::CreditAlphanum12(AlphaNum12 { asset_code, issuer })
                     }
                 }
+            } else if matches!(asset.as_str(), "XLM" | "xlm" | "native") {
+                Err(Error::NativeAsset)?
             } else {
                 Err(Error::InvalidAsset(asset.clone()))?
             };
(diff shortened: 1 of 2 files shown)


cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs
```diff
@@ -43,6 +43,8 @@ pub enum Error {
     Locator(#[from] locator::Error),
     #[error(transparent)]
     TryFromSliceError(#[from] TryFromSliceError),
+    #[error("native asset (XLM) does not have a trustline, use `stellar ledger entry fetch account` instead")]
```
leighmcculloch: Nice error text.

cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs
```diff
@@ -57,8 +59,9 @@ impl Cmd {
     fn insert_asset_keys(&self, ledger_keys: &mut Vec<LedgerKey>) -> Result<(), Error> {
         let acc = self.muxed_account(&self.account)?;
         for asset in &self.asset {
-            let asset = if asset.eq_ignore_ascii_case("XLM") {
-                TrustLineAsset::Native
+            let asset = if asset.eq_ignore_ascii_case("XLM") || asset.eq_ignore_ascii_case("native")
```
leighmcculloch: I think we should remove this check. It's valid for someone to name an asset code xlm or native. We just need to require asset code and issuer to limit the lookup to issued assets that have an issuer.

cmd/soroban-cli/src/commands/ledger/entry/fetch/trustline.rs
```diff
@@ -54,9 +54,7 @@ impl Cmd {
     fn insert_asset_keys(&self, ledger_keys: &mut Vec<LedgerKey>) -> Result<(), Error> {
         let acc = self.muxed_account(&self.account)?;
         for asset in &self.asset {
-            let asset = if asset.eq_ignore_ascii_case("XLM") {
-                TrustLineAsset::Native
```
mootz12, line 58: [picky] Can we will still error here with the helpful "native assets don't have trustlines" error message you had before if they provide `XLM` `xlm` or `native` exactly?

leighmcculloch: If we do, without the issuer, because XLM, xlm, native, are valid codes to use with an issuer.

Issue #2390, opened 2026-02-12
Add `stellar network id` command to output the network ID


leighmcculloch:
### What problem does your feature solve?

Getting the network ID (the SHA-256 hash of the network passphrase) currently requires piping the passphrase through `sha256sum` manually. For example, in [quickstart](https://github.com/stellar/quickstart/blob/da350e3976c5f7454defde6cfa82f5d7df532760/start#L273):

```sh
NETWORK_ID=$(printf "$NETWORK_PASSPHRASE" | sha256sum | cut -f 1 -d " ")
```

This is a common operation needed in scripts and tooling that deal with network setup and transaction signing, and it would be convenient to have it built into the CLI.

Related: #2389

### What would you like to see?

Add a `stellar network id` command that outputs the network ID for a given network. For example:

```sh
$ stellar network id --network testnet
cee0302d59844d32bdca915c8203dd44b33fbb7edc19051ea37abedf28ecd472
```

### What alternatives are there?

- Compute it manually with shell tools: `printf "$NETWORK_PASSPHRASE" | sha256sum | cut -f 1 -d " "`. This works but requires knowing the passphrase and is easy to get wrong (e.g. trailing newline issues).

PR #2407, opened 2026-02-20
Add `stellar network id` to return the network id.

fnando:
### What

```console
# Defaults to testnet, as every other command
$ stellar network id
cee0302d59844d32bdca915c8203dd44b33fbb7edc19051ea37abedf28ecd472

# Respects default network
$ stellar network use local
ℹ️ The default network is set to `local`

$ stellar network id
baefd734b8d3e48472cff83912375fedbc7573701912fe308af730180f97d74a

# Explicit usage
$ stellar network id --network testnet
cee0302d59844d32bdca915c8203dd44b33fbb7edc19051ea37abedf28ecd472

# Formatted JSON
$ stellar network id --output json-formatted
{
  "id": "baefd734b8d3e48472cff83912375fedbc7573701912fe308af730180f97d74a",
  "network_passphrase": "Standalone Network ; February 2017"
}

# One-line JSON
$ stellar network id --output json
{"id":"baefd734b8d3e48472cff83912375fedbc7573701912fe308af730180f97d74a","network_passphrase":"Standalone Network ; February 2017"}
```

### Why

Implement #2390.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/config/network.rs
+++ b/cmd/soroban-cli/src/config/network.rs
@@ -126,8 +126,8 @@ impl Args {
         ) {
             (None, None, None) => Ok(DEFAULTS.get(DEFAULT_NETWORK_KEY).unwrap().1.to_string()),
             (_, Some(_), None) => Err(Error::MissingNetworkPassphrase),
-            (_, _, Some(network_passphrase)) => Ok(network_passphrase),
             (Some(network), None, None) => Ok(locator.read_network(network)?.network_passphrase),
+            (_, _, Some(network_passphrase)) => Ok(network_passphrase),
         }
     }
 }


cmd/soroban-cli/src/commands/network/id.rs
```diff
@@ -0,0 +1,71 @@
...
+use crate::commands::global;
+use crate::config;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Config(#[from] config::Error),
+
+    #[error(transparent)]
+    Network(#[from] config::network::Error),
```
mootz12, line 13: This includes prompting the user for `rpc-url`. To avoid confusion, could we just include args for network and network-passphrase?

The PR created to add `network root-account` does this: https://github.com/stellar/stellar-cli/pull/2402

mootz12 (changes requested):

cmd/crates/soroban-test/tests/it/integration/network.rs
```diff
@@ -1,6 +1,45 @@
...
+        .arg("testnet")
+        .assert()
+        .stdout(predicate::str::starts_with(
+            "cee0302d59844d32bdca915c8203dd44b33fbb7edc19051ea37abedf28ecd472",
+        ))
+        .success();
+}
+
+#[tokio::test]
+async fn network_id_json() {
```
mootz12, line 21: [picky] can these tests live in `soroban-test/test/config`? This way, they are run everytime `make test` is run, instead of only when `rpc-tests` are run.

fnando: it could, but then it wouldn't be integration test. I tend to favor those, but I guess this is no big deal as unit tests.

cmd/soroban-cli/src/commands/network/id.rs
fnando: You mean explicitly listing the props? That's not great. 🤔 

Maybe we should introduce a new `ArgsLocatorAndNetworkIdentifier` struct that doesn't require RPC url.

fnando: Actually, I think the proper fix should be making rpc url fully optional (right now it is `Option<String>`, but required by the `get` function), and validated only when it's actually used.

fnando: I added a reader for the passphrase, so you don't need the whole network object (which requires rpc url). See 12683ef3d795690367921e397d139ef00d2ac8cd.

```console
$ stellar network id --network-passphrase "Standalone Network ; February 2017"
baefd734b8d3e48472cff83912375fedbc7573701912fe308af730180f97d74a
```

cmd/crates/soroban-test/tests/it/integration/network.rs
fnando: given that it outputs to stdout, I don't see much benefit into moving these tests to unit tests. Wdyt?

PR #2415, opened 2026-02-25
Set more restrictive permission when creating the config directory and key files on unix.

fnando:
### What

As the title describes it.

### Why

So config dir and key files are visible only to the owner.

### Known limitations

Windows doesn't have something similar that's easily appliable (we may be able to use ACLs, but it's way more complex).

### DIFF
--- a/cmd/soroban-cli/src/config/locator.rs
+++ b/cmd/soroban-cli/src/config/locator.rs
@@ -4,7 +4,7 @@ use serde::de::DeserializeOwned;
 use std::{
     ffi::OsStr,
     fmt::Display,
-    fs::{self, create_dir_all, OpenOptions},
+    fs::{self, OpenOptions},
     io::{self, Write},
     path::{Path, PathBuf},
     str::FromStr,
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/config/locator.rs
```diff
@@ -495,6 +495,11 @@ impl Pwd for Args {
 pub fn ensure_directory(dir: PathBuf) -> Result<PathBuf, Error> {
     let parent = dir.parent().ok_or(Error::HomeDirNotFound)?;
     std::fs::create_dir_all(parent).map_err(|_| dir_creation_failed(parent))?;
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        std::fs::set_permissions(parent, std::fs::Permissions::from_mode(0o700)).ok();
```
leighmcculloch: The `.ok()` silently discards any error from `set_permissions`. In contrast, the file permission change in `KeyType::write` (line 571-578) properly propagates the error via `map_err`. This inconsistency is worth addressing — if the directory permission change fails, the key files inside it may be exposed.

At minimum, consider logging a warning. Ideally, propagate the error the same way the file permission change does:

```rust
std::fs::set_permissions(parent, std::fs::Permissions::from_mode(0o700))
    .map_err(|_| dir_creation_failed(parent))?;
```

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -495,6 +495,11 @@ impl Pwd for Args {
...
+    }
```
leighmcculloch, line 589: `create_dir_all` may create multiple nested directories (e.g., `~/.config/stellar/identity/`), but `set_permissions` is only called on `parent` (the immediate parent of `dir`). Any intermediate directories created above `parent` will retain the default umask permissions. If the intent is to protect the entire config tree, consider also setting permissions on ancestor directories (at least the stellar config root). If only the leaf directory matters, this is fine — but worth a comment clarifying the intent.

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -563,6 +568,16 @@ impl KeyType {
...
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            std::fs::set_permissions(&filepath, std::fs::Permissions::from_mode(0o600)).map_err(
+                |error| Error::IdCreationFailed {
+                    filepath: filepath.clone(),
+                    error,
+                },
+            )?;
+        }
```
leighmcculloch, line 681: The file is first written with the default umask, then permissions are tightened afterward. There is a brief window where the file exists with wider permissions. For key material, a more robust approach is to open the file with explicit mode using `std::os::unix::fs::OpenOptionsExt::mode()`:

```rust
use std::os::unix::fs::OpenOptionsExt;
std::fs::OpenOptions::new()
    .write(true)
    .create(true)
    .truncate(true)
    .mode(0o600)
    .open(&filepath)?;
```

This ensures the file is **never** world-readable, even momentarily, and will ensure if the file permission can't be set that it will be written with wider permissions.

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -495,6 +495,11 @@ impl Pwd for Args {
 pub fn ensure_directory(dir: PathBuf) -> Result<PathBuf, Error> {
     let parent = dir.parent().ok_or(Error::HomeDirNotFound)?;
     std::fs::create_dir_all(parent).map_err(|_| dir_creation_failed(parent))?;
```
leighmcculloch: Similar to the file race condition — `create_dir_all` creates directories with the default umask, then permissions are set after. For directories, you could use `DirBuilder` with mode:

```rust
#[cfg(unix)]
{
    use std::os::unix::fs::DirBuilderExt;
    std::fs::DirBuilder::new()
        .recursive(true)
        .mode(0o700)
        .create(parent)?;
}
```

This creates directories with the correct permissions from the start, avoiding the window of wider permissions.

leighmcculloch (approved):
Looks good to me. Some things to consider inline which would make it more secure.

PR #2426, opened 2026-03-04
Warn on missing and duplicate types in contract spec

### LINKED ISSUE #2425: Warn on missing types in contract spec during `contract build`

There are edge cases today where a type used in a contract function parameter or return value may not end up in the contract spec. Nothing prevents a user from using such a type in a contract fn, but the resulting spec will be incomplete — downstream tooling (bindings generators, explorers) won't be able to fully interpret the contract.

As the SDK evolves how it decides which specs to include (e.g. spec shaking in [stellar/rs-soroban-sdk#1672](https://github.com/stellar/rs-soroban-sdk/pull/1672)), this kind of check becomes even more valuable as a way to surface bugs early in the spec inclusion logic.

Add a verification step to `contract build` that:

1. Checks that every type referenced in contract function parameters or return values is defined elsewhere in the contract spec.
2. Checks that every type referenced by event topic or data fields is defined elsewhere in the contract spec.
3. Checks that any type in the spec that references other types can resolve those referenced types within the spec (transitive completeness).
4. Emits a **warning** for any missing type, but does **not** stop the build.
5. Ignores SDK built-in types that won't appear in the spec (e.g. the `checkauth` interface types and other types built into the SDK).

- Relying on downstream tooling to report missing types, but that shifts the problem later and makes root cause harder to trace.
- A standalone `contract verify` subcommand, but integrating into `build` catches issues at the earliest point in the workflow.

leighmcculloch:
### What
Add `Spec::verify()` to `soroban-spec-tools` that checks every UDT referenced in function signatures, event params, and type definitions is defined within the spec. Integrate the check into `contract build` to emit warnings after a successful build. Add a hidden `contract spec-verify` subcommand for standalone verification of a WASM file.

### Why
Contracts can compile and deploy with incomplete specs when referenced types are missing — for example, due to spec shaking in the SDK. Downstream tooling such as bindings generators and explorers silently fail when types are unresolvable. Surfacing this at build time gives developers an early, actionable signal.

This will also act as a backup safety to help alert us and developers if changes we make and experiment with wrt spec shaking in #2353 cause specs to disappear that were otherwise needed.

Close #2425

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1039,7 +1039,7 @@ checksum = "fe6d2e5af09e8c8ad56c969f2157a3d4238cebc7c55f0a517728c38f7b200f81"
 dependencies = [
  "serde",
  "termcolor",
- "unicode-width 0.2.1",
+ "unicode-width",
 ]
 
 [[package]]
@@ -1668,12 +1668,6 @@ checksum = "3d248bdd43ce613d87415282f69b9bb99d947d290b10962dd6c56233312c2ad5"
 dependencies = [
  "log",
 ]
-
-[[package]]
-name = "encode_unicode"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "34aa73646ffb006b8f5147f3dc182bd4bcb190227ce861fc4a4844bf8e3cb2c0"
 
 [[package]]
 name = "encoding_rs"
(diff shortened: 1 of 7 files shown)


cmd/soroban-cli/src/commands/contract/spec_verify.rs
```diff
@@ -0,0 +1,44 @@
+use clap::Parser;
+use std::fmt::Debug;
```
mootz12:
```suggestion
```

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -337,6 +337,16 @@ impl Cmd {
...
                 Self::print_build_summary(&print, &final_path, wasm_bytes, optimized_wasm_bytes);
+
+                // Verify spec references after build
+                if let Ok(final_wasm_bytes) = fs::read(&final_path) {
+                    if let Ok(spec) = soroban_spec_tools::Spec::from_wasm(&final_wasm_bytes) {
+                        for w in spec.verify() {
+                            print.warnln(format!("{}: {w}", p.name));
+                        }
+                    }
+                }
```
mootz12: Can we include this in `print_build_summary`? Would maybe make the output cleaner, as the `bytes` is already parsed and the warning would be placed alongside the rest of the build details.

PR #2428, opened 2026-03-05
feat: use serde to build package json and check npm package name

mootz12:
### What

Use serde to modify package.json file during typescript binding generation. And check that the npm package name is valid.

### Why

Fixes the issue where users can create invalid package.json files

### Known limitations

None

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
+++ b/cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
@@ -85,7 +85,7 @@ impl Cmd {
             .to_str()
             .ok_or_else(|| Error::NotUtf8(file_name.to_os_string()))?;
         soroban_spec_typescript::validate_npm_package_name(contract_name)
-            .map_err(|reason| Error::InvalidContractName(contract_name.to_string(), reason))?;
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
```diff
@@ -82,6 +84,8 @@ impl Cmd {
         let contract_name = &file_name
             .to_str()
             .ok_or_else(|| Error::NotUtf8(file_name.to_os_string()))?;
+        soroban_spec_typescript::validate_npm_package_name(contract_name)
```
fnando, line 87: validate_npm_package_name is called here and then again in `boilerplate.rs::init()`. Not a big deal, but we'll hit it twice.

cmd/crates/soroban-spec-typescript/src/boilerplate.rs
```diff
@@ -99,6 +112,33 @@ impl Project {
             })
     }
 
+    fn replace_package_json(&self, root: &Path, contract_name: &str) -> std::io::Result<()> {
```
fnando: This doesn't use `self`, so maybe we could extract it as a static function?

PR #2448, opened 2026-03-11
Publish docker image.

fnando:
### What

Update workflow so it publishes docker images on both workflow_dispatch and release events.

### Why

So we can have docker images that anyone could use.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -100,3 +100,16 @@ jobs:
           platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ env.DOCKER_TAGS }}
+
+      - name: Update Docker Hub description
+        run: |
+          TOKEN=$(curl -s -X POST "https://hub.docker.com/v2/users/login/" \
+            -H "Content-Type: application/json" \
+            -d '{"username":"${{ secrets.DOCKERHUB_USERNAME }}","password":"${{ secrets.DOCKERHUB_TOKEN }}"}' \
+            | jq -r .token)
+
+          jq -n --arg desc "$(cat ./docker/README.md)" '{"full_description": $desc}' | \
+            curl -s -X PATCH "https://hub.docker.com/v2/repositories/stellar/stellar-cli/" \
+              -H "Content-Type: application/json" \
+              -H "Authorization: Bearer ${TOKEN}" \
+              -d @-
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,53 @@
+# Stellar CLI
+
+Command-line interface for building and deploying smart contracts on the [Stellar](https://stellar.org) network.
+
+For full documentation, visit [https://developers.stellar.org](https://developers.stellar.org).
+
+## Quick Start
+
+```sh
+docker run --rm -it -v "$(pwd)":/source stellar/stellar-cli version
+```
+
+## Usage
+
+The container expects your project files to be mounted at `/source` (the default working directory). Any `stellar` subcommand can be passed directly:
+
+```sh
+# Build a contract
+docker run --rm -it -v "$(pwd)":/source stellar/stellar-cli contract build
+
+# Deploy a contract
+docker run --rm -it \
+  -v "$(pwd)":/source \
+  -e STELLAR_RPC_URL=https://soroban-testnet.stellar.org:443 \
+  -e STELLAR_NETWORK_PASSPHRASE="Test SDF Network ; September 2015" \
+  stellar/stellar-cli contract deploy --wasm target/wasm32v1-none/release/my_contract.wasm --source <key>
+```
+
+### Persisting Configuration
+
+Configuration and data are stored inside the container by default and lost when it exits. Mount volumes to keep them across runs:
+
+```sh
+docker run --rm -it \
+  -v "$(pwd)":/source \
+  -v stellar-config:/config \
+  -v stellar-data:/data \
+  stellar/stellar-cli contract build
+```
+
+## Container Paths
+
+| Path | Description |
+| --- | --- |
+| `/source` | Working directory where project files should be mounted. |
+| `/config` | CLI configuration directory (`STELLAR_CONFIG_HOME`). Mount a volume to persist networks and keys across runs. |
+| `/data` | CLI data directory (`STELLAR_DATA_HOME`). Mount a volume to persist cached contract specs and data. |
+
+## Image Tags
+
+- `latest` — most recent release.
+- `X.Y.Z` — specific release version (e.g. `22.6.0`).
+- `<commit-sha>` — build from a specific commit.


.github/workflows/docker.yml
```diff
@@ -13,4 +21,57 @@ jobs:
...
+
+      # Use the published tag as the ref; set both the versioned and `latest` tags.
+      - name: Setup vars (release)
+        if: github.event_name == 'release'
+        run: |
+          ref="${{ github.ref_name }}"
+          version="${ref#v}"
+          echo "STELLAR_CLI_REF=${ref}" >> $GITHUB_ENV
+          echo "DOCKER_TAGS=stellar/cli:${version},stellar/cli:latest" >> $GITHUB_ENV
+
```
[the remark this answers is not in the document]
fnando: Addressed

.github/workflows/docker.yml
```diff
@@ -13,4 +21,57 @@ jobs:
...
     steps:
-      - run: echo "Building and pushing Docker image..."
+      # Check out the repository at the specified ref so that git ls-remote can
+      # resolve branch names and git refs correctly. For workflow_dispatch, this
+      # uses the user-provided ref; for release events, this uses the published
+      # tag.
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref_name }}
+
```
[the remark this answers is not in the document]
fnando: Addressed

.github/workflows/docker.yml
```diff
@@ -13,4 +21,57 @@ jobs:
     runs-on: ubuntu-latest
     permissions: {}
     steps:
-      - run: echo "Building and pushing Docker image..."
+      # Check out the repository at the specified ref so that git ls-remote can
+      # resolve branch names and git refs correctly. For workflow_dispatch, this
+      # uses the user-provided ref; for release events, this uses the published
+      # tag.
```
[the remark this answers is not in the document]
fnando: Addressed

Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get update && \
 
 RUN git clone https://github.com/stellar/stellar-cli.git /tmp/stellar-cli && \
     cd /tmp/stellar-cli && \
-    git fetch origin "${STELLAR_CLI_REF}" && \
+    git fetch origin && \
     git checkout "${STELLAR_CLI_REF}" && \
     cargo install --locked --path cmd/stellar-cli && \
```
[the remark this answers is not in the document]
fnando: Addressed

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+
+      - name: Cache Rust build
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-
```
leighmcculloch: We should use the below action if we want to do caching of the Rust build because it does caching of the build artifacts using sccache which is better at handling growth over time. Caching the target directory directly will otherwise grow to infinity size over time.

- https://github.com/stellar/actions/tree/main/rust-cache

But, because this is part of a release, in general we should never cache release builds or any build that has exposure to release tokens because of the risk of cache poisoning as an injection vector. Cache poisoning on GitHub Actions was exploited recently ([ref](https://grith.ai/blog/clinejection-when-your-ai-tool-installs-another)).

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+
+      - name: Build binary
+        run: cargo build --package stellar-cli --release
+
+      - name: Copy binary for Docker context
+        run: cp target/release/stellar stellar
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
```
leighmcculloch: Can we replace the v4 with the commit sha that matches the latest v4 release? See https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+
+      - name: Log in to Docker Hub
+        if: github.event_name != 'push'
```
leighmcculloch: This step will run if someone changes the workflow to run on pull requests, so I think we should replace this condition with the event names we want to support publishing on rather than rely on the negative condition.

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+        with:
+          context: .
+          platforms: ${{ matrix.platform }}
+          push: false
+          cache-from: type=gha,scope=${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.arch }}
+
+      # Publish — push by digest so the merge job can assemble the manifest.
+      - name: Build and push by digest
+        if: github.event_name != 'push'
```
leighmcculloch: Same comment as https://github.com/stellar/stellar-cli/pull/2448/changes#r2921137155.

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          platforms: ${{ matrix.platform }}
+          outputs: type=image,name=stellar/cli,push-by-digest=true,name-canonical=true,push=true
+          cache-from: type=gha,scope=${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.arch }}
+
+      - name: Upload digest
+        if: github.event_name != 'push'
```
leighmcculloch: Same comment as https://github.com/stellar/stellar-cli/pull/2448/changes#r2921137155.

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+      # Compute Docker tags from the ref.
+      # - Version tag (e.g. v1.2.3): push versioned + latest tags.
+      # - Any other ref: push a tag for the resolved commit SHA.
+      - name: Compute tags
+        run: |
+          ref="${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref_name }}"
+
+          if [[ "$ref" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            version="${ref#v}"
+            echo "DOCKER_TAGS=stellar/cli:${version},stellar/cli:latest" >> $GITHUB_ENV
```
leighmcculloch: Nice!

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+          platforms: ${{ matrix.platform }}
+          push: false
+          cache-from: type=gha,scope=${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=${{ matrix.arch }}
+
+      # Publish — push by digest so the merge job can assemble the manifest.
+      - name: Build and push by digest
+        if: github.event_name != 'push'
+        id: build
+        uses: docker/build-push-action@v7
```
leighmcculloch: Same comment as https://github.com/stellar/stellar-cli/pull/2448/changes#r2921132567.

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
 name: Docker
 
 on:
+  push:
+    branches: [main]
+    paths: [Dockerfile]
```
leighmcculloch: The workflow appears to support publishing every commit to main. Is the intent for every commit to main to get a release and published docker image? Because this line means it will only run if the Dockerfile changes.

.github/workflows/docker.yml
```diff
@@ -2,15 +2,164 @@
...
+      - name: Upload digest artifact
+        if: github.event_name != 'push'
+        uses: actions/upload-artifact@v4
+        with:
+          name: digest-${{ matrix.arch }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
```
leighmcculloch: I haven't used this before, but if you're using the `docker/build-push-action` action, then I don't think you need this and should be able to push a multi-platform merged manifest using the action: https://docs.docker.com/build/ci/github-actions/manage-tags-labels/

We have a separate manifest step for some other repos like quickstart because those repos build process isn't compatible with the build-push-action unfortunately.

fnando: fixed!

.github/workflows/docker.yml
```diff
@@ -3,14 +3,106 @@ name: Docker
...
+      - name: Download binaries
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
+        with:
+          pattern: stellar-*
+          merge-multiple: false
+
+      - name: Stage binaries for Docker context
+        run: |
+          mv stellar-amd64/stellar stellar-amd64
+          mv stellar-arm64/stellar stellar-arm64
```
mootz12: This might fail given the dir and file are same name, I think `mv` assumes the user is trying to put it in the same dir. (at least it does on my mac)

We could just copy instead, and leave `stellar` binaries
```
cp stellar-amd64/stellar ./stellar-amd64
cp stellar-arm64/stellar ./stellar-arm64
```

fnando: I don't follow… what do you mean by "same dir"? In this case, `mv` will move that binary from `stellar-/*` into `$PWD/stellar-*`, which is correct. Mac behaves the same way (unix-based).

mootz12: I assumed the following structure:

```
$PWD
   /stellar-amd64
       -> stellar
   /stellar-arm64
       -> stellar
```

In this case, if `mv stellar-amd64/stellar stellar-amd64` is called, `stellar-amd64` matches the dir `stellar-amd64`, so `mv` assumes you want to put the file in the dir with the same name, which results in an error.

e.g.

```
~ % mkdir test-mv
~ % touch test-mv/my-file
~ % mv test-mv/my-file test-mv
mv: test-mv/my-file and test-mv/my-file are identical
```

mootz12: But if this works then feel free to resolve, I didn't run the action locally to test.

fnando: ohhhhhhhhhhhhhhhh… the problem is not same name… is having dir and file matching the name. Now I see!

PR #2461, opened 2026-03-23
Fix SkipWhitespace returning early EOF on whitespace-only chunks

### LINKED ISSUE #2431: xdr decode commands fails with XDR decode error when base64 input contains whitespace

25.1.0

### What did you do?

Piped a base64-encoded transaction envelope containing line breaks into `stellar tx hash` via stdin:

```bash
{
  printf 'AAAAAgAAAADwSUp9CwmVlPN40mKX1I1j39y6DmYc36QS1aK2x6eYVQ'
  sleep 0.5
  printf '\n\n\n\n'
  sleep 0.5
  printf 'AAAGQAEcMsAAAAAQAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAACTMkzn1TwPo8SIhnKvnyuv9K2/aWjpX9NTYfyiA7vXaAAAAAAX14QAAAAAAAAAAAA=='
} | stellar tx hash --network testnet
```

This is the same base64 envelope as the single-line version below, just with line breaks inserted in the middle (as is common with PEM-style formatting, multi-line terminal paste, or copying from formatted output).

The same issue affects `tx sign`, `tx send`, `tx simulate`, and `tx op add`.

The correct transaction hash, identical to the result of passing the envelope without line breaks:

```bash
printf 'AAAAAgAAAADwSUp9CwmVlPN40mKX1I1j39y6DmYc36QS1aK2x6eYVQAAAGQAEcMsAAAAAQAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAACTMkzn1TwPo8SIhnKvnyuv9K2/aWjpX9NTYfyiA7vXaAAAAAAX14QAAAAAAAAAAAA==' \
  | stellar tx hash --network testnet
```

### What did you see instead?

An XDR decode error. The input is silently truncated at the whitespace, so the decoder either fails to parse the incomplete XDR or produces a wrong result.

### Links

- https://github.com/stellar/rs-stellar-xdr/issues/505

gabrielrondon:
### What

Fix `SkipWhitespace::read()` signaling EOF when a read chunk contains only whitespace bytes. Add unit tests for the reader.

### Why

When base64 XDR input contains whitespace (e.g., newlines from terminal paste or multiline file), `read()` returns `Ok(0)` for whitespace-only chunks. The `Read` trait interprets `Ok(0)` as EOF, causing silent truncation and subsequent XDR decode errors.

The fix loops until at least one non-whitespace byte is found or the inner reader returns true EOF.

Affects all commands using `tx_envelope_from_input()`: `tx hash`, `tx sign`, `tx send`, `tx simulate`, `tx op add`.

### Known limitations

The upstream `stellar-xdr` crate has the same bug in its own `SkipWhitespace` (used by `tx decode`). This PR only fixes the `stellar-cli` copy. The upstream fix should be tracked separately in `stellar/rs-stellar-xdr`.

Closes #2431


gabrielrondon:
Hi — friendly ping on this one. CI is green after the reorder fix on Apr 3. Happy to address any feedback.

fnando (approved):
lgtm

PR #2471, opened 2026-04-07
Verify auth entries can't be used outside of the contract transaction

mootz12:
### What

The CLI currently relies on the RPC to check that no non-root auth is included in the simulation results. This PR adds an explicit check to verify that all non-source account authorizations are strictly tied to the root invocation. The check ensures that any signed authorization entries cannot be submitted outside the context of the invocation.

If any auth entry is detected that could be submitted outside of the transaction, all flagged entries are logged as follows:

```
$ contract invoke --source samwise --id CBKTFN4ATKHJMMKQFYWEUIPVE2FSRCQINHQNTID477G3247XLLE3KJAJ -- diff_auth_sub_auth --addr alice --val "Test" --subcall CCU2ER5XESXLT34MJTGY6GYMXHHJ3UB6NMTMLHSELT5WILZ7DNHJKOPV
⚠️  Authorization entries detected that could be submitted outside the context of this transaction:
  Auth Entry:
    Signer: GBXFXNDLV4LSWA4VB7YIL5GBD7BVNR22SGBTDKMO2SBZZHDXSKZYCP7L
    Function #0:
      Contract: CBKTFN4ATKHJMMKQFYWEUIPVE2FSRCQINHQNTID477G3247XLLE3KJAJ
      Fn: diff_auth_sub_auth
      Args:
        "1"
        "2"
      Function #0:
        Contract: CCU2ER5XESXLT34MJTGY6GYMXHHJ3UB6NMTMLHSELT5WILZ7DNHJKOPV
        Fn: do_auth
        Args:
          "GBXFXNDLV4LSWA4VB7YIL5GBD7BVNR22SGBTDKMO2SBZZHDXSKZYCP7L"
          Test
 
❌ error: Signing authorization entries that could be submitted outside the context of the transaction is not supported in the CLI    
 ```

If an auth entry is detected where the source account is used directly instead of with `Credentials::SourceAccount`, the error text changes to:

```
~ entry
❌ error: Source account detected with Address credentials. This requires an extra signature and is not expected.
```

### Why

The CLI eagerly signs authorization entries returned from the user specified RPC. If an unsafe auth entry was included, the user might unexpectedly sign for something they did not intend.

This adds a check to ensure that everything the CLI signs aligns with the transaction that got submitted.

#### `require_auth_with_args` for non-source accounts

The change also blocks contracts that use `require_auth_with_args` for non-source accounts. Consider a transaction where a user submits from a source account UserA, and also controls UserB `custom_token.transfer(from: UserB, to: UserC, amount: 100)`. If a malicious RPC intercepts this request, the could inject a bad auth such that `custom_token.transfer(from: UserC, to: UserB, amount: 100`. This technically satisfies the root auth requirement as the auth entry exists for the root contract and function invocation, but the arguments are different.

#### No bypass flag

This check currently cannot be bypassed. Most Stellar devtools (like the CLI and Stellar Lab) simulate in recording mode, which blocks non-root authorizations by default. If we add a flag to bypass this, we should also make the change to simulate in `AuthMode::RecordAllowNonRoot`. However, we should keep this consistent between Lab and CLI to prevent confusion, so this change was omitted for now.


mootz12:
Closing in favor of https://github.com/stellar/stellar-cli/pull/2530

PR #2479, opened 2026-04-14
Add decompressed size limits to snapshot downloads.

fnando:
### What

Add decompressed size limits to snapshot downloads.

### Why

To prevent unlimited decompression.

### Known limitations

N/A


mootz12 (approved):
LG2M

PR #2504, opened 2026-04-20
Reject STELLAR_SECRET_KEY when --secure-store is requested.

### LINKED ISSUE #2491: `keys add --secure-store` silently downgrades env-key imports to plaintext files
# 001: `keys add --secure-store` silently downgrades env-key imports to plaintext files

**Date**: 2026-04-17
**Severity**: High
**Impact**: secrets exposed in files
**Subsystem**: keys

## Summary

`stellar keys add <name> --secure-store` does not honor the user's secure-store request when `STELLAR_SECRET_KEY` is present. Instead, it silently imports the raw secret key from the environment and writes it into `identity/<name>.toml` as plaintext.

This is not just a documentation gap about secure-store only supporting seed phrases. Once the user explicitly asks for secure-store persistence, the incompatible env-key input should be rejected; silently downgrading to on-disk plaintext storage exposes signing material in a place the user specifically tried to avoid.

## Root Cause

`commands::keys::add::Cmd::read_secret()` checks `STELLAR_SECRET_KEY` before it checks `self.secrets.secure_store`. If the env var exists, the function returns `Secret::SecretKey { secret_key }` immediately and never enters the secure-store path.

`Cmd::run()` then converts that `Secret` into `Key::Secret(...)` and passes it to `locator::Args::write_key()`, which serializes the raw secret into TOML. The secure-store implementation only supports seed phrases, but the command handles that incompatibility by falling back to plaintext file storage instead of surfacing an error.

## Reproduction

Set `STELLAR_SECRET_KEY` to a valid Stellar secret and run `stellar keys add alice --secure-store --config-dir <tmpdir>`. The command succeeds without warning and creates `<tmpdir>/identity/alice.toml` containing `secret_key = "<env secret>"`.

This uses only supported CLI inputs under normal operation: `STELLAR_SECRET_KEY` is a recognized environment variable, and `--secure-store` is the documented persistence option for `keys add`.

## Affected Code

- `stellar-cli/cmd/soroban-cli/src/commands/keys/add.rs:71-79` — `run()` converts the resolved `Secret` into a `Key` and writes it with no storage-mode validation.
- `stellar-cli/cmd/soroban-cli/src/commands/keys/add.rs:84-100` — `read_secret()` gives `STELLAR_SECRET_KEY` precedence over `--secure-store`, bypassing the secure-store branch entirely.
- `stellar-cli/cmd/soroban-cli/src/config/secret.rs:98-101` — `Secret` is wrapped into `Key::Secret` without preserving the user's secure-store intent.
- `stellar-cli/cmd/soroban-cli/src/config/locator.rs:196-198` — `write_key()` routes the imported key to identity-file persistence.
- `stellar-cli/cmd/soroban-cli/src/config/locator.rs:666-705` — `KeyType::write()` TOML-serializes the secret to disk, producing a plaintext `secret_key = "S..."` file.

## PoC

- **Target test file**: `stellar-cli/poc/secure-store-env-bypass`
- **Test name**: `secure-store-env-bypass`
- **Test language**: bash
- **How to run**:
  1. Run `cargo build` from the repo root.
  2. Copy the script below to `poc/secure-store-env-bypass`.
  3. Run: `bash poc/secure-store-env-bypass`

### Test Body

```bash
#!/usr/bin/env bash
set -euo pipefail

STELLAR="$(cd /repo/stellar-cli && pwd)/target/debug/stellar"

TMPDIR=$(mktemp -d)
trap 'rm -rf "$TMPDIR"' EXIT

TEST_SECRET="SBO3J6Z3UNYRBG4Z33YLG3RNVLH4M2YDKFQRY66JQ6DCFXLLVKRXUZ54"
IDENTITY_NAME="poc_test_identity"

echo "=== PoC: --secure-store bypassed by STELLAR_SECRET_KEY ==="
echo ""
echo "Step 1: Run 'stellar keys add' with --secure-store AND STELLAR_SECRET_KEY env var"
echo "  Expected: secure-store should be used, OR command should reject conflicting inputs"
echo "  Actual:   env var silently overrides --secure-store, writing plaintext to disk"
echo ""

STELLAR_SECRET_KEY="$TEST_SECRET" "$STELLAR" keys add "$IDENTITY_NAME" \
  --secure-store \
  --config-dir "$TMPDIR" 2>&1 || true

echo ""
echo "Step 2: Check what was written to the identity file"
IDENTITY_FILE="$TMPDIR/identity/${IDENTITY_NAME}.toml"

if [ ! -f "$IDENTITY_FILE" ]; then
  echo "FAIL: Identity file not created at $IDENTITY_FILE"
  exit 1
fi

echo "  Identity file location: $IDENTITY_FILE"
echo "  Identity file contents:"
cat "$IDENTITY_FILE"
echo ""

if grep -q "secret_key" "$IDENTITY_FILE"; then
  echo "Step 3: FINDING CONFIRMED"
  echo "  The identity file contains 'secret_key' in plaintext!"
  echo "  Despite --secure-store being explicitly requested, the STELLAR_SECRET_KEY"
  echo "  environment variable caused the secret to be written as plaintext TOML."
  echo ""
  if grep -q "$TEST_SECRET" "$IDENTITY_FILE"; then
    echo "  VERIFIED: The exact secret key from the env var is present in the file."
    echo ""
    echo "RESULT: VULNERABILITY DEMONSTRATED"
    exit 0
  fi
fi

echo "RESULT: Could not demonstrate vulnerability"
exit 1
```

- **Expected**: `keys add --secure-store` should either store only a secure-store reference or reject incompatible `STELLAR_SECRET_KEY` input because secure store supports seed phrases only.
- **Actual**: `STELLAR_SECRET_KEY` silently wins, and the resulting identity file stores the raw secret key in plaintext TOML.

## Adversarial Review

1. Exercises claimed bug: YES — the PoC runs the real CLI command and inspects the identity file produced by that execution.
2. Realistic preconditions: YES — `STELLAR_SECRET_KEY` is a supported CLI environment variable, and inherited shell/CI environments make accidental conflicts with `--secure-store` realistic.
3. Bug vs by-design: BUG — the seed-phrase-only limitation explains why the command cannot satisfy the request with a secret key, but it does not justify silently downgrading an explicit secure-store request to plaintext persistence.
4. Final severity: High — the command writes live signing material into a plaintext file, matching the objective's "secrets exposed in files" high-severity bucket.
5. In scope: YES — the issue is concrete, reproducible, and requires no privileged machine access.
6. Test correctness: CORRECT — the script uses a valid secret key, invokes the supported CLI path, and checks for the exact env-supplied secret in the resulting file.
7. Alternative explanations: NONE — the plaintext file content is produced by the `STELLAR_SECRET_KEY` short-circuit in `read_secret()` followed by normal identity-file serialization.
8. Novelty: NOVEL

## Suggested Fix

When `--secure-store` is set, reject `STELLAR_SECRET_KEY` input with an explicit error such as "secure store only supports seed phrases; unset STELLAR_SECRET_KEY or provide a seed phrase instead." More generally, enforce the requested storage mode before calling `write_key()` so incompatible inputs fail closed rather than silently falling back to plaintext persistence.

fnando:
### What

Reject STELLAR_SECRET_KEY when --secure-store is requested.

### Why

Close https://github.com/stellar/stellar-cli/issues/2491

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/keys/add.rs
+++ b/cmd/soroban-cli/src/commands/keys/add.rs
@@ -33,6 +33,12 @@ pub enum Error {
 
     #[error("An identity with the name '{0}' already exists")]
     IdentityAlreadyExists(String),
+
+    #[error(
+        "--secure-store only supports seed phrases; \
+         unset STELLAR_SECRET_KEY or provide a seed phrase instead"
+    )]
+    SecureStoreRequiresSeedPhrase,
 }
 
 #[derive(Debug, clap::Parser, Clone)]
@@ -82,9 +88,15 @@ impl Cmd {
     }
 
     fn read_secret(&self, print: &Print) -> Result<Secret, Error> {
-        if let Ok(secret_key) = std::env::var("STELLAR_SECRET_KEY") {
-            Ok(Secret::SecretKey { secret_key })
-        } else if self.secrets.secure_store {
+        if self.secrets.secure_store {
+            if std::env::var("STELLAR_SECRET_KEY").is_ok() {
+                return Err(Error::SecureStoreRequiresSeedPhrase);
+            }
+        } else if let Ok(secret_key) = std::env::var("STELLAR_SECRET_KEY") {
+            return Ok(Secret::SecretKey { secret_key });
+        }
+
+        if self.secrets.secure_store {
             let prompt = "Type a 12/24 word seed phrase:";
             let secret_key = read_password(print, prompt)?;
             if secret_key.split_whitespace().count() < 24 {
--- a/cmd/crates/soroban-test/tests/it/config.rs
+++ b/cmd/crates/soroban-test/tests/it/config.rs
@@ -211,7 +211,10 @@ fn secure_store_rejects_env_secret_key() {
         .arg("alice")
         .arg("--secure-store")
         .assert()
-        .failure();
+        .failure()
+        .stderr(predicate::str::contains(
+            "--secure-store only supports seed phrases",
+        ));
 
     // The identity file must not exist — no plaintext fallback.
     assert!(
--- a/cmd/crates/soroban-test/tests/it/config.rs
+++ b/cmd/crates/soroban-test/tests/it/config.rs
@@ -196,6 +196,31 @@ fn seed_phrase() {
 }
 
 #[test]
+fn secure_store_rejects_env_secret_key() {
+    let sandbox = TestEnv::default();
+
+    // --secure-store with STELLAR_SECRET_KEY set should be rejected rather than
+    // silently writing the raw secret to a plaintext identity file.
+    sandbox
+        .new_assert_cmd("keys")
+        .env(
+            "STELLAR_SECRET_KEY",
+            "SDIY6AQQ75WMD4W46EYB7O6UYMHOCGQHLAQGQTKHDX4J2DYQCHVCQYFD",
+        )
+        .arg("add")
+        .arg("alice")
+        .arg("--secure-store")
+        .assert()
+        .failure();
+
+    // The identity file must not exist — no plaintext fallback.
+    assert!(
+        !sandbox.config_dir().join("identity/alice.toml").exists(),
+        "identity file should not be created when --secure-store is rejected"
+    );
+}
+
+#[test]
 fn use_env() {
     let sandbox = TestEnv::default();
 


cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -33,6 +33,12 @@ pub enum Error {
 
     #[error("An identity with the name '{0}' already exists")]
     IdentityAlreadyExists(String),
+
+    #[error(
+        "--secure-store only supports seed phrases; \
+         unset STELLAR_SECRET_KEY or provide a seed phrase instead"
+    )]
+    SecureStoreRequiresSeedPhrase,
```
mootz12, line 41: While not related to the issue, is there a reason we don't support S... keys in secure store?

If there is no reason, I can create a feature request for it.

mootz12 (approved):

cmd/soroban-cli/src/commands/keys/add.rs
fnando: It's probably an oversight. I don't see why S keys shouldn't be acceptable.

PR #2509, opened 2026-04-21
Escape control characters in message sign preview.

fnando:
### What

Escape control characters in message sign preview.

### Why

Close https://github.com/stellar/stellar-cli-internal/issues/8

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/message/sign.rs
+++ b/cmd/soroban-cli/src/commands/message/sign.rs
@@ -4,6 +4,8 @@ use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
 use clap::Parser;
 use sha2::{Digest, Sha256};
 
+use soroban_spec_tools::contract::sanitize;
+
 use crate::{
     commands::global,
     config::{locator, secret},
@@ -90,7 +92,7 @@ impl Cmd {
         } else {
             String::from_utf8_lossy(&message_bytes).to_string()
         };
-        print.infoln(format!("Message: {message_display}"));
+        print.infoln(format!("Message: {}", sanitize(&message_display)));
         println!("{signature_base64}");
         Ok(())
     }
--- a/cmd/soroban-cli/src/commands/message/sign.rs
+++ b/cmd/soroban-cli/src/commands/message/sign.rs
@@ -4,13 +4,12 @@ use base64::{engine::general_purpose::STANDARD as BASE64, Engine};
 use clap::Parser;
 use sha2::{Digest, Sha256};
 
-use soroban_spec_tools::contract::sanitize;
-
 use crate::{
     commands::global,
     config::{locator, secret},
     print::Print,
     signer::{self, Signer},
+    utils::strip_control_escapes,
 };
 
 use super::SEP53_PREFIX;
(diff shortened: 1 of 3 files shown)


cmd/soroban-cli/src/commands/message/sign.rs
```diff
@@ -90,7 +92,7 @@ impl Cmd {
         } else {
             String::from_utf8_lossy(&message_bytes).to_string()
         };
-        print.infoln(format!("Message: {message_display}"));
+        print.infoln(format!("Message: {}", sanitize(&message_display)));
```
leighmcculloch: It's odd to call the `soroban_spec_tools::sanitize` fn from a package that provides contract spec utilities when the message sign command has nothing to do with contracts or contract specs. The sanitize function is just a light wrap around the `escape_bytes` crate, so I think we can just use that here?


```suggestion
        let escaped = escape_bytes::escape(s.as_bytes()).map(char::from).collect();
        print.infoln(format!("Message: {}", sanitize(&message_display)));
```

Given the byte <> str shuffling it'd be nice if the escape-bytes crate had a fn that took on the responsibility of guaranteeing that you can convert bytes or a string to a utf8 safe string:
- https://github.com/stellar/escape-bytes/issues/9

leighmcculloch (approved):

cmd/soroban-cli/src/commands/message/sign.rs
mootz12: I think this package might be too strict, as it obfuscates the output string when using chars that take up more than 1 byte.

For example:
```bash
$ stellar message sign '¡Hola Mundo!' --sign-with-key SAKICEVQLYWGSOJS4WW7HZJWAHZVEEBS527LHK5V4MLJALYKICQCJXMW
ℹ️  Signer: GBXFXNDLV4LSWA4VB7YIL5GBD7BVNR22SGBTDKMO2SBZZHDXSKZYCP7L
ℹ️  Message: \xc2\xa1Hola Mundo!
sH3cuFuzVmfgXuMsJEH+iTryYo1zeQdNbHUFBZry3MRSNgEP2kg2SYuChVmoWMjgJdGyKQKK3tIzGPk3ft9WDw==
```

leighmcculloch: Yeah the `escape-bytes` package outputs printable ASCII only, not printable UTF8. Maybe not what you want here.

fnando: I'm going with this:

```rust
pub fn strip_control_escapes(s: &str) -> String {
    s.chars().filter(|c| !c.is_control()).collect()
}
```

cmd/soroban-cli/src/utils.rs
```diff
@@ -171,6 +171,10 @@ pub fn is_hex_string(s: &str) -> bool {
     s.chars().all(|s| s.is_ascii_hexdigit())
 }
 
+pub fn strip_control_escapes(s: &str) -> String {
+    s.chars().filter(|c| !c.is_control()).collect()
```
leighmcculloch, line 175: We should display the escaped control characters, not filter them. Filtering them could be exploitable. For example, someone could get someone to sign a message, but the preview has characters hidden.

fnando: Makes sense. I'll follow up this with another pr.

leighmcculloch: For example:

```suggestion
    let mut result = String::with_capacity(s.len());
    for c in s.chars() {
        if c.is_control() {
            let mut buf = [0u8; 4];
            for &byte in c.encode_utf8(&mut buf).as_bytes() {
                write!(result, "\\x{:02X}", byte).unwrap();
            }
        } else {
            result.push(c);
        }
    }
    result
```

PR #2513, opened 2026-04-22
Harden file permissions when migrating legacy config.

fnando:
### What

Harden file permissions when migrating legacy config.

### Why

Close https://github.com/stellar/stellar-cli-internal/issues/31

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/test_utils.rs
+++ b/cmd/soroban-cli/src/test_utils.rs
@@ -10,8 +10,19 @@ impl Default for CwdGuard {
 }
 
 impl CwdGuard {
+    #[must_use]
     pub fn new() -> Self {
         Self::default()
+    }
+}
+
+/// Runs `f` with the current working directory automatically restored on return or panic.
+pub fn with_cwd_guard<F: FnOnce()>(f: F) {
+    let saved = std::env::current_dir().unwrap();
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f));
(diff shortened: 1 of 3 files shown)


cmd/soroban-cli/src/test_utils.rs
```diff
@@ -1,3 +1,26 @@
...
+pub struct CwdGuard(std::path::PathBuf);
+
+impl Default for CwdGuard {
+    fn default() -> Self {
+        Self(std::env::current_dir().unwrap())
+    }
+}
+
+impl CwdGuard {
+    pub fn new() -> Self {
```
leighmcculloch, line 14: We should add a must_use here, so that if someone does `let _ = ...` they get a compiler warning, because _ will cause an immediate drop when _cwd will drop at end of scope.

```suggestion
    #[must_use]
    pub fn new() -> Self {
```

e.g. 
```
warning: unused `CwdGuard` that must be used
  --> src/lib.rs:10:5
   |
10 |     let _ = CwdGuard::new();
   |     ^^^^^^^^^^^^^^^^^^^^^^^
```

leighmcculloch: Alternatively a common pattern that is harder to use incorrectly is a closure, e.g.:
```rust
cwdguard(|| {
    // Do the test logic
}); // current dir will be reverted at end
```

leighmcculloch (approved):
The must_use comment above, otherwise lgtm.

PR #2525, opened 2026-04-27
Add container build backend and build verify command

### LINKED ISSUE #2506: Add `--docker` option and `stellar contract verify` for reproducible builds

Contract builds are not reproducible across different machines and environments. Different host OS, architecture, toolchain versions, and system libraries can produce different `.wasm` outputs from the same source code. This makes it impossible for third parties to independently verify that a deployed contract was built from a given source.

Related: https://github.com/stellar/stellar-protocol/discussions/1802 explored reproducing builds without Docker, but that approach hasn't proven reliable.

#### Building Reproducibly

Add a `--docker` option to `stellar contract build` that runs the `cargo rustc` build step inside a Docker container using an official Rust image pinned to a specific digest, always targeting `amd64` architecture to eliminate host architecture differences.

- `--docker` (no value): Pulls the latest official Rust image (`docker.io/library/rust:latest`) as `--platform linux/amd64`. After pulling, the resolved `@sha256:...` digest is recorded in contract metadata so the exact image can be replayed later.
- `--docker=<image>` (with value): Uses the specified Docker image, allowing users to reproduce past builds by passing the exact image with digest (e.g. `docker.io/library/rust@sha256:abcdef...`).
- Only the `cargo rustc` compilation runs inside Docker. The CLI's own post-processing of the `.wasm` file (optimization, metadata, etc.) continues to run on the host, outside Docker.
- The fully qualified Docker image name including its `@sha256:...` digest is recorded as a contract metadata entry, alongside the existing CLI version metadata.

Defaults are powerful and after sometime it might make sense to make this the default build option, meaning the docker build process is used automatically if a docker host is detected, and if the non-docker build process is used a warning is displayed indicating that reproducibility will be limited.

#### Verifying

**`stellar contract verify` command:**

Add a `stellar contract verify` command that automates verification of a contract build. It accepts the same wasm inputs as other contract commands:

- `--wasm <path>` — a local `.wasm` file
- `--wasm-hash <hash>` — fetch the wasm by its hash from the network
- `--contract-id <id>` — fetch the wasm for a deployed contract from the network

The command:
1. Reads the Docker image and CLI version from the contract's metadata.
2. Errors if the CLI version in metadata does not match the currently running `stellar-cli` version (the user must install the matching version first).
3. Rebuilds the contract from source using `stellar contract build --docker=<image>` with the Docker image from metadata.
4. Compares the rebuild output to the original `.wasm` and reports whether they match.

**Manual verification:**

The same verification can also be done manually:
1. Install the same version of `stellar-cli` (recorded in contract metadata).
2. Run `stellar contract build --docker=<image>` using the same Docker image (also recorded in contract metadata).
3. Compare the resulting `.wasm` output.

```mermaid
flowchart TD
    A[stellar contract build] --> B{--docker flag?}
    B -- No --> C[cargo rustc on host]
    B -- "--docker" --> D[Pull latest rust image\namd64, record resolved digest]
    B -- "--docker=image" --> E[Pull specified image\namd64]
    D --> F[cargo rustc inside Docker container]
    E --> F
    C --> G[CLI post-processes .wasm on host]
    F --> G
    G --> H[Embed metadata:\n- CLI version\n- Docker image + digest]
    H --> I[Final .wasm output]

    style D fill:#ccffcc,stroke:#00ff00
    style E fill:#ccffcc,stroke:#00ff00
    style F fill:#ccffcc,stroke:#00ff00
    style H fill:#ccffcc,stroke:#00ff00
```

**Legend:**
- Green = New/proposed components
- Default = Existing/unchanged components

**Why not include the CLI in the Docker image?**

Bundling the CLI into a custom Docker image would create a single blessed image that SDF must maintain and host. But instead using community images (like the official `rust` image) and allowing any image to be specified, builds stay decentralised. No single entity controls the build environment, and users aren't locked into a specific image provider.

- Pinning a specific image version in each CLI release, but this ties CLI releases to image updates and adds maintenance burden. Pulling latest and recording the digest is simpler and always gives the newest toolchain.
- Reproducing builds without Docker by pinning toolchain versions (explored in https://github.com/orgs/stellar/discussions/1802, but has not proven reliable due to host OS and system library differences).
- A fully self-contained Docker image with both Rust and the CLI baked in, but this centralises control and limits flexibility.

### Thanks

This idea is not unique or mine, a few people over time have mentioned ideas very much like this, and other blockchain ecosystems rely on docker to isolate and reproduce as well. Thanks to everyone who has raised this ideas at different times.

leighmcculloch:
### What

Add `--backend docker[=<image>]` to `stellar contract build` (and `deploy`/`upload`) that runs the entire build pipeline inside a container whose entrypoint is `stellar`. Add a `stellar contract build verify` subcommand that reads everything it needs from the wasm's metadata, rebuilds, and reports which (if any) rebuilt artifact is byte-identical to the original. Add a mainnet warning on `stellar contract deploy` when the wasm is missing the meta entries needed for independent verification.

### Why

Contract builds vary across host OS, architecture, and toolchain, preventing third parties from independently confirming a deployed contract was built from given source. Pinning the build to a docker image plus the rust toolchain version makes builds reproducible, recording the source repo + commit + per-package build options lets verifiers rebuild the exact same artifact, and the new `verify` subcommand automates the rebuild-and-compare check.

Closes #2506.

### How it works

Three parts: build-time recording, deploy-time warning, and verify-time reproduction.

#### Build

```
stellar contract build --backend local                          # default; host build
stellar contract build --backend docker                         # build inside docker.io/stellar/stellar-cli@sha256:...
stellar contract build --backend docker=stellar/stellar-cli:26.0.0
stellar contract build --backend docker=quay.io/myorg/myimage@sha256:...
```

For all backends (including `local`), the build:
- Detects whether the workspace is a clean git checkout. If clean and there's an `origin` remote, embeds `source_repo` (URL canonicalized to `https://…`), `source_rev` (full HEAD SHA), and per-package build options (`bldopt_manifest_path` relative to git root, `bldopt_package`, `bldopt_profile`, optional `bldopt_optimize`). The manifest path is auto-inserted whether or not `--manifest-path` was passed on the CLI.
- If the working tree has uncommitted changes, prints a warning and omits `source_repo` / `source_rev` / `bldopt_*`.
- If not a git repo, silently omits.

For `--backend docker`, additionally:
1. Resolves the requested image. Default is `docker.io/stellar/stellar-cli@sha256:cb2fc3116a6ace37a77ca6bb88afb4bee57fc746cd556a4373f2c3ee95d4e917` — pinned by digest so the recorded `bldimg` is reproducible from day one and we sidestep the longstanding Apple Silicon docker quirk where pulling a multi-arch tag with `--platform=linux/amd64` leaves `RepoDigests` empty after pull.
2. Pulls the image (skipping the pull if it's already locally present, since digest-pinned references are immutable).
3. Bind-mounts on the container:
   - `<git_root or workspace_root>` → `/source` (rw, source — also where cargo writes its target dir, shared with the host)
   - host `~/.cargo/registry` → `/usr/local/cargo/registry` (rw, cached crate downloads)
4. The container runs as the host uid:gid, so files written to the bind mount are readable/writable by the host user.
5. Overrides the image's entrypoint to invoke `stellar` directly, bypassing the official image's `entrypoint.sh` (which launches `dbus` + `gnome-keyring` and trips when running under a host UID with no `/etc/passwd` entry — see #2543). `contract build` doesn't use the keyring, so the wrapper is irrelevant here.
6. Runs `stellar contract build --manifest-path /source/<rel> --profile <p> --locked --meta bldimg=<digest> [forwarded args]` inside the container. The args use only flags that exist in published `stellar/stellar-cli` images today; no new flags are added, and `--backend local` is deliberately *not* passed (it's a flag added in this PR and isn't recognized by published images).
7. The in-container cli does cargo + meta injection + spec filtering + optional `wasm-opt` itself; the host only orchestrates and copies outputs to `--out-dir` if requested.

The wasm's `contractmetav0` custom section is populated with up to nine entries:

| key                     | value                                                | regex (validation) | injected by |
|-------------------------|------------------------------------------------------|--------------------|-------------|
| `cliver`                | `26.0.0#abc1234…` (CLI version + git rev)            | `^\d+\.\d+\.\d+(-[A-Za-z0-9.+-]+)?#([0-9a-f]{40}(-dirty)?)?$` | stellar-cli |
| `bldimg`                | `docker.io/stellar/stellar-cli@sha256:…`             | `^[^@\s]+@sha256:[0-9a-f]{64}$` | stellar-cli (this PR; only with `--backend docker`) |
| `rsver`                 | `1.83.0` (resolved rustc version)                    | `^\d+\.\d+\.\d+(-[A-Za-z0-9.+-]+)?$` | soroban-sdk |
| `source_repo`           | `https://github.com/user/repo` (clean repo's origin) | `^https?://\S+$` | stellar-cli (this PR) |
| `source_rev`            | full 40-char HEAD SHA                                | `^[0-9a-f]{40}$` | stellar-cli (this PR) |
| `bldopt_manifest_path`  | e.g. `contracts/foo/Cargo.toml` (relative to git)    | `^([^/\s]+/)*Cargo\.toml$` | stellar-cli (this PR) |
| `bldopt_package`        | cargo package name being built                       | `^[A-Za-z][A-Za-z0-9_-]*$` | stellar-cli (this PR) |
| `bldopt_profile`        | cargo profile (e.g. `release`)                       | `^[A-Za-z][A-Za-z0-9_-]*$` | stellar-cli (this PR) |
| `bldopt_optimize`       | `true` (only present when `--optimize` was used)     | `^true$` | stellar-cli (this PR) |

The presence of `bldimg` is what distinguishes a docker build from a local one — there's no separate `bldbkd` field. For full reproducibility from day one, pin to a specific image with `--backend docker=<name>@sha256:…` and commit before building.

`--backend` and `--docker-host` are also exposed on `stellar contract deploy` and `stellar contract upload` (which auto-build when no `--wasm` / `--wasm-hash` is given), so the same flags work end-to-end.

#### Deploy

`stellar contract deploy` against mainnet now warns when the wasm is missing any of `cliver`, `bldimg`, `rsver`, `source_repo`, `source_rev`, `bldopt_manifest_path`, `bldopt_package`, `bldopt_profile`:

```
⚠ the wasm being deployed is missing reproducibility meta entries: ["bldimg", "source_repo", "source_rev", "bldopt_manifest_path", "bldopt_package", "bldopt_profile"]. The deployed wasm may not be independently verifiable. To make it reproducible, build with `stellar contract build --backend docker` in a clean git repository.
```

The check is mainnet-only (matches network passphrase against `Public Global Stellar Network ; September 2015`); on testnet/futurenet/local the wasm deploys silently.

#### Verify

`verify` is a subcommand of `build` — it lives at `stellar contract build verify`, and works on multi-contract workspaces by rebuilding and finding the match.

```
stellar contract build verify --contract-id CXXX… --network mainnet
stellar contract build verify --wasm-hash <hash>  --network mainnet
stellar contract build verify --wasm contract.wasm
```

1. Fetches the original wasm (file path, hash, or contract id, same flags as `contract info`).
2. Reads `cliver`, `bldimg` (optional), `rsver`, and `bldopt_*` (optional, best-effort) from the wasm's meta. Missing `bldopt_*` entries trigger a warning rather than an error and the build falls back to its defaults — verify still runs, just with the caveat that the rebuild may not be reproducible.
3. Picks the rebuild backend from the meta:
   - `bldimg` present → `Backend::Docker { image: bldimg }`. The image's pinned digest pulls the same in-container cli that produced the original.
   - `bldimg` absent → `Backend::Local`. Best-effort rebuild on the host.
4. Forwards the wasm's `rsver` to the rebuild as `RUSTUP_TOOLCHAIN` (in-container) or `cargo +<rsver>` (local). For docker the toolchain inside the image is fixed by whoever built it; passing `RUSTUP_TOOLCHAIN` lets rustup-managed cargo switch toolchains if the image carries multiple ones.
5. Resolves `bldopt_manifest_path` against the cwd's git top-level (via `git rev-parse --show-toplevel`) so verify works from anywhere inside the checkout.
6. Hashes every rebuilt artifact and looks for a match against the original. Prints ✅ on match (with the matching crate's name); ⚠ + non-zero exit on mismatch (with each rebuilt artifact's name + hash).

The user is responsible for checking out the matching commit before running verify; verify rebuilds from the working tree. (`source_repo` and `source_rev` are embedded in meta to help users find the right commit, but verify itself doesn't clone — that would add a separate trust path.)

#### End-to-end example

```console
$ stellar contract build --backend docker
ℹ Pulling from stellar/stellar-cli
   Digest: sha256:cb2fc3116a6ace37a77ca6bb88afb4bee57fc746cd556a4373f2c3ee95d4e917
   Status: Image is up to date for stellar/stellar-cli@sha256:cb2fc3...
ℹ contract build --manifest-path /source/contracts/foo/Cargo.toml --profile release --locked --meta bldimg=docker.io/stellar/stellar-cli@sha256:cb2fc3...
   Compiling foo v…
    Finished `release` profile [optimized] target(s) in 1.09s
ℹ Build Summary:
   Wasm File: target/wasm32v1-none/release/foo.wasm (907 bytes)
   Wasm Hash: 9f86d081…
✅ Build Complete

$ stellar contract info meta --wasm target/wasm32v1-none/release/foo.wasm
cliver=26.0.0#abc1234
bldimg=docker.io/stellar/stellar-cli@sha256:cb2fc3...
rsver=1.83.0
source_repo=https://github.com/user/my-contract
source_rev=abc1234567890abcdef…
bldopt_manifest_path=contracts/foo/Cargo.toml
bldopt_package=foo
bldopt_profile=release

# Later, on a different machine, with the matching commit checked out:
$ stellar contract build verify --wasm-hash <hash> --network mainnet
ℹ Loading contract from network...
ℹ Loading meta from contract...
   Original wasm hash: 9f86d081…
   stellar-cli version: 26.0.0#abc1234
   rust version: 1.83.0
   Docker image: docker.io/stellar/stellar-cli@sha256:cb2fc3...
   Manifest path: contracts/foo/Cargo.toml
   Package: foo
   Profile: release
ℹ contract build --manifest-path /source/contracts/foo/Cargo.toml --profile release --locked --meta bldimg=docker.io/stellar/stellar-cli@sha256:cb2fc3...
   Compiling foo v…
✅ Build Complete
✅ Verified: rebuilt foo wasm matches 9f86d081…
```

The host CLI's version is irrelevant for verifying a docker-built wasm — whatever cli is in the image is what built (and rebuilds) the wasm.

### Notes

- **Communication with the daemon**: bollard's HTTP API over the docker socket (`/var/run/docker.sock`, or whatever `--docker-host` / `DOCKER_HOST` points at). Same `connect_to_docker` helper used by `stellar container start/stop/logs`, with the same Docker Desktop fallback (`$HOME/.docker/run/docker.sock`). No shell-out to the `docker` CLI. A podman socket exposing the Docker API would also work (untested).
- **Default image is digest-pinned**: `--backend docker` (no `=...`) defaults to `docker.io/stellar/stellar-cli@sha256:cb2fc3...`, *not* `stellar/stellar-cli:latest`. Recording a digest immediately makes builds reproducible day one and avoids the Apple Silicon `RepoDigests`-after-cross-platform-pull quirk. Bumping the default is a single-line const change in `build.rs` (see comments there for the recipe). Users who want a different image specify `--backend docker=...`.
- **Entrypoint override**: the official `stellar/stellar-cli` image's entrypoint runs `entrypoint.sh`, which launches `dbus` + `gnome-keyring`. That setup fails when the container runs as a host UID without an `/etc/passwd` entry — see #2543. We override the entrypoint to point straight at the `stellar` binary, which is fine because `contract build` doesn't touch the keyring.
- **Caching**: the bind-mount of host `~/.cargo/registry` lets the container reuse crate downloads the host already has.
- **Wasm target installation**: deferred to the image. The official image has `wasm32v1-none` pre-installed for its default toolchain; if `RUSTUP_TOOLCHAIN` selects a different one (verify on a wasm built with another rust version), the cli/cargo handle target installation themselves.
- **Toolchain pinning**: verify sets `RUSTUP_TOOLCHAIN=<rsver>` inside the container (and `cargo +<rsver>` for local rebuilds) so the rust version matches whatever the original build used.
- **Image fully-qualified**: `bldimg` is normalized to `<registry>/<path>@sha256:<digest>` (e.g. `stellar/stellar-cli:latest` → `docker.io/stellar/stellar-cli@sha256:…`) so verify can resolve it without relying on the local registry config.
- **Source URL canonicalization**: `source_repo` is normalized to `https://…` form (e.g. `git@github.com:user/repo.git` → `https://github.com/user/repo`).
- **Build options auto-recorded**: `bldopt_manifest_path` is recorded relative to the git repo root regardless of whether `--manifest-path` was passed on the CLI. Verify resolves it against the cwd's git top-level so the command works from anywhere inside the checkout.
- **No new in-container flags**: the host invokes `stellar contract build` inside the image with only flags that exist in published `stellar/stellar-cli` images today (`--manifest-path`, `--profile`, `--locked`, `--meta`, `--package`, `--features`, `--all-features`, `--no-default-features`, `--optimize`). `bldimg` is forwarded via `--meta bldimg=<digest>`, not a new flag.
- **No `bldbkd` field**: presence of `bldimg` is the only signal needed to distinguish a docker build from a local one.
- **Aborted container runs**: may leave a stopped container; clean with `docker container prune`.

### Performance/runtime caveats

Building inside an `amd64` container on a non-amd64 host (Apple Silicon, Linux/arm64) runs under emulation. For small contracts the difference is negligible; for workspaces with heavy dep trees the emulated build can be substantially slower than a native host build. Container runtimes that don't ship qemu/binfmt support won't run amd64 containers on arm64 hosts at all. See https://github.com/stellar/stellar-cli/issues/2506#issuecomment-4358015255.

### Related issues

- #2535 — normalize `cliver` rendering across install paths so it's a single shape.
- #2543 — the official `stellar/stellar-cli` image's `entrypoint.sh` fails under non-root UIDs, motivating the entrypoint override here.

### Status

This is an experiment in validating the ideas in #2506. May or may not be destined for merging — at this moment it's an experiment in validating the approach.


leighmcculloch:
Each meta field needs a very detailed specification of exactly what valid values are expected, and the format(s) those values can take.

For example, depending on how you install the `stellar-cli` today, the `cliver` field can contain any of the following formats:

Homebrew:
```
cliver: 26.0.0#
```

Cargo `crates.io` install:
```
cliver: 26.0.0#60f7458e7ecffddf2f2d91dc6d0d2db4fab03ecc
```

Cargo git install:
```
cliver: 26.0.0#v20.0.0-836-gfe07b3678833e07c43235a6caaeccff81e146856
```

Without a precise spec for each field, downstream tooling (verifiers, registries, indexers) can't reliably parse or validate these values.

leighmcculloch:
Currently only the `cargo build` step runs inside the container. The image is a stock `rust:latest` (with the wasm target installed at runtime) — everything the stellar-cli does after the build (meta injection, post-processing, etc.) runs on the host as part of the `stellar` binary, against the wasm output that comes back through the bind mount.

That works, and the host-side stellar-cli version is captured in the wasm via the `cliver` meta, so `verify` does have what it needs to detect a mismatch. But it still splits the build pipeline across two environments: a pinned, reproducible one (the image digest) and a host-resolved one (the user's installed `stellar`). To reproduce a build, a verifier needs both the right image *and* the right `stellar` on their host — the image digest alone isn't sufficient. To `verify` a build produced by a different `stellar` version than the one installed locally, the user has to install that other version first.

A way to get the best of both:

Embed a Dockerfile in the stellar-cli that, at build time, layers on top of whatever rust base image is requested:
1. `FROM` the user-chosen rust base image (still flexible — users can pin to any `rust:<sha>` they trust)
2. Add the wasm target
3. Install the requested stellar-cli version

Both the rust version and the stellar-cli version are specified at build (and verify) time, the image gets built locally, and the entire build pipeline runs inside it. The user on the host doesn't need to install a matching stellar-cli to verify a build produced by a different version — that version is installed into the image instead.

This avoids the supply-chain cost of us owning and publishing a bespoke stellar build image, keeps the rust base image flexible, and still lets the image digest capture the whole pipeline.

It also resolves the `cliver`-format issue raised in https://github.com/stellar/stellar-cli/pull/2525#issuecomment-4353511610 — since the embedded Dockerfile installs the stellar-cli exactly one way, `cliver` will be rendered exactly one way too, instead of varying by host install method (homebrew vs. crates.io vs. cargo-git).

It also keeps the door open for SDF or someone else to host prebuilt images later as a convenience — but we don't have to figure that out for the first iteration. The embedded-Dockerfile approach defers that decision without foreclosing it.

leighmcculloch:
Opened an issue about the cliver inconsistency here:
- https://github.com/stellar/stellar-cli/issues/2535

leighmcculloch:
Heads up on a devx tradeoff worth documenting: building inside an `amd64` container on Apple Silicon (or any non-amd64 host) runs under emulation — qemu via Docker Desktop / OrbStack — which is a real performance hit on large rust compilations.

Small contracts won't notice it. Workspaces with many contracts, or contracts with heavy dep trees, will see noticeably slower builds inside docker (sometimes a multiple of the native build time). This is a fundamental cost of pinning the build to a single arch for reproducibility, not something the `--backend docker[-all]` impl can avoid.

It's also not only a perf concern: users on container runtimes that don't ship qemu/binfmt support (some minimal Linux setups, certain rootless or stripped-down podman configs, etc.) won't be able to run amd64 containers on arm64 hosts at all — they'll see `exec format error` or similar and the build will simply fail.

Worth surfacing in docs so users aren't caught off guard, and worth keeping `--backend local` as the default for the dev inner loop.

leighmcculloch:
Another scaling tradeoff worth flagging between the two backends:

`--backend docker` just pulls a stock rust image and lets the host CLI do post-processing — no per-(rust, cli) image to build.

`--backend docker-all` needs an image with both rust *and* a specific stellar-cli rev installed. Ideally that image would be prebuilt for known (rust, cli) combinations and just pulled from a registry — same speed as the docker backend. As implemented in this PR it's layered locally instead, which means for each new (rust, cli) pair the user encounters, the image has to be built via a full `cargo install stellar-cli` (a real cli compile, minutes). After that, docker's layer cache hits on the same pair across builds.

In practice this is an edge case — most users build with a small set of (rust, cli) combinations and the layered image gets cached the first time.

leighmcculloch:
Opened an issue about dbus creating problems with using the image for the build for the verification step:
- https://github.com/stellar/stellar-cli/issues/2543

fnando:
@leighmcculloch I just tried this, but I'm getting a warning, even though there are no unstaged files.

```console
$ git status
On branch main
nothing to commit, working tree clean

$ stellar contract build --backend docker
⚠️ git working tree has uncommitted changes; source_repo/source_rev/bldopt_* not embedded in contract metadata. Commit changes for a reproducible build.
```

cmd/soroban-cli/src/commands/contract/verify.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+        if let Some(p) = &bldopt_profile {
+            print.blankln(format!("Profile: {p}"));
+        }
+        if bldopt_optimize {
+            print.blankln("Optimize: true");
+        }
+
+        let backend = match bldimg {
+            Some(image) => build::Backend::Docker { image },
```
mootz12: `bldimg` needs some form of allowlist, e.g. `docker.io/stellar/stellar-cli@sha256:*`, otherwise I can inject a docker image that bypasses verification

cmd/soroban-cli/src/commands/contract/verify.rs
```diff
@@ -0,0 +1,221 @@
...
+            if p.is_absolute() {
+                p
+            } else if let Some(root) = git_top_level() {
+                root.join(p)
+            } else {
+                p
+            }
+        });
+
+        let build_cmd = build::Cmd {
```
mootz12, line 120: Do we need feature flags here? While not extremely common, I have used them in the past to reduce code duplication between similar contracts. Ref -> https://github.com/script3/soroban-governor/tree/main/contracts/votes

cmd/soroban-cli/src/commands/contract/build_docker.rs
```diff
@@ -0,0 +1,500 @@
...
+    //   image's default (typical at verify time), the image's pre-installed
+    //   `wasm32v1-none` target is associated with the *other* toolchain,
+    //   not the selected one — `cargo build --target=wasm32v1-none` would
+    //   fail. `rustup target add` is idempotent (and quick, when the target
+    //   is already present) so always running it is safe.
+    // - The official `stellar/stellar-cli` image's stock entrypoint is a
+    //   wrapper script that launches dbus + gnome-keyring before exec-ing
+    //   `stellar`; that setup is irrelevant for `contract build` and dbus
+    //   refuses to start when the container runs as a host UID with no
+    //   `/etc/passwd` entry. Skipping it keeps the host UID mapping intact.
```
mootz12: This was cleaned up in a recent PR. Does this simplify anything?

cmd/soroban-cli/src/commands/contract/build_docker.rs
```diff
@@ -0,0 +1,500 @@
...
+        image: Some(image.to_string()),
+        entrypoint: Some(entrypoint),
+        cmd: Some(argv),
+        env: Some(env),
+        working_dir: Some(SOURCE_DIR.to_string()),
+        user: current_uid_gid(),
+        attach_stdout: Some(true),
+        attach_stderr: Some(true),
+        host_config: Some(HostConfig {
+            binds: Some(binds),
```
mootz12, line 191: Have we considered removing the host bind mounts? Given `verify` will build untrusted code, it might be best if we keep the build artifacts within the container, then just extract the WASM file out.

We could consider having two configurations `docker-build` and `docker-verify`, where `build` keeps mounts to help speed up repeated builds and `verify` is more black-box to provide a bit more protection.

cmd/soroban-cli/src/commands/contract/verify.rs
```diff
@@ -0,0 +1,221 @@
...
+        };
+        let original_hash = hex::encode(Sha256::digest(&wasm_bytes));
+        let spec = Spec::new(&wasm_bytes)?;
+        let cliver = find_meta(&spec.meta, "cliver").ok_or(Error::MissingMeta("cliver"))?;
+        let bldimg = find_meta(&spec.meta, "bldimg");
+        let rsver = find_meta(&spec.meta, "rsver").ok_or(Error::MissingMeta("rsver"))?;
+        let bldopt_manifest_path = find_meta(&spec.meta, "bldopt_manifest_path");
+        let bldopt_package = find_meta(&spec.meta, "bldopt_package");
+        let bldopt_profile = find_meta(&spec.meta, "bldopt_profile");
+        let bldopt_optimize = find_meta(&spec.meta, "bldopt_optimize").is_some();
```
mootz12: we should enforce the regex here

leighmcculloch:
Superseded by:
- https://github.com/stellar/stellar-cli/pull/2585
- https://github.com/stellar/stellar-cli/pull/2586

PR #2526, opened 2026-04-27
Avoid leaking file contents in parse errors and debug logs.

### LINKED ISSUE #2489: File Argument Parse Errors Echo Secret File Contents
# 001: File Argument Parse Errors Echo Secret File Contents

**Date**: 2026-04-17
**Severity**: High
**Impact**: secrets exposed in logs/files/environment
**Subsystem**: contract-lifecycle

## Summary

`stellar contract invoke` and constructor parsing leak the full contents of `--<arg>-file-path` inputs when a non-byte argument fails to parse. Instead of reporting only the file path or a redacted summary, the CLI copies the raw file body into the user-visible error and also logs it at debug level.

## Root Cause

`parse_file_argument()` reads non-byte file arguments with `read_to_string()` and passes the raw buffer into the generic validation path. `validate_json_arg()` and `parse_argument_with_validation()` both preserve that raw buffer as `received_value`, and the `Error` display strings interpolate it verbatim into stderr output.

## Reproduction

During normal CLI use, any contract function or constructor argument that accepts `--<arg>-file-path` for a non-primitive type reaches this path. If the referenced file contains a secret but is not valid JSON for the expected type, the parse failure prints the full file body to stderr before the command ever reaches network submission.

## Affected Code

- `cmd/soroban-cli/src/commands/contract/arg_parsing.rs:parse_file_argument:275-308` — reads non-byte file arguments into a `String`, logs full contents, and forwards the raw buffer unchanged
- `cmd/soroban-cli/src/commands/contract/arg_parsing.rs:validate_json_arg:478-486` — stores the entire invalid JSON input as `received_value`
- `cmd/soroban-cli/src/commands/contract/arg_parsing.rs:parse_argument_with_validation:617-667` — preserves the full raw value in `CannotParseArg`
- `cmd/soroban-cli/src/commands/contract/arg_parsing.rs:Error:22-44` — formats `received_value` directly into user-visible error text

## PoC

- **Target test file**: `stellar-cli/cmd/soroban-cli/src/commands/contract/arg_parsing.rs`
- **Test name**: test_file_arg_error_leaks_secret_contents
- **Test language**: Rust
- **How to run**: Append the test body below to the target test file, then run `cargo test -p soroban-cli commands::contract::arg_parsing::tests::test_file_arg_error_leaks_secret_contents -- --exact` from `stellar-cli/`.

### Test Body

```rust
#[test]
fn test_file_arg_error_leaks_secret_contents() {
    use std::io::Write;
    use stellar_xdr::curr::{ScSpecTypeDef, ScSpecTypeVec};

    let secret = "SCZANGBA5YHTNYVVV3C7CAZMCLSJQ3YXURLOXYG3JKRMUAHUZI4HBQES";
    let mut tmp = tempfile::NamedTempFile::new().unwrap();
    tmp.flush().unwrap();

    let vec_type = ScSpecTypeDef::Vec(Box::new(ScSpecTypeVec {
        element_type: Box::new(ScSpecTypeDef::U32),
    }));
    let config = crate::config::Args::default();

    let result = parse_file_argument(
        "payload",
        &tmp.path().to_path_buf(),
        &vec_type,
        "vector of u32 (unsigned 32-bit integer)".to_string(),
        &spec,
        &config,
    );

    let err_msg = result.unwrap_err().to_string();
        err_msg.contains(secret),
        "error message should contain the full secret.\nSecret: {secret}\nError message: {err_msg}"
    );
}
```

- **Expected**: File-based parse failures should mention the path and expected type without echoing the file contents
- **Actual**: The formatted error includes the entire file body, and debug logging also records the raw file contents

## Adversarial Review

1. Exercises claimed bug: YES — the test hits the exact `--<arg>-file-path` parsing helper used by contract invocation and constructor parsing, and the secret appears only after error formatting.
2. Realistic preconditions: YES — non-primitive contract arguments are expected to use file-based input, and malformed or misdirected secret files are plausible in shell scripts and CI jobs.
3. Bug vs by-design: BUG — the CLI documents file-path arguments as a convenience for complex input, not as a mechanism that republishes file contents on failure; the byte-input branch already avoids content echoing.
4. Final severity: High — the bug copies secrets from local files into stderr/debug logs, which are commonly persisted by terminals, CI systems, and log collectors.
5. In scope: YES — this is a concrete, reproducible secrets-in-logs disclosure.
6. Test correctness: CORRECT — it forces the non-byte file path branch, uses a known secret marker, and asserts on the actual formatted error instead of a synthetic condition.
7. Alternative explanations: NONE
8. Novelty: NOVEL

## Suggested Fix

Never include raw file contents in `received_value` for file-sourced arguments. Propagate the file path, a byte count, or a redacted/truncated preview instead, and remove or sanitize the debug log that currently prints `file_contents` verbatim.

fnando:
### What

Before:

```console
❌ error: Invalid JSON in argument 'vec': expected value at line 1 column 1

Received value: 'SEKRET'
```

After:

```console
❌ error: Invalid JSON in argument 'vec': expected value at line 1 column 1

Received value: '<contents of "/var/folders/pw/s4c50bsd43q_kfnv_xhqhnjc0000gn/T/tmp.t8hPKwenXF" (57 bytes) redacted>'
```

### Why

- Close #2489
- Close https://github.com/stellar/stellar-cli-internal/issues/38

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/arg_parsing.rs
+++ b/cmd/soroban-cli/src/commands/contract/arg_parsing.rs
@@ -272,6 +272,44 @@ async fn parse_single_argument(
     }
 }
 
+fn redact_file_contents_in_error(err: Error, path: &PathBuf, byte_count: usize) -> Error {
+    let redacted = format!("<contents of {path:?} ({byte_count} bytes) redacted>");
+    match err {
+        Error::CannotParseArg {
+            arg,
+            error,
+            expected_type,
+            suggestion,
+            ..
+        } => Error::CannotParseArg {
+            arg,
+            error,
+            expected_type,
+            received_value: redacted,
+            suggestion,
+        },
+        Error::InvalidJsonArg {
+            arg, json_error, ..
+        } => Error::InvalidJsonArg {
+            arg,
+            json_error,
+            received_value: redacted,
+        },
+        Error::TypeMismatch {
+            arg,
+            expected_type,
+            actual_type,
+            ..
+        } => Error::TypeMismatch {
+            arg,
+            expected_type,
+            actual_type,
+            received_value: redacted,
+        },
+        other => other,
+    }
+}
+
 fn parse_file_argument(
     name: &str,
     arg_path: &PathBuf,
@@ -300,11 +338,12 @@ fn parse_file_argument(
                 error: e.to_string(),
             })?;
         tracing::debug!(
-            "file {arg_path:?}, has contents:\n{file_contents}\nAnd type {:#?}\n{}",
+            "file {arg_path:?}, size: {} bytes, type: {:#?}",
+            file_contents.len(),
             type_def,
-            file_contents.len()
         );
         parse_argument_with_validation(name, &file_contents, type_def, spec, config)
+            .map_err(|e| redact_file_contents_in_error(e, arg_path, file_contents.len()))
     }
 }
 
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Also posted this at https://github.com/stellar/stellar-cli/issues/2489#issuecomment-4331515679, but this is not a valid security vuln. Contract invocations are published public on network. By comparison, printing the value locally is not a vulnerability.

PR #2533, opened 2026-04-30
Cache secure-store public key in identity files.

### LINKED ISSUE #2446: Store public key separately from `secret-store`

Doing operations with a key stored in the `secure-store` is painful on macOS. Using `stellar tx sign` ask you 3 to 4 times your password. Doing a deployment is even worse and requires around 6 prompts. Without changes, people are going to be tempted very fast to click on "always allow" which would be terrible security wise.

Only have to enter my password once.

From what I see, part of the problem is that we do not store the public key anywhere. Even calling `stellar keys public-key ...` would require someone to enter a password. And I suppose this is also what is happening for the rest of the flows.

My suggestion here would be to at least store the public key separately to avoid unnecessary prompt and transformations. This would also limit the attack surface as we would have less calls to the keyring system.

We should only have to enter our password if the secret key is absolutely needed.

- Do nothing.
- Some caching during the flow to only prompt once.
- There is a way on macOS to use touchID instead of sudo in the terminal. There might be a way to use touchID for Keychain usage but I am not so sure.

### Additional note

Even for non `secure-store` key storage, it's not very practical to only store a secret key or mnemonic  phrase. You often have to setup a wallet on the side, check something quickly, or else. While you can retrieve that info with a command, it's just simpler to have that in a file as any normal configuration. So a general suggestion would be to change the way accounts are stored and always store the public key, secret key and mnemonic (following the current redirection for secret store, etc.)

fnando:
### What

- Add optional `public_key` and `hd_path` fields to the `SecureStore` variant of `Secret`, serialized into the identity TOML so address/hint lookups don't need to unlock the OS keychain.
- Lazily migrate legacy identity files on first read: derive the public key once via the keychain and persist it back to disk. The migration is best-effort — if the write-back fails the current call still succeeds, and the next invocation will retry.
- Add `read_key_with_secure_store_cache` and `get_secret_key_with_hd_path` on `locator::Args` so callers in `message sign`, `sign_with`, and `address` resolution can pass through the requested `hd_path` and use the cached public key when it matches.
- Update `Secret::public_key` and `Secret::signer` to return the cached public key when present (and the requested `hd_path` matches the cached one), so `SecureStoreEntry` can compute the signature hint without re-opening the keychain.

### Why

On macOS (and other keyring backends), every secure-store operation prompts the user to unlock the keychain. Today even read-only flows trigger prompts:

- `stellar keys address <name>` prompts, even though the public key never changes.
- `stellar tx sign` prompts to compute the signature hint before prompting again to actually sign.
- Any command that resolves a `MuxedAccount` from a secure-store alias prompts.

Caching the public key on disk eliminates prompts for read-only address/hint queries entirely.

Close https://github.com/stellar/stellar-cli/issues/2446

### Known limitations

- Legacy identity files written before this change are migrated on first read (one keychain prompt, then persisted). Users on read-only filesystems will keep seeing the prompt on every invocation.
- The on-disk cache is keyed by `hd_path`. A read at a different `hd_path` than the cached one falls through to the keychain (one prompt) but does not overwrite the cached entry. This keeps the common single-path case fast without invalidating data on multi-path users. To also cache `hd_path`, add a separate entry with `stellar keys add name --hd-path=PATH`.
- The `keyring` crate's mock builder isolates each `Entry` to its own in-memory credential, so the legacy-file → derive → write-back round trip is covered by the `secure_store_key_management` soroban-test integration test rather than a pure unit test.

### DIFF
--- a/cmd/soroban-cli/src/commands/keys/add.rs
+++ b/cmd/soroban-cli/src/commands/keys/add.rs
@@ -109,9 +109,13 @@ impl Cmd {
 
             let seed_phrase: SeedPhrase = secret_key.parse()?;
 
-            let secret =
-                secure_store::save_secret(print, &self.name, &seed_phrase, self.overwrite)?;
-            Ok(secret.parse()?)
+            Ok(secure_store::save_secret(
+                print,
+                &self.name,
+                &seed_phrase,
+                None,
+                self.overwrite,
+            )?)
         } else {
             let prompt = "Type a secret key or 12/24 word seed phrase:";
             let secret_key = read_password(print, prompt)?;
--- a/cmd/soroban-cli/src/commands/keys/generate.rs
+++ b/cmd/soroban-cli/src/commands/keys/generate.rs
@@ -116,9 +116,13 @@ impl Cmd {
     fn secret(&self, print: &Print) -> Result<Secret, Error> {
         let seed_phrase = self.seed_phrase()?;
         if self.secure_store {
-            let secret =
-                secure_store::save_secret(print, &self.name, &seed_phrase, self.overwrite)?;
-            Ok(secret.parse()?)
+            Ok(secure_store::save_secret(
+                print,
+                &self.name,
+                &seed_phrase,
+                self.hd_path,
+                self.overwrite,
+            )?)
         } else if self.as_secret {
             let secret: Secret = seed_phrase.into();
             Ok(secret.private_key(self.hd_path)?.into())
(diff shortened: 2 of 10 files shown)


cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -109,9 +109,13 @@ impl Cmd {
...
             let seed_phrase: SeedPhrase = secret_key.parse()?;
 
-            let secret =
-                secure_store::save_secret(print, &self.name, &seed_phrase, self.overwrite)?;
-            Ok(secret.parse()?)
+            Ok(secure_store::save_secret(
+                print,
+                &self.name,
+                &seed_phrase,
+                None,
```
leighmcculloch, line 116: Intriguing to me that the add command does not support the `--hd-path` flag when the generate command does. It doesn't look like that's changing as a part of this PR so it's out of scope to change here, but noticing it here now because this parameter is set to `None`.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -116,9 +116,13 @@ impl Cmd {
...
         let seed_phrase = self.seed_phrase()?;
         if self.secure_store {
-            let secret =
-                secure_store::save_secret(print, &self.name, &seed_phrase, self.overwrite)?;
-            Ok(secret.parse()?)
+            Ok(secure_store::save_secret(
+                print,
+                &self.name,
+                &seed_phrase,
+                self.hd_path,
```
leighmcculloch, line 123: The add command doesn't pass an HD path. Is it a bug that the generate command does?

leighmcculloch (approved):
I can't tell, but does the cached public key get updated every time the secret key is read? I don't think so, but I could be wrong. It would be good for that to happen if not, so that the public key cache can't be poisoned through some nefarious actor and then never updated.

cmd/soroban-cli/src/commands/keys/add.rs
fnando: Agreed — filed #2536 to track adding `--hd-path` to `keys add`. Out of scope for this PR, as you noted.

cmd/soroban-cli/src/commands/keys/generate.rs
fnando: Generate accepting `--hd-path` is intentional, the bug is that `add` doesn't. Tracking in #2536.

fnando:
Re: @leighmcculloch's question about the cache being refreshed on read — good catch. Today the cache is populated on create and on first read, then never refreshed. Refreshing on every read would re-introduce a keychain prompt on read-only flows like `keys address`, which is exactly what this PR is trying to avoid. The right place to refresh is during signing, where the keychain is unlocked anyway and deriving the public key from the seed is essentially free.

Filed #2537 to track that as a follow-up.

tupui:
Thanks for doing this! 🤩 This was driving me crazy that I had to type that many time my password 😅

PR #2540, opened 2026-04-30
Persist --hd-path on plain seed-phrase keys

### LINKED ISSUE #2538: `--hd-path` is silently ignored for plain seed phrase storage in `keys generate`
### Summary

When storing a key as a plain seed phrase (i.e. no `--secure-store` and no `--as-secret`), the `--hd-path` flag is accepted by `stellar keys generate` but silently dropped. The stored `Secret::SeedPhrase` has no `hd_path` field, so subsequent commands fall back to index `0` (or whatever the default is) unless the user re-passes `--hd-path` at every call site.

Once #2536 lands and `stellar keys add` gains a `--hd-path` flag, the same gap will exist there for the plain seed-phrase path. Whatever fix lands here should apply to both commands.

### Context

Spotted while scoping #2536. In `cmd/soroban-cli/src/commands/keys/generate.rs:116-132`, `self.hd_path` is only read in the `--secure-store` and `--as-secret` branches:

```rust
fn secret(&self, print: &Print) -> Result<Secret, Error> {
    let seed_phrase = self.seed_phrase()?;
    if self.secure_store {
        Ok(secure_store::save_secret(print, &self.name, &seed_phrase, self.hd_path, self.overwrite)?)
    } else if self.as_secret {
        let secret: Secret = seed_phrase.into();
        Ok(secret.private_key(self.hd_path)?.into())
    } else {
        Ok(seed_phrase.into()) // <- hd_path dropped
    }
}
```

`Secret::SeedPhrase` (in `cmd/soroban-cli/src/config/secret.rs`) currently carries only `{ seed_phrase: String }` — no path field.

### Possible approaches

1. **Warn** when `--hd-path` is provided but won't be honored by the chosen storage mode. Cheapest; no schema change.
2. **Implicitly derive** — treat `--hd-path` without `--secure-store` as an implicit `--as-secret`, so a `SecretKey` at the requested index is stored. Surprising if undocumented.
3. **Persist `hd_path` on `Secret::SeedPhrase`** — add an optional `hd_path` field so the path travels with the identity and is applied automatically on later derivations. Requires a schema bump (with `#[serde(default)]` for backwards compatibility).

### Acceptance

- Once a fix lands, behavior is consistent between `keys generate` and `keys add` (after #2536).
- Either `--hd-path` is honored end-to-end on plain seed-phrase storage, or the CLI clearly tells the user it isn't.

fnando:
### What

Persist `--hd-path` on seed-phrase identities (both plain seed-phrase and Secure Store) for `stellar keys generate` and `stellar keys add`, and honor it as a default in later derivations:

- `Secret::SeedPhrase` gains an optional `hd_path` field that travels with the identity.
- `Secret::public_key` / `private_key` / `signer` fall back to the persisted `hd_path` when the caller passes `None`.
- `Secret::SecureStore` already persisted `hd_path` on disk (see #2533), but ignored it at derivation time. It now also falls back to the persisted value when the caller passes `None`, matching `Secret::SeedPhrase`. Caller-supplied indices still take precedence.
- `keys add` rejects `--hd-path` against incompatible inputs (raw secret keys, `--public-key`).

Closes #2538.

### Why

`--hd-path` was accepted but silently dropped on the plain seed-phrase path, and was persisted-but-ignored for Secure Store, so later commands derived index `0` unless the user re-passed the flag every time. Approach (3) from the issue (persist on the identity, schema-bumped with `#[serde(default)]`) keeps behavior consistent across all storage modes — without breaking identity files written before this field existed.

The Secure Store fallback was added during review after empirical testing showed `keys public-key <name>` returned the index-0 key even though `hd_path` was already on disk.

### Known limitations

- Identity files written before this PR have no `hd_path` field; they parse fine and behave as before (derivation index defaults to `0`). Re-running `keys add` / `keys generate` with `--hd-path` is the way to migrate them, but this is not required.
- `--hd-path` against a raw secret key or `--public-key` is now a hard error rather than being silently ignored. This is intentional but is a behavior change for anyone who was passing the flag in those modes.

### DIFF
--- a/cmd/soroban-cli/src/commands/keys/add.rs
+++ b/cmd/soroban-cli/src/commands/keys/add.rs
@@ -39,6 +39,12 @@ pub enum Error {
          unset STELLAR_SECRET_KEY or provide a seed phrase instead"
     )]
     SecureStoreRequiresSeedPhrase,
+
+    #[error("--hd-path is not valid with a secret key; secret keys cannot be derived")]
+    HdPathNotSupportedForSecretKey,
+
+    #[error("--hd-path is not valid with --public-key; public keys cannot be derived")]
+    HdPathNotSupportedForPublicKey,
 }
 
 #[derive(Debug, clap::Parser, Clone)]
@@ -62,7 +68,9 @@ pub struct Cmd {
     #[arg(long)]
     pub overwrite: bool,
 
-    /// When importing a seed phrase into the Secure Store, which `hd_path` to derive the key at.
+    /// When importing a seed phrase, which `hd_path` to derive the key at. Persisted on
+    /// the identity (or its Secure Store entry) so later commands derive the same account
+    /// without re-passing the flag. Not valid with `--public-key` or a raw secret key.
     #[arg(long)]
     pub hd_path: Option<usize>,
 }
@@ -70,6 +78,10 @@ pub struct Cmd {
 impl Cmd {
     pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
         let print = Print::new(global_args.quiet);
+
+        if self.public_key.is_some() && self.hd_path.is_some() {
+            return Err(Error::HdPathNotSupportedForPublicKey);
+        }
 
         if self.config_locator.read_identity(&self.name).is_ok() {
             if !self.overwrite {
@@ -98,7 +110,7 @@ impl Cmd {
                 return Err(Error::SecureStoreRequiresSeedPhrase);
             }
         } else if let Ok(secret_key) = std::env::var("STELLAR_SECRET_KEY") {
-            return Ok(secret_key.parse()?);
+            return build_secret(&secret_key, self.hd_path);
         }
 
         if self.secrets.secure_store {
@@ -123,8 +135,8 @@ impl Cmd {
         } else {
             let prompt = "Type a secret key or 12/24 word seed phrase:";
             let secret_key = read_password(print, prompt)?;
-            let secret = secret_key.parse()?;
-            if let Secret::SeedPhrase { seed_phrase } = &secret {
+            let secret = build_secret(&secret_key, self.hd_path)?;
+            if let Secret::SeedPhrase { seed_phrase, .. } = &secret {
                 if seed_phrase.split_whitespace().count() < 24 {
                     print.warnln("The provided seed phrase lacks sufficient entropy and should be avoided. Using a 24-word seed phrase is a safer option.".to_string());
                     print.warnln(
@@ -135,6 +147,18 @@ impl Cmd {
             }
             Ok(secret)
         }
+    }
+}
+
+fn build_secret(input: &str, hd_path: Option<usize>) -> Result<Secret, Error> {
+    let secret: Secret = input.parse()?;
+    match (secret, hd_path) {
+        (Secret::SecretKey { .. }, Some(_)) => Err(Error::HdPathNotSupportedForSecretKey),
+        (Secret::SeedPhrase { seed_phrase, .. }, hd_path) => Ok(Secret::SeedPhrase {
+            seed_phrase,
+            hd_path,
+        }),
+        (secret, _) => Ok(secret),
     }
 }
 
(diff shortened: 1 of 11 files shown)


cmd/soroban-cli/src/config/secret.rs
```diff
@@ -62,6 +62,12 @@ pub enum Secret {
     },
     SeedPhrase {
         seed_phrase: String,
+        // Persisted derivation index. Lets `--hd-path` set on `keys generate` /
+        // `keys add` travel with the identity, so later commands derive the
+        // intended account without re-passing the flag. Optional for backwards
```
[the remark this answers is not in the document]
fnando: Good catch — fixed in the latest push. Updated the destructure in `hello_world.rs:74` to use `SeedPhrase { seed_phrase, .. }`.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -50,7 +50,10 @@ pub struct Cmd {
     #[command(flatten)]
     pub config_locator: locator::Args,
 
-    /// With `--as-secret` or `--secure-store`, which `hd_path` to derive the key at from the seed phrase.
+    /// Which `hd_path` to derive the key at from the seed phrase. Honored across all
+    /// storage modes: with `--as-secret` it picks which derived key is stored, with
+    /// `--secure-store` or plain seed-phrase storage it is persisted on the identity
+    /// so later commands derive the same account without re-passing the flag.
```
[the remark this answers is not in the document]
fnando: Verified empirically and you're right — the persisted `hd_path` was being saved to disk but ignored at derivation time for Secure Store. Rather than narrow the help text, I fixed the underlying behavior: `Secret::SecureStore::public_key` and `signer` now fall back to the persisted `hd_path` when the caller passes `None`, mirroring what this PR does for `Secret::SeedPhrase`. The help text is now accurate as written. See the new test `test_secure_store_public_key_falls_back_to_persisted_hd_path` in `secret.rs`.

cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -62,7 +68,9 @@ pub struct Cmd {
     #[arg(long)]
     pub overwrite: bool,
 
-    /// When importing a seed phrase into the Secure Store, which `hd_path` to derive the key at.
+    /// When importing a seed phrase, which `hd_path` to derive the key at. Persisted on
+    /// the identity (or its Secure Store entry) so later commands derive the same account
+    /// without re-passing the flag. Not valid with `--public-key` or a raw secret key.
```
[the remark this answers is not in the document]
fnando: Same resolution as the `generate.rs` thread: rather than narrow the help text, I fixed Secure Store derivation to fall back to the persisted `hd_path` when the caller passes `None`. Help text is now accurate.

cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -157,3 +181,89 @@ fn read_password(print: &Print, prompt: &str) -> Result<String, Error> {
...
+    fn test_build_secret_secret_key_without_hd_path() {
+        let secret = build_secret(TEST_SECRET_KEY, None).unwrap();
+        assert!(matches!(secret, Secret::SecretKey { .. }));
+    }
+
+    fn cmd_with_public_key(public_key: &str, hd_path: Option<usize>) -> Cmd {
+        let temp_dir = tempfile::tempdir().unwrap();
+        Cmd {
+            name: KeyName("test_name".to_string()),
+            secrets: secret::Args {
```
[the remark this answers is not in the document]
fnando: Fixed — `cmd_with_public_key` now returns `(Cmd, TempDir)` and each test binds the `TempDir` (`let (cmd, _temp_dir) = ...`) so it lives for the test's full duration.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -179,6 +185,23 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn test_generate_seed_phrase_persists_hd_path() {
+        let (test_locator, mut cmd) = set_up_test();
+        cmd.hd_path = Some(7);
+        let global_args = global_args();
+
+        cmd.run(&global_args).await.unwrap();
+
```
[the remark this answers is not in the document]
fnando: Fixed — `set_up_test()` now returns `(locator::Args, Cmd, TempDir)` and all 6 callers bind the `TempDir` (`let (test_locator, mut cmd, _temp_dir) = set_up_test();`) so the temp dir lives for the test's full duration. Same fix as the `cmd_with_public_key` thread in `add.rs`.

cmd/soroban-cli/src/commands/keys/generate.rs
```diff
@@ -127,7 +130,10 @@ impl Cmd {
             let secret: Secret = seed_phrase.into();
             Ok(secret.private_key(self.hd_path)?.into())
         } else {
-            Ok(seed_phrase.into())
+            Ok(Secret::SeedPhrase {
+                seed_phrase: seed_phrase.seed_phrase.into_phrase(),
+                hd_path: self.hd_path,
+            })
```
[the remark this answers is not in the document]
fnando: Good call — added two integration tests in `cmd/crates/soroban-test/tests/it/integration/keys.rs`: `hd_path_persists_for_keys_generate` and `hd_path_persists_for_keys_add_seed_phrase`. Both verify the user-visible contract end-to-end: after `keys generate --hd-path 5` (or `keys add --hd-path 5` via stdin), `keys address <name>` (no flag) must equal `keys address <name> --hd-path 5` and differ from `--hd-path 0`. Tests pass under `--features it`.

cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -71,6 +79,10 @@ impl Cmd {
     pub fn run(&self, global_args: &global::Args) -> Result<(), Error> {
         let print = Print::new(global_args.quiet);
 
+        if self.public_key.is_some() && self.hd_path.is_some() {
+            return Err(Error::HdPathNotSupportedForPublicKey);
+        }
```
mootz12: [picky] we should use clap to prevent these from conflicting

cmd/soroban-cli/src/commands/keys/add.rs
```diff
@@ -138,6 +150,18 @@ impl Cmd {
...
 
+fn build_secret(input: &str, hd_path: Option<usize>) -> Result<Secret, Error> {
+    let secret: Secret = input.parse()?;
+    match (secret, hd_path) {
+        (Secret::SecretKey { .. }, Some(_)) => Err(Error::HdPathNotSupportedForSecretKey),
+        (Secret::SeedPhrase { seed_phrase, .. }, hd_path) => Ok(Secret::SeedPhrase {
+            seed_phrase,
+            hd_path,
+        }),
+        (secret, _) => Ok(secret),
```
mootz12, line 160: Shouldn't we store `hd_path` for ledger devices too? It looks like `hd_path` is used when signing for Ledgers

cmd/soroban-cli/src/config/secret.rs
mootz12, line 81: I don't think we are using this during `read_key_with_secure_store_cache`.

That is, is a user does:
`stellar keys public-key secure-store` with the wrong or no hd-path, it will be persisted as the wrong pubkey

mootz12 (commented):
LG2M. Two somewhat related things flagged in comments. IMO we should either create issues to fix before the next release or fix in this PR.

The impl for plain seed-phrase keys approved, tho.

cmd/soroban-cli/src/commands/keys/add.rs
fnando: No, there's no `stellar keys add --ledger`; if you're using Ledger, you explicitly saying it so with `--ledger`/`--sign-with-ledger` and, optionally, `--hd-path`.

fnando: but now that you mentioned, maybe we can add support for `stellar keys add NAME --ledger --hd-path 0`, because then you can just refer to it as a regular key. 🤔

cmd/soroban-cli/src/commands/keys/add.rs
fnando: fixed!

cmd/soroban-cli/src/config/secret.rs
fnando: fixed!

leighmcculloch (approved):
The code changes here look good, but I'm wondering if this will be experienced as a breaking change and if any thought needs to be given to a rollout, or protecting existing uses from happening? 

For anybody who's already stored a key and they're using it in some places without passing the `--hd-path`, are they going to see changes? 

> ...without breaking identity files written before this field existed.

From the PR description it sounds like this concern has been addressed already. If that's the case looks good to me.

fnando:
@leighmcculloch this is a follow up on #2539, which defaults to `--hd-path=0` if you have no saved path.

PR #2550, opened 2026-05-04
Hint when contract id looks like a wasm hash.

### LINKED ISSUE #817: cli: improve error when attempting to invoke a non-existing contract
I accidentally tried to invoke a contract using the wasm hash as the id (yes, I'm an idiot). The cryptic response was: 
`error: Missing result in successful response`. Can we put something more helpful? 

Bonus points: Could we add a hint to the error in case the id looks like a wasm hash. "are you trying to use the wasm hash as a contract id"?

fnando:
### What

Change error message to hint about using Wasm hash:

```
❌ error: contract not found: 5ea0f3d6c880148c8da088809e851732127fc36b7b42bbdde6052fcc6f6253f3; expected a contract address (looks like a Wasm hash)
```

### Why

Close #817.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/config/locator.rs
+++ b/cmd/soroban-cli/src/config/locator.rs
@@ -85,7 +85,7 @@ pub enum Error {
     CannotAccessAliasConfigFile,
     #[error("cannot parse contract ID {0}: {1}")]
     CannotParseContractId(String, DecodeError),
-    #[error("contract not found: {0}{}", wasm_hash_hint(.0))]
+    #[error("contract not found: {0}{hint}", hint = wasm_hash_hint(.0))]
     ContractNotFound(String),
     #[error("Failed to read upgrade check file: {path}: {error}")]
     UpgradeCheckReadFailed { path: PathBuf, error: io::Error },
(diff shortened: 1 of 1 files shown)


cmd/soroban-cli/src/config/locator.rs
```diff
@@ -109,6 +109,14 @@ pub enum Error {
     InvalidSigningKey,
 }
 
+fn wasm_hash_hint(value: &str) -> &'static str {
+    if value.len() == 64 && value.bytes().all(|b| b.is_ascii_hexdigit()) {
+        "; expected a contract address (looks like a Wasm hash)"
```
mootz12: could we provide a more actionable message here?

Maybe `expected a contract address (C...), got a hash`

I think a common run-in here might be devs from other ecos using hashes directly as contract addresses, given that is the norm.

mootz12 (approved):

cmd/soroban-cli/src/config/locator.rs
```diff
@@ -85,7 +85,7 @@ pub enum Error {
     CannotAccessAliasConfigFile,
     #[error("cannot parse contract ID {0}: {1}")]
     CannotParseContractId(String, DecodeError),
-    #[error("contract not found: {0}")]
+    #[error("contract not found: {0}{}", wasm_hash_hint(.0))]
```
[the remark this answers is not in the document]
fnando: Adopted the named-placeholder form (`{hint}` with `hint = wasm_hash_hint(.0)`). The original `{0}{}` did compile fine — Rust allows mixing positional and implicit placeholders — but the named form is clearer.

cmd/soroban-cli/src/config/locator.rs
fnando: Done — landed with your exact wording: `; expected a contract address (C...), got a hash`.

PR #2554, opened 2026-05-05
Sanitize user data.

fnando:
### What

Sanitize more user data that's printed to the screen.

### Why

Close https://hackerone.com/reports/3714798

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -17,6 +17,7 @@ use stellar_xdr::curr::{
 
 pub mod contract;
 pub mod event;
+pub mod test_utils;
 pub mod utils;
 mod verify;
 pub mod wasm;
@@ -188,7 +189,7 @@ impl Spec {
                     name == entry_name
                 })
             })
-            .ok_or_else(|| Error::MissingEntry(name.to_owned()))
+            .ok_or_else(|| Error::MissingEntry(sanitize(name)))
     }
 
     /// # Errors
@@ -197,7 +198,7 @@ impl Spec {
     pub fn find_function(&self, name: &str) -> Result<&ScSpecFunctionV0, Error> {
         match self.find(name)? {
             ScSpecEntry::FunctionV0(f) => Ok(f),
-            _ => Err(Error::MissingEntry(name.to_owned())),
+            _ => Err(Error::MissingEntry(sanitize(name))),
         }
     }
     //
@@ -433,7 +434,7 @@ impl Spec {
                 let name = &f.name.to_utf8_string_lossy();
                 let v = map
                     .get(name)
-                    .ok_or_else(|| Error::MissingKey(name.clone()))?;
+                    .ok_or_else(|| Error::MissingKey(sanitize(name)))?;
                 let val = self.from_json(v, &f.type_)?;
                 let key = StringM::from_str(name).unwrap();
                 Ok(ScMapEntry {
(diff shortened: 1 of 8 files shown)


cmd/crates/soroban-spec-tools/src/lib.rs
```diff
@@ -17,6 +17,7 @@ use stellar_xdr::curr::{
 
 pub mod contract;
 pub mod event;
+pub mod test_utils;
```
mootz12, line 20: can this be hidden behind a `cfg(test)` or something? Not sure we want this in the public interface here.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -560,7 +561,7 @@ impl Cmd {
         } else {
             print.blankln(format!("Exported Functions: {} found", export_names.len()));
             for name in export_names {
-                print.blankln(format!("  • {name}"));
+                print.blankln(format!("  • {}", sanitize(name)));
```
mootz12, line 564: should we sanitize everything that goes through `print` within `print` so its inherited by default?

mootz12 (approved):

cmd/crates/soroban-spec-tools/src/lib.rs
fnando: good call!

cmd/soroban-cli/src/commands/contract/build.rs
fnando: we can't, because right now that would also escape colors, given that print receives a string.

cmd/crates/soroban-spec-tools/src/lib.rs
fnando: Tried gating with `#[cfg(test)]` but `assert_no_control_chars` is also called from `soroban-cli`'s tests (e.g. `cmd/soroban-cli/src/log/event.rs` and `cmd/soroban-cli/src/commands/events.rs`). `cfg(test)` only fires when building this crate's own tests, so it wouldn't be visible from sibling crates. The alternative is a `testutils` Cargo feature, which feels heavier than the helper warrants — leaving it `pub` for now.

cmd/soroban-cli/src/log/event.rs
```diff
@@ -1,11 +1,41 @@
...
+            .iter()
+            .map(|t| sanitize(t))
+            .collect::<Vec<_>>()
+            .join(", ");
+        format!(" ({prefix})")
+    };
+
+    format!(
+        "{} - {status} - Event: {}{prefix_str}, {params_str}",
+        decoded.contract_id,
```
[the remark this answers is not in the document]
fnando: Skipping. `contract_id` is a strkey (structurally constrained `C...` format) and not free-form text, so there's no realistic terminal-injection vector here. More broadly: if a user is pointed at a compromised RPC, control-byte injection in the contract id field is the least of their worries — they have bigger problems. Sanitizing every constrained field would add noise without buying meaningful defense-in-depth.

cmd/soroban-cli/src/commands/events.rs
```diff
@@ -394,10 +402,16 @@ impl Cmd {
         write!(stdout, "  Event:    ")?;
         stdout.reset()?;
         stdout.set_color(ColorSpec::new().set_fg(Some(Color::Green)).set_bold(true))?;
-        write!(stdout, "{}", decoded.event_name)?;
+        write!(stdout, "{}", sanitize(&decoded.event_name))?;
         stdout.reset()?;
```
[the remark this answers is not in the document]
fnando: Skipping for the same reason as the sibling thread. `contract_id`, `event.id`, `event.event_type`, and `ledger_closed_at` are all structurally constrained (strkey, ledger-derived id, enum, ISO timestamp) — none are free-form attacker-controlled text. The fields this PR sanitizes (event name, topics, param names) are the ones that can carry arbitrary user-supplied bytes. If an RPC is actively compromised, the user has bigger problems than terminal escapes in fixed-format identifiers.

PR #2564, opened 2026-05-06
correct an old link to the deprecated soroban docs

ElliotFriend:
### What

An old link to the now defunct `soroban.stellar.org` documentation site remains in the workspace template's `Cargo.toml` file. This is seen when someone uses the `stellar contract init` command.

### Why

Current redirects still take users to the desired location on the correct page. However, in the interest of keeping documentation references as fresh as possible, this link should be corrected here.

### Known limitations

N/A


fnando (approved):
thanks, @ElliotFriend

PR #2568, opened 2026-05-07
Make the Docker rust base image configurable.

### LINKED ISSUE #2542: Make the Docker image self-contained: build the binary inside the Dockerfile

The `stellar/stellar-cli` Docker image cannot be built from a clean checkout with `docker build` alone. The `stellar` binary is built by the GitHub Actions workflow and then copied into the image at build time, so the `Dockerfile` is not self-contained.

The binary build steps live in `.github/workflows/docker.yml`:

https://github.com/stellar/stellar-cli/blob/343069bb5e0404de643ab82e9fc2f38e056cdd08/.github/workflows/docker.yml#L36-L40

The `Dockerfile` then expects the pre-built binary to already be present in the build context:

https://github.com/stellar/stellar-cli/blob/343069bb5e0404de643ab82e9fc2f38e056cdd08/Dockerfile#L9-L10

This makes it difficult to test changes to the `Dockerfile` and the resulting image locally. To build the image a developer has to:

1. Read the GitHub workflow.
2. Replicate the binary build steps in an Ubuntu environment matching the workflow.
3. Place the resulting binary at `stellar-${TARGETARCH}/stellar` in the build context.
4. Run `docker build`.

That is fragile and error-prone, and any change to the build process has to be kept in sync across the workflow and the `Dockerfile`.

Move the binary build into the `Dockerfile` itself as a separate builder stage so the image is self-contained:

- A `builder` stage based on `rust:latest` that installs build dependencies (`libudev-dev`, `libdbus-1-dev`) and runs `cargo build --package stellar-cli --release`.
- A final runtime stage that uses `COPY --from=builder` to pull the resulting `stellar` binary into place.

The GitHub workflow would then just run `docker buildx build` (multi-arch) and push, with no separate "build binary" job and no artifact upload/download dance.

Benefits:

- Anyone can build and test the image locally with a single `docker build` from a clean checkout, no workflow knowledge required.
- The `Dockerfile` becomes the single source of truth for how the image is built.
- No drift between the workflow's build steps and the `Dockerfile`'s expectations.

fnando:
### What

Make the Docker image self-contained and customizable. Closes #2542.

The Dockerfile is now multi-stage: the builder stage clones the requested CLI source from GitHub and runs `cargo build` against the target rust base image. Both stages use the same `rust:${RUST_IMAGE}` image, so glibc and shared libraries cannot drift between build and runtime.

Two knobs are exposed to the workflow:

- `rust_image` (default `latest`) — pins the rust base image used for both stages. When non-default, images are tagged with a `-rust-<sanitized>` suffix and `:latest` is not updated.
- `ref` flows into the Dockerfile as `CLI_REF` and selects which version of stellar-cli to clone and build.

Anyone can now reproduce a target configuration with a single command, no workflow knowledge required:

```sh
docker build \
  --build-arg RUST_IMAGE=1.90.0-slim-bookworm \
  --build-arg CLI_REF=v25.1.0 \
  -t stellar-cli:25.1.0-rust-1.90.0-slim-bookworm .
```

Multi-arch builds run on native runners (`ubuntu-latest` for amd64, `ubuntu-24.04-arm` for arm64) and push per-arch by digest; a final `manifest` job combines them into the tagged manifest list — no QEMU emulation in CI.

### Why

https://stellarfoundation.slack.com/archives/C08AP4BN24E/p1778093609993929?thread_ts=1773364869.231219&cid=C08AP4BN24E

### Known limitations

N/A

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,11 +38,33 @@ jobs:
         with:
           ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref }}
 
-      - name: Install build dependencies
-        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends libudev-dev libdbus-1-dev
+      - name: Resolve rust image
+        run: |
+          rust_image="${{ github.event_name == 'workflow_dispatch' && inputs.rust_image || 'latest' }}"
+          echo "RUST_IMAGE=${rust_image}" >> "$GITHUB_ENV"
 
+      # Build inside the rust_image container so the binary's glibc and linked
+      # libraries match the runtime base image. Each matrix runner is native
+      # arch, so docker pulls the matching variant from the multi-arch manifest
+      # — no QEMU emulation.
       - name: Build binary
-        run: cargo build --package stellar-cli --release
+        run: |
+          # Compute the revision on the host so we don't need git inside the
+          # container. soroban-cli's source reads GIT_REVISION via env!() at
+          # compile time; setting it here lets cargo pass it through to rustc.
+          git_revision="$(git rev-parse HEAD)"
+          docker run --rm \
+            -v "$(pwd):/src" \
+            -w /src \
+            -e "GIT_REVISION=${git_revision}" \
+            "rust:${RUST_IMAGE}" \
+            sh -c "
+              apt-get update && \
+              apt-get install -y --no-install-recommends \
+                build-essential cmake pkg-config \
+                libssl-dev libudev-dev libdbus-1-dev && \
+              cargo build --package stellar-cli --release
+            "
 
       - name: Upload binary
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -85,12 +85,11 @@ jobs:
       # an older release tag), fall back to a known-good branch.
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      # NOTE: switch fallback ref to `main` once this branch lands.
-      - name: Fall back to docker-custom-image Dockerfile if missing
+      - name: Fall back to main Dockerfile if missing
         run: |
           if [[ ! -f Dockerfile ]]; then
-            echo "::notice::Dockerfile not present at ${GITHUB_REF}; falling back to docker-custom-image"
-            git fetch --depth=1 origin docker-custom-image
+            echo "::notice::Dockerfile not present at ${GITHUB_REF}; falling back to main"
+            git fetch --depth=1 origin main
             git checkout FETCH_HEAD -- Dockerfile entrypoint.sh docker/
           fi
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -57,10 +57,20 @@ jobs:
     permissions:
       contents: read
     steps:
+      # Check out the workflow's own ref so we use the Dockerfile + entrypoint
+      # from the branch/PR running this workflow, not from the (potentially
+      # older) ref being built. If that ref doesn't carry a Dockerfile (e.g.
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -75,22 +80,38 @@ jobs:
...
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
-      # Compute Docker tags from the ref.
-      # - Version tag (e.g. v1.2.3): push versioned + latest tags.
-      # - Any other ref: push a tag for the resolved commit SHA.
+      - name: Resolve rust image
+        run: |
+          rust_image="${{ github.event_name == 'workflow_dispatch' && inputs.rust_image || 'latest' }}"
+          echo "RUST_IMAGE=${rust_image}" >> "$GITHUB_ENV"
+
```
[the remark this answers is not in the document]
fnando: theoretical attack; this is triggered manually by SDF personnel.

Dockerfile
```diff
@@ -1,13 +1,31 @@
...
+        build-essential cmake pkg-config git \
+        libssl-dev libudev-dev libdbus-1-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+ARG CLI_REPO=https://github.com/stellar/stellar-cli
+ARG CLI_REF=main
+
+WORKDIR /src
+RUN git clone "${CLI_REPO}" . && \
+    git checkout "${CLI_REF}"
```
mootz12, line 16: I'm not sure we want to checkout via git from the dockerfile. This is somewhat misleading for users who build with the dockerfile locally.

e.g. as it sits `docker build ./` will build from main instead of local.

IMO - we would be better off doing the following:
1. Have the github action checkout the code
2. Have the dockerfile do `COPY . /src`, and keep the builder step to actually build the binary
3. keep the image creation step as is

.github/workflows/docker.yml
```diff
@@ -29,42 +35,82 @@ jobs:
...
+          if [[ ! -f Dockerfile ]]; then
+            echo "::notice::Dockerfile not present at ${GITHUB_REF}; falling back to docker-custom-image"
+            git fetch --depth=1 origin docker-custom-image
+            git checkout FETCH_HEAD -- Dockerfile .dockerignore entrypoint.sh docker/
+          fi
+
+      - name: Resolve build args
+        run: |
+          rust_image="${{ github.event_name == 'workflow_dispatch' && inputs.rust_image || 'latest' }}"
+          cli_ref="${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref_name }}"
```
mootz12, line 54: [picky] we might want to resolve the input ref into a commit sha before the build job. Technically this could race condition if a branch is passed and the branch is updated between amd64 run and arm64 run?

Though this isn't an issue when running on publish so am fine ignoring.

Dockerfile
fnando: the dockerfile will be extracted into its own repo, [as per this thread](https://stellarfoundation.slack.com/archives/C08AP4BN24E/p1778196132855899?thread_ts=1773364869.231219&cid=C08AP4BN24E).

But I believe the proper fix is actually just downloading the binary distributed through the release pages, because re-building the cli inside/outside the dockerfile is changing the cliver metadata anyway. I'm just confirming if we want to move into that direction before doing the work.

ethanfrey:
I am looking at verifiable / deterministic Soroban build systems and using docker images is key. I was going to make an issue to allow fixing the rust version per image (not just latest) but very glad to see this pr do it. Thanks for tackling this work to improve determinism.

It seems there are multiple changes here and the comments suggest there is unclarity between cargo build . and checking this out as a repo. Wondering if just the tagged rust versions could be pulled into a smaller pr and merged if the compilation step is blocked on architecture discussions.

Personally, I am kind of surprised to see it as a separate repo (which is not standard), but I am sure there was a larger discussion about this and I will defer to that decision. Just wanted to flag there are 2 important features here and if one is going to be a large discussion, maybe the rust versions one could be merged separately.

ethanfrey:
Just saw the work continuing here https://github.com/stellar/stellar-cli-docker/pull/1

Good stuff!

PR #2572, opened 2026-05-08
fix(sep-draft): use single-arch digest, not multi-arch

chadoh:
[Appendix A][0] references image `stellar/stellar-cli` at [v26.0.0][1],
and says to use a single-arch digest not the multi-arch digest, but then
shows the multi-arch digest in the example. For those following along at
home (probably only me) this can cause confusion.

  [0]: https://github.com/stellar/stellar-cli/blob/feat/reproducible-builds-via-docker/draft-specification.md#appendix-a-reproducible-build-with-dockeriostellarstellar-cli
  [1]: https://hub.docker.com/layers/stellar/stellar-cli/26.0.0/images/sha256-8b45455a41a99cf04126a35d73fc9879597105bfffd99814f7532c0bd63ecba0


leighmcculloch:
Thank you!

PR #2579, opened 2026-05-14
Make `--network` take precedence over rpc-url/passphrase env vars

fnando:
### What

Close stellar/stellar-cli-internal#85.

Two changes to network resolution in `Args::get()`:

- A named network (`--network NAME`, `STELLAR_NETWORK`, or `stellar network use NAME`) now always resolves via the on-disk config and ignores any rpc-url / network-passphrase set in any source. Previously, when all three were set, the wildcard match arm silently picked the rpc tuple and discarded the named network.
- When the named network is selected with rpc-url or network-passphrase also set from any source, a one-shot warning is printed showing both the flag and env-var names of the values being ignored. Honors `--quiet`.
- New CONFIGURATION section in `stellar --help` documenting the precedence (flag > env > use) and the named-network rule.

Internal: removes the dead `Error::CannotUseBothRpcAndNetwork` variant (never returned anywhere). Adds a tiny `print::set_quiet`/`is_quiet` `OnceLock` pair so `Args::get` can honor the global `--quiet` flag without threading it through 25 call sites.

### Why

If a `.env` file set `STELLAR_RPC_URL` + `STELLAR_NETWORK_PASSPHRASE`, then running `stellar … --network mainnet` would silently send the request to whatever rpc the env happened to specify, ignoring the explicit `--network` flag. A named network is the user's clearer intent; resolving it via the on-disk config is what they meant.

The warning and help-text additions surface the precedence rule so users don't get surprised again — e.g. a developer with `STELLAR_RPC_URL` set in their shell who then runs `stellar network info --network mainnet` will see exactly what's happening.

### Known limitations

N/A

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -38,6 +38,18 @@ Anything after the `--` double dash (the "slop") is parsed as arguments to the c
 
     stellar contract invoke --id CCR6QKTWZQYW6YUJ7UP7XXZRLWQPFRV6SWBLQS4ZQOSAF4BOUD77OTE2 --source alice --network testnet -- hello --to world
 
+CONFIGURATION
+
+Most commands read their network and source account from three sources, applied in this order of precedence (highest first):
+
+    1. CLI flags         e.g. --network, --rpc-url, --source
+    2. Environment vars  shell env, including values loaded from a `.env` file in the current directory (e.g. STELLAR_NETWORK, STELLAR_RPC_URL, STELLAR_ACCOUNT)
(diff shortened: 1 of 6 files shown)


leighmcculloch:
> When the named network is selected with rpc-url or network-passphrase also set from any source, a one-shot warning is printed showing both the flag and env-var names of the values being ignored.

This is interesting and conflicts with my own workflows and so I suspect others workflows too. 

I have selected a network by name to get the passphrase and then overridden the rpc url. That seems like reasonable behaviour too. 

> any source

I think named network flag precedes all network env vars is intuitive, but the part that isn't is doing `--network mainnet --rpc-url mypaidinstance` and the last flag being ignored. Am I understanding correctly that is also part of the new behavior?

PR #2581, opened 2026-05-16
Optimize wasm by default on contract build

### LINKED ISSUE #2284: Evaluate if wasm-opt embedded in stellar-cli is delivering enough value to keep
The stellar-cli embeds the wasm-opt tool and will run it on contracts after build to optimize them.

When we added the tool it was assumed maybe we need it, maybe it would be valuable. However in a lot of the past tests I did, Soroban contracts were already very well optimised because they don't pull in cruft from the Rust stdlib, which in early testing was where the biggest bloat arose from.

To my knowledge we've never done an assessment of real contracts to see if the size gains are actually meaningful. If the gains are not meaningful I will propose we remove the feature to reduce the surface area of what touches binaries between code and the wasm that runs on chain.

cc @brson

### LINKED ISSUE #2565: Make `--optimize` the default behavior of `stellar contract build`, or remove it

The `stellar contract build` command currently has an `--optimize` flag (gated behind the `additional-libs` feature) that is opt-in. This creates two problems:

1. If optimization is meaningful for on-chain contracts, then everyone should be using it, and it should be on by default. Making it opt-in means most users ship un-optimized wasm by accident.
2. If optimization is not meaningful enough to be on by default, then it doesn't need to be in the CLI at all and could be removed entirely, leaving users to install `wasm-opt` and run it independently.

Either way, the current state (an off-by-default first-class flag) is the worst of both worlds.

This also matters for build verification. As discussed in [stellar/stellar#1923 (comment)](https://github.com/orgs/stellar/discussions/1923#discussioncomment-16818349).

Pick one of the following:

**Option A (preferred): make `--optimize` the default.**
- `stellar contract build` optimizes the wasm by default.
- `--optimize=false` (or `--no-optimize`) disables it for edge cases.
- The `additional-libs` feature gating is removed, so optimization is always available in published binaries.
- Build verification metadata drops the optimize knob and assumes optimization is on.

**Option B: remove optimization from the CLI entirely.**
- The `--optimize` flag on `stellar contract build` is removed.
- The `stellar contract optimize` subcommand is also removed.
- Users who want optimization install `wasm-opt` themselves and run it independently.
- Build verification metadata drops the optimize knob.

### How to decide between Option A and Option B

1. Collect data on how beneficial optimization is for most production contracts (e.g. size reduction, instruction count, fee impact across a representative sample of deployed contracts).
2. Get feedback from developers in the ecosystem on whether they currently use `--optimize`, whether they proactively do not use it, and why.

cc @oceans404 @janewang 

- Leave the flag as-is (status quo).

### Relevant code

- `cmd/soroban-cli/src/commands/contract/build.rs` (the `--optimize` flag on `BuildArgs` and its conditional invocation).
- `cmd/soroban-cli/src/commands/contract/optimize.rs` (the standalone `optimize` subcommand).

fnando:
### What

`stellar contract build` now runs `wasm-opt` on the generated wasm by default. The previous behaviour required passing `--optimize`. To skip optimization, pass `--optimize=false`.

When `stellar-cli` is built without the `additional-libs` feature (which gates `wasm-opt`, OS keyring, and Ledger hardware-wallet support), the build still succeeds with an un-optimized wasm and prints a warning suggesting reinstalling with `--features additional-libs`.

### Why

Optimized wasm is what users almost always want to ship, and forgetting `--optimize` produced larger artifacts than necessary. Making it the default removes a footgun and matches what `stellar contract optimize` (already deprecated in favour of `build --optimize`) was encouraging.

Close https://github.com/stellar/stellar-cli/issues/2565
Close https://github.com/stellar/stellar-cli/issues/2284

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/Cargo.toml
+++ b/cmd/soroban-cli/Cargo.toml
@@ -31,7 +31,7 @@ doctest = false
 
 [features]
 default = ["additional-libs"]
-additional-libs = ["dep:keyring", "dep:stellar-ledger"]
+additional-libs = ["dep:wasm-opt", "dep:keyring", "dep:stellar-ledger"]
 emulator-tests = ["stellar-ledger/emulator-tests"]
 
 [dependencies]
@@ -80,7 +80,7 @@ reqwest = { version = "0.12.7", default-features = false, features = [
     "stream",
 ] }
 regex = "1.6.0"
-wasm-opt = { version = "0.116.1" }
+wasm-opt = { version = "0.116.1", optional = true }
 chrono = { version = "0.4.27", features = ["serde"] }
 rpassword = "7.2.0"
 toml = { workspace = true }
--- a/cmd/soroban-cli/Cargo.toml
+++ b/cmd/soroban-cli/Cargo.toml
@@ -31,7 +31,7 @@ doctest = false
 
 [features]
 default = ["additional-libs"]
-additional-libs = ["dep:wasm-opt", "dep:keyring", "dep:stellar-ledger"]
+additional-libs = ["dep:keyring", "dep:stellar-ledger"]
 emulator-tests = ["stellar-ledger/emulator-tests"]
 
 [dependencies]
@@ -80,7 +80,7 @@ reqwest = { version = "0.12.7", default-features = false, features = [
     "stream",
 ] }
 regex = "1.6.0"
-wasm-opt = { version = "0.116.1", optional = true }
+wasm-opt = { version = "0.116.1" }
 chrono = { version = "0.4.27", features = ["serde"] }
 rpassword = "7.2.0"
 toml = { workspace = true }
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -116,6 +116,8 @@ pub struct BuildArgs {
     pub optimize: bool,
 }
 
+// Manual impl so `optimize` defaults to `true`, matching the CLI default.
+// `#[derive(Default)]` would set it to `false`.
 impl Default for BuildArgs {
     fn default() -> Self {
         Self {
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -17,10 +17,8 @@ use std::{
 };
 use stellar_xdr::curr::{Limited, Limits, ScMetaEntry, ScMetaV0, StringM, WriteXdr};
 
-#[cfg(feature = "additional-libs")]
-use crate::commands::contract::optimize;
 use crate::{
-    commands::{global, version},
+    commands::{contract::optimize, global, version},
     print::Print,
     wasm,
 };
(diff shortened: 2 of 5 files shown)


cmd/soroban-cli/src/commands/contract/optimize.rs
```diff
@@ -22,31 +21,19 @@ pub enum Error {
...
-    #[cfg(not(feature = "additional-libs"))]
-    pub fn run(&self) -> Result<(), Error> {
-        Err(Error::Install)
-    }
-
-    #[cfg(feature = "additional-libs")]
     pub fn run(&self) -> Result<(), Error> {
         optimize(false, self.wasm.clone(), self.wasm_out.clone())
     }
 }
```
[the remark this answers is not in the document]
fnando: `stellar contract optimize` is already deprecated before this PR — the doc comment on the subcommand starts with "⚠️ Deprecated, use `build --optimize`" (`cmd/soroban-cli/src/commands/contract/mod.rs:85`) and the dispatcher prints a runtime deprecation warning at `mod.rs:197`. The PR body wording was imprecise ("now deprecated") and has been updated. No additional notice needed in `optimize.rs`.

fnando:
Thanks for the review. Replies to each item:

1. **`require_equals = true`** — Skipping. clap already validates bool values, so `--optimize garbage` fails with a clear "invalid value" error today; `--optimize mypackage` is not a real footgun because `build` takes no positionals. Adding `require_equals` would only restrict the space-separated form, which is a needless syntax restriction for a marginal error-message UX gain.

2. **`BuildArgs::default()` semantics** — Addressed. Added a comment on the manual `Default` impl noting that `optimize: true` matches the CLI default and `#[derive(Default)]` would give `false`. The `deploy/wasm.rs:320` callsite never reads `optimize` (the wasm is already built), so no behaviour change there.

3. **`wasm-opt` is now unconditional** — Updated the PR description to call out that `--no-default-features` builds now also pull in `wasm-opt`/`binaryen`. This is the intended trade-off.

4. **`doctor.rs` tautology** — Fixed. Removed the unconditional `print.checkln("Wasm optimization")` from `check_optional_features`. Wasm optimization is no longer an optional feature, so it does not belong in that listing.

5. **Removed `Error::OptimizeFeatureNotEnabled` variant** — Not concerned. `soroban-cli` is a binary crate, not a stable library API, so internal error variants are not part of a contract.

6. **Regression test for `BuildArgs::default().optimize == true`** — Skipping. Asserting the default in a unit test essentially restates the source line. A parse test for `--optimize=false` is meaningful but low-leverage for a single clap attribute. Manual verification with both forms passed.

leighmcculloch:
> wasm-opt out of additional-libs

I expect this breaks something. I think we should keep the additional lib inside additional libs. For the binary releases leaving it there will have no impact so no impact for most people.

leighmcculloch:
We should discuss:
- https://github.com/stellar/stellar-cli/issues/2284#issuecomment-3522594858

cmd/soroban-cli/Cargo.toml
```diff
@@ -31,7 +31,7 @@ doctest = false
 
 [features]
 default = ["additional-libs"]
-additional-libs = ["dep:wasm-opt", "dep:keyring", "dep:stellar-ledger"]
+additional-libs = ["dep:keyring", "dep:stellar-ledger"]
```
leighmcculloch: See comment above. I think it should stay in here. The intent with additional-libs is it's possible to build with a default rust toolchain.

leighmcculloch (commented):
One comment otherwise lgtm 🎉

cmd/soroban-cli/Cargo.toml
fnando: done!

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -101,18 +101,34 @@ pub struct Cmd {
...
 
-    /// Optimize the generated wasm.
-    #[cfg_attr(feature = "additional-libs", arg(long))]
-    #[cfg_attr(not(feature = "additional-libs"), arg(long, hide = true))]
+    /// Optimize the generated wasm. Enabled by default; pass `--optimize=false` to disable.
+    #[arg(
+        long,
+        default_value_t = true,
+        default_missing_value = "true",
+        num_args = 0..=1,
```
leighmcculloch, line 115: Am I correct in understanding that existing users and scripts that already pass `--optimize` will continue to work because of the num args being allowed to be `0`?

leighmcculloch (approved):
🎉

leighmcculloch:
I linked the PR as closing:
- https://github.com/stellar/stellar-cli/issues/2565
- https://github.com/stellar/stellar-cli/issues/2284

cmd/soroban-cli/src/commands/contract/build.rs
fnando: Correct! This is a neat trick and it works due to `default_missing_value`! Additionally, the following cases are also covered:

- `--optimize true` or `--optimize false` it also works, because they're the allowed values. 
- `--optimize ANYTHING_OTHER_THAN_TRUE_OR_FALSE` fails and you'll get a validation error.

PR #2585, opened 2026-05-22
Add `--verifiable` flag to `stellar contract build`

fnando:
### What

Adds a `--verifiable` flag to `stellar contract build` that performs a reproducible build inside a digest-pinned Docker container and stamps SEP-58 metadata (`bldimg`, `source_uri`, `source_sha256`, `bldopt`) into the resulting WASM so third parties can re-run the build and verify the output byte-for-byte.

New flags, all grouped under a `Verifiable` help section:

- `--verifiable` — opt in to the reproducible build mode; implies `--locked`, infers `--package` (the default-member cdylib contracts) when omitted, and requires a clean git working tree.
- `--image` — override the auto-selected container image. Must be digest-pinned (`...@sha256:...`); tag-only refs are rejected so `bldimg` stays content-addressed.
- `--source-sha256` — SEP-58 source identification: 64-char SHA-256 of the source. Optional — the archive is always generated and its SHA-256 computed for you. When supplied it acts as a pin: the build fails if it doesn't match the generated archive.
- `--source-uri` — SEP-58 source identification: URI where the source can be obtained (any scheme, e.g. `https://example.com/src.tar.gz`). Optional.
- `-d`/`--docker-host` (also reads `DOCKER_HOST`) — override the docker daemon endpoint.

A `--verifiable` build always generates the reproducible source archive, records its SHA-256 as `source_sha256`, writes a content-addressed copy to the data dir's `archives/<sha256>.tar.gz`, and builds from the extracted (permission-hardened) copy so the WASM comes from exactly the bytes that were hashed.

Each contract — explicit `--package` or inferred — is built with its own `--package`, which is forwarded to the build *and* recorded as a `bldopt`, so every WASM is independently reproducible and stable even if the workspace's default members change later. Multi-contract workspaces build every contract in a single container so the crates download, compiled dependencies, and `target/` are shared.

Every `bldopt` is recorded as valid shell syntax: each build option is shell-escaped once at the source, quoting only the value side, so a verifier can join the recorded `bldopt`s and replay the exact invocation through a shell. A flag like `--env B='this is very nice'` is stored as `--env=B='this is very nice'` (flag and key outside the quotes), which round-trips back to the original argument. Run with `--verbose` to print the full `docker run …` command the build executes (the same command surfaced in the error message if the container build fails).

### Source archives

The source archive is built by walking the working directory and tarring it, honoring the project's own `.gitignore`/`.ignore` files; the `.git` directory itself is always skipped. The archive is rooted at the current working directory — run `contract build --verifiable` / `contract archive` from the project (or workspace) root you want archived, so a workspace member's build still gets the whole workspace (its root `Cargo.toml`/`Cargo.lock`).

Selection depends only on the in-tree files plus the `.gitignore`/`.ignore` files inside the archived tree — never on machine-specific state (the global gitignore, `.git/info/exclude`, and parent-directory ignore files are not consulted) — so the same tree always hashes to the same `source_sha256` across machines. Paths that usually shouldn't ship but weren't ignored (`.env`, `target/`, `.idea`, …) are listed in a warning so you can add an ignore rule.

Because the archive is the working tree rather than a committed snapshot, both `contract build --verifiable` and `contract archive` refuse a dirty git tree (with a warning explaining why), so a recorded `source_sha256` always corresponds to a committed state. When the source isn't a git repo, there's nothing to check and they proceed.

### `stellar contract archive`

A standalone command to generate — or inspect — the same reproducible archive a verifiable build uses, sharing the exact byte-generation logic so the output is identical:

- `-o`/`--out-file <PATH>` — write the gzipped tarball. Must end in `.tar.gz` or `.tgz`. Required unless `--dry-run`.
- `--dry-run` — list the entries that would be archived plus the computed `source_sha256`, without writing anything. Handy for confirming contents before a verifiable build, or for producing the artifact to host at a `--source-uri`.

### Why

SEP-58 defines how to verify that a deployed contract WASM came from a specific source built with a specific toolchain image. Until now the CLI had no built-in way to produce such a build — users had to assemble the docker invocation, run cargo inside it, and stamp the custom sections by hand. This makes it a first-class option on `stellar contract build`, and the source archive (auto-generated on build, or produced/inspected via `stellar contract archive`) closes the loop by being the exact artifact that `source_sha256` refers to.

### Known limitations

- The verify side (download `source_uri`, check `source_sha256`, rebuild, byte-compare) is not part of this PR.
- A purely local docker image can't be used directly; reference it by digest (e.g. via a local registry).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2382,9 +2382,9 @@ checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
 
 [[package]]
 name = "globset"
-version = "0.4.16"
+version = "0.4.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "54a1028dfc5f5df5da8a56a73e6c153c9a9708ec57232470703592a3f18e49f5"
+checksum = "e47d37d2ae4464254884b60ab7071be2b876a9c35b696bd018ddcc76847309cd"
 dependencies = [
  "aho-corasick",
  "bstr",
@@ -2959,9 +2959,9 @@ dependencies = [
 
 [[package]]
 name = "ignore"
-version = "0.4.23"
+version = "0.4.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d89fd380afde86567dfba715db065673989d6253f42b88179abd3eae47bda4b"
+checksum = "d4ffa3a0547a138e59ddd6fa3b7c672ed47e6ad6a3cd177984ff1116aa5ba742"
 dependencies = [
  "crossbeam-deque",
  "globset",
@@ -5410,6 +5410,7 @@ dependencies = [
  "hex",
  "home",
  "humantime",
+ "ignore",
  "indexmap 2.11.0",
  "itertools 0.10.5",
  "jsonrpsee-types",
@@ -5464,7 +5465,6 @@ dependencies = [
  "tracing-subscriber",
  "ulid",
  "url",
- "walkdir",
  "wasm-gen",
  "wasm-opt",
  "wasmparser 0.116.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2791,7 +2791,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.4",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -4305,7 +4305,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash 2.1.1",
  "rustls",
- "socket2 0.5.10",
+ "socket2 0.6.4",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -4342,7 +4342,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.5.10",
+ "socket2 0.6.4",
  "tracing",
  "windows-sys 0.60.2",
 ]
@@ -5450,6 +5450,7 @@ dependencies = [
  "strsim",
  "strum 0.17.1",
  "strum_macros 0.17.1",
+ "tar",
  "tempfile",
  "termcolor",
  "termcolor_output",
@@ -6087,6 +6088,17 @@ name = "tap"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "55937e1799185b12863d447f42597ed69d9928686b8d88a1df17376a097d8369"
+
+[[package]]
+name = "tar"
+version = "0.4.46"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f6221d9a6003c78398e3b239969f352578258df48c8eb051caadae0015bc840"
+dependencies = [
+ "filetime",
+ "libc",
+ "xattr",
+]
 
 [[package]]
 name = "temp-dir"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2379,9 +2379,9 @@ checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
 
 [[package]]
 name = "globset"
-version = "0.4.16"
+version = "0.4.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "54a1028dfc5f5df5da8a56a73e6c153c9a9708ec57232470703592a3f18e49f5"
+checksum = "07c34a9410465b45bd9787443bc7370f37735bad04b0f0cd57ff1a3186c98988"
 dependencies = [
  "aho-corasick",
  "bstr",
@@ -2949,9 +2949,9 @@ dependencies = [
 
 [[package]]
 name = "ignore"
-version = "0.4.23"
+version = "0.4.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d89fd380afde86567dfba715db065673989d6253f42b88179abd3eae47bda4b"
+checksum = "00b69833ed729dc5aa7d19541d96d6cf8e9137194207a04916d658e43168402f"
 dependencies = [
  "crossbeam-deque",
  "globset",
@@ -4473,9 +4473,9 @@ dependencies = [
 
 [[package]]
 name = "regex-automata"
-version = "0.4.10"
+version = "0.4.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b9458fa0bfeeac22b5ca447c63aaf45f28439a709ccd244698632f9aa6394d6"
+checksum = "ad8553b9b26413251cbf30e620595c7a41b3887f03da04579c0e6b0d6a06b4b2"
 dependencies = [
  "aho-corasick",
  "memchr",
@@ -5383,6 +5383,7 @@ dependencies = [
  "hex",
  "home",
  "humantime",
+ "ignore",
  "indexmap 2.11.0",
  "itertools 0.10.5",
  "jsonrpsee-types",
@@ -5425,6 +5426,7 @@ dependencies = [
  "strsim",
  "strum 0.17.1",
  "strum_macros 0.17.1",
+ "tar",
  "tempfile",
  "termcolor",
  "termcolor_output",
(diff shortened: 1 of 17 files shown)


saimeunt:
@fnando Thanks for updating this PR to the latest SEP-58 revision, this is really helping prototyping a verification service.

While toying around with `stellar contract build --verifiable` I found it works great when overriding the auto-selected container image, but it failed at finding a correct default image for me.
The ideal path where things just work out of the box is preferred so we should be careful to have sane defaults.

The current approach to select the default image uses `CARGO_PKG_VERSION` from stellar-cli itself which will always be the stellar-cli version executing the `stellar contract build --verifiable` command, and it uses `rustc_version` which basically returns `rustc --version` which will be the local rust compiler version when building the contract.

I think we should try to parse the contract being built `Cargo.toml` instead and try to find appropriate values for the CLI version and rust version, for example in the [increment soroban example](https://github.com/stellar/soroban-examples/blob/main/increment/Cargo.toml) it would return `25.1.0` (pick `soroban-sdk` version under `[dependencies]`) and `1.89.0` (read `rust-version` under `[package]`).

When these values are not available or we can't find an appropriate image we can default to the latest image (and warn the user).

After fetching the correct values from the package being built `Cargo.toml`, we can select the appropriate image by rebuilding the tag and trying to find a corresponding image on Docker Hub (at the moment it will fail if the tag isn't a perfect match).

When listing images from Docker Hub using the `list_published_tags` function, there's an issue with the RegEx used to capture potential tag names (https://github.com/stellar/stellar-cli/pull/2585/changes#diff-2b0cca9e11d4682a5fb7c67d5a12fc8ec010305a77fce2cf5147b11e74bc112bR641): problematic regex is `^(\d+\.\d+\.\d+)-rust(\d+\.\d+\.\d+)$` (anchored `$`), so it won't match the latest available image which is `27.0.0-rust1.96.0-slim-trixie`.

My suggestion: wire the published tags search from Docker Hub not only when the wanted image isn't found, use it to find the correct image having the built package `Cargo.toml` CLI version and rust version found in its tag name using a fixed RegEx (to account for tags being suffixed eg. `-slim-trixie`).

This will fix the current behavior where not passing `--image` always results in a failure (failing to fetch `27.0.0-rust1.96.0` as it doesn't exist) and will improve the sane defaults of trying to find an appropriate image matching the built package CLI/rust versions when not specifying an image.

fnando:
@saimeunt the auto image selection still needs to be worked on. I'm focusing on explicit setting the image for now on my tests. We didn't have the images ready while I started doing this draft.

saimeunt:
@fnando I've tried using `stellar contract build --verifiable` in a GitHub action workflow to handle building an SEP-58 compliant contract WASM with its `source_uri` (also built in the same workflow using `stellar contract archive`) and `source_sha256` correctly set, but at the moment the contract build fails: https://github.com/walnuthq/soroban-increment-contract/actions/runs/28268028935/job/83759251347#step:5:22

```sh
ℹ️  Implying --locked because --verifiable was passed
ℹ️  Pulling image stellar/stellar-cli
ℹ️  Image digest: sha256:c1297d0c2c6790dda6afaa3edd39a959ec12edd6ebe30282dd1d7a663e7c4109
ℹ️  Downloaded newer image for stellar/stellar-cli@sha256:c1297d0c2c6790dda6afaa3edd39a959ec12edd6ebe30282dd1d7a663e7c4109
ℹ️  Inferred packages: soroban-increment-contract
ℹ️  Running verifiable build in docker.io/stellar/stellar-cli@sha256:c1297d0c2c6790dda6afaa3edd39a959ec12edd6ebe30282dd1d7a663e7c4109 (mount /home/runner/.local/share/stellar-cli/tmp/verifiable-src-gvLglZ/source:/source)
ℹ️  Running: docker run --rm -v /home/runner/.local/share/stellar-cli/tmp/verifiable-src-gvLglZ/source:/source -e RUSTUP_TOOLCHAIN=1.96.0-x86_64-unknown-linux-gnu docker.io/stellar/stellar-cli@sha256:c1297d0c2c6790dda6afaa3edd39a959ec12edd6ebe30282dd1d7a663e7c4109 contract build --locked --package=soroban-increment-contract --optimize --meta 'bldimg=docker.io/stellar/stellar-cli@sha256:c1297d0c2c6790dda6afaa3edd39a959ec12edd6ebe30282dd1d7a663e7c4109' --meta 'source_uri=https://github.com/walnuthq/soroban-increment-contract/releases/download/v0.0.1/soroban-increment-contract-v0.0.1-source.tar.gz' --meta source_sha256=cd59f1ed2a5b7674af6d4a1a3f63e6f28663188b23382d2c81d704cf64273e50 --meta bldopt=--locked --meta bldopt=--package=soroban-increment-contract --meta bldopt=--optimize
    ❌ error: `cargo metadata` exited with an error: error: could not find `Cargo.toml` in `/source` or any parent directory
```

My agent reported the root cause, here's an excerpt:

> Root cause (in the context of the CI run)
> 
> The relevant flow in the contract-verify branch:
> 
> 1. Materialize + harden — extract_into_hardened_tempdir() (source_archive.rs) unpacks the archive into a tempdir, then enforce_hardened_tree() chmods it to dirs 0o700, files 0o600, owned by whoever runs stellar (commit 7fb44cab).
> 2. Bind-mount — run_in_container() (verifiable.rs) builds the container with:
>   - binds: Some(vec!["<tempdir>:/source"]) — no :ro
>   - working_dir: Some("/source")
>   - User field not set, so the container runs as the image's default user.
> 3. The build runs cargo in /source and writes outputs to /source/target, which collect_built_contracts() later copies back out.
> 
> Now apply the CI environment:
> 
> - ubuntu-latest runs the stellar process as the runner user — uid ≠ 1000 (the GitHub runner is uid 1001). So the hardened tree is owned by 1001 with 0o700/0o600: no group/other bits at all.
> - The pinned image stellar/stellar-cli@sha256:c129… defaults to a non-root user (uid 1000), and bollard doesn't override User.
> - On native-Linux Docker, bind mounts pass host uid/perms straight through. uid 1000 is not the owner (1001) and the mode grants nothing to non-owners → EACCES traversing /source → cargo can't see Cargo.toml, and it can't create /source/target to write back.
> 
> Why it passed locally but failed in CI: Docker Desktop (macOS) remaps bind-mount ownership to the container user via virtiofs/gRPC-FUSE, so the uid mismatch is invisible. The GitHub-hosted Linux runner does no such remapping, so the 0o700/0o600 + uid-divergence combination bites. It's a host-vs-container uid mismatch that only surfaces on native-Linux bind mounts.

It also suggested potential fixes:

> Potential fixes
> 
> Since this is your PR branch, fixes can live in stellar-cli, in the workflow, or both.
> 
> A. Fix in stellar-cli — make the container user match the tree owner (most correct)
> - In run_in_container(), set user: Some(format!("{uid}:{gid}")) from geteuid()/getegid(). The container then is the owner, so 0o700/0o600 hardening is fully preserved and readable/writable.
> - Caveat: the image's pre-populated CARGO_HOME/HOME (owned by uid 1000) won't be writable by uid 1001, so cargo may fail. Mitigate by also passing HOME/CARGO_HOME env pointing at a writable dir (e.g. under /source or a tmpfs).
> 
> B. Fix in stellar-cli — chown the materialized tree to the image's default uid
> - docker inspect the image's Config.User, then chown the tempdir to that uid:gid before mounting. Keeps the container running as its native 1000 with its working toolchain; only the source ownership changes.
> - Caveat: chowning to a different uid needs root/sudo on the host (the CI runner can sudo, a plain dev may not).
> 
> C. Fix in stellar-cli — decouple the hardened verification copy from the mounted build copy
> - Keep the owner-only hardened tree for hashing/verification, but mount a separate copy with widened modes (dirs 0o755, files 0o644) for the container. The mounted bytes are already-public archived source, so widening read access there doesn't defeat the intent of 7fb44cab (which guards the local extraction). Simplest, needs no root and no uid juggling.
> 
> D. Fix in stellar-cli — read-only source + separate writable target
> - Mount /source as :ro and redirect cargo output via CARGO_TARGET_DIR=/target to a separate writable mount owned by the container user. Solves the write-back half cleanly, but does not solve the read half on its own — still needs A, B, or C for the EACCES on read.
> 
> E. Fix in stellar-cli — pass-through --user / build option
> - Add a flag so the caller chooses the container user, letting the workflow set --user 1000:1000 (or matching). Gives control to the caller; combine with the workflow change in F.
> 
> F. Workflow-only mitigation (no CLI change)
> - Run the build step as a uid that matches the image default (1000) so the hardened tree is owned by 1000 — e.g. create/sudo -u a uid-1000 user to invoke stellar contract build, or run the job in a container: whose user is 1000. Awkward (the runner is 1001, docker-in-docker concerns) and brittle, but unblocks without touching stellar-cli.
> 
> Recommendation (not implementing yet)
> 
> Option A is the canonical fix for bind-mounted build dirs and keeps the 7fb44cab hardening fully intact — provided we also redirect CARGO_HOME/HOME so cargo has a writable home under the matched uid. C is the lowest-risk, root-free fallback if reconciling the image's CARGO_HOME ownership proves fiddly. D is worth pairing with whichever read-fix you choose, so the target write-back is robust regardless of mount semantics.

It looks like Option A would be the correct way to go here, wdyt?

fnando:
Closing in favor of a fresh PR with a clean commit history. Superseded by a new PR from the `contract-verifiable-build` branch.

fnando:
Superseded by #2709 (fresh branch `contract-verifiable-build`, clean single-commit history).

PR #2589, opened 2026-05-26
Upgrade to stable dependencies.

fnando:
### What

Upgrades workspace dependencies to their stable releases: `stellar-xdr` to `26.0.1`, `stellar-strkey` to `0.0.16`, the `soroban-*` crates (`soroban-spec`, `soroban-spec-rust`, `soroban-sdk`, `soroban-env-host`, `soroban-token-sdk`, `stellar-asset-spec`, `soroban-ledger-snapshot`) to `26.0.0`, and `stellar-rpc-client` to `26.0.0`. Adapts call sites for two breaking API changes: `stellar-strkey`'s inherent `to_string()` now returns `heapless::String<N>`, so call sites that need `std::string::String` use `format!("{x}")` to go through `Display`; and `soroban-spec-rust::generate_without_file` now returns `Result<TokenStream, GenerateError>`, propagated via `?` with a new variant on the local error enum.

### Why

Move the workspace off pre-release/RC versions onto the stable releases.

### Known limitations

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1194,7 +1194,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


leighmcculloch (approved):
Fyi most of the strkey changes are going to break in the next release of that lib due to large changes made to the APIs, the existing usage may be easier to change, maybe.

fnando:
@leighmcculloch that's fine for now. We just released stellar-rpc-client using that same version, so we can revisit when the time comes. :)

PR #2605, opened 2026-06-04
Build Docker images from source instead of downloading binaries.

fnando:
### What

Build the `stellar/stellar-cli` Docker image from source via `cargo install` inside the Dockerfile, and produce a multi-arch image by building each architecture on its own native runner before merging them into a single manifest.

The `docker.yml` workflow now runs a matrix `build` job — `linux/amd64` on `ubuntu-latest` and `linux/arm64` on `ubuntu-24.04-arm` — that pushes each platform image by digest, then a `merge` job assembles those digests into one manifest list (`docker buildx imagetools create`) under the computed tags. QEMU emulation is no longer used.

### Why

The current Docker image is broken (see https://github.com/stellar/stellar-cli-docker/issues/16). The move to building the CLI from source means the image is compiled during the Docker build rather than copied in from a prebuilt binary. Compiling under QEMU emulation for a foreign architecture is prohibitively slow, so each architecture is built on a matching native runner and the results are combined into a single multi-arch manifest at publish time.

This is a temporary solution until `stellar/stellar-cli-docker` publishes images.

### Known limitations

This is a stopgap; image publishing is expected to move to `stellar/stellar-cli-docker`.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -16,19 +16,69 @@ defaults:
   run:
     shell: bash
 
+env:
+  REGISTRY_IMAGE: stellar/stellar-cli
+
 jobs:
-  docker:
+  # Resolve the ref to a single immutable SHA and compute the Docker tags once,
+  # so both platform builds and the published manifest refer to one commit even
+  # if the branch advances while the workflow runs.
+  prepare:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    outputs:
+      sha: ${{ steps.resolve.outputs.sha }}
+      tags: ${{ steps.resolve.outputs.tags }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref }}
           fetch-depth: 0
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@06116385d9baf250c9f4dcb4858b16962ea869c3 # v4
+      # Resolve the ref to a SHA and compute Docker tags.
+      # - Version tag (e.g. v1.2.3): push versioned + latest tags.
+      # - Any other ref: push a tag for the resolved commit SHA.
+      - name: Resolve ref and tags
+        id: resolve
+        run: |
+          ref="${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref_name }}"
+          sha="$(git rev-parse HEAD)"
+          echo "sha=${sha}" >> $GITHUB_OUTPUT
+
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -16,37 +16,70 @@ defaults:
...
-      - name: Install build dependencies
-        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends libudev-dev libdbus-1-dev
+      - name: Build and push by digest
+        id: build
+        uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7
+        with:
+          context: .
+          platforms: ${{ matrix.platform }}
+          build-args: |
+            STELLAR_CLI_REV=${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref_name }}
```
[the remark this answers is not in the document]
fnando: Good catch. Addressed in the latest push: a new `prepare` job resolves the ref to a single immutable SHA (`git rev-parse HEAD`) once, and both the matrix builds and the `merge` job consume that SHA — for `STELLAR_CLI_REV`, the checkout, and the SHA tag. So both arch images and the published tag always refer to one commit, even if the branch advances mid-run. Passing a SHA also avoids the unreliable `--rev <branch>` resolution.

Dockerfile
```diff
@@ -3,15 +3,25 @@ FROM rust:latest
...
+    pkg-config && \
     rm -rf /var/lib/apt/lists/*
 
-ARG TARGETARCH
-COPY stellar-${TARGETARCH}/stellar /usr/local/bin/stellar
-RUN chmod +x /usr/local/bin/stellar
+ARG STELLAR_CLI_REV
+RUN cargo install --locked \
+        --git https://github.com/stellar/stellar-cli.git \
+        --rev "${STELLAR_CLI_REV}" \
```
mootz12, line 19: I still prefer if this built from local, but not enough to hold anything up

.github/workflows/docker.yml
mootz12: unrelated, but noticed we hardcode these secrets into the curl command. Probably good cleanup to use env vars here. Github Actions log stream hides these but probably better safe than sorry.

mootz12 (approved):
LG2M.

Two notes inline

.github/workflows/docker.yml
fnando: Good call. Switched the description step to pass the credentials via `env:` and build the login payload with `jq` from `"$DOCKERHUB_USERNAME"`/`"$DOCKERHUB_TOKEN"`, so the secrets are no longer interpolated into the command string.

Dockerfile
fnando: Agreed it would be cleaner to build from the local tree. Keeping `--git` for now since this workflow is a temporary stopgap until `stellar/stellar-cli-docker` publishes images — happy to revisit if it outlives that.

PR #2614, opened 2026-06-17
Update stellar-strkey to 0.0.18

leighmcculloch:
### What

Bump the `stellar-strkey` workspace dependency from 0.0.16 to 0.0.18. This change targets the stellar-strkey release line for protocol 28. The workspace-direct crates now resolve to strkey 0.0.18, while published transitive consumers pinned to `^0.0.16` (sep5, soroban-sdk, stellar-rpc-client) and the transitive 0.0.13 via stellar-xdr remain unchanged, so the lockfile carries three strkey nodes.

### Why

Adopt the latest stellar-strkey release for the protocol 28 line. Version 0.0.17 replaced the single catch-all `DecodeError::Invalid` variant with a set of specific variants, so the two hex-fallback code paths in `contract_id_from_str` (in soroban-spec-tools and soroban-cli) were updated to map their failures to `DecodeError::InvalidPayloadLength`, preserving the prior single-error behavior. It also redacts private keys behind `Unredacted`, so the CLI's secret-printing paths and tests now use `.as_unredacted()`, and removes positional input for the embedded `strkey encode`/`decode` commands in favour of stdin.

0.0.18 specifically fixes the bug that blocked the 0.0.17 bump: the embedded `strkey encode` command rejected owned JSON keys when deserializing `Decoded<Strkey>`, failing the `strkey` integration test. stellar/rs-stellar-strkey#125 fixes it.

### Known limitations

N/A

---

## Protocol 28 coordination

⚠️ This change targets versions of software that will ship for **Protocol 28**, and should **not** be merged unless this repository is ready to accept Protocol 28 changes.

This is one of a coordinated set of PRs bumping `stellar-strkey` to **0.0.18** across the Protocol 28 component stack. It is preferred that all Protocol 28 releases of these components depend on the **same** version of `stellar-strkey`:

- stellar/rs-stellar-xdr#547
- stellar/rs-soroban-env#1694
- stellar/rs-soroban-sdk#1913
- stellar/rs-stellar-rpc-client#99
- stellar/stellar-cli#2614

`stellar-strkey` 0.0.18 raises the minimum supported Rust version to 1.87 and updates/adds transitive dependencies (`heapless` 0.9, `zeroize`), so MSRV and supply-chain policy (cargo-deny / cackle) updates may be required in some repositories before CI is green.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ version = "26.0.0"
 
 # Dependencies from elsewhere shared by crates:
 [workspace.dependencies]
-stellar-strkey = "0.0.16"
+stellar-strkey = "0.0.17"
 sep5 = "0.1.0"
 base64 = "0.21.2"
 thiserror = "1.0.46"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ version = "26.0.0"
 
 # Dependencies from elsewhere shared by crates:
 [workspace.dependencies]
-stellar-strkey = "0.0.17"
+stellar-strkey = "0.0.18"
 sep5 = "0.1.0"
 base64 = "0.21.2"
 thiserror = "1.0.46"
@@ -137,3 +137,9 @@ lto = true
 [profile.release-with-panic-unwind]
 inherits = "release"
 panic = "unwind"
+
+# Temporary: stellar-strkey 0.0.18 is not yet published to crates.io. Point at
+# the release/v0.0.18 branch so this PR can be validated against the actual
+# 0.0.18 code ahead of the release. Remove once 0.0.18 is published.
+[patch.crates-io]
+stellar-strkey = { git = "https://github.com/stellar/rs-stellar-strkey", rev = "73bf43778a4acad5e7b3589f54a5ec3607c4b7b1" }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -76,7 +76,7 @@ version = "28.0.0-rc.1"
 
 # Dependencies from elsewhere shared by crates:
 [workspace.dependencies]
-stellar-strkey = "0.0.16"
+stellar-strkey = "0.0.18"
```
leighmcculloch: Re Copilot's pinning comment on this line: keeping as is. For `0.0.x` versions Cargo's caret requirement `0.0.18` already means `>=0.0.18, <0.0.19`, so it can only resolve to 0.0.18.

---

Cargo.toml
leighmcculloch: CI note: the `Test with soroban-examples` / `build-example (*, true)` jobs fail, but not because of this PR. They fail identically on `main` at 5e60f63 (run 35973405490). The workflow's `experimental_spec_shaking_v2: true` leg adds that feature to each example's `soroban-sdk` dependency, and soroban-examples `main` is now on soroban-sdk 28.0.0, which no longer has that feature (`package ... depends on soroban-sdk with feature experimental_spec_shaking_v2 but soroban-sdk does not have that feature`). I couldn't find an existing fix to port. The fix belongs in `.github/workflows/test-with-soroban-examples.yml` (drop or rework the `true` leg for SDK >= 28), outside this PR.

---

sisuresh (approved):

PR #2618, opened 2026-06-22
Fix descriptions for contract id commands

fnando:
### What

Corrects the help text for the `stellar contract id asset` and `stellar contract id wasm` subcommands. They previously read "Deploy builtin Soroban Asset Contract" and "Deploy normal Wasm Contract", which is wrong — neither deploys anything. They now read "Derive the contract id for a builtin Stellar Asset Contract" and "Derive the contract id for a Wasm contract". `FULL_HELP_DOCS.md` was regenerated to match.

### Why

These subcommands derive and print a contract ID; they do not deploy. The old descriptions were inaccurate.

### Known limitations

N/A

### DIFF
--- a/FULL_HELP_DOCS.md
+++ b/FULL_HELP_DOCS.md
@@ -549,12 +549,12 @@ Generate the contract id for a given contract or asset
 
 ###### **Subcommands:**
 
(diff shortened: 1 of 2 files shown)


mootz12 (approved):
Good catch!

PR #2629, opened 2026-06-26
Shell out to docker instead of using bollard

fnando:
### What

Replaces the `bollard` Docker API client with direct `docker` CLI invocations in the `stellar container start|stop|logs` commands. The `--docker-host` flag (and the `DOCKER_HOST` env var) is now applied by setting `DOCKER_HOST` on the spawned `docker` process; host resolution is otherwise left to docker itself. The unused `bollard` dependency in `stellar-ledger` is also removed.

### Why

The bollard connection code copied unreleased upstream source, pinned a hardcoded API version, and reimplemented host/socket resolution plus a Docker Desktop fallback that the `docker` CLI already handles natively. It also rejected any host scheme it didn't recognize — e.g. `ssh://`, which docker supports — with `URI scheme is not supported`. Shelling out to `docker` removes that hand-rolled logic and lets docker resolve the connection, including remote `ssh://` hosts.

```
$ stellar container start --docker-host ssh://fnando@fnando-server.local
ℹ️ Starting local network
ℹ️ testing: Pulling from stellar/quickstart
ℹ️ Digest: sha256:3b98cc624a235215bbb2dedd18bd3ec3a498053b6b6533533b6a0d94354a7885
ℹ️ Status: Image is up to date for stellar/quickstart:testing
✅ Started container
🔎 Watch logs with `stellar container logs local --docker-host ssh://fnando@fnando-server.local`
ℹ️ Stop the container with `stellar container stop local --docker-host ssh://fnando@fnando-server.local`

$ stellar container start --docker-host ssh://fnando@fnando-server.local
ℹ️ Starting local network
ℹ️ testing: Pulling from stellar/quickstart
ℹ️ Digest: sha256:3b98cc624a235215bbb2dedd18bd3ec3a498053b6b6533533b6a0d94354a7885
ℹ️ Status: Image is up to date for stellar/quickstart:testing
❌ error: a container named "stellar-local" already running

$ stellar container logs --docker-host ssh://fnando@fnando-server.local
Starting Stellar Quickstart
versions:
  quickstart: 16a57418b555ab859e133550c449ff15c2f5c06e
  xdr:
    stellar-xdr 27.0.0 (5262803470be965e42f80023d12fba12808c774a)
    xdr: 68fa1ac55692f68ad2a2ca549d0a283273554439
    features: <none>
  ...

$ stellar container stop --docker-host ssh://fnando@fnando-server.local
ℹ️ Stopping local container
✅ Container stopped

$ stellar container stop --docker-host ssh://fnando@fnando-server.local
ℹ️ Stopping local container
❌ error: container local not found
```

### Known limitations

The `docker` binary must now be installed and on the user's `PATH`; if it isn't, commands fail with a "is docker installed and on your PATH?" hint.


cmd/soroban-cli/src/commands/container/shared.rs
```diff
@@ -48,61 +39,15 @@ impl Args {
...
-                    Err(e)?
-                }
-            }
+    /// Builds a `docker` command, setting `DOCKER_HOST` in the process environment when a
+    /// `--docker-host` (or `DOCKER_HOST` env) override is provided. Host resolution is otherwise
+    /// left to the docker CLI itself.
+    pub(crate) fn docker_command(&self) -> Command {
+        let mut cmd = Command::new("docker");
+        if let Some(host) = &self.docker_host {
+            cmd.env("DOCKER_HOST", host);
```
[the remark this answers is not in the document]
fnando: Good catch. Fixed in 24c61ca7 — `docker_command()` now passes the override as `-H <host>` on the command line instead of setting the `DOCKER_HOST` env var. The `-H` flag outranks `DOCKER_CONTEXT`, so the override is honored even when a context is active.

cmd/soroban-cli/src/commands/container/start.rs
```diff
@@ -85,83 +80,69 @@ impl Runner {
...
         Ok(())
     }
 
+    async fn pull_image(&self, image: &str) {
+        let mut cmd = self.args.container_args.docker_command();
+        cmd.args(["pull", image]).stdout(Stdio::piped());
+
+        let Ok(mut child) = cmd.spawn() else {
+            return self.warn_image_fetch(image);
+        };
```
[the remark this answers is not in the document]
fnando: Fixed in 24c61ca7 — `pull_image` now returns quietly when `docker pull` fails to spawn with `ErrorKind::NotFound`, letting the main `docker run` invocation surface the proper "is docker installed and on your PATH?" hint.

cmd/soroban-cli/src/commands/container/start.rs
```diff
@@ -85,83 +80,69 @@ impl Runner {
...
-                _ => Error::CreateContainerFailed(e),
-            })?;
+        let output = cmd.output().await.map_err(ConnectionError::from)?;
+        if !output.status.success() {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            if stderr.contains("already in use") {
+                return Err(Error::ContainerAlreadyRunning(container_name));
+            }
+            return Err(Error::CreateContainerFailed(stderr.trim().to_string()));
+        }
```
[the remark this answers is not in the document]
fnando: Agreed these are worth covering. The stderr-matching paths are the brittle bits — I will add fake-`docker`-on-`PATH` tests for DOCKER_HOST propagation, already-running, and not-found in a follow-up rather than expanding this dependency-removal PR.

cmd/soroban-cli/src/commands/container/logs.rs
```diff
@@ -28,24 +22,22 @@ pub struct Cmd {
...
-        );
-
-        while let Some(log) = logs_stream.try_next().await? {
-            print!("{log}");
+
+        // Stream logs straight to the terminal by inheriting stdio.
+        let status = self
+            .container_args
+            .docker_command()
+            .args(["logs", "-f", "--tail", "all", &container_name])
```
mootz12, line 32: Just flagging this has a slight difference in where stuff gets logged.

The bollard version logs both the containers `stderr` and `stdout` to `stdout`. This one preserves it, which probably seems better.

cmd/soroban-cli/src/commands/container/start.rs
```diff
@@ -85,83 +80,73 @@ impl Runner {
...
-            .start_container(&create_container_response.id, None::<StartContainerOptions>)
-            .await?;
         self.print.checkln("Started container");
         self.print_instructions();
         Ok(())
     }
 
+    async fn pull_image(&self, image: &str) {
+        let mut cmd = self.args.container_args.docker_command();
+        cmd.args(["pull", image]).stdout(Stdio::piped());
```
mootz12, line 115: Do we want any `stderr` to log out if the `--quiet` flag is passed?

mootz12 (approved):
A few notes in-line, looks good tho

cmd/soroban-cli/src/commands/container/logs.rs
fnando: Yeah, intentional. Inheriting stdio keeps the container's stdout and stderr separated on the CLI side (pipeable, greppable), which felt like the better default. Left it as-is.

cmd/soroban-cli/src/commands/container/start.rs
fnando: Good catch. `docker run`'s stderr is already captured and only surfaced on error, so `--quiet` is silent on success, and `docker logs` inherits stdio by design. `docker pull` was the gap — it writes progress to stderr, which leaked under `--quiet`. Fixed in 6eb74bb0 by redirecting the pull's stderr to null when `--quiet` is set.

Issue #2620, opened 2026-06-24
stellar token: typed SEP-41 + SAC client


aristidesstaffieri:
Part of the **Improve Agent UX for Wallets** epic.

## Summary
Add `stellar token`, a typed SEP-41 + SAC client. Each subcommand is a thin wrapper over the existing `contract invoke` pipeline with token-aware argument parsing, asset/contract resolution, and decimal-aware amount formatting. Because the underlying call is a contract invocation, every state-changing subcommand inherits the full pipeline (simulate → auth-entry signing → tx signing → fee-bump → submit → poll) and returns a decoded receipt. Sane default are included for settings like the transaction fee.

## Subcommands
| Group | Subcommands |
| :---- | :---- |
| Transfer | `transfer`, `transfer-from` |
| Read | `balance`, `allowance`, `name`, `symbol`, `decimals` |
| Allowance | `approve` |
| Supply | `burn`, `burn-from` |
| SAC admin | `mint`, `set-admin`, `clawback`, `set-authorized` |

## Target resolution
- `--id C…` or an alias targets any SEP-41/SAC contract directly, reusing `UnresolvedContract` from `config/alias.rs`.
- Special-case `native` — `--id native` resolves the native SAC.
- Reuse `contract id asset` resolution + `utils::contract_id_hash_from_asset`; asset parsing in `tx/builder/asset.rs`.
- SEP-41 reference implementation: `cmd/crates/soroban-test/tests/fixtures/test-wasms/token/src/contract.rs`.

## New files
`cmd/soroban-cli/src/commands/token/{mod.rs, transfer.rs, balance.rs, approve.rs, allowance.rs, burn.rs, mint.rs, admin.rs, args.rs}` — subcommands plus shared resolution/formatting in `args.rs`.

## Error paths
- **SAC not deployed** → structured error pointing at `contract deploy asset`.
- **Missing trustline / authorization** → named error; provisioning left to a skill using `tx new change_trust`.

## Acceptance criteria
- [ ] `stellar token transfer --id native --to <dst> --amount <amt>` on testnet resolves the SAC, simulates, signs, submits, polls, and returns a JSON receipt with tx hash + decoded result — no caller-supplied seq/fee.
- [ ] Issued-asset transfer succeeds with a deployed SAC + trustlines; returns the deploy-pointer error when the SAC is missing; returns the named trustline error when the trustline is missing.
[...]

PR #2644, opened 2026-07-13
Add `stellar token transfer` for SEP-41 and Stellar Asset Contracts

fnando:
### What

Adds `stellar token`, starting with the `transfer` subcommand — a typed SEP-41 / Stellar Asset Contract (SAC) transfer that wraps the existing `contract invoke` pipeline. It resolves a token target from `--id` (`native` → native SAC, `CODE:ISSUER` → that asset's SAC, or a `C…` contract id / alias), builds a `transfer(from, to, amount)` invocation, and runs it through the full pipeline (simulate → auth-sign → sign → fee-bump → submit → poll). `--from` names the sending account (required; it signs and authorizes), `--to` is the destination, and `--amount` is the raw stroop amount. There are no caller-supplied fee/sequence flags — sane defaults are used. With `--output json` the command returns a machine-readable receipt `{ "tx_hash", "result" }`; in JSON mode the pipeline runs quietly so stdout stays clean. Missing SACs return a structured error pointing at `contract asset deploy`.

To surface the transaction hash, `contract invoke` gains a minimal `execute_with_receipt` returning `{ tx_hash, output }`; its existing `execute`/`invoke` entry points are unchanged.

```console
$ $stellar token transfer --id foo --from dist --to fnando --amount 10000000
ℹ️ Simulating transaction…
ℹ️ Signing transaction: 0b18bae5f344e223e79576e58ec8a5b3f27bb447b42e87f2d09c493fae7b642c
🌎 Sending transaction…
📅 CDF7U4HZVRRONHY7IADHAQPMH2LX2CJHEA7S3IGAPETIJ7DXP7IVIFYF - Success - Event: TransferWithAmountOnly (transfer), from: "GAZVY4EBCVZB2Q622J7OKGR3YYC524PPKREYPYAPHGAUOOUMDLVTNJXP", to: "GAF4UUODFGAAMYRTF5QKUZCCZPXF3S4PRU5NS2BBRVJGX4WLRVI4ZI4Z", amount: "10000000"
0b18bae5f344e223e79576e58ec8a5b3f27bb447b42e87f2d09c493fae7b642c

$ $stellar token transfer --id foo --from dist --to fnando --amount 10000000 --output=json-formatted
{
  "tx_hash": "9339aecba00925b701e36fba5ab5352c0e8e0f594cb59558f4c211aec040a8c2",
  "result": null
}
```

### Why

Part of #2620 (the "Improve Agent UX for Wallets" epic). Agents and wallets need an ergonomic, typed way to move tokens and get back a decoded receipt with a tx hash, instead of hand-crafting `contract invoke` slop and scraping stderr. This lands the `transfer` slice plus the shared `token` module foundation (target resolution + output format) that the remaining subcommands will reuse.

### Known limitations

- Amounts are raw stroops (`i128`); decimal-aware parsing is not included.
- Only `transfer` is implemented; `balance`, `allowance`, `approve`, `burn`, `mint`, and the SAC-admin subcommands are follow-ups.
- JSON-clean output is achieved by running the invoke pipeline quietly; migrating `contract invoke` itself to the `Output` system is a separate follow-up.
- Transferring to an account that has no trustline for the asset fails, as it should, but we don't wrap it in a custom "missing trustline" error. The SAC reports this only as a generic `Error(Contract, #13)`; the human-readable reason (`"trustline entry is missing for account G…"`) comes through the transaction's diagnostic events, which the command already prints. We rely on that diagnostic rather than matching on `#13`, since that number is the SAC contract's internal error enum — not a stable API — and hard-coding it would break silently if it ever changed.

```console
$ stellar token transfer --id foo --from dist --to fnando --amount 10000000
❌ error: transaction simulation failed: HostError: Error(Contract, #13)

  Event log (newest first):
     0: [Diagnostic Event] contract:CDF7U4HZVRRONHY7IADHAQPMH2LX2CJHEA7S3IGAPETIJ7DXP7IVIFYF, topics:[error, Error(Contract, #13)], data:["trustline entry is missing for account", GAF4UUODFGAAMYRTF5QKUZCCZPXF3S4PRU5NS2BBRVJGX4WLRVI4ZI4Z]
     1: [Diagnostic Event] topics:[fn_call, CDF7U4HZVRRONHY7IADHAQPMH2LX2CJHEA7S3IGAPETIJ7DXP7IVIFYF, transfer], data:[GAZVY4EBCVZB2Q622J7OKGR3YYC524PPKREYPYAPHGAUOOUMDLVTNJXP, GAF4UUODFGAAMYRTF5QKUZCCZPXF3S4PRU5NS2BBRVJGX4WLRVI4ZI4Z, 10000000]
```

### DIFF
--- a/cmd/crates/soroban-test/tests/it/integration/token/mod.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/token/mod.rs
@@ -1,0 +1,81 @@
+pub mod transfer;
+
+use soroban_test::{AssertExt, TestEnv};
+
+/// Establish (and auto-authorize) a trustline from `source` to `asset`.
+pub fn add_trustline(sandbox: &TestEnv, source: &str, asset: &str) {
+    sandbox
+        .new_assert_cmd("tx")
+        .args(["new", "change-trust", "--line", asset, "--source", source])
+        .assert()
+        .success();
+}
+
+/// Pay `amount` of `asset` from its issuer to `destination`.
+pub fn issuer_pays(sandbox: &TestEnv, issuer: &str, destination: &str, asset: &str, amount: i128) {
+    sandbox
+        .new_assert_cmd("tx")
+        .args([
+            "new",
+            "payment",
+            "--destination",
+            destination,
+            "--asset",
+            asset,
+            "--amount",
+            &amount.to_string(),
+            "--source",
+            issuer,
+        ])
+        .assert()
+        .success();
+}
+
+/// Deploy the Stellar Asset Contract for `asset`, tolerating the case where a
+/// prior run already deployed it (the SAC is global to the network).
+pub fn deploy_sac(sandbox: &TestEnv, asset: &str, source: &str) {
+    sandbox
+        .new_assert_cmd("contract")
+        .args([
+            "asset",
+            "deploy",
+            "--asset",
+            asset,
+            "--source-account",
+            source,
+        ])
+        .output()
+        .expect("failed to run contract asset deploy");
+}
+
+/// The contract id of the SAC for `asset`.
+pub fn sac_id(sandbox: &TestEnv, asset: &str) -> String {
+    sandbox
+        .new_assert_cmd("contract")
+        .args(["id", "asset", "--asset", asset])
+        .assert()
+        .success()
+        .stdout_as_str()
+}
+
+/// Read a token's balance for `account` through its Stellar Asset Contract,
+/// returning the raw stroop amount.
+pub fn sac_balance(sandbox: &TestEnv, contract_id: &str, account: &str) -> i128 {
+    let stdout = sandbox
+        .new_assert_cmd("contract")
+        .args([
+            "invoke",
+            "--id",
+            contract_id,
+            "--source-account",
+            "test",
+            "--",
+            "balance",
+            "--id",
+            account,
+        ])
+        .assert()
+        .success()
+        .stdout_as_str();
+    stdout.trim().trim_matches('"').parse().unwrap()
+}
--- a/cmd/crates/soroban-test/tests/it/integration/token/mod.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/token/mod.rs
@@ -34,7 +34,7 @@ pub fn issuer_pays(sandbox: &TestEnv, issuer: &str, destination: &str, asset: &s
 /// Deploy the Stellar Asset Contract for `asset`, tolerating the case where a
 /// prior run already deployed it (the SAC is global to the network).
 pub fn deploy_sac(sandbox: &TestEnv, asset: &str, source: &str) {
-    sandbox
+    let output = sandbox
         .new_assert_cmd("contract")
         .args([
             "asset",
@@ -46,6 +46,18 @@ pub fn deploy_sac(sandbox: &TestEnv, asset: &str, source: &str) {
         ])
         .output()
         .expect("failed to run contract asset deploy");
+
+    // A clean deploy succeeds; the only failure we tolerate is the SAC already
+    // existing on this (persistent) network. Any other failure is a real bug —
+    // surface the captured stderr instead of swallowing it and misdirecting the
+    // failure two steps downstream.
+    if !output.status.success() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        assert!(
+            stderr.contains("contract already exists"),
+            "contract asset deploy failed: {stderr}"
+        );
+    }
 }
 
 /// The contract id of the SAC for `asset`.
--- a/cmd/crates/soroban-test/tests/it/integration/token/transfer.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/token/transfer.rs
@@ -58,6 +58,38 @@ async fn transfer_native_returns_receipt_and_moves_funds() {
 }
 
 #[tokio::test]
+async fn transfer_to_contract_destination_succeeds() {
+    let sandbox = &TestEnv::new();
+
+    deploy_sac(sandbox, "native", "test");
+    let native_id = sac_id(sandbox, "native");
+
+    // Any deployed contract can hold SAC balances (no trustline needed), so use
+    // another SAC's contract id as a `C…` destination — this exercises the
+    // contract-address path that `--to` must accept.
+    let issuer = new_account(sandbox, "issuer");
+    let asset = format!("USDC:{issuer}");
+    deploy_sac(sandbox, &asset, "issuer");
+    let contract_dest = sac_id(sandbox, &asset);
+
+    let amount: i128 = 5_000_000;
+    let before = sac_balance(sandbox, &native_id, &contract_dest);
+
+    let receipt = transfer_json(sandbox, "native", &contract_dest, amount);
+    assert!(
+        receipt["tx_hash"].as_str().is_some_and(|h| !h.is_empty()),
+        "expected a non-empty tx_hash, got: {receipt}"
+    );
+
+    let after = sac_balance(sandbox, &native_id, &contract_dest);
+    assert_eq!(
+        after,
+        before + amount,
+        "contract recipient balance should increase"
+    );
+}
+
+#[tokio::test]
 async fn transfer_issued_asset_succeeds_with_deployed_sac_and_trustlines() {
     let sandbox = &TestEnv::new();
     let test = test_address(sandbox);
--- a/cmd/crates/soroban-test/tests/it/integration/token/transfer.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/token/transfer.rs
@@ -189,6 +189,29 @@ async fn transfer_rejects_negative_amount_before_any_rpc() {
 }
 
 #[tokio::test]
+async fn transfer_rejects_muxed_source_with_clear_error() {
+    let sandbox = &TestEnv::new();
+    let recipient = new_account(sandbox, "recipient");
+
+    deploy_sac(sandbox, "native", "test");
+
+    // Muxed (M…) source accounts aren't supported by the invoke pipeline yet
+    // (see #2645). Until then the command must reject them up front with a clear
+    // message rather than a raw strkey decode error deep in the pipeline.
+    let muxed = "MA3D5KRYM6CB7OWQ6TWYRR3Z4T7GNZLKERYNZGGA5SOAOPIFY6YQGAAAAAAAAAPCICBKU";
+    sandbox
+        .new_assert_cmd("token")
+        .args([
+            "transfer", "--id", "native", "--to", &recipient, "--amount", "1", "--from", muxed,
+        ])
+        .assert()
+        .failure()
+        .stderr(predicates::str::contains(
+            "muxed (M…) source accounts are not yet supported",
+        ));
+}
+
+#[tokio::test]
 async fn transfer_json_failure_returns_error_envelope_on_stdout() {
     let sandbox = &TestEnv::new();
     let test = test_address(sandbox);
--- a/cmd/crates/soroban-test/tests/it/integration/token/transfer.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/token/transfer.rs
@@ -130,3 +130,65 @@ async fn transfer_fails_when_recipient_trustline_missing() {
         .failure()
         .stderr(predicates::str::contains("trustline entry is missing"));
 }
+
+#[tokio::test]
+async fn transfer_rejects_negative_amount_before_any_rpc() {
+    let sandbox = &TestEnv::new();
+
+    // A negative amount is nonsensical and is rejected at the clap layer before
+    // any account resolution or RPC work, so this needs no funded accounts and
+    // no network — a literal destination address is enough to parse the args.
+    let recipient = "GAF4UUODFGAAMYRTF5QKUZCCZPXF3S4PRU5NS2BBRVJGX4WLRVI4ZI4Z";
+    sandbox
+        .new_assert_cmd("token")
+        .args([
+            "transfer",
+            "--id",
+            "native",
+            "--to",
+            recipient,
+            "--amount=-1",
+            "--from",
+            "test",
+        ])
+        .assert()
+        .failure()
+        .stderr(predicates::str::contains("amount must not be negative"));
+}
+
+#[tokio::test]
+async fn transfer_json_failure_returns_error_envelope_on_stdout() {
+    let sandbox = &TestEnv::new();
+    let test = test_address(sandbox);
+    let issuer = new_account(sandbox, "issuer");
+    let recipient = new_account(sandbox, "recipient");
+    let asset = format!("USDC:{issuer}");
+
+    // Same missing-recipient-trustline failure as above, but in JSON mode: the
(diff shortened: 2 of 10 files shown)


cmd/soroban-cli/src/commands/token/transfer.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub from: UnresolvedMuxedAccount,
+
+    /// Account to transfer the tokens to.
+    #[arg(long)]
+    pub to: UnresolvedMuxedAccount,
+
+    /// Amount to transfer, in the token's smallest unit (stroops for a Stellar
+    /// Asset Contract).
+    #[arg(long)]
+    pub amount: i128,
```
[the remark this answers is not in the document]
fnando: Good catch — fixed. `--amount` now uses a custom `value_parser` (`parse_nonneg_i128`) that rejects negatives at the clap layer, so `--amount=-1` fails immediately with `amount must not be negative` before any RPC work. (An upper bound isn't needed — the field is `i128`, so anything that parses is already in range, and clap's `.range()` helper doesn't support `i128`.) Added `transfer_rejects_negative_amount_before_any_rpc` to cover it.

cmd/soroban-cli/src/commands/token/transfer.rs
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+
+    pub async fn run(&self, global_args: &global::Args) -> Result<(), Error> {
+        let output = Output::new(self.output.into(), global_args.quiet);
+        // In JSON mode the underlying invoke pipeline's human-readable status
+        // logging (which writes to stderr) would still fire; run it quietly so
+        // machine consumers get clean output without needing `--quiet`.
+        let quiet = global_args.quiet || output.is_json();
+        let config = self.config();
```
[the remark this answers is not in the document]
fnando: Thanks — I dug into this, and the diagnostic isn't actually suppressed. `quiet` only silences the pipeline's human-readable *status* lines (e.g. "Simulating transaction…"); it doesn't touch the error. On a failed simulation the RPC error string already carries the `HostError` and its full diagnostic event log (including `"trustline entry is missing for account …"`), and in JSON mode that's rendered into `{ "error": { "message": … } }` via `error_json`, so the reason is machine-readable today.

A *structured/named* trustline error is intentionally deferred (see Known limitations): `Error(Contract, #13)` is the SAC's internal error enum, not stable API, so matching on it would break silently if it changed. I added `transfer_json_failure_returns_error_envelope_on_stdout` to lock in that the JSON failure envelope is parseable on stdout and contains the trustline diagnostic.

cmd/crates/soroban-test/tests/it/integration/token/transfer.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    sandbox
+        .new_assert_cmd("token")
+        .args([
+            "transfer", "--id", &asset, "--to", &recipient, "--amount", "1", "--from", "test",
+        ])
+        .assert()
+        .failure()
+        .stderr(predicates::str::contains("trustline entry is missing"));
+}
```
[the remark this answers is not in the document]
fnando: Added — `transfer_json_failure_returns_error_envelope_on_stdout` runs a failing transfer with `--output json`, asserts the command fails, that stdout is valid JSON, and that it's an `{ "error": { "message": … } }` envelope carrying the trustline diagnostic.

cmd/soroban-cli/src/commands/token/args.rs
```diff
@@ -0,0 +1,126 @@
...
+    Asset(#[from] builder::asset::Error),
+}
+
+impl FromStr for TokenTarget {
+    type Err = builder::asset::Error;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        // `native` and `CODE:ISSUER` are the two shapes an asset can take; both
+        // route to a SAC. Everything else is a contract id or alias.
+        if value == "native" || value.contains(':') {
```
aristidesstaffieri: I guess this doesn't consider aliases right? Would it be better resolve to the first if there is one?

cmd/soroban-cli/src/commands/token/transfer.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub id: TokenTarget,
+
+    /// Account to transfer tokens from. Signs and authorizes the transfer, so it
+    /// must be an identity or secret key you control.
+    #[arg(long)]
+    pub from: UnresolvedMuxedAccount,
+
+    /// Account to transfer the tokens to.
+    #[arg(long)]
+    pub to: UnresolvedMuxedAccount,
```
aristidesstaffieri: `--to` is typed as `UnresolvedMuxedAccount`, so a contract address destination (`C...`) — valid for SEP-41 `transfer`, e.g. depositing into a vault or protocol contract — fails `xdr::MuxedAccount::from_str`, falls into the `AliasOrSecret` branch, and then dies in `resolve_muxed_account` with a misleading "identity not found" locator error. `contract invoke -- transfer --to C...` handles this today, so the new command is strictly less capable and the error misdirects users toward identity configuration.

Suggestion: give `--to` a type that admits G/M/C destinations (an `UnresolvedScAddress`-style resolver that tries contract strkey → muxed account → alias) and pass the resolved address through to the `transfer` arg.

cmd/soroban-cli/src/commands/token/transfer.rs
```diff
@@ -0,0 +1,212 @@
...
+pub struct Cmd {
+    /// The token to transfer from: a contract id or alias, `native`, or a
+    /// classic asset as `CODE:ISSUER`.
+    #[arg(long = "id")]
+    pub id: TokenTarget,
+
+    /// Account to transfer tokens from. Signs and authorizes the transfer, so it
+    /// must be an identity or secret key you control.
+    #[arg(long)]
+    pub from: UnresolvedMuxedAccount,
```
aristidesstaffieri, line 29: SEP-41 supports muxed addresses (CAP-67 `ScAddress::MuxedAccount`), so `--from M...` should work — and it almost does: `UnresolvedMuxedAccount` resolves it, arg parsing and simulation accept it. It then fails at `invoke.rs:349`, where `client.get_account(&config.source_account()?.to_string())` hands the `M...` strkey to a G-only ed25519 parser and aborts with a raw strkey `DecodeError`.

Direction for supporting it: `stellar-xdr` already provides `xdr::MuxedAccount::account_id()` to extract the underlying `AccountId`. The invoke pipeline should use that (e.g. `config.source_account()?.account_id().to_string()`) for the sequence-number lookup and the classic transaction source in `build_invoke_contract_tx`, while the muxed form keeps flowing into the SEP-41 `transfer(from, ...)` argument so the mux id is preserved on-chain/in events. Worth adding an integration test with an `M...` `--from` (signing via an identity/`--sign-with-*`, since an M-literal carries no secret).

cmd/soroban-cli/src/commands/token/transfer.rs
```diff
@@ -0,0 +1,212 @@
...
+            "transfer", "--from", &from, "--to", &to, "--amount", &amount,
+        ]
+        .into_iter()
+        .map(OsString::from)
+        .collect();
+
+        let invoke_cmd = invoke::Cmd {
+            contract_id: UnresolvedContract::Resolved(contract_id),
+            slop,
+            config: config.clone(),
```
aristidesstaffieri, line 170: If simulation classifies the invocation as read-only (`ShouldSend::DefaultNo` — a token whose `transfer` records no auth, writes, or events), this command exits 0 without submitting anything. Since `invoke::Cmd` is built with `..Default::default()`, `send` stays `Send::Default`, and the pipeline takes the simulation-only early return with `tx_hash: None`.

Reproduced against testnet with a no-op SEP-41 `transfer` (contract `CAUCUECQHXCKUIK5FCKCMVJ5ZFMGEWC56JYYSFI3PFQOMSCYGMKW3OSL`):

```console
$ stellar token transfer --id CAUC... --from probe --to G... --amount 100 --network testnet
ℹ️  Simulation identified as read-only. Send by rerunning with `--send=yes`.
$ echo $?
0
```

Two problems compound here:

1. The stderr hint suggests `--send=yes`, but `token transfer` does not expose `--send` — following the hint yields `unexpected argument '--send'`, so there is no way to force submission.
2. In `--output json` the read-only info line is suppressed (JSON mode runs the pipeline quiet), so a script gets `{"tx_hash":null,"result":null}` with exit 0 — indistinguishable from an intentional success, when no transaction was submitted at all (confirmed via Horizon: no new txs on the account).

Suggestion: a transfer command should never silently resolve as read-only. Either force `Send::Yes` in the constructed `invoke::Cmd`, or make the `DefaultNo` path a hard error for this command (with the option of exposing `--send` if simulate-only is a use case worth keeping).

cmd/crates/soroban-test/tests/it/integration/token/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        .new_assert_cmd("contract")
+        .args([
+            "asset",
+            "deploy",
+            "--asset",
+            asset,
+            "--source-account",
+            source,
+        ])
+        .output()
```
aristidesstaffieri, line 47: `deploy_sac`'s only check cannot see a failed deploy: `Command::output()` returns `Err` only when the child fails to *spawn* — a non-zero exit comes back inside `Ok`, and the `Output` (status + stderr) is discarded. So while the doc comment says this tolerates "the case where a prior run already deployed it", it actually tolerates every failure. The other helpers in this file (`add_trustline`, `sac_id`, `sac_balance`) all use `.assert().success()`.

Demonstrated by pointing the same invocation at an unreachable RPC:

```console
$ stellar contract asset deploy --asset native --source-account test --rpc-url http://localhost:1 ...
❌ error: client error (Connect)
$ echo $?
1
```

wrapped in this exact `.output().expect(...)` pattern:

```
deploy_sac pattern PASSED SILENTLY — the expect() did not trip
child exit status it ignored: ExitStatus(unix_wait_status(256))   ← exit code 1
child stderr it threw away:   "❌ error: client error (Connect)"
```

The failure then surfaces two steps downstream, pointing at the wrong code: `sac_id` still succeeds (`contract id asset` computes the id locally, no network check), so `transfer_native_returns_receipt_and_moves_funds` first dies at `sac_balance` ("Contract not found") or at the transfer itself (`SacNotDeployed`) — sending an investigator into the new `token transfer` code when the actual regression is in `contract asset deploy`.

Suggestion: make the tolerance specific — check `output.status.success()`, and on failure allow only stderr matching the already-deployed error (the simulation's `ExistingValue` / "already exists" message), panicking with the captured stderr otherwise.

cmd/soroban-cli/src/commands/token/args.rs
fnando: Aliases are handled, just resolved later rather than in `from_str` (which has no `locator`, so it can't consult saved aliases). A bare name takes the `else` branch as `UnresolvedContract::Alias` and is resolved in `resolve_contract_id`; for `CODE:ISSUER`, the issuer resolves through `UnresolvedMuxedAccount`, which also accepts aliases. The `native`/`:` check here only picks asset-vs-contract shape, so there's no missed alias case — happy to expand the doc comment to make that explicit if it'd help.

cmd/soroban-cli/src/commands/token/transfer.rs
fnando: Fixed in d5e94709. `--to` is now typed as `UnresolvedScAddress`, so `G…`/`M…`/`C…` and aliases all resolve, and the strkey is passed through to the `transfer` arg. Added `transfer_to_contract_destination_succeeds` covering a `C…` recipient.

cmd/soroban-cli/src/commands/token/transfer.rs
fnando: Agreed this is a real gap. The failure is in the shared `contract invoke` pipeline (`invoke.rs:349` hands the `M…` strkey to a G-only parser), so `contract invoke --source-account M…` hits it too — pre-existing rather than introduced here. The proper fix (thread `account_id()` through the sequence lookup and classic tx source while preserving the muxed form in the `transfer` arg) touches every invoke path, so I filed it as a focused follow-up with its own `M…` coverage: #2645. Happy to add a clear "muxed source not yet supported" guard error in the meantime if you'd prefer that over the raw `DecodeError` while the follow-up lands.

cmd/soroban-cli/src/commands/token/transfer.rs
fnando: Fixed in d60ad5f1. The constructed `invoke::Cmd` now sets `send: Send::Yes`, so a transfer always submits and can never resolve read-only / exit 0 without moving funds (which also removes the misleading `--send=yes` hint path). Real SAC transfers always wrote anyway; this closes the degenerate-SEP-41 case you reproduced.

cmd/crates/soroban-test/tests/it/integration/token/mod.rs
fnando: Good catch — fixed in f975fa8c. `deploy_sac` now checks `output.status.success()` and only tolerates the already-deployed case (stderr containing `contract already exists`), panicking with the captured stderr on any other failure. Confirmed the marker against the sandbox (`Error(Storage, ExistingValue)` / "contract already exists").

cmd/soroban-cli/src/commands/token/transfer.rs
fnando: Landed an interim guard in f1a09c27: `token transfer` now rejects a muxed (`M…`) `--from` up front with a clear message (*"muxed (M…) source accounts are not yet supported for `token transfer`; use the underlying G… account as `--from` instead"*) instead of the opaque host error. Interestingly, writing the test showed the real current failure is `HostError: Error(Value, UnexpectedType)` at simulation — the muxed strkey is rejected in the `transfer` arg even before the `get_account` line. Full end-to-end muxed support stays tracked in #2645.

cmd/soroban-cli/src/commands/token/args.rs
aristidesstaffieri: gotcha yeah I think an expanded comment would be good but also do you think we should start rejecting the "native" string as an alias(or at least emit a warning when this happens)? It seems like a footgun to be able to make an alias that would also match this case.

Right now you get both of these cases:
 - contract invoke --id native → UnresolvedContract::from_str("native") → not a C-strkey → Alias("native") → resolves to the aliased contract.
 - token transfer --id native → args.rs:60 short-circuits before the else branch can ever produce Alias("native") → the XLM SAC

fnando: I'm adding `native` as a built-in alias in https://github.com/stellar/stellar-cli/pull/2646. Once this lands, we can remove the special case "native".

fnando: Done in `9ee7a6b1`. Now that the built-in reserved `native` alias landed (#2646), I removed the special case in `token/args.rs` — `native` now flows through as a contract alias and resolves to the native-asset SAC via `resolve_reserved`, consistent with `contract invoke`.

aristidesstaffieri (approved):

PR #2651, opened 2026-07-20
Add `stellar token balance` command

fnando:
### What

Add a `stellar token balance` command that reads the token balance of an account or contract. The token can be specified as a contract id or alias, `native`, or a classic asset as `CODE:ISSUER`. By default the balance is returned in the token's raw smallest unit (stroops for a Stellar Asset Contract); `--decimal` formats it using the token's `decimals`. Output is available as text (default) or JSON via `--output`.

```console
$ $stellar token balance --id native --account fnando --output=json
{"balance":"100000000000"}

$ $stellar token balance --id native --account fnando
100000000000

$ $stellar token balance --id usdc --account fnando --network testnet
2000000

$ $stellar token balance --id USDC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5 --account fnando --network testnet
2000000

$ $stellar token balance --id USDC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5 --account fnando --network testnet --decimal
0.2

$ http https://horizon-testnet.stellar.org/accounts/$($stellar keys address fnando) | jq '.balances | first | {asset_code, asset_issuer, balance}'
{
  "asset_code": "USDC",
  "asset_issuer": "GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5",
  "balance": "0.2000000"
}
```

### Why

Reading a token balance previously required a manual `stellar contract invoke ... -- balance` call and hand-decoding the result. This adds a first-class, read-only command that resolves the token (including classic assets and `native`) and returns a clean value, with token-aware errors when a Stellar Asset Contract isn't deployed yet.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/token/args.rs
+++ b/cmd/soroban-cli/src/commands/token/args.rs
@@ -1,8 +1,11 @@
 use std::str::FromStr;
 
 use crate::{
+    commands::contract::invoke,
     config::{alias::UnresolvedContract, locator},
+    get_spec,
     output::Format,
+    rpc,
     tx::builder,
     utils::contract_id_hash_from_asset,
 };
@@ -50,6 +53,28 @@ pub enum Error {
     Locator(#[from] locator::Error),
     #[error(transparent)]
     Asset(#[from] builder::asset::Error),
+
+    #[error(
+        "the Stellar Asset Contract {0} is not deployed on this network.\n\
+         Deploy it first with `stellar contract asset deploy --asset <ASSET>`, then retry."
+    )]
+    SacNotDeployed(String),
+
+    #[error("contract {0} was not found on this network")]
+    ContractNotFound(String),
+}
+
+impl Error {
+    /// Machine-readable discriminator for the JSON error envelope's `type` field.
+    #[must_use]
+    pub fn error_type(&self) -> &'static str {
+        match self {
+            Error::Locator(_) => "config",
+            Error::Asset(_) => "invalid_asset",
+            Error::SacNotDeployed(_) => "sac_not_deployed",
+            Error::ContractNotFound(_) => "contract_not_found",
+        }
+    }
 }
 
 impl FromStr for TokenTarget {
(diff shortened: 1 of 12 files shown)


cmd/soroban-cli/src/commands/token/balance.rs
```diff
@@ -0,0 +1,224 @@
...
+            let decimals = self
+                .read_u32(
+                    &config,
+                    quiet,
+                    global_args.no_cache,
+                    contract_id,
+                    vec![OsString::from("decimals")],
+                )
+                .await?;
+            (print::format_number(raw, decimals), Some(decimals))
```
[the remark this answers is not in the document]
fnando: fixed

cmd/soroban-cli/src/commands/token/args.rs
```diff
@@ -85,6 +110,31 @@ impl TokenTarget {
...
+        contract_id: &stellar_strkey::Contract,
+    ) -> Option<Error> {
+        let invoke::Error::GetSpecError(get_spec::Error::Rpc(rpc::Error::NotFound(kind, _))) = err
+        else {
+            return None;
+        };
+        if kind != "Contract" {
+            return None;
+        }
+        Some(if matches!(self, TokenTarget::Asset(_)) {
```
leighmcculloch, line 132: Does "native" get parsed as `TokenTarget::Contract`, meaning it doesn't show the right error here?

cmd/soroban-cli/src/commands/token/balance.rs
```diff
@@ -0,0 +1,223 @@
...
+                    contract_id,
+                    vec![OsString::from("decimals")],
+                    "decimals",
+                )
+                .await?;
+            // `format_number` computes `10.pow(decimals)`, which overflows i128
+            // once decimals exceeds 38. The value comes from the contract, so
+            // reject an implausible scale rather than panic (debug) or wrap
+            // silently (release). Real SACs use 7.
+            if decimals > 38 {
```
leighmcculloch: Maybe we could use an arbitrary decimal type to render the string instead so as not to have to handle an edge case like this explicitly? Is fine though.

leighmcculloch (approved):
Suggestions, but defer to you and looks good otherwise. The UX looks good with defaulting to raw values and offering decimal encoding if specified. I like how the flag is named `--decimal` which aligns with the `decimals` function name.

cmd/soroban-cli/src/commands/token/args.rs
fnando: Good catch — yes, `native` was parsing as `TokenTarget::Contract`, so a missing native SAC reported `contract_not_found` without the deploy guidance. Fixed by routing `native` (like `CODE:ISSUER`) to `TokenTarget::Asset(Asset::Native)` in `from_str`, so it's now classified as a SAC and reports `sac_not_deployed`. Same resolved contract id as before.

cmd/soroban-cli/src/commands/token/balance.rs
fnando: Agreed the explicit guard didn't scale. Introduced a reusable `FixedPoint { value, decimals }` type (new `fixed_point` module) that formats by placing the decimal point in the digit string — no `10.pow`, so it can't overflow for any `decimals`, no dependency. `format_number` now delegates to it and the `> 38` guard is gone. Also opened #2652 to centralize asset/token resolution as a follow-up.

PR #2654, opened 2026-07-22
Centralize asset/token resolution behind a single resolver

### LINKED ISSUE #2652: Centralize asset/token resolution behind a single resolver
### Problem

Resolving a user-supplied token/asset reference to a contract id — and knowing whether it's a Stellar Asset Contract (and for which asset) — is spread across several overlapping places:

- `utils::contract_id_hash_from_asset` — the primitive (`xdr::Asset` → id)
- `config::alias::resolve_reserved` — the reserved `native` alias → native SAC id
- `commands::token::args::TokenTarget` — `--id` string → id, with an asset-vs-contract split
- `commands::contract::deploy::asset` / `commands::contract::id::asset` — deploy/id from `builder::Asset`
- `config::locator` and `commands::snapshot::create` — more ad-hoc `builder::Asset` → id

No single type answers both "what contract id?" and "is this a SAC, and which asset?" — the exact knowledge needed to produce a good `sac_not_deployed` vs `contract_not_found` error. This surfaced in #2651, where `native` had to be special-cased to be classified as a SAC.

### Proposal

Introduce a centralized resolver producing something like `ResolvedToken { contract_id, kind: Sac(Asset) | Contract }`, then migrate the call sites above onto it. This removes the duplicated asset→id derivations and makes SAC-awareness consistent everywhere (errors, deploy, id, snapshot).

### Context

Follow-up from PR #2651 review discussion.

fnando:
### What

Introduces a single token resolver in a new `config::token` module and migrates the scattered asset/token → contract-id derivations onto it, resolving #2652.

- `UnresolvedToken` — a token reference parsed from a user string (`native`/`CODE:ISSUER` → asset; `C…` strkey or alias → contract).
- `ResolvedToken { contract_id, kind }` where `TokenKind` is `Sac(xdr::Asset)` or `Contract` — the single answer to both "what contract id?" and "is this a SAC, and for which asset?". Helpers: `is_sac()` / `asset()`.

Migrated call sites:
- `commands::token::args` — dropped the old `TokenTarget`; `not_deployed_error` is now keyed off `ResolvedToken::kind`, keeping the `sac_not_deployed` vs `contract_not_found` split. `args::Error` now carries only the two token-aware variants it actually produces.
- `commands::token::{balance,transfer}` — `--id` is now `UnresolvedToken`; both resolve once and thread the `ResolvedToken` through.
- `commands::contract::{id,deploy}::asset` — derive the id (and, for deploy, the wrap-tx asset) via the resolver. Deploy resolves up front, before any RPC, so an invalid asset fails fast.
- `config::alias::resolve_reserved` — the reserved `native` alias now resolves through the shared resolver.

`utils::contract_id_hash_from_asset` remains the low-level primitive the resolver builds on.

### Why

Resolving a token/asset reference to a contract id — and knowing whether it's a Stellar Asset Contract and for which asset — was spread across several overlapping places, with no single type answering both questions. This surfaced in #2651, where `native` had to be special-cased as a SAC. The resolver makes SAC-awareness consistent everywhere (errors, deploy, id).

### Known limitations

`snapshot::create` is intentionally left unmigrated: that site parses the asset to extract the **issuer** for account search, not to derive a contract id, so `ResolvedToken`'s shape doesn't fit.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
@@ -53,9 +53,6 @@ pub enum Error {
     Builder(#[from] builder::Error),
 
     #[error(transparent)]
-    Asset(#[from] builder::asset::Error),
-
-    #[error(transparent)]
     Locator(#[from] locator::Error),
 
     #[error(transparent)]
--- a/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
+++ b/cmd/soroban-cli/src/commands/contract/deploy/asset.rs
@@ -19,10 +19,9 @@ use crate::{
         txn_result::{TxnEnvelopeResult, TxnResult},
         HEADING_TRANSACTION,
     },
-    config::{self, data, network},
+    config::{self, data, network, token::UnresolvedToken},
     rpc::Error as SorobanRpcError,
     tx::builder,
-    utils::contract_id_hash_from_asset,
 };
 
 use crate::config::address::AliasName;
@@ -58,6 +57,9 @@ pub enum Error {
 
     #[error(transparent)]
     Locator(#[from] locator::Error),
+
+    #[error(transparent)]
+    Token(#[from] config::token::Error),
 
     #[error(transparent)]
     Fee(#[from] fetch::fee::Error),
@@ -144,11 +146,19 @@ impl Cmd {
         quiet: bool,
         no_cache: bool,
     ) -> Result<TxnResult<stellar_strkey::Contract>, Error> {
-        // Parse asset
         let print = Print::new(quiet);
-        let asset = self.asset.resolve(&config.locator)?;
 
         let network = config.get_network()?;
+        // Resolve the asset up front, before any RPC, so an invalid asset or
+        // unresolvable issuer alias fails fast instead of after network calls.
+        let token = UnresolvedToken::Asset(self.asset.clone())
+            .resolve(&config.locator, &network.network_passphrase)?;
+        let contract_id = token.contract_id;
+        let asset = token
+            .asset()
+            .expect("an asset reference resolves to a SAC")
+            .clone();
+
         let client = network.rpc_client()?;
         client
             .verify_network_passphrase(Some(&network.network_passphrase))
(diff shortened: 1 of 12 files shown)


cmd/soroban-cli/src/commands/contract/id/asset.rs
```diff
@@ -23,6 +23,8 @@ pub enum Error {
     Xdr(#[from] crate::xdr::Error),
     #[error(transparent)]
     Asset(#[from] builder::asset::Error),
```
leighmcculloch: Is `Error::Asset` used anymore?

cmd/soroban-cli/src/commands/contract/deploy/asset.rs
leighmcculloch: Is `Error::Asset` used anymore?

leighmcculloch (approved):

cmd/soroban-cli/src/commands/token/args.rs
```diff
@@ -68,111 +41,30 @@ impl Error {
...
+        return None;
+    };
+    if kind != "Contract" {
+        return None;
     }
+    let contract_id = token.contract_id;
+    Some(if token.is_sac() {
+        Error::SacNotDeployed(format!("{contract_id}"))
+    } else {
+        Error::ContractNotFound(format!("{contract_id}"))
```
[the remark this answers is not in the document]
fnando: Added integration coverage for this branch in `bd1d8fdc`: `balance_fails_when_contract_not_found` and `transfer_fails_when_contract_not_found` invoke a well-formed but undeployed `C…` id and assert the `contract_not_found` error type. Both pass against the sandbox.

cmd/soroban-cli/src/commands/contract/id/asset.rs
fnando: Good catch — it isn't. Asset parsing now happens at arg-parse time and resolution errors surface as `Token(config::token::Error)`, so this variant was dead. Removed in `e05f95c7`.

cmd/soroban-cli/src/commands/contract/deploy/asset.rs
fnando: Same here — dead since resolution moved to the shared resolver. Removed in `e05f95c7`.

PR #2657, opened 2026-07-27
Fix `contract extend` panic when extending a missing entry

### LINKED ISSUE #2599: `stellar contract extend` panics with "index out of bounds" when the extension is a no-op

fnando:
### What

Fixes a panic in `stellar contract extend`. When run against a well-formed but non-existent contract entry, the transaction goes through as a no-op and the CLI then crashes with `index out of bounds: the len is 0 but the index is 0`. It now exits cleanly with a "Ledger entry not found" error instead.

### Why

On a no-op extend the code fetched the ledger entries and unconditionally indexed `entry.entries[0]`. For a non-existent entry that vec is empty, so the index panicked. This replaces the raw index with `.first().ok_or(Error::LedgerEntryNotFound)?`, reusing the existing error variant already returned for the analogous no-op cases in the same function. Fixes #2599.

### Known limitations

N/A

### DIFF
--- a/cmd/crates/soroban-test/tests/it/integration/hello_world.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/hello_world.rs
@@ -1,3 +1,4 @@
+use predicates::prelude::PredicateBooleanExt;
 use soroban_cli::{
     commands::{
         contract::{self, fetch},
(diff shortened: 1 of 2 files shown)


mootz12 (approved):
Good catch

PR #2675, opened 2026-08-07
Add depth limit to xdr encoding and decoding

mootz12:
### What

Adds a depth limit to all XDR encoding and decoding sites.

### Why

This also prevents stack overflow crashes if intentionally deep XDR structs are encountered.

### Known limitations

None

### DIFF
--- a/cmd/crates/soroban-spec-typescript/src/lib.rs
+++ b/cmd/crates/soroban-spec-typescript/src/lib.rs
@@ -15,6 +15,11 @@ use types::{Entry, ErrorEnumCase};
 
 use soroban_spec::read::{from_wasm, FromWasmError};
 
+/// Depth limit when encoding and decoding XDR.
+///
+/// 500 matches `soroban-env-host`'s `DEFAULT_XDR_RW_LIMITS`.
+const XDR_DEPTH_LIMIT: u32 = 500;
+
 pub mod boilerplate;
 mod types;
 
@@ -86,7 +91,12 @@ fn generate_class(
         .join(",\n        ");
     let spec = spec
         .iter()
-        .map(|s| format!("\"{}\"", s.to_xdr_base64(Limits::none()).unwrap()))
+        .map(|s| {
+            format!(
+                "\"{}\"",
+                s.to_xdr_base64(Limits::depth(XDR_DEPTH_LIMIT)).unwrap()
+            )
+        })
         .join(",\n        ");
     format!(
         r#"export interface Client {{{method_types}
--- a/cmd/soroban-cli/src/commands/tx/fetch/envelope.rs
+++ b/cmd/soroban-cli/src/commands/tx/fetch/envelope.rs
@@ -5,6 +5,7 @@ use crate::{
 use clap::Parser;
 
 use super::args;
+use crate::utils::XDR_DEPTH_LIMIT;
 
 #[derive(Parser, Debug, Clone)]
 #[group(skip)]
@@ -36,7 +37,9 @@ impl Cmd {
                     println!("{}", serde_json::to_string(&envelope)?);
                 }
                 args::OutputFormat::Xdr => {
-                    let envelope_xdr = envelope.to_xdr_base64(Limits::none()).unwrap();
+                    let envelope_xdr = envelope
+                        .to_xdr_base64(Limits::depth(XDR_DEPTH_LIMIT))
+                        .unwrap();
                     println!("{envelope_xdr}");
                 }
                 args::OutputFormat::JsonFormatted => {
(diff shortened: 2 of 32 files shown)


cmd/soroban-cli/src/log/event.rs
```diff
@@ -39,7 +40,9 @@ fn format_decoded_event(decoded: &DecodedEvent, status: &str) -> String {
 pub fn all(events: &[DiagnosticEvent]) {
     for (index, event) in events.iter().enumerate() {
         let json = serde_json::to_string(event).unwrap();
-        let xdr = event.to_xdr_base64(xdr::Limits::none()).unwrap();
+        let xdr = event
+            .to_xdr_base64(xdr::Limits::depth(XDR_DEPTH_LIMIT))
```
leighmcculloch, line 44: I think the rpc client decodes the xdr from the response without a depth limit, so introducing a depth limit here without first introducing it there is probably not going to have the affect intended.

cmd/soroban-cli/src/commands/tx/fetch/meta.rs
```diff
@@ -36,7 +37,7 @@ impl Cmd {
                     println!("{}", serde_json::to_string(&meta)?);
                 }
                 args::OutputFormat::Xdr => {
-                    let meta_xdr = meta.to_xdr_base64(Limits::none()).unwrap();
+                    let meta_xdr = meta.to_xdr_base64(Limits::depth(XDR_DEPTH_LIMIT)).unwrap();
```
leighmcculloch, line 40: Same here.

cmd/soroban-cli/src/commands/tx/fetch/envelope.rs
```diff
@@ -36,7 +37,9 @@ impl Cmd {
                     println!("{}", serde_json::to_string(&envelope)?);
                 }
                 args::OutputFormat::Xdr => {
-                    let envelope_xdr = envelope.to_xdr_base64(Limits::none()).unwrap();
+                    let envelope_xdr = envelope
+                        .to_xdr_base64(Limits::depth(XDR_DEPTH_LIMIT))
```
leighmcculloch, line 41: Same here.

cmd/soroban-cli/src/commands/tx/fetch/result.rs
```diff
@@ -35,7 +36,9 @@ impl Cmd {
                     println!("{}", serde_json::to_string(&result)?);
                 }
                 args::OutputFormat::Xdr => {
-                    let result_xdr = result.to_xdr_base64(Limits::none()).unwrap();
+                    let result_xdr = result
```
leighmcculloch, line 39: Same here.

leighmcculloch (approved):
This seems fine, and also relatively low risk making this change in the cli since if a problem does arise where we have set it too low then we can quickly ship a new release and the release channels are all very accessible.

See inline comments as there are some places that I think we won't see the intended affect because the rpc client crate already decodes them without a depth limit (iirc we have not added depth limits there).

Something else I wonder about is will the depth limit actually prevent stack overflow in some of the cases like claimable balances. But in any case this is an improvement.

cmd/soroban-cli/src/log/event.rs
leighmcculloch: Ah I see https://github.com/stellar/rs-stellar-rpc-client/pull/109. 👏🏻

cmd/crates/soroban-spec-typescript/src/lib.rs
```diff
@@ -86,7 +91,12 @@ fn generate_class(
         .join(",\n        ");
     let spec = spec
         .iter()
-        .map(|s| format!("\"{}\"", s.to_xdr_base64(Limits::none()).unwrap()))
+        .map(|s| {
+            format!(
+                "\"{}\"",
+                s.to_xdr_base64(Limits::depth(XDR_DEPTH_LIMIT)).unwrap()
+            )
```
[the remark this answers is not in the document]
mootz12: This is an API change. We can opt to fix this in a followup of we want to update this. For now, the error at least is clear to the user rather than a stack overflow.

cmd/soroban-cli/src/commands/tx/fetch/result.rs
```diff
@@ -35,7 +36,9 @@ impl Cmd {
...
+                        .to_xdr_base64(Limits::depth(XDR_DEPTH_LIMIT))
+                        .unwrap();
```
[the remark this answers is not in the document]
mootz12: The effort to prevent CLI panics in general is separate from this one, which is preventing weird panics like stack overflow.

PR #2678, opened 2026-08-12
Add container support to `stellar contract build`

fnando:
### What

Adds `stellar contract build --image <ref>`, which builds a contract inside a container image instead of compiling locally. The working tree is bind-mounted at `/source`, the image's own `stellar contract build` runs there, and the resulting wasm lands on the host — including copies to `--out-dir` when set. Any tag or digest ref is accepted.

It reuses the existing container-engine abstraction, so `--engine`, `--docker-host`, and the default engine set by `stellar container use` all apply, along with resource limits (`--cpus`, `--memory`). Workspaces with several cdylibs build each package, sharing one container.

Before forwarding flags, the image is probed once (a single throwaway container) for the CLI binary name (`soroban` vs `stellar`), its version, and the default rustup toolchain. Forwarded flags (`--locked`, `--optimize`, `--optimize=false`) are gated by the image's version so an older image doesn't fail on an unknown flag, and `RUSTUP_TOOLCHAIN` is pinned to the image's own toolchain so a `rust-toolchain.toml` in the mounted source can't redirect the build. `--no-image-pull` builds against an image already present locally (offline/air-gapped, digest-pinned, or never pushed), and `--print-commands-only` emits a copy-pasteable reproduce line.

On Linux the build container is deliberately run as the invoking user's uid:gid (`--user`), so wasm written into the bind-mounted `target/` is owned by that user rather than root. Docker Desktop (macOS) and Apple's `container` already map ownership to the host user, so this is Linux-only. It assumes the image keeps `CARGO_HOME`/`RUSTUP_HOME` writable by non-root users, as the official image does.

### Why

Lets you build a contract inside a container image with a fixed CLI and Rust toolchain, without needing a matching Rust build toolchain locally — no local wasm target and no pinned `rustc`. A local `cargo` is still used for workspace discovery (package selection and locating the workspace root); fully removing that dependency by running discovery inside the image is left as a follow-up.

### Known limitations

The combined image probe requires `/bin/sh` and `rustup` in the image (as the official rust-based image provides). With `--print-commands-only` nothing is probed, so the reproduce line assumes a current `stellar` image and omits the toolchain pin. Running an arbitrary image with root-owned toolchain dirs may fail the Linux uid:gid build.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -113,13 +113,13 @@ pub struct Cmd {
     #[arg(long, help_heading = HEADING_CONTAINER)]
     pub image: Option<String>,
 
-    /// Don't pull `--image` before building; use the copy already present locally.
+    /// Pull `--image` before building to refresh a moving tag.
     ///
-    /// Lets you build against a locally-built (never pushed) image or a
-    /// digest-pinned image already on disk, and to work offline — e.g. air-gapped
-    /// verification against a pinned digest. Fails if the image isn't present.
+    /// By default the build uses the image already present locally and doesn't
+    /// pull (matching `docker run`), so a locally-built or digest-pinned image is
+    /// used as-is. Pass `--pull` to fetch the newest image for the tag first.
     #[arg(long, requires = "image", help_heading = HEADING_CONTAINER)]
-    pub no_image_pull: bool,
+    pub pull: bool,
 
     #[command(flatten)]
     pub build_args: BuildArgs,
@@ -263,7 +263,7 @@ impl Default for Cmd {
             locked: false,
             print_commands_only: false,
             image: None,
-            no_image_pull: false,
+            pull: false,
             build_args: BuildArgs::default(),
             container_args: ContainerArgs::default(),
             run_args: ContainerRunArgs::default(),
@@ -922,18 +922,18 @@ mod tests {
     }
 
     #[test]
-    fn no_image_pull_requires_image() {
+    fn pull_requires_image() {
         let cmd = Cmd::try_parse_from([
             "build",
             "--image",
             "docker.io/stellar/stellar-cli:latest",
-            "--no-image-pull",
+            "--pull",
         ])
-        .expect("--no-image-pull with --image must parse");
-        assert!(cmd.no_image_pull);
+        .expect("--pull with --image must parse");
+        assert!(cmd.pull);
 
         // Without --image the flag is rejected rather than silently ignored.
-        assert!(Cmd::try_parse_from(["build", "--no-image-pull"]).is_err());
+        assert!(Cmd::try_parse_from(["build", "--pull"]).is_err());
     }
 
     #[test]
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -21,10 +21,15 @@ use stellar_xdr::{Limited, Limits, ScMetaEntry, ScMetaV0, StringM, WriteXdr};
 use crate::commands::contract::optimize;
 use crate::utils::XDR_DEPTH_LIMIT;
 use crate::{
-    commands::{global, version},
+    commands::{
+        container::shared::{Args as ContainerArgs, RunArgs as ContainerRunArgs},
+        global, version, HEADING_CONTAINER,
+    },
     print::Print,
     wasm,
 };
+
+pub mod container;
 
 /// A built WASM artifact with its package name and file path.
 #[derive(Debug, Clone)]
@@ -97,8 +102,39 @@ pub struct Cmd {
     #[arg(long, conflicts_with = "out_dir", help_heading = "Other")]
     pub print_commands_only: bool,
 
+    /// Build inside this container image (e.g.
+    /// `docker.io/stellar/stellar-cli:latest`). When set, the build runs in the
+    /// container against the bind-mounted working tree instead of locally. Any
+    /// tag or digest ref is accepted.
+    ///
+    /// On Linux the container runs as your uid:gid so built wasm isn't
+    /// root-owned; this assumes the image keeps CARGO_HOME/RUSTUP_HOME writable
+    /// by non-root users, as the official image does.
+    #[arg(long, help_heading = HEADING_CONTAINER)]
+    pub image: Option<String>,
+
+    /// Don't pull `--image` before building; use the copy already present locally.
+    ///
+    /// Lets you build against a locally-built (never pushed) image or a
+    /// digest-pinned image already on disk, and to work offline — e.g. air-gapped
+    /// verification against a pinned digest. Fails if the image isn't present.
+    #[arg(long, requires = "image", help_heading = HEADING_CONTAINER)]
+    pub no_image_pull: bool,
+
     #[command(flatten)]
     pub build_args: BuildArgs,
+
+    // Declared after `build_args` so their `next_help_heading` groups them under
+    // the Container heading without leaking it onto the ungrouped flags above.
+    /// Container connection options (`--engine`, `--docker-host`) used when
+    /// `--image` is set. `--docker-host` is honored only by the docker engine.
+    #[command(flatten, next_help_heading = HEADING_CONTAINER)]
+    pub container_args: ContainerArgs,
+
+    /// Container resource limits (`--cpus`, `--memory`) applied to the
+    /// `--image` build container.
+    #[command(flatten, next_help_heading = HEADING_CONTAINER)]
+    pub run_args: ContainerRunArgs,
 }
 
 /// Shared build options for meta and optimization, reused by deploy and upload.
@@ -205,10 +241,13 @@ pub enum Error {
 
     #[error("wasm parsing error: {0}")]
     WasmParsing(String),
+
+    #[error(transparent)]
+    Container(#[from] container::Error),
 }
 
-const WASM_TARGET: &str = "wasm32v1-none";
-const WASM_TARGET_OLD: &str = "wasm32-unknown-unknown";
+pub(crate) const WASM_TARGET: &str = "wasm32v1-none";
+pub(crate) const WASM_TARGET_OLD: &str = "wasm32-unknown-unknown";
 const META_CUSTOM_SECTION_NAME: &str = "contractmetav0";
 
 impl Default for Cmd {
@@ -223,7 +262,11 @@ impl Default for Cmd {
             out_dir: None,
             locked: false,
             print_commands_only: false,
+            image: None,
+            no_image_pull: false,
             build_args: BuildArgs::default(),
+            container_args: ContainerArgs::default(),
+            run_args: ContainerRunArgs::default(),
         }
     }
 }
@@ -231,8 +274,14 @@ impl Default for Cmd {
 impl Cmd {
     /// Builds the project and returns the built WASM artifacts.
     #[allow(clippy::too_many_lines)]
-    pub fn run(&self, global_args: &global::Args) -> Result<Vec<BuiltContract>, Error> {
+    pub async fn run(&self, global_args: &global::Args) -> Result<Vec<BuiltContract>, Error> {
         let print = Print::new(global_args.quiet);
+
+        // When an image is given, build inside that container instead of locally.
+        if self.image.is_some() {
+            return container::run(self, global_args, &print).await;
+        }
+
         let working_dir = env::current_dir().map_err(Error::GettingCurrentDir)?;
         let metadata = self.metadata()?;
         let packages = self.packages(&metadata)?;
@@ -716,7 +765,7 @@ fn get_rustflags() -> Option<Vec<String>> {
     None
 }
 
-fn get_wasm_target() -> Result<String, Error> {
+pub(crate) fn get_wasm_target() -> Result<String, Error> {
     let Ok(current_version) = version() else {
         return Ok(WASM_TARGET.into());
     };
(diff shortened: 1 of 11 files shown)


cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+                continue;
+            }
+        } else if !md.workspace_default_members.contains(&p.id) {
+            continue;
+        }
+
+        let file = format!("{}.wasm", p.name.replace('-', "_"));
+        // The container may build for either wasm target depending on its rust
+        // version; fall back to the current host default for the reported path.
+        let src = [WASM_TARGET, WASM_TARGET_OLD]
```
leighmcculloch: `collect_built_contracts` probes `WASM_TARGET` before `WASM_TARGET_OLD` and takes the first path that exists, so a stale wasm from an earlier build with a different target triple can be reported/copied instead of the one the container just built.

**Failure scenario:** Build once locally/in-container with an older Rust toolchain producing `target/wasm32-unknown-unknown/release/foo.wasm`; later run `stellar contract build --image <old-image>` (also `wasm32-unknown-unknown`) so it overwrites that file, but a leftover `target/wasm32v1-none/release/foo.wasm` from a still-earlier newer build remains on disk. `find(|path| path.exists())` checks `WASM_TARGET` (`wasm32v1-none`) first, finds the stale file, and reports/copies it as the freshly built contract — the caller (including `contract deploy`) silently gets an old wasm instead of the just-built one.

---

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+        mc.manifest_path(p);
+    }
+    mc.exec()
+}
+
+/// Resolve the packages to build. An explicit `--package` wins; otherwise the
+/// default-member crates that build a cdylib, mirroring the local build's
+/// package selection. May be empty (no cdylib default members), in which case
+/// the caller falls back to a single no-`--package` build.
+fn resolve_packages(cmd: &Cmd) -> Result<Vec<String>, Error> {
```
leighmcculloch: `resolve_packages`/`collect_built_contracts` don't replicate the local build's manifest-path-based single-package selection (`build.rs`'s `packages()`), so `--image` builds a different package set than a local build with the same flags.

**Failure scenario:** In a workspace with several default-member cdylibs (e.g. `contracts/a`, `contracts/b`), run `stellar contract build --manifest-path=contracts/a/Cargo.toml --image ` with no `--package`. Locally this builds only package `a` (`build.rs`'s `packages()` matches by manifest path). With `--image`, `resolve_packages` ignores `manifest_path` entirely and returns every default-member cdylib (`a` and `b`), and `forwarded_build_args` then forwards the same `--manifest-path=contracts/a/Cargo.toml` alongside `--package=b` for b's build — an inconsistent flag pairing that doesn't match what was requested. The same divergence occurs when running from a workspace subdirectory instead of using `--manifest-path`, which `build.rs`'s doc comment says should build just that crate.

---

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+/// Collect the built wasm from the mounted `target/`. Because the working tree
+/// was bind-mounted, the container writes artifacts straight to the host under
+/// `<workspace>/target/<triple>/<profile>/`. The container's rust toolchain
+/// decides the target triple, so both known triples are probed. Copies to
+/// `--out-dir` when set.
+fn collect_built_contracts(
+    cmd: &Cmd,
+    workspace_root: &Path,
+) -> Result<Vec<BuiltContract>, super::Error> {
+    let md = metadata(cmd).map_err(Error::from)?;
```
leighmcculloch: `container::run` invokes `cargo metadata` (via `metadata(cmd)`) up to three separate times per build — in `resolve_workspace_root`, `resolve_packages`, and `collect_built_contracts` — instead of computing it once and reusing it.

On a large workspace or a slow/networked filesystem, each `cargo metadata` subprocess call can add noticeable latency.

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+    // With `--print-commands-only` nothing runs, so don't pull, probe, or build;
+    // just render the run command against a current image below.
+    let print_only = cmd.print_commands_only;
+
+    // Pull unless printing (nothing runs) or the user opted out with
+    // `--no-image-pull` to use a locally present image (offline / air-gapped
+    // verification, or a locally-built image that was never pushed).
+    if !print_only && !cmd.no_image_pull {
+        pull_image(&docker, image, print).await?;
```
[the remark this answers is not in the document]
fnando: Good catch. Rather than a per-engine "never pull" flag on `run` (there's no portable one — Apple's `container` has no `--pull`), `--no-image-pull` now enforces itself: it runs `image inspect` up front and fails fast when the image isn't present locally, so `run` can't silently pull a missing image. The default path still does an explicit `pull` to refresh a moving tag (`run`'s default only fetches a *missing* image, never re-pulling an existing one). Verified against Apple's `container`: present image proceeds, missing image errors before any build.

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+/// Resolve the packages to build. An explicit `--package` wins; otherwise the
+/// default-member crates that build a cdylib, mirroring the local build's
+/// package selection. May be empty (no cdylib default members), in which case
+/// the caller falls back to a single no-`--package` build.
+fn resolve_packages(cmd: &Cmd) -> Result<Vec<String>, Error> {
+    if let Some(pkg) = &cmd.package {
+        return Ok(vec![pkg.clone()]);
+    }
+    let md = metadata(cmd)?;
```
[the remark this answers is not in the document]
fnando: This already selects only the target package — `cargo metadata` narrows `workspace_default_members` to the package matching `--manifest-path` (or cwd), and `resolve_packages` filters by that set, mirroring the local build. Verified: `--image --manifest-path contracts/add/Cargo.toml` and running from a member dir both build only that package; the full-workspace chain only happens from the root. Added a regression test to lock this in.

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+            continue;
+        }
+
+        let file = format!("{}.wasm", p.name.replace('-', "_"));
+        // The container may build for either wasm target depending on its rust
+        // version; fall back to the current host default for the reported path.
+        let src = [WASM_TARGET, WASM_TARGET_OLD]
+            .iter()
+            .map(|triple| target_root.join(triple).join(&cmd.profile).join(&file))
+            .find(|path| path.exists())
```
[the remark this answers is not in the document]
fnando: Good catch. Now selecting the freshest of the candidate triple dirs by mtime instead of the first that exists, so a stale wasm left in the other triple's dir from an earlier build can't be picked up. Added a unit test proving the fresher file wins over list position.

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+        let reproduce = format!(
+            "{prefix} run --rm{run_flags} -v {bind} -w /source{env_flags} --entrypoint /bin/sh {image} -c {}",
+            shell_escape::escape(chain.clone().into())
+        );
+        (Some("/bin/sh"), vec!["-c".to_string(), chain], reproduce)
+    } else {
+        let cmd = container_cmds.first().cloned().unwrap_or_default();
+        let reproduce = format!(
+            "{prefix} run --rm{run_flags} -v {bind} -w /source{env_flags} {image} {}",
+            escape_args(&cmd)
```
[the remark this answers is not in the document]
fnando: The single-package path intentionally runs the image's own entrypoint, which for the official image is the CLI binary. Overriding it to the probed `bin` would help an image whose CLI is on `PATH` but not the entrypoint, while breaking the opposite case (CLI is the entrypoint but not on `PATH`). Since it's a deliberate design choice and the official image works either way, keeping it as-is.

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -0,0 +1,1083 @@
...
+    Ok(metadata(cmd)?.workspace_root.into_std_path_buf())
+}
+
+fn metadata(cmd: &Cmd) -> Result<cargo_metadata::Metadata, cargo_metadata::Error> {
+    let mut mc = MetadataCommand::new();
+    mc.no_deps();
+    if let Some(p) = &cmd.manifest_path {
+        mc.manifest_path(p);
+    }
+    mc.exec()
```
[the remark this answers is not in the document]
fnando: You're right — discovery still shells out to host `cargo metadata`, so a local `cargo` is required. Moving discovery into the image is a larger change (it needs the workspace root to set up the bind mount before it can run metadata there, plus host-cwd→container mapping for member selection), so I've scoped it out of this PR. Updated the description to drop the "no local Rust setup" wording: you don't need a matching Rust *build* toolchain locally (no wasm target, no pinned rustc), but `cargo` is still used for workspace discovery. Tracking the in-container discovery as a follow-up.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -97,8 +102,39 @@ pub struct Cmd {
...
+    #[arg(long, help_heading = HEADING_CONTAINER)]
+    pub image: Option<String>,
+
+    /// Don't pull `--image` before building; use the copy already present locally.
+    ///
+    /// Lets you build against a locally-built (never pushed) image or a
+    /// digest-pinned image already on disk, and to work offline — e.g. air-gapped
+    /// verification against a pinned digest. Fails if the image isn't present.
+    #[arg(long, requires = "image", help_heading = HEADING_CONTAINER)]
+    pub no_image_pull: bool,
```
leighmcculloch: It would be more consistent with docker's build command, which I think defaults to no pull if the image is already local.

Anotehr thing to consider is that an airgapped feature probably needs to be about more than just the docker image, but also dependencies for cargo, etc. So we could chase a feature like this, but doing it just for the docker image seems incomplete so I'd chase it separately to the initial add of support.

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -97,8 +102,39 @@ pub struct Cmd {
...
+    // the Container heading without leaking it onto the ungrouped flags above.
+    /// Container connection options (`--engine`, `--docker-host`) used when
+    /// `--image` is set. `--docker-host` is honored only by the docker engine.
+    #[command(flatten, next_help_heading = HEADING_CONTAINER)]
+    pub container_args: ContainerArgs,
+
+    /// Container resource limits (`--cpus`, `--memory`) applied to the
+    /// `--image` build container.
+    #[command(flatten, next_help_heading = HEADING_CONTAINER)]
+    pub run_args: ContainerRunArgs,
```
leighmcculloch, line 137: `--run-args` seems a bit too generic so ambiguous what part of the build process it applies to, `--container-run-args`?

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -97,8 +102,39 @@ pub struct Cmd {
...
+
     #[command(flatten)]
     pub build_args: BuildArgs,
+
+    // Declared after `build_args` so their `next_help_heading` groups them under
+    // the Container heading without leaking it onto the ungrouped flags above.
+    /// Container connection options (`--engine`, `--docker-host`) used when
+    /// `--image` is set. `--docker-host` is honored only by the docker engine.
+    #[command(flatten, next_help_heading = HEADING_CONTAINER)]
+    pub container_args: ContainerArgs,
```
leighmcculloch, line 132: Should this be cotainer build args or this applies for both?

cmd/soroban-cli/src/commands/contract/build.rs
```diff
@@ -97,8 +102,39 @@ pub struct Cmd {
...
+    /// Build inside this container image (e.g.
+    /// `docker.io/stellar/stellar-cli:latest`). When set, the build runs in the
+    /// container against the bind-mounted working tree instead of locally. Any
+    /// tag or digest ref is accepted.
+    ///
+    /// On Linux the container runs as your uid:gid so built wasm isn't
+    /// root-owned; this assumes the image keeps CARGO_HOME/RUSTUP_HOME writable
+    /// by non-root users, as the official image does.
+    #[arg(long, help_heading = HEADING_CONTAINER)]
+    pub image: Option<String>,
```
leighmcculloch, line 114: How can I build with container without choosing the container? Is there a way to do that?

I'm thinking about how we move to auto building with a container if docker is available.

leighmcculloch (commented):
Couple questions about scope and how this will evolve to be the default way to build to make sure what we build now will evolve well into that.

There are also a couple things claude noticed that I've let pass through verbatim because they appear meaningful problems to consider.

cmd/soroban-cli/src/commands/contract/build.rs
fnando: this doesn't create `--run-args`, it flats the container args, which brings things like `--cpus` and `--memory`.

cmd/soroban-cli/src/commands/contract/build.rs
fnando: same thing, it just flattens the args (i.e. `--docker_host` and `--engine`).

cmd/soroban-cli/src/commands/contract/build.rs
fnando: not at this moment. I made the conscious decision of postponing auto image selection.

cmd/soroban-cli/src/commands/contract/build.rs
fnando: The idea is that images must be 100% runnable by this feature. That means `stellar`/`soroban`/`sh` must be available, rust + wasm target must be installed, and the image is ready to be ran.

We can decide later on how to tackle images that don't have the proper dependencies, but I don't think this should be a concern at this moment.

As for pulling images, maybe we should reverse the logic. We can have `--pull`, optionally, to actually fetch the latest image. Without it, `docker run` will only download the image if not present. Wdyt?

cmd/soroban-cli/src/commands/contract/build/container.rs
fnando: This is handled now — `collect_built_contracts` no longer takes the first existing path. `newest_existing_artifact` picks the candidate with the most recent mtime, so a stale wasm left in the other triple's dir from an earlier build can't win over the one the container just wrote. Covered by the `newest_existing_artifact_prefers_freshest_not_first` unit test.

cmd/soroban-cli/src/commands/contract/build/container.rs
fnando: is,

fnando: `resolve_packages` does forward `--manifest-path`: `metadata(cmd)` passes it to `cargo metadata`, which narrows `workspace_default_members` to the member matching the manifest (or the cwd when run from a member dir), and both `resolve_packages` and `collect_built_contracts` filter on that set — so `--image` builds the same package as a local build. The `build_with_image_selects_package_by_manifest_path` integration test locks this in. Happy to add a from-a-subdirectory case if you'd like it covered explicitly too.

cmd/soroban-cli/src/commands/contract/build/container.rs
fnando: Good call — resolved in 70db2886. `cargo metadata` now runs once at the top of `run` and the single `Metadata` is threaded into `resolve_packages` and `collect_built_contracts` (and used for the workspace root), so a large or networked workspace pays one subprocess instead of three.

cmd/soroban-cli/src/commands/contract/build.rs
fnando: Went with the `--pull` reversal in 891fd047: the default now matches `docker run` — it uses the image already present locally and only pulls a missing one, keeping a locally-built or digest-pinned image as-is — and `--pull` opts in to refreshing a moving tag. Agreed that true airgapped support needs more than the image (cargo deps, etc.), so I've left that out of this PR to chase separately.

leighmcculloch: Sounds good. The alignment with existing tools will be less surprising.

leighmcculloch (approved):

Issue #2469, opened 2026-04-06
CLI panics when formatting output after successful contract invocation when return type is BytesN<32>


mugane-dj:
### stellar version
stellar 25.2.0 (28484880988199233a7e8e87c97cb12dac323cb3)
stellar-xdr 25.0.0 (dc9f40fcb83c3054341f70b65a2222073369b37b)
xdr curr (0a621ec7811db000a60efae5b35f78dee3aa2533)

### What did you do?
Invoked a schedule operation
```
stellar contract invoke \
    --id $TIMELOCK_ID \
    --source deployer \
    --network-passphrase "$NETWORK" \
    -- \
    schedule_op \
    --target $TIMELOCK_ID \
    --function upgrade \
    --args "[{\"bytes\":\"055ef816223ee5216b18c2df00d615ee089a8ef11b927a764e4f5d6c0cb4f4c7\"}]" \
    --predecessor 0000000000000000000000000000000000000000000000000000000000000000 \
    --salt 10561faae77b148a71ecd7f4bb22b653975b46e1430e19280e62978acaa39814 \
    --delay 600 \
    --proposer $DEPLOYER
```

### What did you expect to see?
Operation ID to be output by the CLI

### What did you see instead?
```
ℹ️ Simulating transaction…
ℹ️ Signing transaction: fa3bb43c15c2da1d721cc1eeb6ae3f4b378120d3119828568d6e3bbd19dae72b
🌎 Sending transaction…
✅ Transaction submitted successfully!
🔗 https://stellar.expert/explorer/testnet/tx/fa3bb43c15c2da1d721cc1eeb6ae3f4b378120d3119828568d6e3bbd19dae72b

thread 'main' (185501) panicked at cmd/crates/soroban-spec-tools/src/lib.rs:593:27:
not yet implemented: Bytes(
    ScBytes(
        BytesM(055ef816223ee5216b18c2df00d615ee089a8ef11b927a764e4f5d6c0cb4f4c7),
    ),
) doesn't have a matching Val
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Issue #2438, opened 2026-03-10
`stellar contract invoke` panics when contract function returns spec type `SC_SPEC_TYPE_VAL`


leighmcculloch:
### What version are you using?

stellar-cli 25.1.0

### What did you do?

When a Soroban contract function has return type `Val` (which maps to `SC_SPEC_TYPE_VAL` in the contract spec), and the function returns `Bool(false)`, calling the function via `stellar contract invoke` panics.

**Minimal reproducing contract:**

```rust
#![no_std]

use soroban_sdk::{contract, contractimpl, Env, Val};

#[contract]
pub struct BoolFalseContract;

#[contractimpl]
impl BoolFalseContract {
    pub fn retval(_env: Env) -> Val {
        Val::from_bool(false).into()
    }
}
```

**Cargo.toml:**

```toml
[package]
name = "bool_false_poc"
version = "0.1.0"
edition = "2021"
publish = false

[lib]
crate-type = ["cdylib"]

[dependencies]
soroban-sdk = "25.1.0"

[dev-dependencies]
soroban-sdk = { version = "25.1.0", features = ["testutils"] }

[profile.release]
opt-level = "z"
overflow-checks = true
debug = 0
strip = "symbols"
debug-assertions = false
panic = "abort"
codegen-units = 1
lto = true
```

**Steps to reproduce:**

```bash
# Build the contract
cargo build --release --target wasm32-unknown-unknown

# Deploy to local network
stellar contract deploy \
  --wasm target/wasm32-unknown-unknown/release/bool_false_poc.wasm \
  --source deployer --network local --alias bool-false-poc

# Invoke the function — this panics
stellar contract invoke --id bool-false-poc --source deployer --network local -- retval
```

### What did you expect to see?

The CLI should successfully display the return value, e.g.:

```
false
```

### What did you see instead?

The CLI panics with:

```
thread 'main' panicked at 'not yet implemented: Bool(false,) doesn't have a matching Val'
```

### Notes

- The contract itself executes correctly — the issue is purely in the CLI's result formatting
- All non-Void `ScVal` variants are affected when the spec type is `SC_SPEC_TYPE_VAL`, not just `Bool`
- A fix would need to handle `ScType::Val` for all `ScVal` variants, likely by recursively inferring the concrete type from the value itself

PR #2686, opened 2026-08-15
test: cover Bool returned as SC_SPEC_TYPE_VAL (#2438)

Galmanus:
### What
Adds a regression test locking the behavior reported in #2438: a contract function returning `Val` (`SC_SPEC_TYPE_VAL`) whose runtime value is `ScVal::Bool(false)` (or `true`) must render via `xdr_to_json` instead of panicking with "doesn't have a matching Val".

### Why this is test-only
The defect itself is **already fixed on `main`** by the generic-`Val` arm added in #2469 (`(val, ScType::Val) => to_json(val)`), which handles every `ScVal` variant. Existing regression coverage for that arm exercises `Bytes`, `Map` and `Vec` — but **not `Bool`**. This adds that missing case so a future refactor of the match can't silently reintroduce the panic.

Note: the `false` value in the original report is incidental — both `Bool(true)` and `Bool(false)` hit the same arm; the match discriminates on the variant, not the payload. #2438 can be closed as resolved by #2469 once this coverage lands.

### Tests
`cargo test -p soroban-spec-tools` and `cargo clippy -p soroban-spec-tools --all-targets` are green.

### DIFF
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -2511,6 +2511,23 @@ mod tests {
     }
 
     #[test]
+    fn test_xdr_to_json_bool_with_val_type() {
+        // Regression test for https://github.com/stellar/stellar-cli/issues/2438
+        // A contract fn returning `Val` (SC_SPEC_TYPE_VAL) whose runtime value is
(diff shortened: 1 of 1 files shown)


Galmanus:

PR #2688, opened 2026-08-18
Fix manifest path separator in container build

leighmcculloch:
### What

Build `--manifest-path` for `contract build --container`/`--image` with a forward-slash join instead of `Path::display()`.

### Why

`rel.display()` renders with the host OS separator, so on Windows it emits backslashes even though the path is passed to `contract build` running inside a Linux container, breaking every Windows CI run (`build-and-test-windows (msrv)` and `(latest)`) since #2678 landed.

### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -282,7 +282,12 @@ fn forwarded_build_args(
             .strip_prefix(workspace_root)
             .map(Path::to_path_buf)
             .unwrap_or(abs);
-        args.push(format!("--manifest-path={}", rel.display()));
+        let rel_posix = rel
```
leighmcculloch: Rebuilding the manifest path from `rel.components()` mishandles `RootDir`/`Prefix` components whenever `strip_prefix(workspace_root)` fails (e.g. manifest reachable via a different symlink/case than cargo_metadata's resolved workspace_root, or genuinely outside the workspace), producing a malformed value instead of a clean POSIX path.

When `abs.strip_prefix(workspace_root)` errors, `rel` falls back to the full absolute path (the `unwrap_or(abs)` above), which still carries `RootDir`/`Prefix` components. `Component::RootDir.as_os_str()` is the OS separator itself (`/` on Unix, `\` on Windows), so joining every component with `/` yields e.g. `//home/me/proj/contracts/add/Cargo.toml` on a Unix host (doubled leading slash) or `C:/\/Users/me/proj/contracts/add/Cargo.toml` on Windows (a literal backslash embedded in what's meant to be a pure-POSIX path) instead of the intended clean forward-slash path. A simpler, correct approach — `rel.to_string_lossy().replace(std::path::MAIN_SEPARATOR, '/')` — string-replaces the rendered path rather than reconstructing it from components, and doesn't hit this case.

---

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -282,7 +282,12 @@ fn forwarded_build_args(
...
+            .components()
+            .map(|c| c.as_os_str().to_string_lossy())
+            .collect::<Vec<_>>()
+            .join("/");
+        args.push(format!("--manifest-path={rel_posix}"));
```
leighmcculloch: This PR fixes Windows path-separator handling but ships with no test exercising an actual backslash-separated input, and CI's Windows job doesn't run on this PR branch to catch a regression either.

`forwarded_build_args_relativizes_manifest_path` (around line 863, and all other tests in this file) construct `manifest_path`/`ws()` using forward-slash strings (e.g. `/tmp/ws/contracts/add/Cargo.toml`), so `rel.components()` only ever sees `Normal` components split on `/` regardless of host OS — the old `rel.display()` and the new components-join produce byte-identical output for every existing test, so the tests pass identically before and after this change and never exercise the backslash-splitting behavior the PR claims to fix. Compounding this, `.github/workflows/rust.yml`'s `build-and-test-windows` job is gated to only run on `main`/`release/*` branches — it does not run for a normal PR branch like `fix-container-manifest-path-sep`, so even a native Windows run of the suite doesn't happen in this PR's CI; the fix's actual behavior is unverified until after merge to `main`.

---

Cargo.lock
leighmcculloch: `cargo-deny (advisories)` is failing on this PR with:

```
error[vulnerability]: h2 unbounded empty DATA frames
RUSTSEC-2026-0258, h2 v0.4.12
```

This is unrelated to this PR's manifest-path-separator fix. `h2 0.4.12` is pulled in transitively via `testcontainers -> bollard -> hyper` (a dev-dependency used only by the container-build integration tests), and that exact pin already existed in `Cargo.lock` on `main` at this PR's base commit (`80cc0f18`), before this PR touched anything. The advisory appears to have been published after that, so it now flags a pre-existing version rather than anything introduced here.

Also, the `advisories` leg of the `cargo-deny` matrix job runs with `continue-on-error: true` in `.github/workflows/rust.yml`, so this doesn't block the `complete` gating job or this PR's mergeability. Not fixing it here since it's out of scope for this change.

---

.github/workflows/rust.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           rust-version: ${{ matrix.rust }}
 
   build-and-test-windows:
-    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/') || startsWith(github.head_ref, 'release/')
+    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/') || startsWith(github.head_ref, 'release/') || github.event.pull_request.number == 2688
```
fnando: do we need to keep this?

cmd/crates/soroban-test/tests/it/build.rs
```diff
@@ -106,7 +106,7 @@ fn build_with_image_selects_package_by_manifest_path() {
         .arg("build")
         .arg("--image")
         .arg("docker.io/stellar/stellar-cli:latest")
-        .arg(manifest_path_arg(&add_path()))
+        .arg(format!("--manifest-path={}", add_path()))
```
fnando, line 109: we should still use manifest_path_arg, because it escapes the value.

leighmcculloch: Re "we should still use manifest_path_arg, because it escapes the value":

`manifest_path_arg`'s `escape()` call is for a different purpose elsewhere in this file — it builds the *expected stdout string* to compare against `--print-commands-only` output (e.g. lines 21-23, 129, 204, 224), matching how the CLI itself shell-escapes arguments when printing a command line it would run in a shell.

Here at line 109, the manifest path is passed straight to `assert_cmd`'s `.arg()`, which becomes a literal `argv` entry with no shell involved — no escaping is needed or correct in that context. Using `manifest_path_arg` here was actually the original Windows CI failure: on Windows, `shell_escape` wraps the backslash-separated path (`contracts\add\Cargo.toml`) in literal quote characters, and since there's no shell to strip them, clap sees a single argument that starts with `"` instead of `--manifest-path=...` and rejects it as unrecognized (`error: unexpected argument ''--manifest-path=...''`). That's the CI failure this PR was created to fix.

I've kept the direct `format!("--manifest-path={}", add_path())` here rather than reverting to `manifest_path_arg`, since reverting would reintroduce that Windows failure. Happy to add a comment on this line calling out the distinction if that'd help future readers.

---

leighmcculloch: Do we need to escape the value here?

As far as I can tell `manifest_path_arg`'s `escape()` call builds the expected stdout string to compare against `--print-commands-only` output matching how the CLI itself shell-escapes arguments when printing a command line it would run in a shell.

Here the manifest path is passed straight to `assert_cmd`'s `.arg()`, which becomes a literal `argv` entry with no shell involved meaning no escaping should be needed or correct in that context.

I think using `manifest_path_arg` here was what is causing the Windows CI failures: on Windows, `shell_escape` wraps the backslash-separated path (`contracts\add\Cargo.toml`) in literal quote characters, and since there's no shell to strip them, clap sees a single argument that starts with `"` instead of `--manifest-path=...`.

What am I missing?

.github/workflows/rust.yml
leighmcculloch: 485617f

cmd/crates/soroban-test/tests/it/build.rs
fnando: yeah, i think here it's fine to drop the escaping. It's also just a test.

fnando (approved):

PR #2692, opened 2026-08-22
Accept Rust-style _ digit separators in numeric contract arguments

### LINKED ISSUE #2447: Allow number separator for argument parsing

When invoking contracts which involve numbers, e.g. an amount field, you might end up having to write large values like 1000000000. At first glance it's hard to parse and error prone. This is a common problem when dealing with asset as you could have to offset by 1e7.

You might be tempted to use handy separators like `_` which currently leads to the following error:

`Context: Expected type i128 (signed 128-bit integer), but received: '100_0000000'`

I would like to be able to define numbers with separators like it can be done in Rust.

Do nothing.

Galmanus:
Closes #2447

## What

`stellar contract invoke -- fn --amount 100_0000000` now parses, instead of failing with `Expected type i128 (signed 128-bit integer), but received: '100_0000000'`.

## How

`Spec::from_string` strips underscores before parsing when both hold:

- the target type is numeric: `u32`/`i32`/`u64`/`i64`/`u128`/`i128`/`u256`/`i256`/`timepoint`/`duration`
- every underscore sits strictly between two digits (Rust-literal style)

Everything else is passed through untouched:

- malformed separators (`_1000`, `1000_`, `1__000`) still fail with the existing error
- non-numeric types keep their underscores — `hello_world` as a `Symbol`, `"1_000"` as a `String`, byte payloads

`Option<numeric>` works via the existing recursion into the inner type.

## Testing

- New unit tests: separator acceptance across i128/u32/i64/u256 and `Option<u32>`, rejection of malformed forms, and non-numeric passthrough
- `cargo test -p soroban-spec-tools`: 104 passed
- `cargo clippy -p soroban-spec-tools --all-targets`: clean

### DIFF
--- a/cmd/crates/soroban-spec-tools/src/lib.rs
+++ b/cmd/crates/soroban-spec-tools/src/lib.rs
@@ -811,6 +811,9 @@ pub fn from_string_primitive(s: &str, t: &ScType) -> Result<ScVal, Error> {
 
 /// Strips Rust-style `_` digit separators (e.g. `100_000_0000`) from `s` when
 /// the target type is numeric and every underscore sits between two digits.
+/// Only plain decimal input (with an optional leading sign) qualifies:
+/// base-prefixed forms such as `0x1_000` are returned unchanged, so the
+/// downstream parser rejects the underscore instead of misreading the value.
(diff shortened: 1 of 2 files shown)


Galmanus:

PR #2693, opened 2026-08-22
fix: don't warn about upgrades from stale cached versions when crates.io fetch fails

Galmanus:
### What
When the crates.io fetch fails, `has_available_upgrade` no longer reports the cached versions as the latest release. The fetch failure is propagated, so no upgrade warning is printed from stale data. The cached check time still advances, keeping the retry throttled to once per check interval.

### Why
Fixes #2464. The cached versions can be arbitrarily old (the cache is only refreshed on a successful fetch), so printing an upgrade warning from them tells the user to upgrade to a version that is itself outdated.

### Testing
- `test_apply_fetch_result_failure_is_propagated_not_swallowed` — regression test for the stale-warning path: failure surfaces, cached versions untouched, check time advances.
- `test_apply_fetch_result_success_replaces_cached_versions` — success path replaces versions wholesale.
- `cargo test -p soroban-cli --lib upgrade_check`: 7 passed. `cargo clippy` and `cargo fmt --check` clean.

### DIFF
--- a/cmd/soroban-cli/src/upgrade_check.rs
+++ b/cmd/soroban-cli/src/upgrade_check.rs
@@ -89,8 +89,10 @@ pub async fn has_available_upgrade(
         // Without a fresh answer from crates.io the cached versions may be
         // arbitrarily old, and reporting them as "the latest release" tells
         // the user to upgrade to a version that is itself outdated (#2464).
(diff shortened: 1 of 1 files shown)


Galmanus:

Issue #2549, opened 2026-05-04
Improve multisig handling


tupui:
### What problem does your feature solve?

I did not manage to easily use the CLI to call my contract in a multisig setup.

E.g. using this argument twice is failing
`error: the argument '--sign-with-key <SIGN_WITH_KEY>' cannot be used multiple times`

### What would you like to see?

We need to promote multisig in all its forms. I am trying to have a G address which is a multisig, using this with the CLI proves to be hard if not impossible. I did not yet manage to do `stellar contract deploy` for instance. I could build the TX with the CLI but then using labs to sign twice does not work and returns an error.

The only way I could find was to initialise first the contract with a normal G, then update the admin with a multisig G. Then we can build a TX XDR with the CLI and call multiple times the CLI to sign with different addresses and send. Or use labs, but here as well multisig is not easy to use and involves copy pasting.

Side note: on the contract initialisation I was faced with `data:["[recording authorization only] encountered authorization not tied to the root contract invocation for an address. Use `require_auth()` in the top invocation or enable non-root authorization.` as I wanted to directly pass the G multisig to be used. But this does not work, I had to then add another admin function to go around the problem. I could not find how to `enable non-root authorization` after looking into the docs and the code of the SDK.

### What alternatives are there?

- Not using a G address with multisig for calling contracts.
- Continue like that. Which to me is a terrible DX.

tupui:
Correction on the deploy part itself, I could do it. The difficulty is that you have to be very careful with the source-account, have to do manual tx signing in chain and doing simulation before signing otherwise signatures are removed.

```bash
override network = testnet
override admin = bob --hd-path 1
override extra_signer = toto

contract_deploy_tx:
	stellar contract deploy \
  		--wasm ... \
  		--source-account $(admin) \
  		--network $(network) \
  		--build-only \
  		-- \
  		--admin $(admin)

contract_deploy_multisig:
	make contract_deploy_tx | sed -n '$$p' | \
	stellar tx simulate --source-account $(admin) | \
	stellar tx sign --sign-with-key $(admin) --network $(network) | \
	stellar tx sign --sign-with-key $(extra_signer) --network $(network) | \
	stellar tx send
```

PR #2696, opened 2026-08-22
fix: warn when tx simulate discards existing signatures

Galmanus:
### What
`tx simulate` silently dropped any signature already attached to the incoming envelope — `unwrap_envelope_v1` destructures it and the re-wrapped result always carries an empty signature list. It now warns with the discarded count and the correct order of operations (simulate first, then sign).

### Why
This is the concrete pain reported in #2549 (the author's follow-up comment): in a multisig flow, collected signatures vanish with no indication. The signatures are genuinely unusable after simulation — it rewrites the fee and Soroban transaction data, which invalidates them — so preserving them would be wrong; the missing piece was telling the user. This PR references #2549 rather than closing it: the issue also asks for repeated `--sign-with-key` and broader multisig UX, which are larger design decisions.

### Testing
- New `signature_count` helper in `tx/xdr.rs` with a unit test covering signed and unsigned envelopes; `cargo test -p soroban-cli --lib tx::xdr::`: 7 passed.
- `cargo clippy` and `cargo fmt --check` clean.

### DIFF
--- a/cmd/soroban-cli/src/commands/tx/simulate.rs
+++ b/cmd/soroban-cli/src/commands/tx/simulate.rs
@@ -66,7 +66,14 @@ impl Cmd {
         // signature on the incoming envelope is invalid on the simulated
         // result. They were silently dropped before; say so, since the correct
         // order (simulate, then sign) is the actual fix on the user's side.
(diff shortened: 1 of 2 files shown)


Galmanus:

PR #2697, opened 2026-08-22
feat: preserve the signed envelope in the action log when a send fails

### LINKED ISSUE #2609: Graceful tx fail with retry

If for some reason sending a tx fails, we need to resubmit it. It might involves recrafting and resigning a tx which can be problematic. E.g. using a Ledger.

I would like to get a better error message and at least get the failed XDR. We could even propose the command to run and retry to send the tx.

- Retry as usual. Implies a high churn on users if they use a secret store.
- Add a specific retry command for the last TX.
- Store XDRs in a tmp file and have a command to fetch and present this log. (I actually like this and good for auditability if the CLI is used in a compliant environment.)

Galmanus:
### What
When `send_transaction_polling` fails, both send paths (`tx.rs::send` and `tx/args.rs::handle_tx`) now save the signed envelope to the action log as a new `Action::SendFailed { envelope_xdr }` entry and print the exact pipeline to resubmit without re-signing:

```
stellar cache actionlog read --id <ULID> | jq -r .action.send_failed.envelope_xdr | stellar tx send
```

### Why
Fixes #2609, following the direction in the issue discussion of extending the existing action-log machinery rather than adding a retry command. Previously the action-log write only ran after a successful submit, so a network hiccup after interactive signing (hardware wallet, multisig) lost the envelope and every signature had to be collected again.

Notes for review:
- The envelope is written through the existing `data::write` path — hardened file permissions, RPC URL redacted — the same guarantees the current action log provides.
- Saving is best-effort: a cache-write failure is logged at debug level and never masks the original RPC error, and `--no-cache` skips it entirely.
- The action log is flagged Experimental, so the new externally-tagged variant is a compatible addition; old entries still deserialize.

### Testing
- New unit test `test_send_failed_round_trips_and_renders_as_failed`: the variant round-trips through `write`/`read` unchanged and `actionlog ls` renders a distinct `SendFail`/`FAILED` row. Uses the existing `STELLAR_DATA_HOME` + `#[serial]` harness.
- `cargo test -p soroban-cli --lib config::data::`: 4 passed. `cargo clippy` and `cargo fmt --check` clean.

### DIFF
--- a/cmd/crates/soroban-test/tests/it/integration/tx/payment.rs
+++ b/cmd/crates/soroban-test/tests/it/integration/tx/payment.rs
@@ -1,8 +1,9 @@
 use soroban_cli::tx::ONE_XLM;
+use soroban_cli::xdr::{Limits, ReadXdr, TransactionEnvelope};
 
-use soroban_test::TestEnv;
+use soroban_test::{AssertExt, TestEnv};
 
-use crate::integration::util::setup_accounts;
+use crate::integration::util::{gen_account_no_fund, setup_accounts};
 
 #[tokio::test]
 async fn payment_with_alias() {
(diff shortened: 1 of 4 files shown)


Galmanus:

PR #2700, opened 2026-08-24
Show diagnostic events when a transaction fails on-chain

### LINKED ISSUE #495: soroban-cli: Print more detailed transaction submission errors
Right now when a transaction submission fails (e.g. when running `soroban invoke` like at https://github.com/stellar/soroban-tools/issues/490 ) the CLI only prints out: `error: transaction submission failed` 

Unfortunately this is not informative nor actionable, we should be adding extra information to the error.

fnando:
### What

When a transaction passes simulation but fails on-chain, the CLI now prints the transaction's diagnostic events (host/contract error, message, and any logs) on stderr, across `contract invoke`, `tx send`, and the `tx` op commands. Previously only the `TransactionResult` result codes were shown. Closes #495.

### Why

The reason for a failure was never displayed — the diagnostic events were decoded and available on the RPC response but discarded before reaching the user.

**Before** (`main`):

```
🌎 Sending transaction…
❌ error: transaction submission failed: Some(
    TransactionResult {
        fee_charged: 10858,
        result: TxFailed(
            VecM([OpInner(InvokeHostFunction(ResourceLimitExceeded))]),
        ),
        ext: V0,
    },
)
```

**After** (this branch):

```
🌎 Sending transaction…
❌ Error event: [{"symbol":"error"},{"error":{"budget":"exceeded_limit"}}] = {"vec":[{"string":"operation instructions exceeds amount specified"},{"u64":"3880"},{"u64":"1"}]}
❌ error: transaction submission failed: Some(
    TransactionResult { ... InvokeHostFunction(ResourceLimitExceeded) ... },
)
```

The new `Error event` line surfaces the host's message (`operation instructions exceeds amount specified`, needs `3880` vs `1` allowed) — absent before.

### Known limitations

Only Soroban transactions have diagnostic events; classic ops print nothing extra. The event line uses raw `ScVal` JSON rather than fully decoded text.

### DIFF
--- a/cmd/soroban-cli/src/log/event.rs
+++ b/cmd/soroban-cli/src/log/event.rs
@@ -131,6 +131,44 @@ pub fn contract_with_spec(events: &[DiagnosticEvent], print: &Print, spec: Optio
     }
 }
 
+/// Render an `error` diagnostic event as a readable line.
+///
+/// `topics` carries the error type/code (an `ScError`) and `data` carries the
+/// message payload. Both are serialized as JSON, which escapes any control
+/// bytes present in the payload.
+fn format_error_event(topics: &[xdr::ScVal], data: &xdr::ScVal) -> String {
+    let topics_json = serde_json::to_string(topics).unwrap();
+    let data_json = serde_json::to_string(data).unwrap();
+    format!("Error event: {topics_json} = {data_json}")
+}
+
+/// Display diagnostic events for a failed transaction.
+///
+/// Reuses [`contract_with_spec`] to render contract events and `log` diagnostic
+/// events, and additionally surfaces the `error` diagnostic event (the host or
+/// contract error), which the success-path renderer intentionally drops. This is
+/// the most actionable information when a submitted transaction fails.
+pub fn failure(events: &[DiagnosticEvent], print: &Print) {
+    contract_with_spec(events, print, None);
+
+    for event in events.iter().cloned() {
+        if let DiagnosticEvent {
+            event:
+                ContractEvent {
+                    body: ContractEventBody::V0(ContractEventV0 { topics, data, .. }),
+                    type_: ContractEventType::Diagnostic,
+                    ..
+                },
+            ..
+        } = event
+        {
+            if topics.first() == Some(&xdr::ScVal::Symbol(str_to_sc_symbol("error"))) {
+                print.errorln(format_error_event(topics.as_slice(), &data));
+            }
+        }
+    }
+}
+
 fn str_to_sc_symbol(s: &str) -> xdr::ScSymbol {
     let inner: xdr::StringM<32> = s.try_into().unwrap();
     xdr::ScSymbol(inner)
@@ -146,6 +184,25 @@ mod tests {
     use indexmap::IndexMap;
     use serde_json::json;
     use soroban_spec_tools::test_utils::assert_no_control_chars;
+
+    #[test]
+    fn format_error_event_renders_code_and_message_without_control_bytes() {
+        let topics = vec![
+            xdr::ScVal::Symbol(str_to_sc_symbol("error")),
+            xdr::ScVal::Error(xdr::ScError::Contract(7)),
+        ];
+        // A message payload carrying an ANSI control sequence to ensure it is escaped.
+        let data = xdr::ScVal::String(xdr::ScString("boom\x1b[31m".try_into().unwrap()));
+
+        let rendered = format_error_event(&topics, &data);
+
+        assert!(rendered.contains("boom"), "message payload should be shown");
+        assert!(
+            rendered.contains('7') || rendered.to_lowercase().contains("contract"),
+            "error code should be shown: {rendered}"
+        );
+        assert_no_control_chars(&rendered);
+    }
 
     #[test]
     fn format_decoded_event_strips_attacker_control_bytes() {
--- a/cmd/soroban-cli/src/tx.rs
+++ b/cmd/soroban-cli/src/tx.rs
@@ -149,14 +149,18 @@ pub async fn send_transaction_polling_with_events(
     match client.send_transaction_polling(signed_tx).await {
         Ok(res) => Ok(res),
         Err(e) => {
-            // Best-effort: recover the diagnostic events the RPC client discarded.
+            // Best-effort: recover the diagnostic events the submission error dropped.
+            //
+            // Since protocol 23 a failed transaction's diagnostic events are
+            // returned in the response's dedicated `diagnosticEventsXdr` field
+            // rather than embedded in the transaction meta, so read them from
+            // `events.diagnostic_events` (which the RPC client populates from that
+            // field) instead of from `result_meta`.
             if let Ok(hash) = transaction_env_hash(signed_tx, network_passphrase) {
                 if let Ok(resp) = client.get_transaction(&Hash(hash)).await {
-                    if let Some(meta) = resp.result_meta {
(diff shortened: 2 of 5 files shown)


cmd/soroban-cli/src/tx.rs
```diff
@@ -122,3 +128,38 @@ where
...
+    match client.send_transaction_polling(signed_tx).await {
+        Ok(res) => Ok(res),
+        Err(e) => {
+            // Best-effort: recover the diagnostic events the RPC client discarded.
+            if let Ok(hash) = transaction_env_hash(signed_tx, network_passphrase) {
+                if let Ok(resp) = client.get_transaction(&Hash(hash)).await {
+                    if let Some(meta) = resp.result_meta {
+                        let events = crate::log::extract_events(&meta);
+                        crate::log::event::all(&events);
+                        crate::log::event::failure(&events, print);
```
leighmcculloch: `send_transaction_polling_with_events` re-fetches and prints diagnostic events as a "failure" whenever `send_transaction_polling` returns `Err`, without checking the re-fetched `GetTransactionResponse.status`. A transient/network error mid-poll (not an actual on-chain `FAILED` result) can therefore print misleading "Error event" lines for a transaction that later actually succeeded.

Concretely: if `send_transaction_polling` times out or hits a transient RPC error while the submitted tx is still pending, and `get_transaction(&hash)` right after already shows `status == SUCCESS` (note `result_meta` is populated on success too, per `invoke.rs:411`), this code doesn't check `status` and unconditionally calls `log::event::failure(&events, print)` — printing stderr lines styled as errors for a transaction that actually succeeded, confusing the user into thinking it failed.

Consider checking `resp.status` before calling `failure()`.

---

cmd/soroban-cli/src/tx.rs
leighmcculloch: The doc comment on `sim_sign_and_send_tx` at line 27 ("Does not handle any logging related to the result, events, or effects of the transaction") is now stale/incorrect given this PR's changes — that function now routes through `send_transaction_polling_with_events`, which does print diagnostic "Error event" / log lines on failure.

A future maintainer reading that comment could assume the function stays silent about events/results and layer their own failure-event printing on top (or rely on it being side-effect-free), leading to duplicate output or wrong debugging conclusions. Worth updating the doc comment to reflect the new behavior.

---

cmd/soroban-cli/src/log/event.rs
```diff
@@ -131,6 +131,44 @@ pub fn contract_with_spec(events: &[DiagnosticEvent], print: &Print, spec: Optio
...
+    format!("Error event: {topics_json} = {data_json}")
+}
+
+/// Display diagnostic events for a failed transaction.
+///
+/// Reuses [`contract_with_spec`] to render contract events and `log` diagnostic
+/// events, and additionally surfaces the `error` diagnostic event (the host or
+/// contract error), which the success-path renderer intentionally drops. This is
+/// the most actionable information when a submitted transaction fails.
+pub fn failure(events: &[DiagnosticEvent], print: &Print) {
```
leighmcculloch: `failure()` does two full passes over `events` — one full clone inside `contract_with_spec` (which does `events.iter().cloned()` over the whole diagnostic-events vector) and a second full clone in this function's own loop just to find the `error`-topic event. For transactions with many diagnostic events this doubles allocation/iteration cost for no real benefit.

Not a correctness issue, just an efficiency one — the error-event check could likely be folded into the same match/loop that `contract_with_spec` already performs over the events, avoiding the second clone+iterate pass.

---

cmd/soroban-cli/src/tx.rs
```diff
@@ -122,3 +128,38 @@ where
...
+///
+/// On failure the RPC client returns only the (decoded) result codes and discards
+/// the transaction's diagnostic events. This helper best-effort re-fetches the
+/// failed transaction to recover and print those events — the host/contract error,
+/// its message, and any contract logs — which explain *why* the transaction
+/// failed. The original error is always preserved and returned unchanged.
+///
+/// # Errors
+/// Returns the underlying submission error if the transaction fails.
+pub async fn send_transaction_polling_with_events(
```
leighmcculloch: No test exercises `send_transaction_polling_with_events` itself, or its wiring into `contract invoke` / `tx send` / `tx op` — only the pure formatter `format_error_event` is unit tested. A future refactor could break the `get_transaction` re-fetch, the hash computation for fee-bump envelopes, or the quiet-mode gating on this failure path and nothing would catch it.

This overlaps with an existing Copilot review comment on this PR asking for `soroban-test` integration coverage of these entry points — worth addressing alongside that.

---

fnando:
This will need an update on stellar/rs-stellar-rpc-client as well.

cmd/soroban-cli/src/tx.rs
leighmcculloch: I think this comment is out-of-date now?

cmd/soroban-cli/src/tx.rs
```diff
@@ -122,3 +128,42 @@ where
...
+    match client.send_transaction_polling(signed_tx).await {
+        Ok(res) => Ok(res),
+        Err(e) => {
+            // Best-effort: recover the diagnostic events the submission error dropped.
+            //
+            // Since protocol 23 a failed transaction's diagnostic events are
+            // returned in the response's dedicated `diagnosticEventsXdr` field
+            // rather than embedded in the transaction meta, so read them from
+            // `events.diagnostic_events` (which the RPC client populates from that
+            // field) instead of from `result_meta`.
```
leighmcculloch, line 160: It depends when the failure occurs. If the failure occurs during submission then it will be returned in the former, otherwise if the tx does get included in a ledger and fails during apply, I believe it should be in the latter. So do we support both?

leighmcculloch (approved):
Couple questions otherwise looks good to me.

cmd/soroban-cli/src/tx.rs
leighmcculloch: Actually this might have changed and my statement might be incorrect.

cmd/soroban-cli/src/tx.rs
```diff
@@ -122,3 +128,38 @@ where
...
+///
+/// # Errors
+/// Returns the underlying submission error if the transaction fails.
+pub async fn send_transaction_polling_with_events(
+    client: &soroban_rpc::Client,
+    signed_tx: &TransactionEnvelope,
+    network_passphrase: &str,
+    print: &print::Print,
+) -> Result<GetTransactionResponse, soroban_rpc::Error> {
+    match client.send_transaction_polling(signed_tx).await {
```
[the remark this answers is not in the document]
fnando: Added integration coverage in `hello_world.rs` (`on_chain_failure_surfaces_diagnostic_events` plus a quiet-mode variant) through `contract invoke`. `tx send`/`tx`-op are classic flows that emit no diagnostic events, and share the same recovery path the invoke tests exercise.

cmd/soroban-cli/src/tx.rs
fnando: Good catch — fixed in 18e34282.

cmd/soroban-cli/src/tx.rs
fnando: P23 moved these to `diagnosticEventsXdr` (`events.diagnostic_events`), so this is the apply-in-ledger case. Submission-time rejections carry none, so the re-fetch is a no-op there.

Uthen242 (approved):

PR #2704, opened 2026-08-26
Find example wasm when target lives at workspace root

fnando:
### What

Fix the `Find wasm` step in the "Test with soroban-examples" workflow so it locates the built `.wasm` when an example is a Cargo workspace, searching the whole checkout instead of only `${matrix.dir}/target`.

### Why

The `soroban-examples` repo recently added the `groth16_verifier` example (containing `bn254_verifier` and `bls12_381_verifier`), which is a Cargo workspace with its contracts as members. For workspaces, build output goes to the workspace-root `target/`, not `${matrix.dir}/target`, so the step failed with `No such file or directory` and broke CI on every open PR and on `main`. Searching from the checkout root matches the pattern already used in `test-with-openzeppelin-contracts.yml`.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/test-with-soroban-examples.yml
+++ b/.github/workflows/test-with-soroban-examples.yml
@@ -91,7 +91,8 @@ jobs:
       - name: Find wasm
         id: find-wasm
         run: |
-          wasm=$(find . -name '*.wasm' -path '*/wasm32v1-none/release/*' -not -path '*/deps/*' | head -1)
+          root=$(dirname "$(cargo locate-project --workspace --message-format plain --manifest-path ${{ matrix.dir }}/Cargo.toml)")
+          wasm=$(find "$root/target" -name '*.wasm' -path '*/wasm32v1-none/release/*' -not -path '*/deps/*' | head -1)
           echo "wasm=$wasm" | tee -a $GITHUB_OUTPUT
       - name: Set artifact name
         id: artifact-name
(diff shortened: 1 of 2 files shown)


.github/workflows/test-with-soroban-examples.yml
```diff
@@ -91,7 +91,7 @@ jobs:
       - name: Find wasm
         id: find-wasm
         run: |
-          wasm=$(find ${{ matrix.dir }}/target -name '*.wasm' -path '*/wasm32v1-none/release/*' -not -path '*/deps/*' | head -1)
+          wasm=$(find . -name '*.wasm' -path '*/wasm32v1-none/release/*' -not -path '*/deps/*' | head -1)
```
leighmcculloch: We should probably rethink this command because it will only pick one .wasm file, but that won't actually support multiple contract workspaces will it?

leighmcculloch: Instead of finding from the root, another approach would be to execute `cargo locate-project --workspace` from within `matrix.dir` to locate the root of the workspace and then `target/` should be in that directory.

fnando: Done — switched to `cargo locate-project --workspace` to find the workspace root and search its `target/`. Applied the same fix to `test-with-openzeppelin-contracts.yml` for consistency.

fnando: Good catch. The matrix runs one contract per job, so we only need a single wasm here, but scoping the lookup to the workspace root (instead of the whole checkout) means we won't grab an unrelated wasm. Broader multi-contract-per-job support would be a larger change we can do as a follow-up if needed.

leighmcculloch (approved):

PR #2708, opened 2026-08-31
Follow SEP-41 argument names in token commands regardless of contract naming

fnando:
### What

`stellar token balance` and `stellar token transfer` now map their arguments to the contract **by position** instead of by the contract's parameter names. Callers supply values in SEP-41's canonical order (`transfer(from, to, amount)`, `balance(id)`, `decimals()`) and the contract's actual parameter names no longer matter.

### Why

The commands used to build an argv like `transfer --from … --to … --amount …` and let the generic invoke parser match those flags to the contract spec's parameter names. That only worked when the deployed contract literally named its parameters `from`/`to`/`amount`/`id`. A SEP-41-compliant contract can name them anything (`sender`, `recipient`, `amt`, `addr`, …), so the commands broke against those contracts. SEP-41 fixes the function names and argument order, not the parameter names — so mapping by position is the correct model. Doing this now keeps the two existing commands correct before more token subcommands (approve, allowance, mint, burn, …) are built on the same pattern.

### Known limitations

Positional mapping requires the contract function's arity to match the SEP-41 signature exactly; a deviating arity fails with a clear error rather than sending a malformed call. Attacker-influenceable spec parameter names are sanitized before appearing in any error message, consistent with the existing invoke path.

### DIFF
--- a/cmd/soroban-cli/src/commands/token/balance.rs
+++ b/cmd/soroban-cli/src/commands/token/balance.rs
@@ -1,5 +1,3 @@
-use std::ffi::OsString;
-
 use clap::Parser;
 
 use crate::{
@@ -11,7 +9,7 @@ use crate::{
     config::{
         self, locator, network, sign_with,
         token::{ResolvedToken, UnresolvedToken},
-        UnresolvedContract, UnresolvedScAddress,
+        UnresolvedScAddress,
     },
     fixed_point::FixedPoint,
     output::Output,
@@ -130,12 +128,8 @@ impl Cmd {
                 quiet,
                 global_args.no_cache,
                 &token,
-                vec![
-                    OsString::from("balance"),
-                    OsString::from("--id"),
-                    OsString::from(&account),
-                ],
                 "balance",
+                vec![account],
             )
             .await?;
 
@@ -149,8 +143,8 @@ impl Cmd {
                     quiet,
                     global_args.no_cache,
                     &token,
-                    vec![OsString::from("decimals")],
                     "decimals",
+                    vec![],
                 )
                 .await?;
             (FixedPoint::new(raw, decimals).to_string(), Some(decimals))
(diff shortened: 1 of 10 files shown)


cmd/soroban-cli/src/commands/token/balance.rs
```diff
@@ -130,12 +128,8 @@ impl Cmd {
                 quiet,
                 global_args.no_cache,
                 &token,
-                vec![
-                    OsString::from("balance"),
-                    OsString::from("--id"),
-                    OsString::from(&account),
-                ],
                 "balance",
+                vec![account],
```
leighmcculloch, line 132: It's a bit unclear to me how this works, is the idea that the fns lose the named options like `--id` and are no longer in the form `stellar token balance --id G...` and become positional, for e.g. `stellar token balance G...`? That deviation from the way that `stellar contract invoke` works seems unnecessary, since the main problem being addressed is that a contract might use different names for fields other than those defined in SEP-41, and we can solve that problem with a change that diverges much less from the `contract invoke` subcommand by playing a game of aliasing.

By aliasing I mean taking the arguments as resolved, and adding an alias to the each field with their name as known by SEP-41 given their ordered position in the fn.

leighmcculloch: The example I use above with 'id' being the account being queried is incorrect, instead using transfer as a better example:

Today I understand the command as:
```
stellar token transfer --id C... --from G... --to G... --amount 2
```

Can we introduce aliasing, so that a contract whose invoke command looks like this because of custom argument names:
```
stellar contract invoke --id C... -- transfer --src G... --dst G... --amount 2
```
Can be used as follows via aliasing:
```
stellar token transfer --id C... --src G... --dst G... --amount 2
or
stellar token transfer --id C... --from G... --to G... --amount 2
```
Or if supporting the custom names is undesirable, simply replacing the names but keeping the use of options/flags:
```
stellar token transfer --id C... --from G... --to G... --amount 2
```

fnando: Because invoking a contract requires knowing the argument names as defined in the spec, we can't assume developers followed the naming as defined in SEP-41. E.g. `fn allowance(env: Env, from: Address, spender: Address) -> i128;` could be defined as `fn allowance(env: Env, addr: Address, consumer: Address) -> i128;`, which would need to be invoked as `--addr <address> --consumer <address>`.

So, instead, this allows us to call the function with positional arguments instead, which then is mapped to the actual contract's names.

The command's name must be stable and will match SEP-41's spec (e.g. `fn allowance(env: Env, from: Address, spender: Address) -> i128;
` will directly translate to `--from <from> --spender <spender>`).

leighmcculloch: One really big downside to positional cli's though, it's easy to mixup and misunderstand the parameters which for a token transfer can be detrimental.

For example, it's difficult to determine in the following positional cli use who is the sender and receiver:

```
stellar token transfer C... C... C... 2
```

Where-as with the use of aliasing, we retain that explicitness still fully addressing the issue of field names:

```
stellar token transfer --id C... --from G... --to G... --amount 2
```

If we want positional arguments, then I think we should do it on all invokes too rather than making this a bespoke behaviour for the token commands. The `stellar contract invoke` command could support both named and positional invocation. And then the token command would benefit from that and we'd have more consistency across the CLI.

Whichever way we go here, it can work for all the invoking commands, either through aliasing (which achieves the same thing) or through supporting positional on both. My preference is aliasing because I think the explicitness makes it safer.

fnando: Oh, positional args are not meant for users. It's just an internal thing that will be used by `stellar token` commands. They're not even exposed to the CLI, so they can't run commands like `stellar token transfer arg1 arg2 arg3`.

leighmcculloch: Ok so this change is essentially the same as the aliasing described above, just without allowing for the original contracts arg names? Sgtm

fnando: Right!

leighmcculloch (approved):

PR #2713, opened 2026-09-04
Add `stellar token decimals` subcommand

fnando:
### What

Adds `stellar token decimals`, a read-only subcommand returning a token's SEP-41 `decimals()` metadata as a `u32`. Resolves a token by `--id` (contract id, alias, `native`, or `CODE:ISSUER`) and prints the value as text or JSON.

### Why

Part of #2620 (typed SEP-41 + SAC client), completing the read-metadata group after `name` and `symbol`. Stacked on the `token-symbol` branch; a thin wrapper over `contract invoke` reusing `args::invoke_by_position` and `args::not_deployed_error`.

### Known limitations

N/A

### DIFF
--- a/cmd/soroban-cli/src/commands/token/decimals.rs
+++ b/cmd/soroban-cli/src/commands/token/decimals.rs
@@ -1,0 +1,123 @@
+use clap::Parser;
+
+use crate::{
+    commands::{
+        contract::invoke,
+        global,
+        token::args::{self, OutputFormat},
+    },
+    config::{self, locator, network, sign_with, token::UnresolvedToken},
+    output::Output,
+};
+
+#[derive(Debug, Parser, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// The token to query: a contract id or alias, `native`, or a classic asset
+    /// as `CODE:ISSUER`.
+    #[arg(long = "id")]
+    pub id: UnresolvedToken,
+
+    /// Format of the output.
+    #[arg(long, default_value = "text")]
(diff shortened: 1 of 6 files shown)


cmd/soroban-cli/src/commands/token/decimals.rs
```diff
@@ -0,0 +1,123 @@
...
+        let quiet = global_args.quiet || output.is_json();
+        let config = self.config();
+        let network = config.get_network()?;
+
+        let token = self
+            .id
+            .resolve(&config.locator, &network.network_passphrase)?;
+
+        // SEP-41 `decimals()` takes no arguments and returns a `u32`.
+        let receipt = args::invoke_by_position(
```
leighmcculloch: The invoke-then-decode boilerplate (`invoke_by_position` + `not_deployed_error` + manual `serde_json` decode with a hardcoded default) is now duplicated a third time (after `name.rs` and `symbol.rs`), and `balance.rs` already has a more general private helper (`read_parsed<T: FromStr>`) doing the same job for both quoted and bare JSON values.

Not a runtime bug, but a maintenance cost: a future fix to the invoke/parse/error-translation logic (e.g. a new failure mode from `invoke_by_position`) has to be applied in four separate places (`name.rs`, `symbol.rs`, `decimals.rs`, `balance.rs`'s `read_parsed`) instead of one shared helper in `args.rs`, and it's easy to update three and miss the fourth.

---

leighmcculloch (approved):

Issue #2645, opened 2026-07-14
Support muxed (M…) source accounts in the contract invoke pipeline


fnando:
### Background

Spun out of review feedback on #2644 (`stellar token transfer`). SEP-41 supports muxed addresses (CAP-67 `ScAddress::MuxedAccount`), and `--from M…` almost works today: `UnresolvedMuxedAccount` resolves it, and arg parsing + simulation accept it.

### Problem

It then fails in the shared `contract invoke` pipeline at `cmd/soroban-cli/src/commands/contract/invoke.rs:349`:

```rust
client.get_account(&config.source_account()?.to_string()).await?
```

The `M…` strkey is handed to a G-only ed25519 parser, aborting with a raw strkey `DecodeError`. Because this lives in the shared pipeline, `contract invoke --source-account M…` hits the same failure — it is not specific to `token transfer`.

### Direction

`stellar-xdr` already provides `xdr::MuxedAccount::account_id()`. The pipeline should use the underlying `AccountId` (e.g. `config.source_account()?.account_id().to_string()`) for the sequence-number lookup and the classic transaction source in `build_invoke_contract_tx`, while the muxed form keeps flowing into the SEP-41 `transfer(from, …)` argument so the mux id is preserved on-chain / in events.

### Acceptance

- `--source-account M…` / `--from M…` works end-to-end through simulate → sign → submit.
- Add an integration test with an `M…` source (signing via an identity / `--sign-with-*`, since an M-literal carries no secret).

### Notes

Pre-existing limitation, not a regression from #2644. Until this lands, `token transfer` may want a clear "muxed source not yet supported" guard error instead of the raw `DecodeError`.

Part of the "Improve Agent UX for Wallets" epic (#2620).

Galmanus:
@fnando quick design question before #2694 gets a review pass, since it deviates from the Direction here on one point and I'd rather settle it in the issue than in review.

The Direction says the muxed form should keep flowing into the SEP-41 `transfer(from, …)` argument so the mux id is preserved. Empirically that path fails: on a protocol-27 quickstart, a muxed `from` fails simulation with `HostError: Error(Value, UnexpectedType)` — consistent with muxed addresses not being accepted by `require_auth` (the same ruling #2534 encodes for auth entries). So #2694 currently resolves a muxed `--from` to the underlying `G…` for the `transfer` argument, while a muxed `--to` still flows through as-is and keeps the mux id in the argument and events.

Two ways to go:

[...]

PR #2714, opened 2026-09-04
Add `stellar token approve` subcommand

fnando:
### What

Adds `stellar token approve`, a write subcommand that sets a SEP-41 allowance. `--from` grants and authorizes the allowance, `--spender` is the delegate, `--amount` is the allowance in smallest units, and `--expiration-ledger` is the ledger after which it expires. Returns a JSON receipt with the tx hash.

### Why

Part of #2620 (typed SEP-41 + SAC client), first of the write commands after the read-metadata group. A thin wrapper over `contract invoke` reusing `args::invoke_by_position` and `args::not_deployed_error`, mirroring `transfer`. Landing `approve` before `allowance` lets the allowance command's tests do a clean approve→read round-trip.

### Known limitations

### DIFF
--- a/cmd/soroban-cli/src/commands/token/approve.rs
+++ b/cmd/soroban-cli/src/commands/token/approve.rs
@@ -1,0 +1,223 @@
+use clap::Parser;
+
+use crate::{
+    commands::{
+        contract::invoke,
+        global,
+        token::args::{self, OutputFormat},
+    },
+    config::{
+        self, locator, network, sign_with, token::UnresolvedToken, UnresolvedMuxedAccount,
+        UnresolvedScAddress,
+    },
+    output::Output,
+};
+
+#[derive(Debug, Parser, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// The token to approve an allowance on: a contract id or alias, `native`,
+    /// or a classic asset as `CODE:ISSUER`.
+    #[arg(long = "id")]
+    pub id: UnresolvedToken,
+
+    /// Account granting the allowance. Signs and authorizes the approval, so it
+    /// must be an identity or secret key you control.
+    #[arg(long)]
+    pub from: UnresolvedMuxedAccount,
+
+    /// Account or contract allowed to spend on `--from`'s behalf. Accepts a
+    /// `G…`/`M…` account, a `C…` contract address, or an alias.
+    #[arg(long)]
+    pub spender: UnresolvedScAddress,
+
+    /// Allowance to grant, in the token's smallest unit (stroops for a Stellar
+    /// Asset Contract). Replaces any existing allowance.
+    #[arg(long, value_parser = parse_nonneg_i128)]
+    pub amount: i128,
+
+    /// Ledger sequence after which the allowance expires. Must be at or beyond
+    /// the current ledger when the amount is positive.
+    #[arg(long = "expiration-ledger")]
+    pub expiration_ledger: u32,
+
+    /// Format of the output.
+    #[arg(long, default_value = "text")]
+    pub output: OutputFormat,
+
+    #[command(flatten)]
+    pub network: network::Args,
+
+    #[command(flatten)]
+    pub locator: locator::Args,
+
+    #[command(flatten)]
+    pub sign_with: sign_with::Args,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Config(#[from] config::Error),
+    #[error(transparent)]
+    Network(#[from] network::Error),
+    #[error(transparent)]
+    Args(#[from] args::Error),
+    #[error(transparent)]
+    Token(#[from] config::token::Error),
+    #[error(transparent)]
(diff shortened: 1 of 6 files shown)


cmd/soroban-cli/src/commands/token/approve.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// `G…`/`M…` account, a `C…` contract address, or an alias.
+    #[arg(long)]
+    pub spender: UnresolvedScAddress,
+
+    /// Allowance to grant, in the token's smallest unit (stroops for a Stellar
+    /// Asset Contract). Replaces any existing allowance.
+    #[arg(long, value_parser = parse_nonneg_i128)]
+    pub amount: i128,
+
+    /// Ledger sequence after which the allowance expires. Must be at or beyond
```
leighmcculloch, line 39: The CLI validates `--amount` is non-negative (via `parse_nonneg_i128`, see the `Error::Args`/`args::Error` layer) to avoid an opaque `HostError`, but this doc comment states the same trap condition applies to `--expiration-ledger` — a positive amount with an expiration ledger already in the past — yet there's no equivalent validation.

cmd/soroban-cli/src/commands/token/approve.rs
```diff
@@ -0,0 +1,223 @@
...
+#[derive(Debug, serde::Serialize)]
+struct Receipt {
+    /// Hex-encoded hash of the submitted transaction.
+    tx_hash: Option<String>,
+    /// The decoded contract return value (`null` for SEP-41 `approve`, which
+    /// returns nothing).
+    result: serde_json::Value,
+}
+
+impl Cmd {
```
leighmcculloch: `Cmd::config()`, the `MuxedSourceNotSupported` check, and the `run()`/`Receipt` scaffolding here are a near line-for-line duplicate of the equivalent code in `transfer.rs`. That's not a bug today, but it means the two commands can silently drift — e.g. a future fix to how muxed sources are rejected, how fees are defaulted, or how the receipt is rendered gets applied to one and missed in the other, since nothing enforces they stay in sync. Might be worth factoring the shared parts (config assembly, muxed-source guard, receipt shape) into a common helper once a third command needs the same pattern.

---

cmd/soroban-cli/src/commands/token/approve.rs
```diff
@@ -0,0 +1,223 @@
...
+        // the source account (which also signs and authorizes), `spender` is the
+        // delegate being granted the allowance.
+        //
+        // The invoke pipeline can't source a transaction from a muxed account
+        // yet (see #2645), and a muxed strkey in the `from` arg is rejected
+        // mid-simulation with an opaque host error; reject it up front with a
+        // clear message instead.
+        let source_account = config.source_account()?;
+        if matches!(source_account, crate::xdr::MuxedAccount::MuxedEd25519(_)) {
+            return Err(Error::MuxedSourceNotSupported);
```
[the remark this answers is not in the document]
fnando: Added `approve_rejects_muxed_source_with_clear_error` in 98aa3df7 — passes an `M…` value as `--from` and asserts the clear "muxed (M…) source accounts are not yet supported" message, mirroring the transfer test. (`mint`/`clawback` carry the same branch and will get the equivalent test in their PRs.)

cmd/soroban-cli/src/commands/token/approve.rs
```diff
@@ -0,0 +1,223 @@
...
+            Error::Invoke(_) => "invoke",
+            Error::Serde(_) => "internal",
+            Error::MuxedSourceNotSupported => "unsupported",
+        }
+    }
+}
+
+/// The machine-readable receipt of a token approval.
+#[derive(Debug, serde::Serialize)]
+struct Receipt {
```
leighmcculloch, line 114: Fwiw receipt is not a common term in the Stellar vocab. I'd call this Result rather than introduce a new concept. It's not a blocker though, as it appears to just be a term being used internal to the cli.

leighmcculloch (approved):

cmd/soroban-cli/src/commands/token/approve.rs
fnando: Fair observation on the asymmetry. The difference is cost: `amount < 0` is a free static check, but validating `expiration_ledger` against "current ledger" needs an RPC round-trip to fetch the latest ledger. A stale expiration with a positive amount fails safely at simulation (nothing submitted), so we deliberately don't add that guard. The doc states the contract's requirement, not a CLI-enforced one — leaving it as-is.

cmd/soroban-cli/src/commands/token/approve.rs
fnando: Good point. "receipt" is only an internal Rust struct name here — the JSON output field is already `result` (`{tx_hash, result}`), so consumers never see the term. Since it's not user-facing and you flagged it as non-blocking, I'll leave it as-is for now rather than rename across all the write commands. Happy to revisit if we want to standardize the internal naming later.

PR #2716, opened 2026-09-04
Add `stellar token mint` subcommand

fnando:
### What

Adds `stellar token mint`, a SAC-admin write subcommand that mints new tokens. `--admin` signs and authorizes the mint (the asset issuer for a Stellar Asset Contract), `--to` is the recipient, and `--amount` is the quantity in smallest units. Returns a JSON receipt with the tx hash.

### Why

Part of #2620 (typed SEP-41 + SAC client), first of the SAC-admin commands. A thin wrapper over `contract invoke` reusing `args::invoke_by_position` and `args::not_deployed_error`. Unlike `transfer`/`approve`, the admin only authorizes — it is not a `mint` argument — so only `[to, amount]` are passed positionally.

### Known limitations

### DIFF
--- a/cmd/soroban-cli/src/commands/token/mint.rs
+++ b/cmd/soroban-cli/src/commands/token/mint.rs
@@ -1,0 +1,211 @@
+use clap::Parser;
+
+use crate::{
+    commands::{
+        contract::invoke,
+        global,
+        token::args::{self, OutputFormat},
+    },
+    config::{
+        self, locator, network, sign_with, token::UnresolvedToken, UnresolvedMuxedAccount,
+        UnresolvedScAddress,
+    },
+    output::Output,
+};
+
+#[derive(Debug, Parser, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// The token to mint: a contract id or alias, `native`, or a classic asset
+    /// as `CODE:ISSUER`.
+    #[arg(long = "id")]
+    pub id: UnresolvedToken,
+
+    /// The token's administrator. Signs and authorizes the mint, so it must be
+    /// an identity or secret key you control (the asset issuer for a Stellar
+    /// Asset Contract).
+    #[arg(long)]
+    pub admin: UnresolvedMuxedAccount,
+
+    /// Account or contract to mint the tokens to. Accepts a `G…`/`M…` account, a
+    /// `C…` contract address, or an alias.
+    #[arg(long)]
+    pub to: UnresolvedScAddress,
+
+    /// Amount to mint, in the token's smallest unit (stroops for a Stellar Asset
+    /// Contract).
+    #[arg(long, value_parser = parse_nonneg_i128)]
+    pub amount: i128,
+
+    /// Format of the output.
+    #[arg(long, default_value = "text")]
+    pub output: OutputFormat,
+
+    #[command(flatten)]
+    pub network: network::Args,
+
+    #[command(flatten)]
+    pub locator: locator::Args,
+
+    #[command(flatten)]
+    pub sign_with: sign_with::Args,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error(transparent)]
+    Config(#[from] config::Error),
+    #[error(transparent)]
+    Network(#[from] network::Error),
+    #[error(transparent)]
+    Args(#[from] args::Error),
+    #[error(transparent)]
+    Token(#[from] config::token::Error),
+    #[error(transparent)]
+    ScAddress(#[from] config::sc_address::Error),
+    #[error(transparent)]
+    Invoke(#[from] invoke::Error),
+    #[error(transparent)]
+    Serde(#[from] serde_json::Error),
+
+    #[error(
+        "muxed (M…) source accounts are not yet supported for `token mint`; \
+         use the underlying G… account as `--admin` instead"
+    )]
+    MuxedSourceNotSupported,
+}
+
+/// Parse `--amount` as a non-negative `i128`. A negative mint amount is always
+/// invalid, so reject it at the clap layer instead of letting it reach the
+/// contract and fail as an opaque `HostError` deep in simulation.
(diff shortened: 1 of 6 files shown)


cmd/soroban-cli/src/commands/token/mint.rs
```diff
@@ -0,0 +1,211 @@
...
+        "muxed (M…) source accounts are not yet supported for `token mint`; \
+         use the underlying G… account as `--admin` instead"
+    )]
+    MuxedSourceNotSupported,
+}
+
+/// Parse `--amount` as a non-negative `i128`. A negative mint amount is always
+/// invalid, so reject it at the clap layer instead of letting it reach the
+/// contract and fail as an opaque `HostError` deep in simulation.
+fn parse_nonneg_i128(value: &str) -> Result<i128, String> {
```
leighmcculloch: `parse_nonneg_i128` is byte-for-byte duplicated across `transfer.rs`, `approve.rs`, and `mint.rs` instead of living once in `token/args.rs`. A future fix to the negative-amount validation message or logic (e.g. adding a max-value check) has to be applied in three places by hand; missing one leaves that command with stale/inconsistent validation behavior.

---

cmd/soroban-cli/src/commands/token/mint.rs
```diff
@@ -0,0 +1,211 @@
...
+
+        let token = self
+            .id
+            .resolve(&config.locator, &network.network_passphrase)?;
+
+        // The admin only authorizes the mint; it is not itself a `mint` argument.
+        // The invoke pipeline can't source a transaction from a muxed account yet
+        // (see #2645), so reject a muxed admin up front with a clear message.
+        let source_account = config.source_account()?;
+        if matches!(source_account, crate::xdr::MuxedAccount::MuxedEd25519(_)) {
```
leighmcculloch: The muxed-source rejection block (resolve `source_account`, `matches!` `MuxedEd25519` check, `MuxedSourceNotSupported` error) is duplicated near-identically across `transfer.rs`, `approve.rs`, and this file. When muxed-account support lands (referenced issue #2645) and this guard needs to be removed or changed, it's easy to update two of the three copies and leave one command still rejecting valid muxed sources.

---

cmd/soroban-cli/src/commands/token/mint.rs
```diff
@@ -0,0 +1,211 @@
...
+        // In JSON mode the underlying invoke pipeline's human-readable status
+        // logging (which writes to stderr) would still fire; run it quietly so
+        // machine consumers get clean output without needing `--quiet`.
+        let quiet = global_args.quiet || output.is_json();
+        let config = self.config();
+        let network = config.get_network()?;
+
+        let token = self
+            .id
+            .resolve(&config.locator, &network.network_passphrase)?;
```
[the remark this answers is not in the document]
fnando: Fixed in 6c37ee31 — added a shared `args::reject_native_admin` guard that rejects `native` up front with a typed `native_not_administered` error, dropped it from the `--id` help, and added a test. Same guard applied to `clawback` (#2717).

cmd/soroban-cli/src/commands/token/args.rs
```diff
@@ -102,3 +109,16 @@ pub fn not_deployed_error(token: &ResolvedToken, err: &invoke::Error) -> Option<
...
 }
+
+/// Guard for admin-only operations (`mint`, `clawback`, `set-admin`,
+/// `set-authorized`): the native asset's SAC is deployed without an
+/// administrator (see `contract deploy asset`, which only writes the `Admin`
+/// entry for non-native assets), so any admin call against it is guaranteed to
+/// fail deep in simulation. Reject `native` up front with a typed error naming
+/// the `operation` instead. Non-native tokens pass through untouched.
+pub fn reject_native_admin(token: &ResolvedToken, operation: &'static str) -> Result<(), Error> {
+    if matches!(token.asset(), Some(crate::xdr::Asset::Native)) {
```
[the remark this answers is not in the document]
fnando: Good catch. But the guard only ever catches the native SAC — any other un-administrable target (custom token with no `mint`, or a SAC you don't admin) still hits the same opaque `HostError(Storage, MissingValue)`, and we can't detect administrability in general (the `Admin` entry is SAC-only). Since it already fails safely at simulation, I've dropped the special-case guard entirely. Reverted in the latest push.

cmd/soroban-cli/src/commands/token/mint.rs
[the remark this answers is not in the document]
fnando: Reverted — the native-only guard was inconsistent (every other un-administrable target still errored opaquely). See the args.rs thread. Latest push.

cmd/soroban-cli/src/commands/token/mint.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// The token to mint: a contract id or alias, `native`, or a classic asset
+    /// as `CODE:ISSUER`.
+    #[arg(long = "id")]
+    pub id: UnresolvedToken,
+
+    /// The token's administrator. Signs and authorizes the mint, so it must be
+    /// an identity or secret key you control (the asset issuer for a Stellar
+    /// Asset Contract).
+    #[arg(long)]
+    pub admin: UnresolvedMuxedAccount,
```
leighmcculloch, line 28: It's surprising that admin is a parameter here. The SAC's mint fn doesn't require it as an input. And I'm not aware of other contracts that provide mint fns requiring it either.

So I think it should be removed, and rely on the standard `--source` for specifying the signer.

Note that the mint fn is not part of SEP-41, it is a SAC fn, so these token commands are a mix of SEP-41 and SAC fns.

leighmcculloch (approved):

cmd/soroban-cli/src/commands/token/mint.rs
leighmcculloch: Hmm, I guess --admin makes it more explicit. I'm on the fence, approving as is and defer to you but worth considering.

PR #2718, opened 2026-09-04
Add `stellar token set-admin` subcommand

fnando:
### What

Adds `stellar token set-admin`, a SAC-admin write subcommand that transfers administration of a token. `--admin` (the current admin) signs and authorizes the change, and `--new-admin` is the successor. Returns a JSON receipt with the tx hash.

### Why

Part of #2620 (typed SEP-41 + SAC client), a SAC-admin command alongside `mint`/`clawback`. A thin wrapper over `contract invoke` reusing `args::invoke_by_position` and `args::not_deployed_error`; like the other admin commands, the admin only authorizes and is not a `set_admin` argument, so only `[new_admin]` is passed positionally. The happy-path test verifies control transferred behaviorally — after `set-admin`, the new admin can `mint`.

### Known limitations

### DIFF
--- a/cmd/soroban-cli/src/commands/token/set_admin.rs
+++ b/cmd/soroban-cli/src/commands/token/set_admin.rs
@@ -1,0 +1,193 @@
+use clap::Parser;
+
+use crate::{
+    commands::{
+        contract::invoke,
+        global,
+        token::args::{self, OutputFormat},
+    },
+    config::{
+        self, locator, network, sign_with, token::UnresolvedToken, UnresolvedMuxedAccount,
+        UnresolvedScAddress,
+    },
+    output::Output,
+};
+
+#[derive(Debug, Parser, Clone)]
+#[group(skip)]
+pub struct Cmd {
+    /// The token to re-administer: a contract id or alias, or a classic asset as
+    /// `CODE:ISSUER`.
+    #[arg(long = "id")]
+    pub id: UnresolvedToken,
+
+    /// The token's current administrator. Signs and authorizes the change, so it
+    /// must be an identity or secret key you control (the asset issuer for a
+    /// Stellar Asset Contract).
+    #[arg(long)]
+    pub admin: UnresolvedMuxedAccount,
+
+    /// The new administrator to hand control to. Accepts a `G…`/`M…` account, a
+    /// `C…` contract address, or an alias.
+    #[arg(long = "new-admin")]
+    pub new_admin: UnresolvedScAddress,
+
+    /// Format of the output.
+    #[arg(long, default_value = "text")]
+    pub output: OutputFormat,
+
+    #[command(flatten)]
+    pub network: network::Args,
+
+    #[command(flatten)]
+    pub locator: locator::Args,
+
+    #[command(flatten)]
+    pub sign_with: sign_with::Args,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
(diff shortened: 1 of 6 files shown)


cmd/soroban-cli/src/commands/token/set_admin.rs
```diff
@@ -0,0 +1,193 @@
...
+        // In JSON mode the underlying invoke pipeline's human-readable status
+        // logging (which writes to stderr) would still fire; run it quietly so
+        // machine consumers get clean output without needing `--quiet`.
+        let quiet = global_args.quiet || output.is_json();
+        let config = self.config();
+        let network = config.get_network()?;
+
+        let token = self
+            .id
+            .resolve(&config.locator, &network.network_passphrase)?;
```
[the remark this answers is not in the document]
fnando: Same settled call as #2716 (mint) and #2717 (clawback): we don't special-case native. A guard only catches the native SAC — every other un-administrable target (a `C…` contract with no `set_admin`, or a SAC you don't admin) fails identically with the same opaque `HostError(Storage, MissingValue)`, and administrability isn't detectable in general (the `Admin` entry is SAC-only). It already fails safely at simulation, so we leave native as an accepted `--id` reference form.

cmd/soroban-cli/src/commands/token/set_admin.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        // The admin only authorizes the change; it is not itself a `set_admin`
+        // argument. The invoke pipeline can't source a transaction from a muxed
+        // account yet (see #2645), so reject a muxed admin up front with a clear
+        // message.
+        let source_account = config.source_account()?;
+        if matches!(source_account, crate::xdr::MuxedAccount::MuxedEd25519(_)) {
```
leighmcculloch: The muxed-source-account rejection is re-implemented identically in every SAC-admin subcommand (`mint`, `clawback`, now `set_admin`) instead of living once in the shared invoke/config pipeline.

A future SAC-admin subcommand (e.g. a planned `freeze`/`unfreeze`) could be added by copying this file again but the author forgets to paste the `matches!(source_account, MuxedEd25519(_))` check; a muxed `--admin` then reaches deep into the invoke pipeline and fails with an opaque strkey/HostError instead of the clear, typed `MuxedSourceNotSupported` message users get from `mint`/`clawback`/`set-admin` today. Centralizing the check in `config::Args::source_account()` or the invoke pipeline itself would fix it once for all current and future admin commands and remove the risk of a missed copy-paste.

---

cmd/soroban-cli/src/commands/token/set_admin.rs
leighmcculloch: `set_admin.rs` duplicates nearly all of `clawback.rs`/`mint.rs` (`Cmd` struct shape, full `Error` enum, `config()`, `Receipt`, and the entire `run()` body) with only field names and the invoked function changed.

Not a crash, but a maintenance cost: any future fix to the shared `run()` logic (e.g. changing how the tx_hash is echoed, or how build-only receipts are handled) now has to be made correctly in three near-identical ~200-line files; a fix applied to one and missed in another silently diverges behavior between `token mint`/`clawback`/`set-admin`. A shared harness in `token/args.rs` (taking the function name, positional args, and doc strings) would remove most of this duplication.

---

leighmcculloch (approved):

PR #2720, opened 2026-09-05
Shake by the recorded version

leighmcculloch:
> [!NOTE]
> Part of a stack of PRs that must merge in this order.
>
> Discussion is here:
> - stellar/stellar-protocol#2014
>
> A first group of PRs deliver const-encoded contract specs, so that contract specs are produced at compile time instead of at proc-macro execution time. This provides the foundation for the capability to construct the specs from information that is not known at proc-macro execution and only known at compile time, like the fully qualified name of a type:
> 1. stellar/rs-stellar-xdr#580
> 1. stellar/rs-soroban-env#1736
> 1. stellar/rs-soroban-sdk#2061
> 1. stellar/rs-soroban-sdk#2078
> 1. stellar/rs-soroban-sdk#2080
> 1. stellar/rs-soroban-sdk#1965
>
> A second group of PRs deliver fully qualified type names in contract specs. Instead of a type having the name `Context` it will have the name `soroban_sdk::auth::Context`. Qualified type names make it possible to uniquely identify types in the spec, even when they have the same name. This resolves several problems with contract specs the type identify problem (https://github.com/stellar/rs-soroban-sdk/issues/1570), type aliases limitations (https://github.com/stellar/rs-soroban-sdk/issues/1857 https://github.com/stellar/rs-soroban-sdk/issues/1063), and optimise spec shaking data section size (https://github.com/stellar/rs-soroban-sdk/issues/1978):
> 1. stellar/stellar-xdr#312
> 1. stellar/rs-stellar-xdr#543
> 1. stellar/rs-soroban-sdk#1970
> 1. stellar/stellar-cli#2674
>
> A third group of PRs use those qualified names to shake specs by reachability, so that the markers a contract carries shrink to the few entries a spec never names:
> 1. stellar/rs-soroban-sdk#2043
> 1. stellar/stellar-cli#2720 ← this PR

### What

Patch in the soroban-sdk change that shakes specs by reachability (stellar/rs-soroban-sdk#2043), announce `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V3` to the SDK alongside the existing v2 var, and pick the shaking rules from the `rssdk_spec_shaking` version the built wasm records rather than from a single fixed set of rules.

### Why

Each var names a set of shaking rules, and this CLI announces every set it can apply, because the SDK refuses to build a contract whose rules the build system has not claimed. The v3 var is required by that SDK: from that version most types carry no marker, and a build system has to shake them out by following the references to them instead, so one announcing only v2 would shake by markers alone and strip every type. The v2 var stays set because it is still true of this CLI, and a contract built against an older published SDK looks for that var alone and refuses to build without it.

Which rules apply to a given wasm is a property of that contract, not of the CLI — it may have been built long before this version — so the CLI reads the version the contract recorded and passes it to `soroban_spec::shaking::filter`: version 2 keeps an entry only if it carries a marker, version 3 also follows the references from functions, events, and types, and a wasm recording no version at all is left unshaken.

### Known limitations

Contract fixtures built by the tests resolve soroban-sdk from crates.io rather than through the workspace patch, so they cover the version 2 path and the SDK's own end-to-end test contract covers version 3. The `[patch.crates-io]` git revisions are replaced with published versions once stellar/rs-stellar-xdr#566 and the soroban-sdk stack land and release.

### DIFF
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -342,11 +342,14 @@ impl Cmd {
                 cmd.env("CARGO_BUILD_RUSTFLAGS", rustflags);
             }
 
-            // Set env var to inform the SDK that this CLI supports spec
-            // optimization. The var says only that this CLI shakes; which
-            // rules it shakes a given contract by comes from the version that
-            // contract records in its meta.
+            // Set env vars to inform the SDK that this CLI supports spec
+            // optimization. Each var names a set of shaking rules, and this
+            // CLI announces every set it can apply, because the SDK refuses to
+            // build a contract whose rules the build system has not claimed.
+            // Which rules a given contract is shaken by comes from the version
+            // that contract records in its meta.
             cmd.env("SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2", "1");
+            cmd.env("SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V3", "1");
 
             let cmd_str = serialize_command(&cmd);
 
--- a/cmd/soroban-cli/src/commands/contract/build.rs
+++ b/cmd/soroban-cli/src/commands/contract/build.rs
@@ -342,14 +342,11 @@ impl Cmd {
                 cmd.env("CARGO_BUILD_RUSTFLAGS", rustflags);
             }
 
-            // Set env vars to inform the SDK that this CLI supports spec
-            // optimization. Version 3 is announced because this CLI shakes
-            // types that carry no marker of their own, by following the
-            // references to them. Version 2 is announced alongside it, and
-            // remains true of this CLI, because a contract on an older SDK
-            // looks for that var alone and refuses to build without it.
+            // Set env var to inform the SDK that this CLI supports spec
+            // optimization. The var says only that this CLI shakes; which
+            // rules it shakes a given contract by comes from the version that
+            // contract records in its meta.
             cmd.env("SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2", "1");
-            cmd.env("SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V3", "1");
 
             let cmd_str = serialize_command(&cmd);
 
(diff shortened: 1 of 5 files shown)


cmd/soroban-cli/src/commands/contract/build.rs
leighmcculloch: CI note on head `5430005` — everything is green except `test generated libraries (ubuntu-latest-8-cores)`, and that one is not this PR's.

The Rust workflow completed successfully across all 12 jobs (`check`, `build-and-test` msrv/latest × x86_64/arm64, `cargo-deny` ×4, `publish-dry-run`), as did every other workflow: the 30-job `build-crate` matrix, `build` ×4, `test`, `emulator-tests`, `docs-validation`, and `action` ×4. Note `check` is red on the base branch (stellar/stellar-cli#2674) but passes here.

`test generated libraries` fails because the JS SDK cannot read the spec entries at all:

```
TypeError: XDR Read Error: saw 63 length String, max allowed is 60
  › String.read (@stellar/js-xdr@4.0.0/src/string.js:17:13)
  › processSpecEntryStream (@stellar/stellar-sdk/src/contract/utils.ts:186:30)
```

and, for the SAC client:

```
TypeError: XDR Read Error: saw 44 length String, max allowed is 32
```

`@stellar/js-xdr@4.0.0` still enforces the old 60-byte UDT name and 32-byte symbol limits, which the qualified names from the base stack (stellar/stellar-xdr#312, stellar/rs-soroban-sdk#1970) exceed. It clears when those length changes reach a published JS SDK. The same job fails on the base PR stellar/stellar-cli#2674 (run 33889813965). Nothing in this PR touches name lengths — it only changes which spec entries survive shaking — so there is no fix to port and I have not widened the PR with one.

Locally against the bumped SDK rev: `cargo check --workspace --all-targets` and `cargo clippy --workspace --all-targets -- -D warnings` both clean, and `cargo test -p soroban-test --test it build::` passes 27/27, including `build_with_spec_shaking_has_feature_meta`, `build_with_spec_shaking_filters_unused_types`, `build_with_spec_shaking_filters_unused_events`, `build_with_spec_shaking_preserves_all_functions`, and `build_without_spec_shaking_preserves_all_entries`.


---

leighmcculloch: CI note on head `001d120` — everything is green except `test generated libraries (ubuntu-latest-8-cores)`, and that one is not this PR's.

Green: the Rust workflow in full (`check`, `build-and-test` msrv/latest × x86_64/arm64, `cargo-deny` ×4, `publish-dry-run`), the 30-job `build-crate` matrix, `build` ×7, `RPC Tests`, and `docs-validation`. Note `check` is red on the base branch (stellar/stellar-cli#2674) but passes here.

**`test generated libraries`** fails because the JS SDK cannot read the spec entries at all:

```
TypeError: XDR Read Error: saw 63 length String, max allowed is 60
  › String.read (@stellar/js-xdr@4.0.0/src/string.js:17:13)
  › processSpecEntryStream (@stellar/stellar-sdk/src/contract/utils.ts:186:30)
```

and, for the SAC client:

```
TypeError: XDR Read Error: saw 44 length String, max allowed is 32
```

`@stellar/js-xdr@4.0.0` still enforces the old 60-byte UDT name and 32-byte symbol limits, which the qualified names from the base stack (stellar/stellar-xdr#312, stellar/rs-soroban-sdk#1970) exceed. It clears when those length changes reach a published JS SDK. The same job fails on the base PR stellar/stellar-cli#2674 (run 33889813965). Nothing in this PR touches name lengths — it only changes which vars are announced to the SDK and which spec entries survive shaking — so there is no fix to port and I have not widened the PR with one.

**One transient failure worth recording:** `test (ubuntu-latest-8-cores)` failed on the first attempt of this head with 143 passed / 1 failed, `integration::auth::non_root_auth_mode_via_env_var` dying at `contract deploy --source=test` with `Account not found: GABZP2TVYQ6IQUHB7OGBZWCHIEWCKDXDAO2OPPZ64DDSHK7DES5OWRDT` — the local standalone network had not funded the test account, before any spec logic ran. Attempt 2 passed. That was the one re-run; a repeat would be treated as real.

Locally against the bumped SDK rev: `cargo check --workspace --all-targets` and `cargo clippy --workspace --all-targets -- -D warnings` both clean, and `cargo test -p soroban-test --test it build::` passes 27/27, including `build_with_spec_shaking_has_feature_meta`, `build_with_spec_shaking_filters_unused_types`, `build_with_spec_shaking_filters_unused_events`, `build_with_spec_shaking_preserves_all_functions`, and `build_without_spec_shaking_preserves_all_entries`. On the SDK side I also confirmed the v3 gate bites: building `tests/spec_shaking` with only `SOROBAN_SDK_BUILD_SYSTEM_SUPPORTS_SPEC_SHAKING_V2=1` fails with `soroban-sdk requires stellar-cli v28.1.0+`, which is what this PR's second env var answers.


---

willemneal:
The title of this PR has popped up a couple times now and I can't not think of it being about dancing to a recorded version.

PR #2724, opened 2026-09-15
Deprecate `stellar contract bindings typescript`

fnando:
### What

Adds a deprecation warning when running `stellar contract bindings typescript`, pointing users to the JavaScript Stellar SDK CLI at https://github.com/stellar/js-stellar-sdk#cli. The warning is printed via `Print::warnln`, so it respects `--quiet` and goes to stderr.

### Why

TypeScript bindings generation is moving to the JavaScript Stellar SDK. This gives users a clear heads-up and directs them to the supported path before the command is eventually removed.

### Known limitations

N/A


cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
```diff
@@ -55,6 +55,11 @@ impl Cmd {
     pub async fn execute(&self, quiet: bool) -> Result<(), Error> {
         let print = Print::new(quiet);
 
+        print.warnln(
+            "`stellar contract bindings typescript` is deprecated. Use the JavaScript Stellar SDK \
+             instead: https://github.com/stellar/js-stellar-sdk#cli",
+        );
```
leighmcculloch, line 61: We should add to the commands doc commenet `/// ⚠️ Deprecated, use X` so that the deprecation shows up in the help, not only the output when running it.

jeesunikim (approved):

Ryang-21 (approved):

leighmcculloch (commented):
One follow up:

cmd/soroban-cli/src/commands/contract/bindings/typescript.rs
fnando: good call! I'll follow up on this.

PR #2734, opened 2026-09-18
Stream image-pull progress for container builds

fnando:
### What

Reworks the container build's image-pull output. `stellar contract build --image <ref> --pull` now runs the pull through a reusable `shared::Args::pull_image` that streams the engine's own progress **to our stderr** (the engine writes pull progress to stdout, which we redirect to stderr so our stdout stays clean; the engine's stderr is inherited and streams live). Under `--quiet` both streams are discarded (`Stdio::null()`), so there's no pull log at all. When the pull fails, it surfaces as `shared::Error::PullImageFailed { image }` — reported to the user as `could not pull image` — which already flows through `Error::Engine`.

The container-local `pull_image`/`PullImageFailed` are removed in favor of the shared versions, and `build::Error: From<shared::Error>` lets the call site use a plain `?`.

### Why

Split out of #2709 (verifiable builds) at review request: the pull helper is independent of `--verifiable` and improves the existing `--image` build on its own. Pulling it into its own PR keeps #2709 focused and lets this land separately.

### Known limitations

N/A


cmd/soroban-cli/src/commands/container/shared.rs
```diff
@@ -261,6 +266,58 @@ impl Args {
...
+            .spawn()
+            .map_err(|e| self.io_error(e))?;
+
+        let stdout = child.stdout.take();
+        let stderr = child.stderr.take();
+
+        let stream_stdout = async {
+            if let Some(stdout) = stdout {
+                let mut lines = BufReader::new(stdout).lines();
+                while let Ok(Some(line)) = lines.next_line().await {
```
leighmcculloch, line 289: This loops stops if there's a read error, but the pipe will still be open and the command keeps running and I think this means that the OS pipe buffer behind the pipe has the risk of filling and blocking the command being run from completing. Essentially a deadlock.

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -95,7 +92,10 @@ pub async fn run(
     // `pull` up front to refresh a moving tag to its newest image. Nothing is
     // pulled when only printing the command, since nothing runs.
     if !print_only && cmd.pull {
-        pull_image(&docker, image, print).await?;
+        docker
+            .pull_image(image, print)
```
leighmcculloch: The unconditional "Pulling image {image}" status message (previously printed at the top of the old free-function `pull_image`) was removed here and not replaced by anything guaranteed to print, since the new `Args::pull_image` only echoes lines matching specific engine-output substrings.

**Failure scenario:** For `--engine apple-container` (a first-class supported engine per this file's `Engine` enum), if Apple's `container image pull` doesn't emit lines containing exactly "Pulling from", "Digest", or "Status" (unverified — the substring set is copied from docker's known output format), the user gets zero output for the entire duration of a `--pull` that can take minutes, whereas before this PR they always saw at least "Pulling image X" up front regardless of engine.

---

cmd/soroban-cli/src/commands/container/shared.rs
```diff
@@ -261,6 +266,58 @@ impl Args {
         };
         cmd
     }
+
+    /// Pull `image`, streaming the engine's high-level status lines ("Pulling
+    /// from", "Digest", "Status") through `print`. Per-layer progress written to
+    /// stderr is captured rather than shown and surfaced only when the pull
+    /// fails, as `PullImageFailed`, so a failed pull can report the engine's own
+    /// error. A missing engine binary surfaces via `io_error` as `NotFound`.
+    pub(crate) async fn pull_image(&self, image: &str, print: &Print) -> Result<(), Error> {
```
leighmcculloch: New `Args::pull_image` duplicates the pre-existing, unmodified `Runner::pull_image` in `container/start.rs` (same stdout-line-filter heuristic) instead of being reused there, and the two now behave differently for the same underlying operation.

**Failure scenario:** `container/start.rs::pull_image` (lines ~125-152, untouched by this PR) inherits stderr straight to the terminal when not `--quiet`, so `stellar container start` still shows docker's raw per-layer progress bars live. The new `shared::Args::pull_image` always captures stderr into memory and shows it only on failure, so `stellar contract build --image X --pull` never shows live per-layer progress even when not quiet. Two call sites doing conceptually the same "pull with filtered status output" now diverge in UX, and a future bug fix (e.g. the UTF-8 issue flagged elsewhere in this review) has to be applied twice.

---

cmd/soroban-cli/src/commands/contract/build/container.rs
```diff
@@ -95,7 +92,10 @@ pub async fn run(
...
+            .await
+            .map_err(Error::Engine)?;
```
leighmcculloch, line 98: I don't think the map err is required here with the way the error is defined from I noticed a #[from] above for it anyway.

cmd/soroban-cli/src/commands/container/shared.rs
```diff
@@ -261,6 +266,58 @@ impl Args {
...
+        let mut child = self
+            .pull_command(image)
+            .stdout(Stdio::piped())
```
leighmcculloch, line 278: How does this work when `--quiet` is used?

leighmcculloch (approved):
Couple minor things but lgtm.

cmd/soroban-cli/src/commands/container/shared.rs
fnando: Fixed in 1704d551: `pull_image` no longer pipes/parses stdout — it redirects the engine's stdout to our stderr (its stderr inherited), so there's no read loop that could stop draining and deadlock the child.

cmd/soroban-cli/src/commands/container/shared.rs
fnando: Fixed in 1704d551: under `--quiet` both the engine's stdout and stderr go to `Stdio::null()` (no pull log at all); otherwise they stream to our stderr.

cmd/soroban-cli/src/commands/contract/build/container.rs
fnando: Fixed in 1704d551: added `build::Error: From<shared::Error>` so plain `?` works, and dropped the `map_err`.

cmd/soroban-cli/src/commands/container/shared.rs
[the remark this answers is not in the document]
fnando: Added `integration::pull::build_image_pull_reports_failure` asserting a failed pull surfaces `could not pull image`; the success/streaming path is exercised by `--pull` in the verifiable-build integration test.

cmd/soroban-cli/src/commands/container/shared.rs
```diff
@@ -261,6 +291,32 @@ impl Args {
...
+    /// Pull `image`, streaming the engine's own progress to our stderr so our
+    /// stdout stays clean (the engine writes pull progress to stdout, which we
+    /// redirect to stderr; its stderr is inherited). Under `quiet` both are
+    /// discarded. A missing engine binary surfaces via `io_error` as `NotFound`.
+    pub(crate) async fn pull_image(&self, image: &str, quiet: bool) -> Result<(), Error> {
+        let (stdout, stderr) = if quiet {
+            (Stdio::null(), Stdio::null())
+        } else {
+            (stderr_as_stdio(), Stdio::inherit())
+        };
```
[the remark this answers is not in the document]
fnando: Pipe/parse was dropped in 1704d551 to fix a deadlock @leighmcculloch flagged. Now: engine stdout→our stderr, stderr inherited, both nulled under `--quiet`, failure = `could not pull image`. The PR description was the stale part — updated.

cmd/crates/soroban-test/tests/it/integration/pull.rs
```diff
@@ -0,0 +1,26 @@
...
+    sandbox
+        .new_assert_cmd("contract")
+        .current_dir(&fixture)
+        .arg("build")
+        .arg("--image")
+        .arg("docker.io/library/stellar-cli-does-not-exist-zzz:latest")
+        .arg("--pull")
+        .assert()
+        .failure()
+        .stderr(predicate::str::contains("could not pull image"));
```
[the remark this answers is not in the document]
fnando: That output contract was the old stdout-filtering design, dropped in 1704d551 — nothing left to filter. Failure is covered by `build_image_pull_reports_failure`; not adding a fake engine.

cmd/soroban-cli/src/commands/container/shared.rs
```diff
@@ -1,10 +1,37 @@
...
+use std::process::Stdio;
 
 use clap::ValueEnum;
 use tokio::process::Command;
 
 use crate::print::Print;
 
+/// The current process's stderr as a `Stdio`, so a child's stdout can be routed
+/// to our stderr — keeping our own stdout clean. Falls back to inheriting on the
+/// rare fd/handle clone failure.
```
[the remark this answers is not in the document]
fnando: Fixed in 93570a65 — the fallback now returns `Stdio::null()` (and same for non-unix/windows targets), so a clone failure drops the progress instead of leaking it onto our stdout.

